PR #16, opened 2012-12-18
Potential fix for issue 9556

mrfr0g:
https://dev.ckeditor.com/ticket/9556

A simple way of reproducing this issue is to type the word, 'Hello' in
to a blank document, press enter, then type the letter 'w'. The
auto-correct mechanism will attempt to join the two text nodes as one,
with the most common suggestion being 'yellow'.

The core of this issue is mobile safari's reliance on detecting when
the enter key has been pressed. It's not mentioned in the ticket, but
when you press enter the next line of the document is also not
auto-capitialized.

I have found only two ways of fixing this issue, the first and most
obvious is to ask the user to disable the auto-correct setting in their
device. The second is to allow the browser to execute the event
normally (not preventDefault), and execute a deferred function. The
deferred function should first; execute the undo command (this will
remove the browser created element), then execute the enter logic
normally.

This does open the door for a potential visual issue. While in debug
(running _source) mode the cursor sometimes shows on the previous node,
then is placed back on the new node. In production mode I do not see
this issue.

It should also be mentioned that this fixes both the auto-correct and
auto-capitalization issues.

Test environment;
I am using the iPad 2 and running iOS 6.0.1.

### DIFF
--- a/core/htmldataprocessor.js
+++ b/core/htmldataprocessor.js
@@ -259,7 +259,7 @@
 				}
 
 				// From IE DOM, at the end of a text block, or before block boundary.
-				if ( CKEDITOR.env.ie && isOutput && ( !atBlockEnd || node.parent.name in textBlockTags ) )
+				if ( CKEDITOR.env.ie && isOutput && ( !atBlockEnd || node.parent && node.parent.name in textBlockTags ) )
 					return true;
 
 				// From the output.
--- a/plugins/enterkey/plugin.js
+++ b/plugins/enterkey/plugin.js
@@ -13,6 +13,10 @@
 				editorFocus: false,
 				exec: function( editor ) {
 					enter( editor );
+					// Return false to not prevent the default behavior 
+					if(CKEDITOR.env.iOS) {
+						return false;
+					}
 				}
 			});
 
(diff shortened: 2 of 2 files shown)


pjasiun:
I have checked these changes with the newest CKEditor (4.5 beta) and the Safari on the iOS 8.3 (iPad mini) and when I press enter the first paragraph is removed so the solution seems not to work. In the same manner, the bug is keep valid on the newest iOS.

mrfr0g:
Not really a surprise, the pull request was made 2 years ago ... a lot changes in two years. At this point I dont think it makes sense to keep this open.

Reinmar:
Yeah... I'm sorry for this situation. We weren't ready to handle PRs at that moment. The [test suite](http://docs.ckeditor.com/#!/guide/dev_tests) wasn't public yet, there was no [contributors guide](http://docs.ckeditor.com/#!/guide/dev_contributing_code) and we didn't know how to handled PRs that we couldn't review at the moment. The problem is that reviewing and finishing patches (writing tests, testing, polishing the code) takes usually a lot of time (in case of complex tickets it's even a day or more) and there's rarely a person who can spare that much, so we were postponing reviewing just like I did. Sometimes... postponing forever.

Fortunately, with time we learnt how to deal with this situation. We opened the tests suite, published more guides, added Grunt tasks for checking code style and linting, etc. Also, we try to give feedback as soon as possible, so the contributor still remembers the patch and may improve it if it's not perfect yet.

So again, I'm very sorry that we wasted your patch. It won't happen again :).

PR #30, opened 2013-02-22
Feature request with attached patch - allow a visual spacer between toolbar groups

plumlee:
Our visual design requires a vertical spacer between toolbar groups. Please see attached code and sample page. We are also looking to be able to use this to do some responsive design, where we can use this as a hook for moving toolbars up and down into available space as page width allows. However, this patch is only for the visual display. 

CSS by [Kevin Parkerson](https://github.com/kevinparkerson)


f1ames:
Hi,

PR #39, opened 2013-05-07
Allow flow elements as children of anchor tags per HTML5 spec

heydenberk:


tiemevanveen:
Any updates on this pr or this issue http://dev.ckeditor.com/ticket/9457? It's been a while but the editor is still rejecting those structures. The test suite is now public right?

AuthorProxy:
Up the same problem with this code:

``` js
<a class="product" href="#">
  <div class="image">
    <img src="someimage.jpg">
  </div>
  <div class="text">
    <span>debet cards</span>
  </div>
</a>
```

PR #56, opened 2013-09-07
Check if iframe exists before clearing custom data

jimmykane:
The issue was introduced when someone uses ckeditor as an AngularJS directive. 

I wrote a directive that before initializing an instance of CKEDITOR if first destroys all existing ones. 
So far so good. But if I have a ckeditor instance initilized and I remove the directive with angularJS (eg inside repear) then the iframe becomes null and thus a error popsup. 

'Uncaught TypeError: Cannot call method 'clearCustomData' of null'

This is the case scenario: 

1. Bind an instance to an element using the CKEDITOR.replace
2. Having the instance open remove the element using ANGULARJS
3. Try to use destroy on all existing instances. 
4. Error. 'Uncaught TypeError: Cannot call method 'clearCustomData' of null'

### DIFF
--- a/plugins/wysiwygarea/plugin.js
+++ b/plugins/wysiwygarea/plugin.js
@@ -497,19 +497,21 @@
 				// Memory leak proof.
 				this.clearCustomData();
 				doc.getDocumentElement().clearCustomData();
-				iframe.clearCustomData();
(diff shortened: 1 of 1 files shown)


Reinmar:
We'll give #200 a try and it's nearly the same patch, so I'm closing this PR.

PR #109, opened 2014-06-27
Add table row properties dialog

Jelle-S:
This pull request allows users to add classes, id and style attributes to a table row.

### DIFF
--- a/plugins/menu/plugin.js
+++ b/plugins/menu/plugin.js
@@ -540,6 +540,6 @@ CKEDITOR.plugins.add( 'menu', {
  */
 CKEDITOR.config.menu_groups = 'clipboard,' +
 	'form,' +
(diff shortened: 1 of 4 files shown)


Reinmar:
Hi,

I'm very sorry for the lack of response. We optimistically assumed that we'll be able to finish your PR and we reported http://dev.ckeditor.com/ticket/12269. But it turned out that we weren't able to work on this since then. We should have replied you with what's missing in the PR instead.

If you're still interested to work on the row properties dialog, which we would also like to see in the core, the most important thing to do are tests. Here you can read about [CKEditor testing environment](http://docs.ckeditor.com/#!/guide/dev_tests). Another thing is code style issues - you can check it with `grunt jscs:all && grunt jshint:all` after installing Grunt (we also use git hooks, but we didn't yet document this - see `grunt --help`).

mlewand:

PR #125, opened 2014-09-10
Make iFrame body have a minHeight equal to the config.height of t...

aljachimiak:
...he editor (#10770)

I ran into an issue while using an iPad.  I found that making the height of the iFrame's body larger make the area focus-able.

This also may make other event focused solutions (see line 213 or 221) to the same problem unnecessary.

### DIFF
--- a/plugins/wysiwygarea/plugin.js
+++ b/plugins/wysiwygarea/plugin.js
@@ -178,6 +178,14 @@
 			body.removeAttribute( 'disabled' );
 		}
 
+		// Fix problem with cursor not appearing in iOS (iPad) when clicking below the body (#10770).
+		if( typeof(this.editor.config.height) == number ) {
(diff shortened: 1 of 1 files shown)


xkem:
Issue is closed and not fixed: https://dev.ckeditor.com/ticket/10770

f1ames:
@xkem Could you provide reproduction steps and details on your environment where you have encountered this issue?

thirdender:
This issue is happening again in 4.11.4 and can be seen on an iPad at https://ckeditor.com/docs/ckeditor4/latest/examples/autogrow.html

Tapping in the white area of the editor does not bring up the on-screen keyboard, the user _must_ tap on editable text (or where the first line would appear if the editor has no text entered yet). Not sure if previous fixes have been reverted or are no longer effective.

PR #129, opened 2014-09-19
#12077 Add support for download attribute to link plugin

sbusse:
This change adds a checkbox to set the HTML5 download attribute in the advanced tab of the link dialog, allowing to force a download.

Please note that setting a custom file name for the download as mentioned in the comments of the related ticket was not implemented this time. Having two fields (checkbox + textfield) managing one single attribute would have led to a much more complicated change.

download-flag

### DIFF
--- a/plugins/link/dialogs/link.js
+++ b/plugins/link/dialogs/link.js
@@ -110,7 +110,7 @@
 		return {
 			title: linkLang.title,
 			minWidth: 350,
-			minHeight: 230,
+			minHeight: 240,
 			contents: [
 				{
 				id: 'info',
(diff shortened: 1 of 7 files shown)


uwolfer:
Any update?

Do you think this option should be in the tab "Target" instead of "Advanced"?

uwolfer:
Any chances to get this into 4.5?

Reinmar:
Oops. I'm very sorry guys, but this PR just got lost somewhere in the PRs list :(((. We're review it soon, so it can land in 4.6.0.

I would normally need to ask for tests... but taken into account our response time, we'll handle this.

szymonkups:
Thank You for the PR!
I merged changes into major branch with 6eaa9e4. Changes will be included in CKEditor 4.6.0.

I've added tests and `requiredContent` field for the checkbox so it can be activated/deactivated by content filter.

PR #134, opened 2014-10-06
Update image.js

eon8:
We have to update preview after the original image was loaded.
If we do that - the preview will have correct size after dialog is opened.
At this moment an image with a reduced size has 100% dimensions in dialog preview block before we make any changes and another update is triggered.

### DIFF
--- a/plugins/image/dialogs/image.js
+++ b/plugins/image/dialogs/image.js
@@ -217,6 +217,8 @@
 
 					this.firstLoad = false;
 					this.dontResetSize = false;
+
+					updatePreview(this);
 				};
 
 			var onImgLoadErrorEvent = function() {


Reinmar:
This is a weeeeird issue. I spent a good time trying to figure out WTH and didn't succeed. I can reproduce it but only on editor loaded from CDN (our demo uses our CDN) what made it impossible to debug.

Your fix looks good though, so I decided to give it a try. However, I'm not able to verify that it indeed fixes the issue, so I closed it conditionally. We'll wait until the release and if the next version (4.4.7) turns out to be ok, we'll ultimately close this issue. This is a ridiculous way of closing a ticket, because we could try to test a RC on the CDN, but I don't want to spend more time on so small detail :D.

I reported a ticket: http://dev.ckeditor.com/ticket/12818 and merged your change to master with 89e2c9b. We'll add a changelog entry when we'll verify that everything works.

Reinmar:
OK, I can confirm it's fixed now. I could not be able to reproduce the same issue on our demo. Thanks again for the patch. We can now close the issue :)

PR #137, opened 2014-10-13
Fix to ensure any ascendant tables found are within editor

Paul-Martin:
When tableresize plugin attempts to determine if mouse is currently
over a table, it performs a search of the ascendants of the event
target.  However this search will continue outside of the editable
container, and thus find tables in which the editable may be
embedded.  This fix adds a check to ensure any found table is in
fact a descendant of the editable container.

### DIFF
--- a/plugins/tableresize/plugin.js
+++ b/plugins/tableresize/plugin.js
@@ -385,6 +385,12 @@
 
 					table = target.getAscendant( 'table', 1 );
 
+					// Make sure the table we found is inside the container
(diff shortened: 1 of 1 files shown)


adelura:
Related ticket on track [#12825](http://dev.ckeditor.com/ticket/12825).

adelura:
Thanks for PR! We merge it in https://github.com/ckeditor/ckeditor-dev/commit/a66138ec790acf6a2777e1ecf352ad640464f987.

PR #138, opened 2014-10-27
Css class/css style edit dialog

mconta:
This simple plugin adds the possibility to edit/add CSS classes to every visible HTML element and to add/edit the css style field definition. I find it useful because it's difficult to edit CSS class definition without opening the HTML editor in source code view.


mlewand:

PR #140, opened 2014-11-06
save plugin inline editor ajax save optional feat.

elcodedocle:
Leaves the current functionality for `CKEDITOR.ELEMENT_MODE_REPLACE` without any modifications, adding a few extra lines for handling `CKEDITOR.ELEMENT_MODE_INLINE` editors: 

If a target is specified on an inline editor container's html5 `data-actionUrl` param and the AJAX plugin is available, the standard save button will show up in the editor. 

The button will show up disabled and it will be enabled only when the editor content is dirty. 

When clicked, it will trigger an AJAX POST request containing the editor id and content (post params `editorID` and `editabledata`, urlencoded).

### DIFF
--- a/plugins/save/lang/en.js
+++ b/plugins/save/lang/en.js
@@ -3,5 +3,6 @@ Copyright (c) 2003-2014, CKSource - Frederico Knabben. All rights reserved.
 For licensing, see LICENSE.md or http://ckeditor.com/license
 */
 CKEDITOR.plugins.setLang( 'save', 'en', {
-	toolbar: 'Save'
+	toolbar: 'Save',
+	ko: 'Error: Save operation has failed.'
 } );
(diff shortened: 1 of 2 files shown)


Reinmar:
Hi, thanks for the PR and sorry for the late reply.

I think that these new features are a candidate for a new plugin. The editor's core features should not interfere too much with the integration layer, because there are nearly infinite possibilities. Your implementation works in specific way and will not satisfy countless other cases. The same with the current CKEditor's integration with forms - it's nice and it's actually good that we have, because it's tricky, but for instance a data validation is a problem.

As for the save plugin itself it exists now in its current for the legacy reasons. If we were designing it today the plugin would only expose a button and a command doing... nothing. The rest should stay in hands of developers integrating CKEditor with their systems. Other, 3rd party plugins could of course provide this integration.

I'm closing this PR, but if you decide to work on your plugin, then it would be a great addition to the [Addons Repository](http://ckeditor.com/addons/plugins/all).

elcodedocle:
Hey! That's fine. I agree more generic is the way to go.

I will follow your advice and consider working on it and send it to the addons repository.

ck-editor is an amazing piece of software, both from a user and a developer point of view, I want to say thank you very much to you and every other contributor working on this project :-)

PR #142, opened 2014-11-07
Added event-facilities to the autoloader

elektronaut1024:
Added event-facilities to the autoloader

This allows to listen for when all required scripts have been loaded asynchronously, when ckeditor itself is loaded on demand after the document is already finished.

Previously there was no way to wait until everything was available and only then initialize the editor.

### DIFF
--- a/core/loader.js
+++ b/core/loader.js
@@ -103,6 +103,35 @@ if ( !CKEDITOR.loader ) {
 			};
 
 		var pendingLoad = [];
+		var handlers = {
+			ready: []
+		};
+		
+		function unregister(event,callback){
+			handlers[event] = handlers[event].filter(function(handler){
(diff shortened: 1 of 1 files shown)


Reinmar:
I've finally looked into this case and I'm still unsure what may be fixed here.

When loading this way a built version of CKEditor, then the `CKEDITOR.stylesSet` and all other core modules are available directly when the `load` event is fired.

And it's impossible to load a development version of CKEditor this way, because it tries to write to the document and that's prohibited:

```
Failed to execute 'write' on 'Document': It isn't possible to write into a document from an asynchronously-loaded external script unless it is explicitly opened.
```

However, that's not a problem that I feel we need to fix in the development version. I don't also think it can be fixed without making the loading process asynchronous and we can't do that.

PR #145, opened 2014-12-02
Custom data support for widget commands

zipp3r:
Just pass the second editor.execCommand argument to Widget instance initialization


Reinmar:
I made a small cleanup (code style, variable name), added API docs (pity that commands are not documented :() and merged changes to the major branch. See 2dbc4f3.

> We should also implement the edit function 

I think it's ok the way you guys proposed it. It makes sense that this data is passed only to the constructor, especially that the property is called `startupData`. It does not have to affect the `edit()` call, because instead of executing the command, one could simply execute the `setData()` method and, if needed, then the command. Basically, the widget creation scenario is the most complex one there, so it makes sense to cover only it, because in the other scenarios a developer can simply call the right method directly.

Thanks for the great contribution, guys!

PR #147, opened 2014-12-03
Stop removing disabled items from menu.

danyaPostfactum:
Fix for http://dev.ckeditor.com/ticket/12723

Context menu now contains permanent cut/copy menu items (disabled when selection is empty). Is it ok?

We could also modify `contextmenu` plugin, so it would hide disabled items. Not sure.


TheDigitalOrchard:
I agree that such options should remain visible, but disabled when they can't be used. It gives added context that a "cut" or a "copy" operation _could_ be performed in the current context when the appropriate conditions are met (i.e. some text is selected).

f1ames:
Hi,
Thank you for supporting Open Source and contributing to the CKEditor 4 project!

PR #148, opened 2014-12-08
Cell split bug 8024

khirl:
Bug fix for http://dev.ckeditor.com/ticket/8024

`Shouldn't the table feature 'split cell horizontal/vertical' be switched?
'Split horizontal' actually splits the cell vertical. 'Split vertical' actually splits the cell horizontal.
`
tab key based navigation ckeditor sample


kevinisagit:
Hi Reinmar, Is there more you need to get this PR ready and merged?

PR #153, opened 2014-12-20
Fixed Uncaught SecurityError for Chrome/IE

thecatontheflat:
This PR fixes the issue when trying to init CKEditor inline within an iframe

### DIFF
--- a/plugins/lineutils/plugin.js
+++ b/plugins/lineutils/plugin.js
@@ -59,8 +59,9 @@
 			win: editor.window
 		}, def, true );
 
-		this.frame = this.win.getFrame();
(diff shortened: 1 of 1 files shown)


Reinmar:
Thanks for the details. I confirmed the issue and reported it on our Trac (http://dev.ckeditor.com/ticket/12812).

To accept the patch we need tests (a manual and an automated one if possible), but in this case it requires a deep knowledge about our testing system, so I've started working on this: https://github.com/cksource/ckeditor-dev/commits/t/12812

Reinmar:
I merged the PR to master with 33557ff.

Thanks!

PR #155, opened 2015-01-12
#12796 - Fix list unwrapping

andrewstucki:
This pull attempts to fix http://dev.ckeditor.com/ticket/12796 by returning immediately when executing indent/outdent on a list where the common ancestor of the selected nodes happens to be the editor itself  or another block element that is in the editor, but not contained within an ol or ul (such as a div).

The while loop without this conditional causes the selection to propagate up to parent list elements outside of an editor in inline mode and modify them, breaking the editor itself.

If some default behavior is desired rather than just returning immediately before calling indent, let me know.

### DIFF
--- a/plugins/indentlist/plugin.js
+++ b/plugins/indentlist/plugin.js
@@ -232,13 +232,13 @@
 			range;
 
 		while ( ( range = iterator.getNextRange() ) ) {
-			var rangeRoot = range.getCommonAncestor(),
-				nearestListBlock = rangeRoot;
+			var nearestListBlock = range.getCommonAncestor();
 
 			while ( nearestListBlock && !( nearestListBlock.type == CKEDITOR.NODE_ELEMENT && context[ nearestListBlock.getName() ] ) ) {
 				// Avoid having plugin propagate to parent of editor in inline mode by canceling the indentation. (#12796)
 				if ( editor.editable().equals( nearestListBlock ) ) {
-					return 0;
+					nearestListBlock = false;
+					break;
 				}
 				nearestListBlock = nearestListBlock.getParent();
 			}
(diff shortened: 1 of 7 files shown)


Reinmar:
Thanks for the PR. I pushed https://github.com/cksource/ckeditor-dev/tree/t/12796 with:

* corrected code style,
* switch to `editor.editable()` for the limiting element,
* more precise test assertions,
* additional assertion which checks that the outdent feature worked.

If you would test an identical scenario in a classic (iframed) editor, then the first list would be outdented. It's not the best result possible, but user partially achieves some result, what's much better than doing nothing (what may be confusing). Additionally, though I haven't checked this, it may happen that if only one list was selected:

```html
<ul>[<li>1</li>]</ul>
```

nothing's gonna happen with the patch that you provided.

There's one more case:

```html
<ul>[<li>1</li>]</ul><ol><li>2</li></ol>
(or similar selection, but only of the first item)
```

which I checked that falls into the same behaviour - the algorithm leaks.

To sum up:

1. At least two more tests are needed.
1. The outdent command should work in all 3 cases.

andrewstucki:
Just saw the update, thanks for the feedback. Will implement sometime in the next few days when I get the chance.

andrewstucki:
Integrated the default behavioral changes and added the requested test cases.

andrewstucki:
@Reinmar: Just wondering if there are any other issues you want me to fix before the merge?

Reinmar:
I haven't managed to check your PR this week and next week I'm on holidays. I'll ask one of my colleagues to help.

Thanks!

pjasiun:
I merged your tests into one file so I can use one editor instead of 3. This way tests execute faster and, regarding that there are more then 3k tests, performance does matter. I checked changes and now everything works the same way in classic and inline editors, so I merged them into master. Good job!

pjasiun:
Changes will be available in 4.4.8.

andrewstucki:
Awesome, thanks so much!

PR #156, opened 2015-01-21
Ticket #12838: Fix for saving undo images to account for firing the 'change' event

Skxiao:
When the change event is fired, if the receiving function changes the text in any way, the UndoManager does not update the save image to the changed version. Therefore, it saves the version before the change was fired, which is the incorrect version for the snapshot.

Steps to reproduce:

1. Create an onChange function that receives the 'change' event and edits the text when executed
2. Make a change that would trigger the event with the correct parameters to edit the text
3. Try to undo the change

Expected: Will undo the change
Actual: Undo only reverts to intermediate step (which may or may not trigger the onChange event again)

Fix: replace old image with a new image after the change event is fired.

### DIFF
--- a/plugins/undo/plugin.js
+++ b/plugins/undo/plugin.js
@@ -379,6 +379,10 @@
 				}
 			}
 
+			// create new content image after change event is fired 
+			// to update the image before pushing it to snapshots
+			var image = new Image( editor );
+
 			// Drop future snapshots.
 			snapshots.splice( this.index + 1, snapshots.length - this.index - 1 );
 


Reinmar:
The `editor#change` event was introduced mostly for informative purposes like auto saving. It was neither meant to be super precise (what the [doc](http://docs.ckeditor.com/#!/api/CKEDITOR.editor-event-change) explains) nor used in such way.

Of course, we could extend its capabilities, but this is a complicated piece of code and may affect performance, therefore:

1. Tests would need to be created. In this case it means looots of tests.
2. Other cases where the `editor#change` event is fired would need to be reviewed.
3. More performant solution would need to be implemented. In your solution second image is created every time what's simply wasteful.

From my experience it may be a month of work, so I really advise to find an alternative solution. For example - fire `editor#updateSnapshot` after a short timeout.

PR #157, opened 2015-01-23
#5547: Prefer attributes vs. inline style for height/width in image plugin

jameswilson:
This is a pull request that properly fixes the trac issue [#5547](https://dev.ckeditor.com/ticket/5547). This work is based in part on the patch from [comment #21](https://dev.ckeditor.com/ticket/5547#comment:21) by user @pivotal on that issue.

The solution does a couple things differently than that patch:
1. first checks to verify that the activeFilter allows width and height _attributes_ using `img[width]` and `img[height]` [allowed content rule](http://docs.ckeditor.com/#!/guide/dev_allowed_content_rules-section-string-format); if so it updates or adds those attributes, _and removes any inline height/width css styles_.
2. if not, it then checks to verify that the activeFilter allows width and height rules in inline css styles using `img{width}` and `img{height}` test; if so it updates/adds the inline css styles and removes the height/width attributes.
3. Finally, if neither the attributes or inline styles are supported, it removes both.

Yes, I realize  [#5547](https://dev.ckeditor.com/ticket/5547) was closed with the suggestion to use `image2` instead, however the `image` plugin is still in the repository which makes this still fair game for a fix.  Image2 is also not supported by the Drupal 7 wysiwyg or media modules (haven't tested ckeditor module because this is for an existing site and we don't have the bandwidth to switch at this point).

There was one line (below) dealing with attribute removal for "Non-internal commits" that I wasn't entirely sure of its purpose, but since since the removal is now handled in other parts of the patch, I opted to simply remove this, because it seemed now out of place.

```
!internalCommit && element.removeAttribute( 'width' );
```

Thanks for your review of this pull request.  I will also add a comment referring here and add an updated diff patch file on the trak issue as well.

### DIFF
--- a/plugins/image/dialogs/image.js
+++ b/plugins/image/dialogs/image.js
@@ -599,20 +599,39 @@
 										commit: function( type, element, internalCommit ) {
 											var value = this.getValue();
 											if ( type == IMAGE ) {
-												if ( value && editor.activeFilter.check( 'img{width,height}' ) )
-													element.setStyle( 'width', CKEDITOR.tools.cssLength( value ) );
-												else
+												if ( value ) {
+													if ( editor.activeFilter.check( 'img[width]' ) ) {
+														element.setAttribute( 'width', value );
+														element.removeStyle( 'width' );
+													} else if ( editor.activeFilter.check( 'img{width}' ) ) {
+														element.setStyle( 'width', CKEDITOR.tools.cssLength( value ) );
+														element.removeAttribute( 'width' );
+													}
+												} else {
+													element.removeAttribute( 'width' );
 													element.removeStyle( 'width' );
-
-												!internalCommit && element.removeAttribute( 'width' );
+												}
 											} else if ( type == PREVIEW ) {
 												var aMatch = value.match( regexGetSize );
 												if ( !aMatch ) {
 													var oImageOriginal = this.getDialog().originalElement;
-													if ( oImageOriginal.getCustomData( 'isReady' ) == 'true' )
-														element.setStyle( 'width', oImageOriginal.$.width + 'px' );
+													if ( oImageOriginal.getCustomData( 'isReady' ) == 'true' ) {
+														if ( editor.activeFilter.check( 'img[width]' ) ) {
+															element.setAttribute( 'width',  oImageOriginal.$.width );
+															element.removeStyle( 'width' );
+														} else if ( editor.activeFilter.check( 'img{width}' ) ) {
+															element.setStyle( 'width',  oImageOriginal.$.width + 'px' );
+															element.removeAttribute( 'width' );
+														}
+													}
 												} else {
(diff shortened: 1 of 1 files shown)


jswiderski:
What about using dissallowedContent (since CKE 4.4) manually as described here: http://docs.ckeditor.com/#!/guide/dev_acf-section-example%3A-disallow-inline-styles-and-use-attributes-instead?

Perhaps I don't see the whole picture here but I don't think that such pull request is needed. if one wants to use attributes we simply configures CKEditor ACF properly.

Reinmar:
> What about using dissallowedContent (since CKE 4.4) manually as described here: http://docs.ckeditor.com/#!/guide/dev_acf-section-example%3A-disallow-inline-styles-and-use-attributes-instead?

See https://dev.ckeditor.com/ticket/5547#comment:32

jswiderski:
Let's match this pull request with an open ticket in our bug tracker - http://dev.ckeditor.com/ticket/12774.

Once this is reviewed, we will close the issue.

Reinmar:
I've finally managed to review the PR.

First of all, I've been trying to understand what `internalCommit` does. And after an hour I ended up with http://dev.ckeditor.com/ticket/13300. We'll remove it, so don't worry about it.

As for the code - generally it is ok, but there are few things that need to be improved before we'll be able to merge it:

* It should be `height` here, not `width`:
  * https://github.com/BluesparkLabs/ckeditor-dev/commit/ab0e56cbcd18562285230a440971307d615870bd#diff-2c9cc095ac642a0862b5b88ed71281c7R671
  * https://github.com/BluesparkLabs/ckeditor-dev/commit/ab0e56cbcd18562285230a440971307d615870bd#diff-2c9cc095ac642a0862b5b88ed71281c7R672
* The code can be simplified, because that "check filter, remove one, add another" part repeats many times. It can be moved to a separate function like `setSize( editor, element, property, value )`.
* We also need to change the validators - now they will accept values with `px` and `%`, when only [`\d+`](http://w3c.github.io/html-reference/datatypes.html#common.data.integer.non-negative) is allowed in attributes.
* And then goes tests - we would need at least one for adding a new image and perhaps also one for editing an image which already had some attrs/styles (but see the next point). You can read about our [testing environment](http://docs.ckeditor.com/#!/guide/dev_tests) and also, recently, we posted a guide about [contributing patches](http://docs.ckeditor.com/#!/guide/dev_contributing_code) were some details about other tools such as linter were included.
* I've also thought that styles should have higher priority than attributes for the sake of backwards compatibility. So if both styles and attributes are allowed, then styles should be used. This can be achieved by changing the order of the conditions.

Reinmar:
The `internalCommit` arg was removed - 10a7105.

f1ames:
Hi,
Thank you for supporting Open Source and contributing to the CKEditor 4 project! The https://dev.ckeditor.com/ticket/5547 issue has been already resolved some time ago so we' re closing this PR. Thank you!

PR #158, opened 2015-01-25
Update tr.js

guncebektas:


wwalc:
Thank you for your patch!

In general, our localization process is done through the [CKEditor UI Translation Center](https://www.transifex.com/projects/p/ckeditor/) at Transifex and is [documented here](http://docs.cksource.com/CKEditor_3.x/Developers_Guide/Localization). I ported your change there and would like to invite you to join the Turkish team. Since the Turkish localization is still incomplete (`image2` plugins and `plugin-a11yhelp-dialogs` translations are incomplete at this moment), we could really use your help!

Thanks again!

PR #161, opened 2015-02-11
XML Declaration should be the first in the output

friedolinfoerder:
The xml declaration should be the first in the output string (see http://www.w3.org/wiki/Doctypes_and_markup_styles), so the if statement has to be the last.
Also see http://stackoverflow.com/questions/28458015/ckeditor-moves-xml-tag for further information

### DIFF
--- a/plugins/wysiwygarea/plugin.js
+++ b/plugins/wysiwygarea/plugin.js
@@ -493,11 +493,11 @@
 						data = data.replace( /<br>(?=\s*(:?$|<\/body>))/, '' );
 
 					data = editor.dataProcessor.toDataFormat( data );
-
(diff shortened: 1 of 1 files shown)


Reinmar:
Hi,

Thanks for the PR - it looks promising.

I'm not sure now when we'll be able to truly review it, because we're now totally devoted to finishing CKEditor 4.5.0, but if not before closing it, then shortly after that.

Also, it would help if you could add a test. The wysiwygarea plugin's tests are scattered over other directories, but you could add one for this change in https://github.com/friedolinfoerder/ckeditor-dev/tree/master/tests/plugins/wysiwygarea

mlewand:

PR #162, opened 2015-02-17
Problem paste(CTL+V) in chrome

kgtstr:
When paste(CTL+V) in chrome,  all of text element is add <span style="font-weight: normal;">XXXX</span>
cause:

plugins/clipboard/plugin.js   L781
----------------------------------------------------
var selPastebin = range.select();
// If non-native paste is executed, IE will open security alert and blur editable.
// Editable will then lock selection inside itself and after accepting security alert
// this selection will be restored. We overwrite stored selection, so it's restored
// in pastebin. (#9552)
if ( CKEDITOR.env.ie ) {
blurListener = editable.once( 'blur', function() {
editor.lockSelection( selPastebin );
} );
}

next code is fixed.
----------------------------------------------------
// If non-native paste is executed, IE will open security alert and blur editable.
// Editable will then lock selection inside itself and after accepting security alert
// this selection will be restored. We overwrite stored selection, so it's restored
// in pastebin. (#9552)
if ( CKEDITOR.env.ie ) {
var selPastebin = range.select();
blurListener = editable.once( 'blur', function() {
editor.lockSelection( selPastebin );
} );
}


Reinmar:
I'm sorry, but this PR is incorrect. Please first open a ticket on http://dev.ckeditor.com (if one doesn't already exist) so we know what your PR fixes. Then please make a new PR with only your commits in it.

PR #163, opened 2015-02-25
Prevent an exception when destroying an editor not fully instantiated

cyril-sf:
@addepar/uie @dylnuge @dinahshi


cyril-sf:
closing

PR #164, opened 2015-02-25
Added Disallowed Content example

tiemevanveen:


Reinmar:
Thanks :). I'm afraid that your patch will be wasted, because we already killed the old samples in 4.5.0. As you can see on the [major branch](https://github.com/ckeditor/ckeditor-dev/tree/major/samples) they are gone. I will merge your changes to master, though, so to make at least this minimal use of them :).

Reinmar:
BTW. The old samples are replaced by the SDK:

* https://github.com/ckeditor/ckeditor-sdk
* http://sdk.ckeditor.com/

As well as by the new ["Functionality Overview" section](http://docs.ckeditor.com/#!/guide/dev_features) in the guides.

tiemevanveen:
I bit off-topic, but I created the examples when I was looking into the disallowContent option to prevent pasted-in widgets, do you know if that is possible? I've added a question on [SO](http://stackoverflow.com/questions/28722121/ckeditor-how-to-use-the-advanced-content-filter-and-widgets-together-in-2-edit).

Reinmar:
I replied on SO.

tiemevanveen:
Saw it, thanks.

PR #165, opened 2015-02-26
Add and fix Korean language translations

misol:
Add Korean language file in which plugins have no Korean translation files.
Fix Korean language file to be read smoothly.


misol:
related to http://dev.ckeditor.com/ticket/12989

PR #167, opened 2015-03-05
Multiple nested editable for widget plugin

zipp3r:
Hello, i need the ability to create multiple nested editable areas of same type in one widget.
See http://dev.ckeditor.com/ticket/12524
In addition I have added the 'editableKey' event.


zipp3r:
Is there anybody out? :)
Maybe you review the code and if everything is OK I'll add tests?

Reinmar:
Hi, sorry for lack of response, but until CKEditor 4.5.0 is released we will not be able to work on non-must-have topics (this includes PRs). 4.5.0 is our absolute priority right now.

As for the PR - it's very hard to tell now, without digging deeper, what obstacles are there. So we'll not be able to tell if it's going to work and how close you are just by reading the code. It looks promising, but after many years in the business I'm not naive :dancer: 

Definitely, the first step to confirm that this approach is realistic will be to update the existing tests - basically what had been passing must be passing. Right now on your branch I have (on http://tests.ckeditor.dev:1030/#tests/is:unit,path:/tests/plugins/widget): `263 passed / 45 failed in 28s 853ms - 100%` . Then, based on working samples of widget with multiple-nested-editables we can check if the new feature works for the users - manual clicking is the quickest way, because quite effortlessly we can check many scenarios.

So this is the must-have for us to start working on this PR. Then of course the missing automated tests must be created, but since you didn't work in a TDD way, it can be done at the end, when we'll be quite sure that the feature is feasible.

PS. You should create a branch in your repository. We name branches after ticket numbers - i.e. `t/12542`.

chaegumi:
I also need this function, but after thinking about it, I think it may not be very necessary. The compromise is that in this layout requirement, there will not be many editable areas. Predefine a few first, and then use the class in the editable area. The predefined class can be deleted when the front end is output later.

```
editables: {
                editable1: {
                    selector: '.jeawin-col-editable1',
                },
                editable2:{
                    selector: '.jeawin-col-editable2'
                },
                editable3 :{
                    selector: '.jeawin-col-editable3',
                },
                editable4: {
                    selector: '.jeawin-col-editable4',
                },
                editable5:{
                    selector: '.jeawin-col-editable5'
                },
                editable6 :{
                    selector: '.jeawin-col-editable6',
                },
                editable7: {
                    selector: '.jeawin-col-editable7',
                },
                editable8:{
                    selector: '.jeawin-col-editable8'
                },
                editable9 :{
                    selector: '.jeawin-col-editable9',
                },
                editable10: {
                    selector: '.jeawin-col-editable10',
                },
                editable11:{
                    selector: '.jeawin-col-editable11'
                },
                editable12:{
                    selector: '.jeawin-col-editable12',
                }
            },
```

```
<div class="jeawin-row">
                <div class="jeawin-col jeawin-col-editable1">12</div>
            </div>

<div class="jeawin-row">
                                            <div class="jeawin-col jeawin-col-6 jeawin-col-editable1">
                                                6
                                            </div>
                                            <div class="jeawin-col jeawin-col-6 jeawin-col-editable2">
                                                6
                                            </div>
                                        </div>

<div class="jeawin-row">
                                            <div class="jeawin-col jeawin-col-4 jeawin-col-editable1">
                                                4
                                            </div>
                                            <div class="jeawin-col jeawin-col-4 jeawin-col-editable2">
                                                4
                                            </div>
                                            <div class="jeawin-col jeawin-col-4 jeawin-col-editable3">
                                                4
                                            </div>
                                        </div>
```

PR #169, opened 2015-03-09
Update de.js

bnw:
The deleted segment translates to "the current content will be deleted!".
This statement is wrong, if templates_replaceContent is set to false. This statement is not present in the english version.


wwalc:
Thank you for your patch!

In general, our localization process is done through the [CKEditor UI Translation Center](https://www.transifex.com/projects/p/ckeditor/) at Transifex and is [documented here](http://docs.cksource.com/CKEditor_3.x/Developers_Guide/Localization). I ported your change there and would like to invite you to join the German team. Since the German localization needs improvements (as your pull requests confirms), we could really use your help!

Thanks again!

PR #170, opened 2015-03-10
Allow the editor to be destroyed before being fully initialized

cyril-sf:
This seems to be related to http://dev.ckeditor.com/ticket/11502

I understand that this is not a solution to all problems ( I'm not familiar enough with the implementation of CKEditor ), but it seems to cover some basic needs without changing the APIs.

### DIFF
--- a/core/editor.js
+++ b/core/editor.js
@@ -180,7 +180,9 @@
 
 		// Return the editor instance immediately to enable early stage event registrations.
 		CKEDITOR.tools.setTimeout( function() {
-			initConfig( this, instanceConfig );
+			if( this.status !== 'destroyed' ) {
+				initConfig( this, instanceConfig );
+			}
 		}, 0, this );
 	}
 
@@ -758,8 +760,10 @@
 
 			this.editable( null );
 
-			this.filter.destroy();
-			delete this.filter;
+			if( this.filter ) {
+				this.filter.destroy();
+				delete this.filter;
+			}
 			delete this.activeFilter;
 
 			this.status = 'destroyed';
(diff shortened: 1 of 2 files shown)


Reinmar:
Hi,

Thanks for the PR. You're right that this is not a complete solution, but indeed it handles an important case when editor is destroyed while editor's config is being retrieved. There are unfortunately at least few other situations when editor still cannot be destroyed cleanly (setting data, changing mode, wysiwygarea initialisation, etc.), but since you made the PR (which looks good), there's no sense in rejecting it. Someone will benefit from it :).

I'm not sure now when we'll be able to truly review it, because we're now totally devoted to finishing CKEditor 4.5.0, but if not before closing it, then shortly after that.

cyril-sf:
Thanks for the feedback. We will definitely benefit from it at Addepar if we could avoid managing our own fork!

cyril-sf:
http://dev.ckeditor.com/ticket/11502#comment:15

> They deal with a very small part of this problem. None of them was also fully complete, so we couldn't just accept them.

I'm surprised to discover somewhere else that this PR can't be accepted, when the only thing you said here is that `since you made the PR (which looks good), there's no sense in rejecting it`.

Reinmar:
You misunderstood me. I haven't written that your PR will not be accepted. I wrote that it's not complete in a way that would allow us to accept it without any additional work. See for example [#13632](http://dev.ckeditor.com/ticket/13632). We need this feature in order to be able to improve your PR.

Anyway, I expect that we'll work on all related issues (including your PR) this week, so let's see where we are then. I hope everything will be already on master :).

szymonkups:
Once again, thanks for the PR. I've added log message when editor is destroyed before is fully initialized and merged pull request with a122cdbfcdf7e4bf2aa04794bc147b6dbf286242.

cyril-sf:
Thanks!

PR #172, opened 2015-03-16
Update de.js

navido:
fix l10n


AnnaTomanek:
Hi Navido,

thanks a lot for this!

In general, we manage the localization process through the [CKEditor UI Translation Center at Transifex](https://www.transifex.com/projects/p/ckeditor/). We do have the [German language group](https://www.transifex.com/projects/p/ckeditor/language/ro/) there, too, so it would be really cool if you wanted to join us and work on it if you find anything missing/wrong.

In the meantime, I copied over your suggestions to the German localization file and the updates will be available in the next CKEditor release. Thanks again for your help!

PR #173, opened 2015-03-17
Fix hungarian translation

nkovacs:
Alt text is NOT tooltip text


wwalc:
Thank you for your patch!

In general, our localization process is done through the [CKEditor UI Translation Center](https://www.transifex.com/projects/p/ckeditor/) at Transifex and is [documented here](http://docs.cksource.com/CKEditor_3.x/Developers_Guide/Localization). I ported your change there and would like to invite you to join the Hungarian team. Since the Hungarian localization is still missing a few translations, we could really use your help!

Thanks again!

nkovacs:
Thanks

PR #174, opened 2015-03-17
The ckeditor produce the duplicate values

jiang01yi:
Hi, i am using the ckEditor in jquery ui dialog, everything is working fine.
But when I am using the inertText/insertHtml, it randomly insert the duplicated value in chrome. firefox and IE is working fine, just Chrome.
Please help me fix that. thank you


Reinmar:
Hi. It's not the right place to report issues. There's already a ticket that describes similar problem - http://dev.ckeditor.com/ticket/12995, however, we couldn't reproduce the issue. If you could provide more details about how to reproduce it, then perhaps it will allow us to do so.

PR #175, opened 2015-03-19
When ckeditor be used in web page which content is refresh with ajax

liuinsect:
I user ckeditor in a web page which content is refresh with ajax,
I can see the ckeditor ,but when I click the plugin button, I get " Error: [CKEDITOR.dialog.openDialog] Dialog "about" failed when loading definition." error , 
even though I config the CKEDITOR.plugins.basePath , it is also wrong,
why ckeditor can not config the plugin path ?????

as a famous tool, why it is so hard to work?


Reinmar:
See http://docs.ckeditor.com/#!/guide/dev_basepath. If it were possible to make it automatic in every case we would do so. But it isn't.

PS. This is not the right place to ask questions. You can:

* report bugs on http://dev.ckeditor.com
* ask questions on http://stackoverflow.com/questions/tagged/ckeditor

liuinsect:
i have found the problem， the reason is i put the "  <script src="/misc/module/ckeditor/ckeditor.js"></script>" in a velocity file which is load  by ajax .
I will try to find the real reason , if necessary  I will repotr a bug

liuinsect:
by the way , I have  use the basePath,but not work

PR #176, opened 2015-03-25
Fix: skip empty nodes in getByAddress to match the logic in getIndex.

melindaweathers:
This is my proposed fix for https://dev.ckeditor.com/ticket/13089 - please see ticket for more information.  Not sure if it is the right way to fix the problem for everyone, but it solved the problem in my application.


Reinmar:
Thanks for the PR. I'm afraid that the bug is in a different place though. 

See https://dev.ckeditor.com/ticket/13089#comment:3

PR #177, opened 2015-03-31
Updated Forms plugin to allow support of HTML5 Form Validation plugin

AWHEELERTWS:


pjasiun:
Hi,

Thank you for the contribution. That is nice to see that another modern web feature may be integrated with CKEditor. Although, I have 3 comments.

First is technical: one plugin should know nothing about any other plugin, that is the rule, so `forms` plugin can not have any referenced to the `html5validation`. Otherwise there would be a mess with checking which plugins combination exists. Fortunately you should be able to extend `allowedContent` rules in the `html5validation` plugin.

Secondly, you did not include any licence information, which is needed to let others use your plugin. I recommend to the same license that CKEditor core and others plugins use, so the users can use it without learning more about licences. CKEditor is distributed under the GPL, LGPL and MPL, you can include header similar to this https://github.com/ckeditor/ckeditor-dev/blob/master/plugins/basicstyles/plugin.js#L1-L4 but with you name in it.

Finally, we try to keep core repository small and include only most needed and most popular plugins here. You can add your plugin on the http://ckeditor.com/addons/plugins/ so we can check how popular it will be and if users want to have such feature.

Because you can add it there without pull request I am closing this one.

AWHEELERTWS:
​Hi,
I agree with you entirely - plugins should not have references to each
other. I did this after a good amount of experimentation trying to extend
the allowedContent inside of the forms plugin. After I was unable to do
this I looked for examples with similar (adding tabs to existing plugins)
and ended up referencing the *dialogadvtab* plugin (the *table* plugin
implements *dialogadvtab* very similarly). Simplify specifying
allowedContent for this plugin didn't work because it needs to specifically
alter the allowedContent of the forms plugin. Could you please point me in
the right direction on this? Is doing something like
CKEDITOR.filters.allow('[required,pattern]', 'forms') supposed to work? Any
insight you can provide on the correct methodology would be very
appreciated.

Also thank's for reminding me about the license. If I am able to resolve
this issue then I'll make sure to add a friendly GPL/LGPL licence onto it
and submit it to the plugins repository for everyone else to enjoy as well.

Thank you,

*Alex Wheeler*
Web Developer @ Total Web Services
Phone: (503) 468-1100 x120
Email: awheeler@totalwebservices.net

AWHEELERTWS:
Hi Piotrek,
Thank you for pointing us in the right direction with the
allowedContent/filters. I misunderstood the API documentation and figured
out why the code I previously tried wasn't working properly. We've
published the plugin now and everything appears to be working properly on
it.

Thanks again,

*Alex Wheeler*
Web Developer @ Total Web Services
Phone: (503) 468-1100 x120
Email: awheeler@totalwebservices.net

PR #178, opened 2015-04-07
Use https cdn when loading from an https page

joncostello:
Use window.location.protocol to determine what protocol to use for the cdn

### DIFF
--- a/plugins/mathjax/plugin.js
+++ b/plugins/mathjax/plugin.js
@@ -11,7 +11,7 @@
 
 ( function() {
 
-	var cdn = 'http:\/\/cdn.mathjax.org\/mathjax\/2.2-latest\/MathJax.js?config=TeX-AMS_HTML';
+	var cdn = window.location.protocol + '\/\/cdn.mathjax.org\/mathjax\/2.2-latest\/MathJax.js?config=TeX-AMS_HTML';
 
 	CKEDITOR.plugins.add( 'mathjax', {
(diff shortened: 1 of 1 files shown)


wwalc:
Thanks for the pull request. The same change can be made in a simpler way: using leading double slash (`//`).

However, there is a little drawback of doing it this way: when someone launches CKEditor straight from his local file system omitting the web server (`file:///`), then the Mathjax plugin will not work if the library is loaded this way.

Although running CKEditor from `file:///` protocol looks odd, it's still important as some people launch CKEditor this way straight after downloading and the first impression is important if it does not cost an extra effort.

My proposal would be to load the library from https if window.location.protocol is https and use http otherwise.

Reinmar:
CKEditor does not support loading from `file:///`. It more or less works, but I remember that there were some issues with it. Therefore the fully sufficient and the cleanest option is to start from `//`.

As for the patch, the MathJax configuration is going to change in CKEditor 4.5.0 due to licensing problems. There will be no default setting for `config.mathJaxLib`, so one will need to set it manually.

Therefore any change in CKEditor 4.4.x does not make sense now. I made sure though, that 4.5.0's documentation uses the double slash - 34c90be15c.

wwalc:
Keep in mind that `file:///` is not only used by people loading CKEditor for evaluation soon after downloading it. Although we do not officially support it, people also load CKEditor e.g. inside desktop applications where CKEditor is loaded form a local file system (e.g. via Webbrowser control etc).

Thus, although I'm not saying that MathJax support is critical there, we shall not ignore completely this environment.

Reinmar:
You're right. I forgot about these cases. We should keep this in mind. But as for MathJax, since this is a config option, then one can set whatever they like.

PR #179, opened 2015-04-17
Fix: querySelector failed when enabling editor on element with numeric id

tkolo:
Assuming you have `<div id="1"></div>` and do `CKEDITOR.inline("1"`), this will fail, because #1 is not a valid query selector. More info here:
http://stackoverflow.com/questions/20306204/using-queryselector-with-ids-that-are-numbers

### DIFF
--- a/core/dom/document.js
+++ b/core/dom/document.js
@@ -288,7 +288,7 @@ CKEDITOR.tools.extend( CKEDITOR.dom.document.prototype, {
 	 * @returns {CKEDITOR.dom.nodeList}
 	 */
 	find: function( selector ) {
-		return new CKEDITOR.dom.nodeList( this.$.querySelectorAll( selector ) );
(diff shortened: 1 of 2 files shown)


Reinmar:
`CKEDITOR.inline()` does not use `querySelector()` (it only uses `getElementById()`). But the problem occurs when using `editable.find/findOne()` which e.g. widgets do, because they use the id of that element to standardise `querySelector()`'s behaviour.

Thanks for the fix. Two things though, before we can marge it:

* tests (a basic one is here: https://github.com/ckeditor/ckeditor-dev/blob/master/tests/core/dom/element/element.js#L893, and here is [testing environment description](http://docs.ckeditor.com/#!/guide/dev_tests),
* code style (you can use `grunt jscs && grunt jshint` if you install them by running `npm install` and if you installed `grunt-cli` globally).

mlewand:

PR #180, opened 2015-04-19
fixed the type error message

z281625810:
Refer to ticket #12847
https://dev.ckeditor.com/ticket/12847#comment:2


Reinmar:
I don't understand your change.

From perfectly reasonable:

```
var childTagName = linkChildren.getItem( 0 ).getName();
```

You left:

```
var childTagName = linkChildren.getItem( 0 );
```

I'm afraid that your change only hides the error, not fixes the issue.

PR #182, opened 2015-04-29
Update ro.js

ngodina:
be better


AnnaTomanek:
Hi Nicolae,

thanks a lot for this!

In general, we manage the localization process through the [CKEditor UI Translation Center at Transifex](https://www.transifex.com/projects/p/ckeditor/). We do have the [Romanian language group](https://www.transifex.com/projects/p/ckeditor/language/ro/) there, but the translation is still incomplete so it would be really cool if you wanted to join us and work on it.

In the meantime, I copied over your suggestions to the Romanian localization file and the updates will be available in the next CKEditor release. Thanks again for your help!

PR #183, opened 2015-05-06
ticket-13254: Fix block outdent failure when using divarea plugin

jcttrll:
Fix block outdent failure when using divarea plugin instead of iframe


Reinmar:
Related ticket: http://dev.ckeditor.com/ticket/13254

Reinmar:
Fixed with d63c062 after some minor improvements in the tests. Thanks @jcttrll! This was a perfect patch.

PR #185, opened 2015-05-06
Fixed protection pattern for attributes

asmforce:
Added support for self-closing tags (otherwise a protection pattern isn't being changed and un-protection will fail)

### DIFF
--- a/core/htmldataprocessor.js
+++ b/core/htmldataprocessor.js
@@ -907,7 +907,7 @@
 		// Different protection pattern is used for those that
 		// live in attributes to avoid from being HTML encoded.
 		// Why so serious? See #9205, #8216, #7805, #11754, #11846.
-		data = data.replace( /<\w+(?:\s+(?:(?:[^\s=>]+\s*=\s*(?:[^'"\s>]+|'[^']*'|"[^"]*"))|[^\s=>]+))+\s*>/g, function( match ) {
+		data = data.replace( /<\w+(?:\s+(?:(?:[^\s=\/>]+\s*=\s*(?:[^'"\s\/>]+|'[^']*'|"[^"]*"))|[^\s=\/>]+))+\s*\/?>/g, function( match ) {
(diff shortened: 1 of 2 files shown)


Reinmar:
Thanks for the patch. We will need two more things to be able to accept it:

* Description of the bug that this patch fixes or a ticket from http://dev.ckeditor.com.
* Tests for the patch. See the [Testing Environment](http://docs.ckeditor.com/#!/guide/dev_tests) guide and check [this file](https://github.com/ckeditor/ckeditor-dev/blob/master/tests/core/htmldataprocessor.js).

asmforce:
I use a source protection inside a tag's attribute:


The self-closing (standalone) tag contains a slash before right angle bracket. We could use a spaces before it (or not use):



The first one is ok but the second causes a wrong behavior. I get the following result after toDataFormat:


My PR fixes that.
Also I added an appropriate changes to tests/core/htmldataprocessor.js (already available in this PR).

Have a look please.

Reinmar:
I confirmed the issue and reported the ticket: http://dev.ckeditor.com/ticket/13292

Reinmar:
I pushed https://github.com/cksource/ckeditor-dev/tree/t/13292 with your patch plus some my tests improvements. Everything looks good, but I've been wondering whether we can make only the last of these 4 changes. AFAICS adding `\/?` at the end solves the issue and doesn't break any other test. Are you aware of any cases where that could fail?

asmforce:
If we left a \/? only at the end, regex would match an invalid HTML tags.
E.g.
`<tagname attr1='val1' / />`

To be honest I'm not sure if that behavior is is better or worse. ;)

Reinmar:
Good point. I think it's better if it matches more tags – even invalid ones. Especially that right after these regexps we pass the data through a native DOM element and browsers fix small glitches. I'm not sure about this one, but that doesn't matter.

So to merge this PR we only need to simplify the change in the regexp and run tests. We'll handle this.

Reinmar:
Merged to master with a2ff7bf. Thanks!

PR #186, opened 2015-05-11
fix documentation issue with creating a text node

benkiefer:
Fixes https://dev.ckeditor.com/ticket/13268

Thanks for the great product!


Reinmar:
Thanks! Merged.

PR #187, opened 2015-05-19
Sharedspace plugin now supports ckeditor elements and not only ids

Undergrounder:
The sharedspace plugin should support passing passing dom elements and not only ids.

In angular js it has no sense to use ids in directives. (Reusable components).

### DIFF
--- a/plugins/sharedspace/plugin.js
+++ b/plugins/sharedspace/plugin.js
@@ -35,9 +35,14 @@
 		}
 	} );
 
-	function create( editor, spaceName, targetId ) {
-		var target = CKEDITOR.document.getById( targetId ),
-			innerHtml, space;
+	function create( editor, spaceName, targetP ) {
+		var target, innerHtml, space;
+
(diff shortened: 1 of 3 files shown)


Reinmar:
Hi. Thanks for the patch. Great to see that it already contains a test :).

Four small things before we can merge it:

* I think that the setting should accept native DOM elements rather than CKEditor ones. The reason is that later on the config object is cloned and cloning non-native types works poorly.
* Extend documentation of this option (native element's type is `{HTMLElement}`).
* Please remove console logs, comment in L132 and code style in both JS files. You can read here about [contributing patches](http://docs.ckeditor.com/#!/guide/dev_contributing_code) where code style checker  and linter are also mentioned.
* You can remove L55-L57 in tests.

Undergrounder:
All changed and tested.

Reinmar:
Great! Your PR was merged to major with git:1d283a5. Thanks again for the work.

PR #188, opened 2015-05-26
Update bg.js

iivanov2:
I've just translated some missing strings in bulgarian language.

### DIFF
--- a/lang/bg.js
+++ b/lang/bg.js
@@ -20,7 +20,7 @@
 CKEDITOR.lang[ 'bg' ] = {
 	// ARIA description.
 	editor: 'Текстов редактор за форматиран текст',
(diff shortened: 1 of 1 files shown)


AnnaTomanek:
Thanks a lot for your contribution!

In general, we manage the localization process through the [CKEditor UI Translation Center at Transifex](https://www.transifex.com/projects/p/ckeditor/). We do have the [Bulgarian language group](https://www.transifex.com/projects/p/ckeditor/language/bg/) there, but the translation is still incomplete so it would be really cool if you wanted to join us and work on it.

In the meantime, I copied over your suggestions to the Bulgarian localization file and the updates will be available in the next CKEditor release. Thanks again for your help!

PR #190, opened 2015-06-01
Fixed: null pointer when try to access inexistent widgetElement

bjrke:
We have a problem with our plugin, which can be solved this way
I also think this check is necessary, because of DOM manipulations which could be done by other plugins or by the frontend user itself.
please apply this patch, thank you

### DIFF
--- a/plugins/widget/plugin.js
+++ b/plugins/widget/plugin.js
@@ -2498,16 +2498,18 @@
 					widget = widgetsRepo.instances[ attrs[ 'data-cke-widget-id' ] ];
 					if ( widget ) {
 						widgetElement = element.getFirst( isParserWidgetElement );
-						toBeDowncasted.push( {
-							wrapper: element,
-							element: widgetElement,
-							widget: widget,
-							editables: {}
-						} );
+						if ( widgetElement ) {
+							toBeDowncasted.push( {
+								wrapper: element,
+								element: widgetElement,
+								widget: widget,
+								editables: {}
+							} );
 
-						// If widget did not have data-cke-widget attribute before upcasting remove it.
-						if ( widgetElement.attributes[ 'data-cke-widget-keep-attr' ] != '1' )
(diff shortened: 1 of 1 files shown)


oleq:
Thank you! I noticed that your patch comes with no tests and we're unable to verify, not to mention merge it at the moment. Could you [provide some](http://docs.ckeditor.com/#!/guide/dev_contributing_code-section-tests)?

bjrke:
I am not sure, what I should test and how testing works in ckeditor.
I tried this code in google chrome debugger after I found the error.
The widgetElement was not found, but the widget was existent. I think extending the existing check is a good pattern.
I also have problems to setup bender.

Reinmar:
> The widgetElement was not found, but the widget was existent. I think extending the existing check is a good pattern.

The question is – why it happens? If we were able that this happens for a valid reason, then we would be able to make decision what to do with that missing widget element. Perhaps we should look for it differently? Or we should rebuild the whole algorithm? Or perhaps your fix is fully justifiable. But without a real case and tests we cannot make a decision.

Therefore,

1. Please report a ticket on http://dev.ckeditor.com where you explain how to reproduce this issue. The ticket must unfortunately contain a working sample or a test. Based on that we'll be able to tell whether we agree that the issue is valid and perhaps we'll be able to guide you what's the correct solution
2. Please read [Contributing Code](http://docs.ckeditor.com/#!/guide/dev_contributing_code) and [Testing Environment](http://docs.ckeditor.com/#!/guide/dev_tests) guides. If you have any troubles with setting up Bender.js please let us know e.g. here. But again – we need to know details to be able to help you.

bjrke:
i've written a bug report with our plugin which is needed to produce the bug:
http://dev.ckeditor.com/ticket/13347

Reinmar:
Closing, because as explained in http://dev.ckeditor.com/ticket/13347#comment:1 an issue is in a different place.

PR #192, opened 2015-06-12
fix variable name typo in dialogui plugin

FlorianLudwig:
as noticed by my colleague @BFriedrichs


Reinmar:
Uh... I wanted to say – how could tests not catch this?! But I realised that `dialogui` is probably the last plugin with so low coverage.

Anyway, thanks for the patch. I'd normally ask for tests for this patch, but taking into account its simplicity and lack of other tests, that would be senseless.

FlorianLudwig:
Actually I have no idea how to trigger this.  It was an "typoed grep" that found this in the first place.

Anyway, many words for a such a small patch :)

Reinmar:
I had to justify why I haven't included/requested tests ;). And well... such bugs are embarrassing :D.

PR #194, opened 2015-06-24
Merge MHE customization (ability to use ckeditor in iframe) with late…

davidsilva:
…st ckeditor code. This is necessary for functioning within the context of iframes, where trying to access the dom outside the iframe is not allowed.


Reinmar:
Hi,

Thanks for the patch. In order to accept it we will need two things:

1. Being able to reproduce a bug. For that we need a more detailed description of an issue.
2. Tests. Preferably automated and manual ones, but if implementing automated one would be a problem, only the manual one will be ok.

You can read more here - [Contributing Code and Providing Patches to CKEditor](http://docs.ckeditor.com/#!/guide/dev_contributing_code).

Reinmar:
Hm... It kinda looks as if this PR was made mistakenly, because more commits are being pushed to it. If I'm mistaken, please make a new one with proposed changes moved to a separate branch.

PR #196, opened 2015-07-02
fix help icon

Heba3a:
Fix help icon for Arabic users 
https://dev.ckeditor.com/ticket/13473

### DIFF
--- a/skins/moono/toolbar.css
+++ b/skins/moono/toolbar.css
@@ -135,6 +135,16 @@ a.cke_button
 	float: right;
 }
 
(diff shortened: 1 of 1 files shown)


Reinmar:
Hi. Thanks for the patch. I'm afraid that it isn't valid though. It would fix only the Arabic language while there are many other RTL languages. Also, we have separate icon set for RTL languages (see [here](https://github.com/ckeditor/ckeditor-dev/blob/master/plugins/list/plugin.js#L816)) and I believe the reason is that flipping an icon may not be enough – that does not take e.g. shadows into consideration.

PR #197, opened 2015-07-03
Added the name of the field in filetools

Katochimoto:
Example:
```js
var loader = editor.uploadRepository.create( file );
loader.loadAndUpload( 'http://foo/bar', 'my_field_name' );
```

### DIFF
--- a/plugins/filetools/plugin.js
+++ b/plugins/filetools/plugin.js
@@ -43,9 +43,10 @@
 
 			editor.on( 'fileUploadRequest', function( evt ) {
 				var fileLoader = evt.data.fileLoader,
+					fieldName = evt.data.fieldName || 'upload',
 					formData = new FormData();
 
-				formData.append( 'upload', fileLoader.file, fileLoader.fileName );
+				formData.append( fieldName, fileLoader.file, fileLoader.fileName );
 				fileLoader.xhr.send( formData );
 			}, null, null, 999 );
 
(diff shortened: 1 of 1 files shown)


pjasiun:
I am not sure about changes in `loadAndUpload` and `upload`. These methods know nothing about the way how data will be uploaded. For example user may want to send data as a stream, without a `form`. Them the `field` name will means nothing. The `fileUploadRequest` is the place there the request is build, so if user want to change the field name he should add the listener of this event and overwrite the way it works. For the same reasone I agree with your changes in `fileUploadRequest`. It would be good if the default listener check if `evt.data.fieldName` is set. Then it would be enough to set this parameter in the listener to change the `fieldName`:

```
	editor.on( 'fileUploadRequest', function( evt ) {
		evt.data.fieldName = `foo`;
	} );
```

If you want to have a different `fields` for different requests you could add the special parameter to the URL which will do nothing but is read in the `fileUploadRequest` event listener and set the filed depends on the parameter value:

 ```
	config.imageUploadUrl = 'example.com?field=image'

	editor.on( 'fileUploadRequest', function( evt ) {
		if( evt.data.url.indexOf( `field=image` ) > 0 ) {
			evt.data.fieldName = `image`;
		} else {
			evt.data.fieldName = `foo`;
		}
	} );
```

It may looks like a hack, but this is the simplest and most flexible API. Adding addition parameter causes problems in the long term.

If you agree with such architecture, limit you solution to changes in the `fileUploadRequest`, write [Bender.js tests](http://docs.ckeditor.com/#!/guide/dev_tests) to it and update `fileUploadRequest` method docs.

Katochimoto:
Thanks for the reply! This solution suits me.

Reinmar:
Related tickets: http://dev.ckeditor.com/ticket/13518 and http://dev.ckeditor.com/ticket/13519 and even http://dev.ckeditor.com/ticket/13520.

PR #199, opened 2015-07-09
Fixes #11376 and other IE11 issues

andreyskv:
Fixes #11376 issue and extra paragraph tags around li tags

### DIFF
--- a/plugins/pastefromword/filter/default.js
+++ b/plugins/pastefromword/filter/default.js
@@ -802,16 +802,18 @@
 						// that doesn't include "mso-list:Ignore" on list bullets,
 						// note it's not perfect as not all list style (e.g. "heading list") is shipped
 						// with this pattern. (#6662)
-						if ( ( /MsoListParagraph/i ).exec( element.attributes[ 'class' ] ) || element.getStyle( 'mso-list' ) ) {
-							var bulletText = element.firstChild( function( node ) {
-								return node.type == CKEDITOR.NODE_TEXT && !containsNothingButSpaces( node.parent );
-							} );
+						
+						//Commented out because it breaks in IE11 (fixes #11376 and extra paragraph tags around li tags ) (Tested with Chrome IE11/10)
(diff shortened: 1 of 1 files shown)


Reinmar:
Hi. Thanks for the patch. Unfortunately it could not be accepted due to lack of tests. But in the meantime we worked on the [#11376](http://dev.ckeditor.com/ticket/11376) as well and closed it with 9b8ff65.

Reinmar:
PS. Check out http://docs.ckeditor.com/#!/guide/dev_contributing_code.

andreyskv:
Up to you, but you could review it and understand the change before shutting it off or give a chance to write a test, because my change could be better/cleaner solution. There is no single new line in the patch, but deleted buggy old code so not even sure what to test as long as current tests work. If you follow the logic there is actually another code which tries to bypass what these lines did, but did not work. Now looks like your code undoes the harm of that hack too. Now I understand why the code base is so messy.

Reinmar:
I'm sorry to reject your patch with so few words, but working on similar tickets in the past I knew that the code you removed was necessary. And you can see it by running our tests.

Run the following ones:

http://tests.ckeditor.dev:1030/#tests/is:unit,tag:pastefromword,path:/tests/plugins/pastefromword

And there's 9 red ones in:

* http://tests.ckeditor.dev:1030/tests/tickets/9456/1
* http://tests.ckeditor.dev:1030/tests/tickets/9456/2

I should have mentioned that when closing your PR. Sorry for that.

I've seen you commented on part of 9b8ff65. If you see an option to simplify that code and keep all tests green on all browsers, then I'll be more than happy to accept your PR. And the more mess your patch removes, the happier I am. But first of all, tests must pass.

PS. In CKEditor 4.6 we'll work on a new MS Word filter, so the truth is that all effort made now will be slightly forgotten in few months. Although, the new filter will be implemented in a separate plugin, so the old one will live for developers who cannot upgrade (there will be differences between them).

andreyskv:
Ok. Thanks for more details. I did no realize there were tests in this project and that the patch failed some of them. It might be support for some older browsers.  Then l agree that we need to understand why tests actually failed and my fix is not safe to apply as is.

PR #200, opened 2015-07-10
[FIX] Allow for the iframe having been removed already

StefanRijnhart:
When integrating ckeditor in Odoo, we encounter TypeError: iframe is null in http://localhost:9069/web_ckeditor4/static/lib/ckeditor/plugins/wysiwygarea/plugin.js:520 when the editor is created in a ajax popup. The editor is destroyed when the popup is closed, but by then the Odoo framework has already cleared out the DOM.

Same as https://github.com/ckeditor/ckeditor-dev/pull/56, but I thought I'd try again.

### DIFF
--- a/plugins/wysiwygarea/plugin.js
+++ b/plugins/wysiwygarea/plugin.js
@@ -487,16 +487,19 @@
 				// Memory leak proof.
 				this.clearCustomData();
 				doc.getDocumentElement().clearCustomData();
-				iframe.clearCustomData();
+				if ( iframe )
+					iframe.clearCustomData();
 				CKEDITOR.tools.removeFunction( this._.frameLoadedHandler );
 
-				var onResize = iframe.removeCustomData( 'onResize' );
-				onResize && onResize.removeListener();
+				if ( iframe ) {
+					var onResize = iframe.removeCustomData( 'onResize' );
+					onResize && onResize.removeListener();
 
-				// IE BUG: When destroying editor DOM with the selection remains inside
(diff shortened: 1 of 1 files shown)


Reinmar:
I'm still a bit against just hiding an evidence of an issue rather than fixing the code (e.g. Odoo which should destroy CKEditor prior to destroying the DOM), but ok... let's see.

I would normally ask you for tests and some other improvements, but I want to have this researched on our side. My main doubt is whether new instances of the editor can be initialised after one of them is so ugly destroyed. Also, I want console logs notifying the developer that they destroyed CKEditor incorrectly, so this issue stays somehow visible.

We'll try to ship this in 4.5.3.

StefanRijnhart:
Thanks @Reinmar for your response. On our side, we see no problems initializing new instances of the editor on the same page after destroying old ones with this patch.

szymonkups:
Thanks for the PR. I've created separate [ticket](https://dev.ckeditor.com/ticket/13790) on Trac.
I've pushed new branch [t/13790](https://github.com/cksource/ckeditor-dev/tree/t/13790), it includes changes made in pull request plus additional test. It will be merged into master branch after the release of CKEditor 4.5.4 (which is currently in progress). I will close this PR after merge to the master branch.

szymonkups:
Merget into master branch with ca91770256d1027901011def7460a77eb55f2697.

hongquanvn1998:
How to call ckeditor 4.x widget for field text in odoo ? thanks

PR #201, opened 2015-07-19
Set toolbarconfigurator button "type" attribute

colemanw:
Love the new configurator! I've had good success integrating it with our [php-based CRM](https://civicrm.org/blogs/colemanw/big-changes-wysiwyg-editing-47) to allow users to configure their editing experience in-app. One little problem was that the missing "type" attribute on the buttons would cause premature form submission. Thanks for considering this change.

### DIFF
--- a/samples/toolbarconfigurator/js/fulltoolbareditor.js
+++ b/samples/toolbarconfigurator/js/fulltoolbareditor.js
@@ -324,6 +324,8 @@ window.ToolbarConfigurator = {};
 
 		$button.addClass( 'button-a' );
 
+		$button.setAttribute( 'type', 'button' );
+
 		if ( typeof cssClasses == 'string' ) {
 			cssClasses = cssClasses.split( ' ' );
 


Reinmar:
Thanks! :) We're glad people like it and that it's also helpful for other projects.

Reinmar:
Changelog entry in a4c9294525faae.

colemanw:
Thanks for the merge @Reinmar. For my future reference is submitting a PR like this the right way to send in fixes or should I open an issue instead?

Reinmar:
This one was ok to be closed this way, because it was not related to the editor itself, but samples and it was a simple thing. But in >99% cases we would need a ticket on http://dev.ckeditor.com, although, if a PR is clear for us (it includes a proper description), we can open that ticket ourselves. Not a big deal.

You can find more details about contributing code in docs.ckeditor.com/#!/guide/dev_contributing_code

PR #203, opened 2015-07-27
Stylesheet parser should accept CSS classes without elements

cr-solutions:
Currently the "Stylesheet Parser" only accept CSS Classes with element like "p.mystyle", but not classes without elements like .table_darkgrey.

CSS Sample:
.table_darkgrey {
background-color: #a8aeb3;
border-bottom: 30px solid #a8aeb3;
border-top: 30px solid #a8aeb3;
}

For more see also the ticket https://dev.ckeditor.com/ticket/9922
It contain also my last plugin.js for the "Stylesheet Parser"

The changes have inline comments with ticket refernce.

### DIFF
--- a/plugins/stylesheetparser/plugin.js
+++ b/plugins/stylesheetparser/plugin.js
@@ -1,4 +1,4 @@
-﻿/**
+/**
  * @license Copyright (c) 2003-2015, CKSource - Frederico Knabben. All rights reserved.
  * For licensing, see LICENSE.md or http://ckeditor.com/license
  */
@@ -32,6 +32,7 @@
 			var selector = aSelectors[ i ];
 
 			if ( validSelectors.test( selector ) && !skipSelectors.test( selector ) ) {
+                                    
 				// If we still don't know about this one, add it
 				if ( CKEDITOR.tools.indexOf( aClasses, selector ) == -1 )
 					aClasses.push( selector );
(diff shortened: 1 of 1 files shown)


Reinmar:
Now the PR includes what it should, thanks. 

However, there are a couple of issues:

1. Incorrect code style and lack of tests. Read more in http://docs.ckeditor.com/#!/guide/dev_contributing_code
2. The patch touches two parts of code. First is [this loop](https://github.com/cr-solutions/ckeditor-dev/commit/76e93ceac6d1fa504e99cb1ebf5a10212d47f003#diff-7fc4016a6833d41d1c3f536a9087f306R74) which perhaps is the place that must be fixed. Second is the [timeout](https://github.com/cr-solutions/ckeditor-dev/commit/76e93ceac6d1fa504e99cb1ebf5a10212d47f003#diff-7fc4016a6833d41d1c3f536a9087f306R108) added later which I don't understand at all. It looks like some legacy code that isn't necessary any more. Please always limit your patches to the minimum.

The last thing is... I don't think this is the right approach. Why using `<span>` by default if element name is not defined? As [Alfonso commented](https://dev.ckeditor.com/ticket/9922#comment:2) the problem lays in the styles system which does not handle generic styles. This means styles which do not have element names defined, so which can be applied to all elements. This is a much bigger feature to code, but this is the only reasonable change that can be do. Defaulting to `<span>`s is just a hack which solves a limited number of cases.

Therefore, I'll to close this PR as a proper patch would not have anything in common with this. I'm also afraid that adding the mentioned feature is very complicated because it would be a pretty deep change in the styles system.

PR #204, opened 2015-07-29
plugins image2: Set default Captioned image = Alternative Text

vuthao:


Reinmar:
Thanks for the PR. Please see our guide about [contributing code](http://docs.ckeditor.com/#!/guide/dev_contributing_code) (especially the part about tests).

However, before you code something, I think we need a discussion whether this is a good change. I remember that it was proposed before but it was rejected because alternative text is not the same thing as a caption so the editor should not propose that. Will post RFC to other guys.

wwalc:
Humm interesting concept, but -1 from me, unfortunately. Alt and caption are two completely different things, which means that quite often alt will be different. Caption can be totally different that what is shown on the image and the best example of it are memes.

mlewand:
Yes, unfortunately figcaption is a different thing. Generally speaking alt is here you give concise description of the image, whereas figcaption is used for giving supplementary text (explaination of non-obvious things, legends etc.). Because of that we shouldn't reuse alt in figcaption.

Reinmar:
Playing devil's advocate role – I know that figcaption and alt are semantically different things and therefore a user has always a chance to type any caption they want. Question is whether we can't "preset" caption to some value and that value could be alt because (perhaps) in some cases it could be the same as alt.

pjasiun:
I also think that there may be cases where caption and alt can be the same and there are no cases where caption should be the `Caption` string, so in my opinion `alt` as a default caption makes more sense then harcoded string. User can change it anyway, does not matter what was default and if, in some cases, we put the caption he wanted to have, it will be nice.

I know that we could educate users by not putting alt as an caption, but I believe that these who know, already know and these who do not know will simple see it as a bad UX.

anhyeuviolet:
I am now using CKeditor with a open source CMS called NukeViet.
About the conversation, I think it will be better if there is a function lets user change the caption of the inserted image.
Because in my experience, when user checked at "This image have a caption", there is   only a string shown as "Label"; but our editor cannot modify this label ( with Mozilla Firefox ).
So I hope you can change something, let user reuse ALT as a CAPTION ( as this PR do ) or fix this error ( I think this is an error ) to make sure every editor with CKeditor can have a best experience.
Thanks !

Reinmar:
The fact that with Firefox the label cannot be edited is a recent Firefox regression (see https://bugzilla.mozilla.org/show_bug.cgi?id=1181130) and Mozilla is working on a solution for it. So if that was a reason for making this change, then it won't be necessary.

wimleers:
I asked some accessibility experts and Drupal people who were involved in this for feedback: https://twitter.com/wimleers/status/628246101674602496.

mgifford:
I agree with @wwalc & @mlewand on this. It doesn't seem like a good default. 

Maybe if the difference was clearly laid out to explain to a user how they are different. 

It might be useful (for ATAG) to start providing information like this to users in the interface:
http://html5doctor.com/the-figure-figcaption-elements/ 

I think that if we show the user by default that they are the same it may introduce confusion...

jessebeach:
Oh, I do love easy defaults that make accessibility ux more automatic. This is an interesting idea @vuthao. But I'm persuaded by the arguments in the thread that captions and alt text are not equivalent. I see two issues:

1. Imagine that this were not an image in the figure, but a blockquote. We wouldn't simply repeat the text of the blockquote in the caption. Presumably, the caption would provide contextualizing commentary on the quote. If we think of the image as content similar to the blockquote, we better see the difference between content and its presentation frame, the figure.
1. I worry that if alt text is printed to the screen automatically, content producers may skip the alt text entirely because they don't want it to be presented visually, and the PR as it stands, does not provide an opt-out.

I could see a checkbox: "Expose image alt text as caption". That might be an acceptable authoring feature that would allow for reuse, which is I think your intent with this proposal.

sorressean:
I'm in full agreement with @@Reinmar here; I like the checkbox and don't really like setting one to default. This will lead to odd workarounds and I feel like caption and alt have different use-cases.`

mlewand:
Semantics aside: keep in mind DRY factor.

Once somebody will enter some goofy value, and wants to edit it, then he's very likely to edit only the caption in CKEditor, without entering image (widget) details.

TC:

* user makes a typo in alt text, e.g. "Old, ruined brifge",
* accepts the dialog (the alt text gets propagated as a caption),
* while reviewing user finds out the typo and changes... the caption only. Alt text remains with the typo.

-----

* user enters some inaccurate text as an alt text, e.g. "Old bridge",
* accepts the dialog (the alt text gets propagated as a caption),
* after he sees the image within his content, he decides that it should have been "Old, ruined bridge"
	* most of users will edit caption and won't bother with alt text (which will be desynchronized from then)

I believe alt text should be provided consciously, not by coincidence.

I see a different need here: if widget gets inserted, then the first editable (if any) should be focused. But it's beyond scope of this issue, it's [#11700](https://dev.ckeditor.com/ticket/11700).

fredck:
I totally agree with @mlewand... this automation will cause confusion. Users will learn that the field in the dialog is there to set the caption and miss the scope of it totally.

IMHO, setting alt and caption should be left as two separate actions that are consciously performed.

PR #206, opened 2015-08-06
Pale Moon 26+ support

dertuxmalwieder:
--> http://palemoon.org/WIP/

### DIFF
--- a/core/env.js
+++ b/core/env.js
@@ -158,7 +158,7 @@ if ( !CKEDITOR.env ) {
 		 *
 		 * @property {Boolean}
 		 */
-		env.gecko = ( navigator.product == 'Gecko' && !env.webkit && !env.ie );
+	        env.gecko = (( navigator.product == 'Gecko' || navigator.product == 'Goanna') && !env.webkit && !env.ie );
 
 		/**
 		 * Indicates that CKEditor is running in a Blink-based browser like Chrome.


Reinmar:
TBH, I've heard about Pale Moon once and I don't know anything specific about this project. I've been reading some [info](http://www.palemoon.org/info.shtml) about the project, but it's oriented around UI, speed, etc. but mentions nothing about how far this engine diverged from Gecko and how it merges changes from Gecko. That would shed some light on "is it still a Gecko or not?" question.

Another thing is - why does the `navigator.product` has been changed to some new string? This is especially worrying since 'Goanna' is not a project name, but milestone name, so it will (I guess) change in the future again. Or won't it?

Finally, this PR breaks the editor instead of fixing it. The editor works fine even with the `env.gecko` property set to `false`. However, it does not work at all with your patch because `env.version` is set incorrectly.

dertuxmalwieder:
Sorry - written in a hurry without much chance to test it. Currently, the name is set to stay but it may or may not diverge from Gecko some more.

fredck:
Q1: Are browsers based on Gecko generally changing `navigator.product` to something which is not `Gecko`?

Q2: If Q1 is confirmed, wouldn't it be better to instead change the way Gecko is sniffed to make it more generic, instead of adding code specific to yet another browser? Otherwise it would allow for all Gecko based browsers to be detectable in the CKEditor code, which is senseless, considering that there is no official support for them in the plans.

dertuxmalwieder:
Is "product sniffing" actually needed? Wouldn't "feature sniffing" be sufficient here?

Reinmar:
We are rarely sniffing for features. Usually we need to sniff for bugs. And sniffing for stuff like:

* will the browser crash if we do something,
* will the browser move selection to element X rather than outside it,
* did user copy only plain text or is it a bug in Firefox

Is either impossible or time consuming so it would create terrible user experience. Therefore we need to base on browser sniffing.

Reinmar:
(I edited my previous comment because it was unclear).

wolfbeast:
Current trunk of Pale Moon has navigator.product changed to "Gecko". This is done because all current browsers (including IE 11, Safari, etc.) now state "Gecko" here, and using anything else apparently breaks a number of sites (blame the browser wars).

Please understand: the name "Goanna" is **not** a milestone name. It is an actual branding of our forked off version of Gecko and will be there to stay. It will also be present in the UA string from that point forward as Goanna/{buildID}. To not break the web for other sites checking for Gecko in the UA, we've also had to add a Gecko compat slice there.

More info:
Some compat info for developers https://forum.palemoon.org/viewtopic.php?f=24&t=9077
General introduction https://forum.palemoon.org/viewtopic.php?f=1&t=8607

If your code from this point forward would officially add support for Pale Moon, that would be great.

Of note: With the previous navigator.product name, your demos on the site would work flawlessly; with navigator.product changed to Gecko, it now complains about not being officially supported...

Reinmar:
Thanks for the clarifications and for reconsidering getting back to `navigator.product == 'Gecko'`. It's not like we don't want to see new products on the market, but we are in a situation where we **must** know the engine. Therefore all major browsers follow this totally broken naming scheme because the Web wouldn't work otherwise.

From what I see if `navigator.product` is `Gecko` and user agent is something like:

```
Mozilla/5.0 (Windows NT 6.1; WOW64; rv:1.0) Goanna/20150801 PaleMoon/26.0.0b1
```

Then the thing that fails is the version recognition. CKEditor reads the `rv:1.0` fragment and thinks that this is Gecko 1.0, which obviously isn't supported.

In my opinion setting `rv` to `1.0` is a mistake because it does not follow the existing scheme. A scheme to which even the biggest companies had to adapt to. It's easier to change one browser than hundreds (if not more) projects.

I'm sorry, but I don't want to complicate our code further in this case.

wolfbeast:
The "existing scheme" unfortunately makes looking at rv: completely pointless. Mozilla has started (with FF4 and later) to make this directly equal to the product version -- the mistake lies there.

The problem there is one of status quo: With Goanna, we want to return to a sane, independent engine version not tied to versions of products that might be based on it. Firefox used to do this as well with gecko 1.8 and 1.9 versions for Firefox 3.*.
Choosing 1.0 is a mistake, indeed, considering the legacy checks for these old Gecko versions still in place in various libraries, so it will be 2.0 and up instead when released to avoid that particular confusion. Gecko 2.0 was never released, they jumped to 4 (=Firefox 4) and as such even if you look at rv: there is definitely plenty of leeway to make a simple check (even apart from checking for Goanna/* which is unique to our engine).

The bottom line is that Gecko 26 *is not* equal to Goanna 26, and if a library assumes such, it will break various things. We must make it clear that they are separate engines. As much as you don't want to complicate your code, we don't want to end up on the heap of confused Firefox rebuilds that aren't actual forks.

If you have any other suggestion as to what would be a sane scheme to use without breaking the intended independence, then I'm all ears.

mlewand:
Definitely switching `navigator.product` to `Gecko` was a good idea. As for `rv` I'd suggest playing along with FF - that will just bring you more interoperability. To mark engine version you can just append it to the UA and anybody interested can extract it. E.g.

`Mozilla/5.0 (Windows NT 6.1; WOW64; rv:31.9) Gecko/20151012 Firefox/31.9 PaleMoon/25.7.3 <engineCode>/25.7`

Unfortunately we can't provide official support for Pale Moon. Each browser is a major effort for us, especially during release testing phase.

As for CKEditor, the good news is that I don't see too much version dependency with FF, so now that PM has friendlier `navigator.product` it should work well.

Reinmar:
> If you have any other suggestion as to what would be a sane scheme to use without breaking the intended independence, then I'm all ears.

That's the thing – I don't believe in the independence. I want browser vendors to make sane decisions, you want me to make them too. But while you may ask us to change the code, you can't ask all other projects. So in order to make good products we all have to bend to each other (yeah...).

Anyway, Pale Moon is a fork of Gecko and so far, let's face it, it's not super popular. Therefore, I would propose imitating Firefox in as many details as possible (what will ensure that projects which try to detect Firefox will succeed) and adding something special (additional fragment to the UA string), so those who want to support Pale Moon in special way will be able to detect it as well. I think that such solution is a win-win.

wolfbeast:
> That's the thing – I don't believe in the independence.

Then I suggest you read up on Pale Moon. Why would I go through branding hell and these kinds of issues unless it was needed?

Marking the browser as "Gecko/26" or "rebranded gecko 26" is a mistake since we are not fully compatible with it, and gives the absolute wrong impression, and only perpetuates the problems there are. rv: is supposed to represent the underlying engine version; projects using that mistakingly when doing UA sniffing (which is a bad idea to begin with when you can do feature detection) to find out the *application* version are at fault, and can't be a compelling reason to perpetuate the status quo.

We already have clear indicators in the UA (and have had these for years) about the different application it is (the PaleMoon/nn.nn.nn slice).

But I'll give it some thought and see if I can find a different solution, at least regarding the UA.

KrzysiekJ:
I’ve experienced this bug myself and found this thread to be interesting. The Mozilla Developer Network says the following about [the `navigator.product` property](https://developer.mozilla.org/en-US/docs/Web/API/NavigatorID/product):

> This feature has been removed from the Web standards. Though some browsers may still support it, it is in the process of being dropped. Do not use it in old or new projects. Pages or Web apps using it may break at any time.

> […]

> Note: Do not rely on this property to return a real product name. All browsers return "Gecko" as the value of this property.

So `navigator.product` is now officially meaningless and cannot be relied upon regarding whatever.

I’ve changed my user agent string to say the following:

```
AFlyingElephant/5.0 (X11; Linux x86_64; rv:2.1) Goanna/20160629 PaleMoon/26.3.3
```

It contains neither “Mozilla” nor “Gecko” nor “Firefox”, so there is no basis upon which it can be assumed that it the browser is somehow Gecko-related. Despite of that, CKEditor does not start with that UA string. It does however start if I change it to the following: 

```
AFlyingElephant/5.0 (X11; Linux x86_64; rv:46.0) Goanna/20160629 PaleMoon/26.3.3
```

It looks like Gecko engine is assumed despite having no legitimate basis for it. This constitutes a bug.

KrzysiekJ:
I’ve opened [a pull request which removes relying on `navigator.product`](https://github.com/ckeditor/ckeditor-dev/pull/286) It won’t solve this particular issue though, as Pale Moon includes “Gecko” in its user agent string. If it however decides to abandon this and include only “Goanna”, then CKEditor should start to work.

wolfbeast:
Unfortunately, abandoning the Gecko slice has proven to break many sites (ranging from presenting mobile versions to not presenting any web content at all). Although we have a preference that can remove this slice, it's not a good idea to leave it out by default. What needs to be done when you insist on sniffing user agents is extend it to include Pale Moon's specific identification and respond accordingly -- or make sure to use proper feature detection that doesn't use UAs (which is always the better option)

KrzysiekJ:
In principle, the fault lies on the side of those sites that dumbly whitelist only “Gecko” browsers and it is those sites that should adjust. On the other side, if a browser puts “Gecko” in its UA string, then it shouldn’t be surprised it is being treated like Gecko.

That being said, Chrome and Edge also put “Gecko” in their UA strings, but there are special cases in the CKEditor’s code which explicitly exclude them as Gecko. If so, why not exclude also Goanna? Not blacklisting something doesn’t imply that it has to be officially supported. A comment in the code about `env.isCompatible` says:

> Since CKEditor 4.5 this property is a blacklist of browsers incompatible with CKEditor. It means that it is set to `false` only in browsers that are known to be incompatible.

As Pale Moon is not known to be incompatible and it doesn’t identify itself in any non-standard way (when compared to IE and Chrome), then it should not be blacklisted.

As a proposal, I’ve changed my pull request accordingly.

f1ames:
Hi,

PR #207, opened 2015-08-11
http://dev.ckeditor.com/ticket/9856

mark-wade:


Reinmar:
Hi. Thanks for the PR, but I'm afraid this isn't the right solution. It will also unblock context menu on editor's UI.

A good solution was described by the author of this ticket: http://dev.ckeditor.com/ticket/9856

`disableContextMenu()` could not only check if the target allows context menus:

```
if ( !event.data.getTarget().hasClass( 'cke_enable_context_menu' ) )
    ...
```

but also if its ancestor allow it:

```
function allowsContextMenu( element ) {
    return element.hasClass( 'cke_enable_context_menu' );
}
if ( !event.data.getTarget().getAscendant( allowsContextMenu, true )
    ...
```

Then, that class needs to be added to the divarea editable and it's done.

If you would like to propose another PR with such solution, please read http://docs.ckeditor.com/#!/guide/dev_contributing_code. This case will require at least a manual tests for the divarea plugin and an automated test for `disableContextMenu()` (it can be tested by mocking the `evt` object passed to `fire( 'contextmenu', evt ) and checking if `preventDefault()` was called.

PR #210, opened 2015-08-23
added support for Arabic replace Success Msg , ticket #13686

ahmedasaleh:
I added support for  ticket #13686.
Basically I modified the message for the replace success in file plugins/find/lang/ar.js .

### DIFF
--- a/plugins/find/lang/ar.js
+++ b/plugins/find/lang/ar.js
@@ -12,7 +12,7 @@ CKEDITOR.plugins.setLang( 'find', 'ar', {
 	notFoundMsg: 'لم يتم العثور على النص المحدد.',
 	replace: 'إستبدال',
 	replaceAll: 'إستبدال الكل',
-	replaceSuccessMsg: 'تم استبدال 1% من الحالات ',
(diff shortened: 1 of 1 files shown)


ahmedasaleh:
link to ticket: https://dev.ckeditor.com/ticket/13686#comment:1

AnnaTomanek:
Hi Ahmed,

Thank you for taking your time to investigate this!

However, "%1" is actually a placeholder to display the number of occurances replaced. It should be replaced by CKEditor with an actual number, just like it works for other language versions, and not be replaced with 100%. For some reason it does not get replaced in Arabic and I created ticket [#13688](https://dev.ckeditor.com/ticket/13688) for this issue.

Additionally, all CKEditor localizations are handled through the [CKEditor UI Translation Center](https://www.transifex.com/projects/p/ckeditor/) at Transifex, not directly in source files. There is an [Arabic localization team](https://www.transifex.com/ckeditor/ckeditor/language/ar/) available there, you are most welcome to join and provide some help as Arabic is still quite incomplete are the moment. This will be much appreciated!

PR #211, opened 2015-08-28
Fixing #13703 by checking if current selection is actually adjacent to widget

dustyjewett:
…before preventing keydown.

With a passing test! I don't know where this test should actually go, as it isn't ACTUALLY testing fake selections, rather it's testing that a fake selection doesn't happen, and that a deletion is happening instead.

Note: Your tests seem to be order-specific.  I initially tried to insert my test prior to the preceding test, and that test failed.  As soon as I moved the test down, they passed.

### DIFF
--- a/core/selection.js
+++ b/core/selection.js
@@ -356,10 +356,41 @@
 
 			var next = range[ keystroke < 38 ? 'getPreviousEditableNode' : 'getNextEditableNode' ]();
 
+			// Special case for widgets.  We want to select it instead of delete it.
 			if ( next && next.type == CKEDITOR.NODE_ELEMENT && next.getAttribute( 'contenteditable' ) == 'false' ) {
-				editor.getSelection().fake( next );
-				evt.data.preventDefault();
-				evt.cancel();
+				var currentBlock = editor.elementPath( range.startContainer ).block,
+					nextBlock = editor.elementPath( next ).block;
+
+				// If the widget is not in a block (hence, the root editable() is it's block), OR
+				// 	  the widget is in the same block as the cursor, select the widget.
+				// We special-case this to allow the browser to handle the merging of blocks when deleting the space between them.
+				if ( !nextBlock || nextBlock.equals( currentBlock ) ) {
+					editor.getSelection().fake( next );
+					evt.data.preventDefault();
+					evt.cancel();
+				} else if ( CKEDITOR.env.gecko && ( keystroke === 8 || keystroke === 48 ) ) {
+					// Firefox has a bug from 2011 that prevents backspace/delete from working as expected around contenteditable=false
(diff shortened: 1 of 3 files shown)


Reinmar:
Thanks for the PR.

I'm afraid that you're entering a minefield, but let's see :).

1. The code style is wrong (see http://docs.ckeditor.com/#!/guide/dev_contributing_code)
2. That test passes on Blink and WebKit only. The reason is that only in these engines CKEditor uses a custom backspace key handler. On other browsers there's no custom handler, so mocking the backspace key gives you nothing. That's why we check in some tests if the event was cancelled. This will unfortunately vary between browsers.
3. Since this vary between browsers, we must be sure that if we let the browser to handle backspace it will do it right. For that we need a manual test.
4. You touched only backspace, what about forward delete?
5. As for the tests order – yes, there seems to be some relation between these tests. I was trying to figure out what is it, but that's nothing obvious unfortunately. I noticed that the selection is set differently in the existing test if you add your test before. These tests are mostly integration tests (because how can you test selection differently?) and stuff like this happens from time to time. Debugging this may be too hard for you so you can place your tests somewhere earlier.
6. As for the fix itself, I'm afraid it's not enough. What will happen with `<b><widget></b><em>^foo</em>` and in similar cases where the range is anchored in a different element, but a widget is the previous editing point? If you add such check, we need to make sure it's complete. Generally speaking, the algorithm should check if [elementPath](http://docs.ckeditor.com/#!/api/CKEDITOR.dom.elementPath) from the current position and from the position that we found, share the same [block](http://docs.ckeditor.com/#!/api/CKEDITOR.dom.elementPath-property-block).

Reinmar:
One more thing – I'm curious how this affects the block widget case.

dustyjewett:
@Reinmar: I've updated the PR:

1: Code style is now correct (installed the pre-commit hook!)
2/3/4/5: Removed the test that I added. Added a manual test instead. While simple, it handles all of the use-cases that I could think of in as few steps as possible.
6: I now check the block, instead of the parent.  

I uncovered a fantastic bug in Firefox that prevents merging of two blocks when a contenteditable=false is the first|last item in one of the blocks, and manually implemented the merger between the blocks for Firefox only.  It would probably work for all browsers, and ensure that functionality is the same across all of them in this specific case, but I didn't really think it was a good idea to open up those floodgates.

Also, I'm really starting to dislike Firefox. :confounded: 

As noted in the commit notes, there is an outstanding CKEDITOR bug that I'm not sure is filed, and is not fixed by this PR.  If the widget is wrapped in an inline tag, it is not identified and is deleted. This code change doesn't affect that case in any way, making it better or worse, but it did affect the scenarios that I tested.

dustyjewett:
Any chance this can get another pass?

PR #212, opened 2015-08-28
bbcode / don't remove adv dialog tabs for all instances

SteveTheTechie:
This makes the dialogDefinition listener installed by the bbcode plugin look at whether the bbcode plugin is actually in the current editor instance that the listener is removing the advanced dialog tabs for.  I have a mix of bbcode editor instances (for text snippets) and non-bbcode editor instances.  While I am moving to get rid of my legacy bbcode based text snippets, being able to use both types of editor instances for now without limiting the "normal" editor instances is helpful.

### DIFF
--- a/plugins/bbcode/plugin.js
+++ b/plugins/bbcode/plugin.js
@@ -5,6 +5,9 @@
 
 ( function() {
 	CKEDITOR.on( 'dialogDefinition', function( ev ) {
+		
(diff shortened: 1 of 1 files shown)


Reinmar:
Thanks for the PR.

Please see how to [contribute code](http://docs.ckeditor.com/#!/guide/dev_contributing_code). We will need a test(s) and corrected code style.

Reinmar:
The solution that you proposed seems to make sense. You can simplify it though by checking for `!ev.editor.plugins.bbcode`.

PR #213, opened 2015-09-08
Fix panel flipping

hgl:
To fix cases like these
screen shot 2015-09-08 at 8 28 24 pm
screen shot 2015-09-08 at 8 30 04 pm

### DIFF
--- a/plugins/floatpanel/plugin.js
+++ b/plugins/floatpanel/plugin.js
@@ -325,7 +325,7 @@ CKEDITOR.plugins.add( 'floatpanel', {
 							if ( spaceAfter < rectWidth ) {
 								// Flip to show on right.
 								if ( spaceBefore > rectWidth )
-									left += rectWidth;
+									left = left + rectWidth - offsetParent.$.offsetWidth;
 								// Align to window left.
 								else if ( viewportSize.width > rectWidth )
 									left = left - rect.left;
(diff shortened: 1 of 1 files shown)


Reinmar:
Thanks for the PR!

Please see http://docs.ckeditor.com/#!/guide/dev_contributing_code. We'll need tests and description how to reproduce this issue.

hgl:
I wanted to add tests, but your tests for floatpanel doesn't contain any flipping tests. So I thought that feature were manually tested?

Reinmar:
I guess a manual test will be enough (the floatpanel API may not be easily testable). To see how our manual tests look like install and start our testing env and open http://tests.ckeditor.dev:1030/#tests/is:manual (e.g. https://github.com/ckeditor/ckeditor-dev/tree/master/tests/plugins/tableresize/manual)

hgl:
I actually manually tested with `samples` in the release folder. Is that enough? Should I upload screenshots to prove it?

hgl:
Also just a quick question, why is the panel iframe based? Any plan to implement with divs directly?

Reinmar:
> Also just a quick question, why is the panel iframe based? Any plan to implement with divs directly?

Because this allows to style the options using a separate stylesheet, which is shared with the content (what's used by the format and styles dropdowns). It might also be required by screen readers to work correctly – there's a lot of logic around a11y and focus there. I don't think this is a requirement today, but that might have been a reason in the past.

There are no plans currently to rewrite them as that would make styling with the same stylesheet impossible, so such change would be backward incompatible. We would need to implement a new plugin that would implement the same API, but would use a simpler structure. It's doable, but it means quite a lot of work for which we don't have resources right now.

Reinmar:
> I actually manually tested with samples in the release folder. Is that enough? Should I upload screenshots to prove it?

We need a manual test which describes exactly this case (or cases) because:

* we need to verify it too,
* we need to verify it on all supported browsers during pre-release testing phase that we always perform,
* we need it as a documentation of the change to be able to reproduce everything in the future if something else will be changed in that code.

f1ames:
Hi,
Thank you for supporting Open Source and contributing to the CKEditor 4 project!

PR #214, opened 2015-09-09
Fix for radio button onChange event in IE9+ (http://dev.ckeditor.com/…

iliyakostadinov:
This is the proposed fix for the radio button onChange even, which currently doesn't work in IE9+
The original ticket can be found here: http://dev.ckeditor.com/ticket/12733

### DIFF
--- a/plugins/dialogui/plugin.js
+++ b/plugins/dialogui/plugin.js
@@ -1290,7 +1290,7 @@ CKEDITOR.plugins.add( 'dialogui', {
 			 */
 			eventProcessors: {
 				onChange: function( dialog, func ) {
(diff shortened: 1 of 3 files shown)


Reinmar:
Thanks for the PR!

Please see http://docs.ckeditor.com/#!/guide/dev_contributing_code. We'll need a test (manual one will be enough I think, as this is browser specific anyway). See this: https://github.com/ckeditor/ckeditor-dev/pull/213#issuecomment-138571467

iliyakostadinov:
I have added a manual test.

szymonkups:
Thank You for the contribution!
I've added some minor code style fixes and merged changes into master with 47e6f7f.

PR #215, opened 2015-09-10
Make non-break space and non-break hyphen available in special chars

gbastien:
Hi,

this is about https://dev.ckeditor.com/ticket/13733

This is my first commit on ckeditor, I tried to follow conventions, do not know about :
- need to update other languages?

Could you please review/comment/merge?

Thank you!

Gauthier

PR #216, opened 2015-09-11
(http://dev.ckeditor.com/ticket/13733): Make non-break space and non-break hyphen available in specialchar

gbastien:
Hi again,

this is a new PR with minimal changes and commits.

Could you please comment/review/...merge ;-)

Gauthier


mlewand:
Note that this is a follow up to #215.

mlewand:
Hi @gbastien, thanks for your willingness in contributing to CKEditor4! First thing, sorry it took us a while to review your PR. We were quite busy and now we have some time to clean up the queue.

Now, I see the reason for the feature, as there are some cases where people would like to use non breaking space and non-break hyphen, however we didn't even got a single feature request related to that on GH, and it's been already a year since we migrated.

The nice thing is that adding new entities is very easy, thanks to [`config.specialChars` property](https://docs.ckeditor.com/ckeditor4/latest/api/CKEDITOR_config.html#cfg-specialChars). Values in this dialog are a subject of opinion and particular use case. We found that current values are good for start, but we know that it won't fit all use cases, thus the config option.

I posted it as a working code here: https://codepen.io/mlewand/pen/wxppYO You could also extend the existing special chars code array, it requires just some more JavaScript. 🙂 

Also you can note that you included a wrong entity for non-breaking dash, it's `&#8209;` rather than `&8209;`.

I'm closing the PR, since as explained, it could be achieved more easily just by adjusting the configuration.

Finally I'm personally sorry, that we failed to let you know this at much earlier stage, and you went trough the hassle of creating two pull requests - sorry for that.

gbastien:
Hi @mlewand 

no problem about delay as we build a little plugin based on https://ckeditor.com/cke4/addon/nbsp

I think it is still relevant to have these 2 characters in the default special characters set, it is useful when it is about "editing" content, I mean the "document editor" way, so for real editors that need to build really accurate composition.

We are working with governement that use CKeditor to write their laws and meeting reports, that is why we need that kind of behavior for CKeditor.

Thank you,

Gauthier

PR #217, opened 2015-09-29
Support for strikethrough / s in bbcode plugin

akahl-owl:
This small change adds support for the `s` tag in BBCode, resulting in strikethrough text. This seems to be widely supported by different BBCode implementations.

### DIFF
--- a/plugins/bbcode/plugin.js
+++ b/plugins/bbcode/plugin.js
@@ -26,9 +26,9 @@
 		}
 	} );
 
-	var bbcodeMap = { b: 'strong', u: 'u', i: 'em', color: 'span', size: 'span', quote: 'blockquote', code: 'code', url: 'a', email: 'span', img: 'span', '*': 'li', list: 'ol', s: 'span' },
-		convertMap = { strong: 'b', b: 'b', u: 'u', em: 'i', i: 'i', code: 'code', li: '*' },
-		tagnameMap = { strong: 'b', em: 'i', u: 'u', li: '*', ul: 'list', ol: 'list', code: 'code', a: 'link', img: 'img', blockquote: 'quote' },
+	var bbcodeMap = { b: 'strong', u: 'u', i: 'em', s: 's', color: 'span', size: 'span', quote: 'blockquote', code: 'code', url: 'a', email: 'span', img: 'span', '*': 'li', list: 'ol' },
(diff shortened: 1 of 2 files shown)


Reinmar:
Hi and thanks for the PR.

I'm not sure though if `[s]` should become a `<span>`. Why not transforming it to `<s>` which the original strikethrough feature uses?

Second thing is that we need tests. Please see http://docs.ckeditor.com/#!/guide/dev_tests and https://github.com/ckeditor/ckeditor-dev/blob/master/tests/plugins/bbcode/bbcode.js

akahl-lshift:
Hi Reinmar,

both `<s>` and `<strike>` are deprecated and their use is discouraged: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/strike

```Usage Note: As with all purely styling elements, <strike> has been deprecated in HTML 4 and XHTML 1, and obsoleted in HTML5. If semantically appropriate, i.e., if it represents deleted content, use the <del> instead; in all other cases use a <span> element and style it with the CSS text-decoration property, with the line-through value.```

I'll quickly write one or more tests and come back.

Reinmar:
The `<s>` element is totally valid: http://www.w3.org/TR/html5/text-level-semantics.html#the-s-element

There's always question of the semantics – whether the user actually means what the element is supposed to mark up. But there are exactly the same problems with `<del>`, `<strong>`, etc. Therefore, these styles are [configurable](http://docs.ckeditor.com/#!/api/CKEDITOR.config-cfg-coreStyles_bold), so the developer integrating CKEditor with his/her website can make the final decision.

In any case, the default strikethrough feature uses `<s>` by default, so bbcode should use it too to be consistent.

akahl-lshift:
I'm a little confused by what Mozilla is writing but take your w3.org reference as having a higher precedence. The two added commits reflect this and add simple tests.

akahl-lshift:
Is this still being considered?

mlewand:
Don't worry, we'll review it soon. Currently we have a lot of tickets in review queue, so as soon as we'll deal with that we'll be back to your PR. Stay tuned!

f1ames:
Thank you for contributing @akahl-owl!

We have moved your changes to a fresh PR (with all files up to date) and will proceed from there, please see https://github.com/ckeditor/ckeditor-dev/pull/3367.

PR #218, opened 2015-10-06
Fix typos and make comments consistent in build.sh

wimleers:
I first noticed `arrgs` instead of `args`, then looked for other nitpicks to fix :)


fredck:
Details matter! Thanks @wimleers ;)

PR #220, opened 2015-10-12
Switch the editor to use readonly=true instead of contenteditable=false for base editable tag.

dustyjewett:
Fixes https://dev.ckeditor.com/ticket/13807

Changing contenteditable to false on the main editable turns off some events like focus. It's also semantically incorrect. This PR addresses this by changing the readonly attribute instead.

The biggest refactor is in range.checkReadOnly, in which the logic was changed to remain consistent in it's assertions. An additional test was added here to ensure adequate coverage.


dustyjewett:
bender was lying to me, I have some failing tests.

PR #222, opened 2015-10-14
Localization of the smiley's titles.

anhr:
Currently localized English and Russian. Any other languages welcome

### DIFF
--- a/config.js
+++ b/config.js
@@ -67,7 +67,33 @@ CKEDITOR.editorConfig = function( config ) {
 		'toolbar,' +
 		'undo,' +
 		'wysiwygarea';
-	// %REMOVE_END%
+    // %REMOVE_END%
+
+    // Toolbar configuration generated automatically by the editor based on config.toolbarGroups.
+	config.toolbar = [
+        { name: 'document', groups: ['mode', 'document', 'doctools'], items: ['Source', '-', 'Save', 'NewPage', 'Preview', 'Print', '-', 'Templates'] },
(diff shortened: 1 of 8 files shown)


mlewand:
Thank you for your pull request. However we'll strip `smileys` plugin as this is beyond the scope your proposal in PR description.

There were few changes needed to be made to improve your PR:

* if no translation is found then we should still use config (backward compatibility, e39760e),
* this change should be also applied for assistive techonologies (accessibility, 97442d069),
* code style.

As for any other lang file than `en.js` you need to contribute thoguh [Transifex](https://www.transifex.com/ckeditor/ckeditor/), more on this in [CKEditor Localization](http://docs.cksource.com/CKEditor_3.x/Developers_Guide/Localization) article. So we'd like to ask you to contribute Russian localization strings there.

Merged with f9b3094.

mlewand:
Unfortunately we had to revert your pull request. Let me explain the reason for that with an example.

Let's say that there is some English blog, and it uses CKE for comments fromatting. If you access it with a Russian browser (assuming that English UI lang is not forced) CKE will automatically switch UI to Russian version. In this case if you'll put some smileys, it will use Russian localization, which will be confusing for different language users.

I'm sorry for not catching this fact during the initial review. As for now the best way to localize your smileys is in fact [config.smiley_descriptions](http://docs.ckeditor.com/#!/api/CKEDITOR.config-cfg-smiley_descriptions) property if you want them to be localized in your site. We can still use your localization for UI-only translations, but it's too late to include it to 4.5.5 now.

PR #224, opened 2015-10-18
Add config to map smileys to one or more BBcode-representations.

mandark:
hi there!

until now the mappings of smileys to their BBcode-representations was hardcoded into the BBcode-plugin. this resulted in errors when adding or modifying the smileys via the configs of the smiley-plugin.

additionally, this mapping was a 1:1 mapping, meaning e.g. the bbcode ":(" would map to the "sad"-smiley, but the bbcode ":-(" wouldn't.

my changes introduce a new config-setting (config.bbode_smiley_map), which allows mapping of a smiley (via it's description defined in config.smiley_descriptions) to one or more BBcode-representations.

i guess this would also close the ticket 8198.

best wishes,
mandark

### DIFF
--- a/plugins/bbcode/plugin.js
+++ b/plugins/bbcode/plugin.js
@@ -3,7 +3,9 @@
  * For licensing, see LICENSE.md or http://ckeditor.com/license
  */
 
+
 ( function() {
+
 	CKEDITOR.on( 'dialogDefinition', function( ev ) {
 		var tab,
 			name = ev.data.name,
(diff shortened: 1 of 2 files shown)


mlewand:
Hello @mandark, thanks for the contribution. Sorry that we're reviewing this PR that late (actually it's one of the oldest PRs here).

Since few years passed by now, the web is now going with Emojis, thus it renders our Smileys plugin deprecated. That's the reason why we introduced new [CKEditor 4 Emoji plugin in 4.10.0](https://ckeditor.com/blog/CKEditor-4.10-released/#emoji).

Also the very feature that you're requesting here has been proposed in [4.11.0 enhancements for Emoji plugin](https://github.com/ckeditor/ckeditor-dev/issues/2062#issue-330198329) (see "Turning Popular Matches into Emoji" section). As such this is where we want to put this feature, rather than the Smileys/BBCode plugin.

If you'd still wish to have it integrated with Smileys/BBCode only, your best option is to create a new plugin. You could also publish it in our [Addons Repository](https://ckeditor.com/cke4/addon)!

PR #225, opened 2015-10-27
Ckeditor on IE8/9 fail to load when a polyfill for classList is included

sleung-ttc:
The classList polyfill does not reach the elements within the editor iframe, but supportsClassLists assumed all elements has it available once the initial test passed. 

This is causing some issue for ie8/9 users.

addresses [#13867](http://dev.ckeditor.com/ticket/13867)

### DIFF
--- a/core/dom/element.js
+++ b/core/dom/element.js
@@ -154,9 +154,15 @@ CKEDITOR.dom.element.clearMarkers = function( database, element, removeFromDatab
 };
 
 ( function() {
-	var testElement = document.createElement( 'span' ),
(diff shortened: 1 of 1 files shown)


mlewand:
Thanks for the contribution! Unfortunately, this time, we can't accept this PR as it contains reference error (check method is called `supportsClassList` while later on code calls `supportsClassLists`) and it does return `undefined` for any call (there's no `return` statement).

We'll work on this issue in [t/13867](http://dev.ckeditor.com/ticket/13867).

PR #226, opened 2015-10-28
added broken test for inline button expectation

trshafer:
This is a broken test for https://github.com/ckeditor/ckeditor-dev/pull/219 . This does not have the code fix, as there are a few ways to fix the test.

addresses http://dev.ckeditor.com/ticket/13802


mlewand:
Sorry for late reply. What we really need is fix + tests, the unit test alone does not give us much. So as much as it is a valid TC I need to close it until we have a complete solution.

PR #228, opened 2015-10-30
null check on listeners

marcpeabody:
We are getting an issue in IE10(and IE11 on Windows 8) where we could trigger the remove function to get called twice and the second call would complain that blur could not be found on a null or undefined object.

It was intermittent but the odds went up significantly when I would click on the background area of the toolbar and then very quickly click the mathjax button in the toolbar (which uses a dialog).

We have found that adding this if-conditional check locally has resolved the issue.

### DIFF
--- a/core/focusmanager.js
+++ b/core/focusmanager.js
@@ -235,8 +235,10 @@
 		remove: function( element ) {
 			element.removeCustomData( SLOT_NAME );
 			var listeners = element.removeCustomData( SLOT_NAME_LISTENERS );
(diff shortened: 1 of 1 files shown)


mlewand:
Hi @marcpeabody!

Before we'll be able to review PR please provide a step-by-step instruction on how to reproduce the issue. Also probide tests for your change for more details on that see [Contributing Code and Providing Patches to CKEditor](http://docs.ckeditor.com/#!/guide/dev_contributing_code) article.

mlewand:
Closing PR due lack of feedback.

PR #230, opened 2015-10-31
Can use the editor on a php script i will sell ?

RachidBourougaa:
I wonder if I can use the editor on a script (a site) that I will sell on Envato market?


mlewand:
Hi @RachidBourougaa,

Pull Requests is not the best place to ask this question but you'll definitely find answer by asking our team through [contact form at CKSource.com](https://cksource.com/contact). :)

PR #231, opened 2015-11-17
T/10682

thomasc72:
Changes made here to make certain a tbody row exists for the table after done with the table properties.
It will add the row if it doesn't exist.
Also fixes an issue with the table properties header dropdown being set to row when no header row exists but the <thead> group exists.
Also adds an option to set a footer row group <tfoot> for the table.
I've tested this in house and it does what is desired.
I couldn't figure out how to bender.js test it.


f1ames:
Thanks for the PR!

I see it aims to solve [#10682](http://dev.ckeditor.com/ticket/10682) which is nice because the issue was reported some time ago.
As you mentioned in the description it fixes 3 issues:

>  tbody row exists for the table after done with the table properties

> fixes an issue with the table properties header dropdown being set to row when no header row exists but the group exists

> adds an option to set a footer row group for the table

First and second issues are connected but the third one is little off topic. It's rather improvement than fix and not related to [#10682](http://dev.ckeditor.com/ticket/10682) so I suggest it should be moved to individual PR.

> I couldn't figure out how to bender.js test it.

It's good that you were also thinking about adding tests. Proper PR should have both unit and manual tests. So I will ask you to provide it too.
Writing tests is easier when you could base on existing tests, in this case I suggest [table plugin tests](https://github.com/ckeditor/ckeditor-dev/tree/master/tests/plugins/table) - there are both manual and unit tests. To set up the testing environment I suggest to follow [Testing Environment Guide](http://docs.ckeditor.com/#!/guide/dev_tests).
If you still have some problems with it do not hesitate to ask, I'm sure we can figure it out.

mlewand:

PR #232, opened 2015-11-19
Change the order of toolbar buttons in BiDi plugin with RTL UI

itgoldman:
reversed "order" of toolbar icons when editor language is RTL so would look the right way

### DIFF
--- a/plugins/bidi/plugin.js
+++ b/plugins/bidi/plugin.js
@@ -236,8 +236,14 @@
 
 			var lang = editor.lang.bidi;
 
-			addButtonCommand( 'BidiLtr', lang.ltr, 'bidiltr', bidiCommand( 'ltr' ), 10 );
-			addButtonCommand( 'BidiRtl', lang.rtl, 'bidirtl', bidiCommand( 'rtl' ), 20 );
+			if (editor.lang.dir=="rtl") {
+				// reversed "order" of toolbar icons when editor language is RTL so would look the right way
+				addButtonCommand( 'BidiRtl', lang.rtl, 'bidirtl', bidiCommand( 'rtl' ), 10 );
+				addButtonCommand( 'BidiLtr', lang.ltr, 'bidiltr', bidiCommand( 'ltr' ), 20 );
+			} else {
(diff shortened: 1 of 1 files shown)


plugins/bidi/plugin.js
```diff
@@ -236,8 +236,14 @@
 
 			var lang = editor.lang.bidi;
 
-			addButtonCommand( 'BidiLtr', lang.ltr, 'bidiltr', bidiCommand( 'ltr' ), 10 );
-			addButtonCommand( 'BidiRtl', lang.rtl, 'bidirtl', bidiCommand( 'rtl' ), 20 );
+			if (editor.lang.dir=="rtl") {
+				// reversed "order" of toolbar icons when editor language is RTL so would look the right way
+				addButtonCommand( 'BidiRtl', lang.rtl, 'bidirtl', bidiCommand( 'rtl' ), 10 );
```
mlewand: This code should be simplified. You don't have to copy and paste both calls to `addButtonCommand`, instead you could just use a ternary operator, by extracting the text direction value:

```js
var isLtr = editor.lang.dir === 'ltr';

addButtonCommand( 'BidiLtr', lang.ltr, 'bidiltr', bidiCommand( 'ltr' ), isLtr ? 10 : 20 );
addButtonCommand( 'BidiRtl', lang.rtl, 'bidirtl', bidiCommand( 'rtl' ), isLtr ? 20 : 10 );
```

plugins/bidi/plugin.js
```diff
@@ -236,8 +236,14 @@
 
 			var lang = editor.lang.bidi;
 
-			addButtonCommand( 'BidiLtr', lang.ltr, 'bidiltr', bidiCommand( 'ltr' ), 10 );
-			addButtonCommand( 'BidiRtl', lang.rtl, 'bidirtl', bidiCommand( 'rtl' ), 20 );
+			if (editor.lang.dir=="rtl") {
```
mlewand: Code style, we use single quotes, spaces in the parenthesis.

mlewand:

PR #234, opened 2015-11-29
Fixed Help icon in RTL layout #13473

AhmedMustafa:
Fixing the help icon in RTL layout.
http://dev.ckeditor.com/ticket/13473
https://github.com/ckeditor/ckeditor-dev/pull/196
https://github.com/ckeditor/ckeditor-dev/pull/209


jacekbogdanski:
Hello,

thanks for the PR. I'm afraid that we cannot provide changes to the editor's `core` code based on plugins sniffing. Core library shouldn't know about external plugins due to among others [Separation of Concerns](https://en.wikipedia.org/wiki/Separation_of_concerns) programming principle.

Although I see that you could expand `getIconStyle` function with special icon mapping and provide different mapping for the Hebrew language. With that improvement, your changes will be more general and we will able to reuse this code in the future.

PR #235, opened 2015-12-01
widget definition insert method now includes editor as argument

pm-marcparmet:
The `insert` member of a widget definition is a no-arg function and the current editor instance is not readily available to any implementation of `insert`, short of a workaround that constructs a new widget definition for every editor instance or does something clever with closures. Revise the definition of `insert` to include the editor as an argument.

This widget definition API change will be fully backwards compatible.

I tried creating a ticket at dev.ckeditor.com for this pull request but got
```
Trac detected an internal error:
error: nothing to repeat
There was an internal error in Trac.
```

### DIFF
--- a/plugins/widget/plugin.js
+++ b/plugins/widget/plugin.js
@@ -1853,7 +1853,7 @@
 				// Otherwise...
 				// ... use insert method is was defined.
 				else if ( widgetDef.insert )
-					widgetDef.insert();
+					widgetDef.insert( editor );
 				// ... or create a brand-new widget from template.
 				else if ( widgetDef.template ) {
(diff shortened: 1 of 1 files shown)


gsmiro:
I'd also add the commandData as an argument since it enables one to integrate an external workflow which passes data to the widget based on page logic.

mlewand:
First sorry for delay, it's been a busy period for us, and now we're back to older PRs.

The proposal looks good, however before merging it into CKE I'd like to resolve issue raised by @gsmiro (as `editor` alone could be obtained from the scope, where you have access to the editor and add widget definition). I have created a dedicated issue on that: #2253.

We'll adjust the solution proposed by @marcparmet and customize it accordingly on a separate branch. Thanks a lot for the contribution!

PR #236, opened 2015-12-01
Error check when getWindow() doesn't exist

Cycododge:
Likely not the best solution. Condition is after initializing CKEditor (as iframe), resizing the window, then removing the editor from the DOM improperly with jquery.

```
Uncaught TypeError: Cannot read property 'frameElement' of undefinedCKEDITOR.editor.resize 
@ ckeditor.js:328resizeEditor 
@ plugin.js?t=F969:139(anonymous function) 
@ plugin.js?t=F969:60
```

### DIFF
--- a/plugins/autogrow/plugin.js
+++ b/plugins/autogrow/plugin.js
@@ -119,6 +119,9 @@
 		}
 
 		function resizeEditor() {
+			//stop here if this doesn't exist yet
+			if(!editor.getWindow){ return; }
+
 			// Hide scroll because we won't need it at all.
 			// Thanks to that we'll need only one resizeEditor() call per change.
 			if ( maxHeightIsUnlimited )


Tade0:
Hello,

could you say something more about this way of removing the editor that you mentioned?

Cycododge:
The editor is no longer implemented in the same way that was causing the issue, however the UX was:

* Click button to show editor
* Editor added to DOM
* Editor slides opens with jQuery
* Use clicks button to close
* Editor container slides closed with jQuery
* Editor instance is destroyed

Honestly it has been a long time, so I don't have any more information than that available.

mlewand:
@Cycododge thank you for going to all the trouble of recalling the steps.

By the looks of the stack trace, it looks as if the window was no longer defined here: https://github.com/ckeditor/ckeditor-dev/blob/056ee77/core/creators/themedui.js#L272

We can't do a fix without a viable repro steps and test coverage.

@Cycododge if you'd be interested in investigating it further, feel free to reopen the ticket as you have more information. For now I'll close the issue. Sorry for that it took us so long to get back to that, but we had a ton of work for long time.

PR #237, opened 2015-12-25
Fix null dereference exceptions during SPA navigations.

kevgrig:
In a single page application (SPA) using JQueryMobile, an instance
of CKEditor is loaded in page1 and during this loading process, a
user navigates to page2. The SPA framework submits an AJAX request,
removes page1 from the DOM and inserts page2 into the DOM into the
"content" section of the page. Various null dereference exceptions
have been observed within CKEditor code during this process. This
commit adds some sanity checks to those places.


elderbas:
I've noticed this error too. I applied your changes to my own branch and I'm not getting any errors but I'm not instantiating a new instance either.  I should be using .destroy(true) on the editor instance before creating a new one, correct?

kevgrig:
Hi @elderbas when you say that you're not instantiating a new instance, do you mean that applying the patch breaks something, or that you're trying to test the change?

In my case, I'm going from a page with an editor to a page without an editor, so I'm simply hitting issues of code running on a no-longer-existing editor, rather than a problem with a new editor.

llexa313:
Will this PR be merged? I have same issue.

f1ames:
Thanks for this PR @kevgrig. This pull request is missing tests, see our [Contributing Code](http://docs.ckeditor.com/#!/guide/dev_contributing_code) guide for more details. If you need some further assistance with it, do not hesitate to ask.

kevgrig:
@f1ames Hi, I don't think it's possible to write tests for this, so, when I find some time, I'll look into the option of submitting an incomplete PR...

f1ames:
I think we need some way to evaluate the patch and be sure if it fixes the issue. Isn't it possible to simulate/reproduce the case you described in your https://github.com/ckeditor/ckeditor-dev/pull/237#issue-123846351 using manual or automatic tests?

kevgrig:
Screwed up when refactoring and accidentally closed.

kevgrig:
@f1ames I created a manual test in commit [3c853d9](https://github.com/ckeditor/ckeditor-dev/pull/237/commits/3c853d90323a2d026e1c667ab21aafa46ff172f8) with JQueryMobile, two pages, a dynamically generated CKEDITOR and navigation between, but I can't reproduce the problem. My only guess is that it might somehow be related to AJAX page replacement rather than static page replacement, but I don't know. How do you recommend we move forward with this PR?

kevgrig:
I keep hitting new exceptions day-by-day - I think ckeditor simply wasn't designed for my use case, so I'm going to switch to catching exceptions and suppressing them. The editor creating the exceptions is basically just trying to die anyway, so it might be too difficult to make it die gracefully.

PR #238, opened 2015-12-30
Allowing editing the link text in the dialog

khaledzaidan:
Premise: If I have a link in the editor, and I want to add text to the beginning or end of the link text, placing the cursor at either end of the link would just add regular text (outside the <a> tag). If select the text of the link and type some text, only the first character I type becomes the text of the link, and the rest of the text is just regular text. The only way I found was to view the source and do the edits there (which is not ideal).

Ideally, I would love the editor to just "know" that I want to add text to the link if there are no spaces between the link and the cursor. But I understand this is complicated (and is honestly beyond my knowledge in ckeditor).

Instead, I've allowed editing the text of the link from the dialog! Seems intuitive to be able to edit the text from there, and doesn't really conflict with the ideal solution I mentioned earlier when/if it gets implemented.

I also added the label "Text" to the lang files.

### DIFF
--- a/plugins/link/dialogs/link.js
+++ b/plugins/link/dialogs/link.js
@@ -116,6 +116,22 @@
 				label: linkLang.info,
 				title: linkLang.info,
 				elements: [ {
+					type: 'text',
+					id: 'text',
+					label: linkLang.text,
+					required: true,
+					setup: function( data ) {
+						if ( data.text )
+							this.setValue( data.text );
+					},
+					commit: function( data ) {
+						if ( !data.text )
+							data.text = '';
+
+						data.text = this.getValue();
+					}
+				},
+				{
 					id: 'linkType',
 					type: 'select',
 					label: linkLang.type,
@@ -795,6 +811,7 @@
 				}
 
 				var data = plugin.parseLinkAttributes( editor, element );
+				data.text = element.getText();
 
 				// Record down the selected element in the dialog.
 				this._.selectedElement = element;
@@ -839,6 +856,7 @@
 
 					element.setAttributes( attributes.set );
 					element.removeAttributes( attributes.removed );
+					element.setText( data.text );
 
 					// Update text view when user changes protocol (#4612).
 					if ( href == textView || data.type == 'email' && textView.indexOf( '@' ) != -1 ) {
--- a/plugins/link/dialogs/link.js
+++ b/plugins/link/dialogs/link.js
@@ -811,7 +811,8 @@
 				}
 
 				var data = plugin.parseLinkAttributes( editor, element );
-				data.text = element.getText();
+				if (element)
+					data.text = element.getText();
 
 				// Record down the selected element in the dialog.
 				this._.selectedElement = element;
--- a/plugins/link/dialogs/link.js
+++ b/plugins/link/dialogs/link.js
@@ -813,6 +813,8 @@
 				var data = plugin.parseLinkAttributes( editor, element );
 				if (element)
 					data.text = element.getText();
+				else
+					data.text = editor.getSelection().getSelectedText();
 
 				// Record down the selected element in the dialog.
 				this._.selectedElement = element;
@@ -845,6 +847,14 @@
 						element: 'a',
 						attributes: attributes.set
 					} );
+
+					// Replace the text in the editor with the text from the dialog.
+					if ( data.text ) {
+						var dialogText = new CKEDITOR.dom.text( data.text, editor.document );
+						range.deleteContents();
+						range.insertNode( dialogText );
+						range.select();
+					}
 
 					style.type = CKEDITOR.STYLE_INLINE; // need to override... dunno why.
 					style.applyToRange( range, editor );
--- a/plugins/link/lang/af.js
+++ b/plugins/link/lang/af.js
@@ -54,6 +54,7 @@ CKEDITOR.plugins.setLang( 'link', 'af', {
 	targetFrameName: 'Naam van doelraam',
 	targetPopup: '<opspringvenster>',
 	targetPopupName: 'Naam van opspringvenster',
+	text: 'Text', // MISSING
 	title: 'Skakel',
 	toAnchor: 'Anker in bladsy',
 	toEmail: 'E-pos',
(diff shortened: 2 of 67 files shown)


k644606347:
good job!

ghidinelli:
This is a long-requested and great feature for usability - it's fairly shocking this isn't already part of CkEditor after so many years. I see that this has a merge conflict but there are no lines marked as conflicting as a viewer. Can we get this to merge and released?

mlewand:
Yes we'd like to have this feature. However I'm not sure how much this PR will help. I can see, just by looks of it, couple of issues:

* missing tests,
* language file changes (only en.js is supposed to be changed directly, rest is handled automatically as explained in our docs),
* it always puts link text field, which should not be the case when you're linking an image or a widget,

Nonetheless it requires a proper review, we'll try to check it soon. Thanks for pinging!

mlewand:
@Tade0 could you review this PR? There is a ticket for it in Trac, so you can search for it and you might find more requirements there.

Tade0:
There's one additional issue I found:
Whenever the selection contains something else than a single line of text things begin to break. I achieved the most spectacular results when trying to create a link when a table or a list was selected.

From the discussion in this ticket:
http://dev.ckeditor.com/ticket/7154
I assume the desired way to handle this would be to deactivate the link text field if the selection is something else than a single line of text.

ghidinelli:
Thanks @mlewand and @Tade0. We're not ckeditor developers but I have someone on my team who could help flesh things out if you need a hand and can provide a bit of direction. I see you're targeting 4.5.8 - cheers!

mlewand:
@ghidinelli that sound great - we're more than happy to cooperate with you to get it done! I'll ask @Comandeer to provide guides on that one.

Comandeer:
The issue caused by converting non-text links into plain text could be fixed by simple checking if the selection contains any non-text elements. The idea how to do it could be borrowed from: http://dev.ckeditor.com/attachment/ticket/7154/7154.patch (`setup` function). That simple improvement should fix most of the issues with the current solution. Mind also that applying such change requires change in [line 870](https://github.com/khaledzaidan/ckeditor-dev/blob/master/plugins/link/dialogs/link.js#L870) (there is no check while setting the new link's text, so there is high possibility that we end up with empty link if the field is disabled!). I'm also not sure if lines [851-857](https://github.com/khaledzaidan/ckeditor-dev/blob/master/plugins/link/dialogs/link.js#L851-L857) would be longer needed.

About the tests: there should be at least one unit test (similar to [existing ones](https://github.com/ckeditor/ckeditor-dev/blob/0f2ef4c49b1b83a6157756f6c39003ed221cd4e0/tests/plugins/link/link.js#L151-L163)) and one manual test (with [`.html` file](https://github.com/ckeditor/ckeditor-dev/blob/0f2ef4c49b1b83a6157756f6c39003ed221cd4e0/tests/plugins/link/manual/parseuppercasedurlparams.html) and [`.md` file](https://github.com/ckeditor/ckeditor-dev/blob/0f2ef4c49b1b83a6157756f6c39003ed221cd4e0/tests/plugins/link/manual/parseuppercasedurlparams.md)). The tests should check if the field is disabled in non-text content (e.g. when we're editing link with line break inside or link inside list), check if the link's text is properly updated and check if the link's content is not deleted when "Text" field is disabled.

There is also one codestyle issue, very easy to fix: just use brackets for every `if` blocks, even the one-line ones ;)

ghidinelli:
Thanks guys, I'm going to have my teammate @ryanguill take a crack at this. How do you feel about adding hover support for following a link or unlinking text (and more specifically, is it possible?)
screen shot 2016-04-13 at 17 49 54

mlewand:
Great to hear that, we're looking forward to it.

@ghidinelli It would be an interesting feature that could be extracted as a stand alone plugin, and published in [CKEditor's Add-ons Repository](http://ckeditor.com/addons/plugins/all).

ryanguill:
A couple questions - I looked at the contribution docs but was unclear on a few things:

1) should I branch off and target master or is there a better branch to target?
2) whats the easiest way to test manually while developing? Is there a url I can hit while ive got the bender server running?

Comandeer:
1) Yes, I think master would be the best.
2) If you got bender server running, its dashboard should be accessible under http://localhost:1030. More info in our docs: http://docs.ckeditor.com/#!/guide/dev_tests

ryanguill:
Thanks @Comandeer, im not sure if I explained what im looking for well though - im looking for a place to test my changes - I have bender server running but I only see how to run the unit tests - im looking for a way to see an example ckeditor so that I can play with my changes as I develop.

Comandeer:
You can use `/samples

Comandeer:
You can use [`/samples`](https://github.com/ckeditor/ckeditor-dev/tree/master/samples) directory inside repo as your playground. Probably you'll just have to add your plugin to [`config.js`](https://github.com/ckeditor/ckeditor-dev/blob/master/config.js) to be loaded inside samples.

ryanguill:
Thanks for that @Comandeer.  So I believe I have the work done, thanks mostly to the patch that you referenced.  Was there a reason that patch wasn't successful 5 years ago? I had to change a few things but the broad strokes seem to work well.

I am having trouble with the tests though - the one test I have tried to write doesn't work the way that I think it should and my changes are also causing the failure of another test in the suite, but I am unsure what it is really testing or how it is supposed to work.  I don't believe I can take over this PR directly (correct?) so would the best thing to do at this point is to create a new PR that we can work from and get feedback on my solution and the tests?

mlewand:
> Was there a reason that patch wasn't successful 5 years ago?

Well the main reason is that it was not just reviewed. We have huge base of tickets, and some patches (especially old ones, where tickets are not updated / requested by the community) just did not get our attention. Also if this PR was based on Garry's patch, then it had some flaws highlighted by @Tade0.

> are also causing the failure of another test in the suite

Please, check what's the reason for this problem. I know that sometimes it takes some time, but you need to understand if it's a regression that new code is adding, or if it's simply a problem with failing test. If it's the latter, you could ofc fix the test. Just make sure to explain inside PR discussion why you think it was a wrong test.

> I don't believe I can take over this PR directly

Yes, you're right - if you create new PR we'll be able to follow it and get updates as you push the code. You could reference this PR in the description, so that new comers have a better background/understanding of the problem. :)

Cheers!

ryanguill:
New PR here: https://github.com/ckeditor/ckeditor-dev/pull/265

@mlewand I dont intend to leave any unfounded regressions, I don't understand what the test is testing so I am unsure what the problem is or how to resolve.

mlewand:
All right, good news: [the feature](http://dev.ckeditor.com/ticket/7154#comment:25) made it's way into upcoming 4.5.11 release! It will be out this week.

Thank you all for the contribution!

PR #239, opened 2016-01-07
allow customization of the dialog for URL entry

JulienGremillotKosmos:
I think this dialog is a little rough, so I'd like to be able to modify the label without impacting other ckeditor messages and add a little explanation :
- allow custom URL label for this plugin
- allow adding custom HTML description on the URL entry dialog, before the text input

### DIFF
--- a/plugins/embedbase/dialogs/embedbase.js
+++ b/plugins/embedbase/dialogs/embedbase.js
@@ -9,6 +9,8 @@ CKEDITOR.dialog.add( 'embedBase', function( editor ) {
 	'use strict';
 
 	var lang = editor.lang.embedbase;
+	var urlLabel = lang.url ? lang.url : editor.lang.common.url;
+	var description = lang.description ? lang.description : '';
 
 	return {
 		title: lang.title,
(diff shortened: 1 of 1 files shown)


mlewand:
@JulienGremillotKosmos sorry for the late response, after quite a busy period we're back to PRs.

Now back to the topic, you don't have to put changes like that in the core code of CKEditor 4, as it exposes convenient API for that already.

This is the code that does exactly what you did during the runtime:

```js
CKEDITOR.on( 'dialogDefinition', function( evt ) {
	var dialogDefinition = evt.data.definition,
		tab = dialogDefinition.getContents( 'info' ),
		lang = evt.editor.lang.embedbase;

	if ( evt.data.name == 'embedBase' && tab ) {
		tab.add( {
			type: 'html',
			id: 'description',
			html: lang.description ? lang.description : ''
		} );

		tab.get( 'url' ).label = lang.url ? lang.url : evt.editor.lang.common.url;
	}
} );
```

Even better would be to put this code in a plugin, so you could easily decide where you want to plug it in!

PR #243, opened 2016-01-19
Check for wildcard edge case when checkApplicable

dok:
In the style definition, when the user has not specified an element property, the element will default to the wildcard '*' string.

When we check whether an element is viable for manipulation, we should include the wildcard as a valid item.


dok:
Closed due to breaking side-effects

PR #244, opened 2016-01-19
Preview image for image2 plugin

jevanlingen:
Made an attempt to return the preview image feature from the original image plugin. For now it just shows the image in the dialog; feature could be expanded to show the alignment and other styling too (as the original image plugin did).

### DIFF
--- a/plugins/image2/dialogs/image2.js
+++ b/plugins/image2/dialogs/image2.js
@@ -16,6 +16,7 @@ CKEDITOR.dialog.add( 'image2', function( editor ) {
 
 		lockButtonId = CKEDITOR.tools.getNextId(),
 		resetButtonId = CKEDITOR.tools.getNextId(),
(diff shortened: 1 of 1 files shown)


jacekbogdanski:
Hello,

at first, I would like to apologize for the late response.  We had a lot of work with CKEditor recently but now we are trying to get back to older pull requests.

I tested your feature and it looks very promising. However, we are trying to keep Enhanced Image plugin UI as easy as possible so it won't confuse our users - especially that it has been shipped some time ago and users are used to the current layout. Also, we didn't get any requests for this feature from our community.

One of the best thing with CKEditor code is it has been build using modules so everyone is able to extend it. I believe that you could create separate plugin extending `image2` with proposed changes so CKEditor users will be able to decide if they want to include this feature into Enhanced Image plugin.

Nevertheless, thank you for your contribution, but this time we have to reject merging this changes into CKEditor codebase.

PR #247, opened 2016-01-21
Don't use document.write if document.readyState is 'interactive'

acusti:
Trying to load CKEditor from source breaks in Chrome, because the conditionals in the main ckeditor.js script and core/loader.js only check if `document.readyState == 'complete'`. However, a `readyState` of `'interactive'` on the document element also indicates that document.write can no longer be used, because `'interactive'` indicates that the document has finished loading and has been parsed (but sub-resources such as images, stylesheets and frames are still loading).

[Reference](https://developer.mozilla.org/en-US/docs/Web/API/Document/readyState)

### DIFF
--- a/ckeditor.js
+++ b/ckeditor.js
@@ -20,7 +20,7 @@ else {
 	CKEDITOR._autoLoad = 'ckeditor';
 
 	// Include the loader script.
-	if ( document.body && ( !document.readyState || document.readyState == 'complete' ) ) {
+	if ( document.body && ( !document.readyState || document.readyState == 'interactive' || document.readyState == 'complete' ) ) {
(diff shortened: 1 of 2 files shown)


mlewand:
Thanks for the PR!

I can see that `CKEDITOR.loader` does not have any tests yet (that's a shame!). The best place for test suite will be:

* Start test suite in `tests/core/loader.js` file (or `tests/core/loader/x.js` if it will require multiple files due to document mocking).
* Create `tests/core/ckeditor/ckeditorload.js` to test loading logic in `ckeditor.js` file.

Looking forward for the tests!

mlewand:

PR #249, opened 2016-01-22
Update node version

dok:
The tests do not run unless the node version is specified to be 4.0 or greater

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,6 +2,9 @@
 	"name": "ckeditor-dev",
 	"version": "4.5.7",
 	"description": "The development version of CKEditor - JavaScript WYSIWYG web text editor.",
+	"engines": {
+		"node": ">=4.0.0"
+	},
 	"devDependencies": {
 		"benderjs-coverage": "^0.2.1",
 		"benderjs": "^0.4.1",


mlewand:
Bender works fine without this property specified. What tests are you referring to?

dok:
Are you sure? It seems like when you do a clean install of the repo, benderjs-sinon wants a node version of 4.0.0 or greater. There is definitely some es6 style string interpolation in benderjs-sinon@0.3.1.

Here is a gist of my terminal https://gist.github.com/dok/40acd3495d92b1ca3261

Reinmar:
You're right – benderjs-sinon requires Node 4+. So this PR would make sense there, not here.

mlewand:
All right, @dok would you mind sending PR to [benderjs-sinon](https://github.com/benderjs/benderjs-sinon)?

Reinmar:
Actually, it's already defined in benderjs-sinon: https://github.com/benderjs/benderjs-sinon/blob/master/package.json#L33

So... that's sad – npm should let you know that something is wrong better than it did it:

```
npm WARN engine benderjs-sinon@0.3.1: wanted: {"node":">=4.0.0"} (current: {"node":"0.12.7","npm":"2.14.2"})
```

But I think that it's a npm's issue more than CKE4's. We shouldn't be forced to check all dependencies' requirements and "hoist" them to our package.json. That's the idea.

mlewand:
Ok so there's no need to send a PR. I'm sorry I didn't check if this prop is already there.

dok:
Npm did warn the user that the dependency is unmet but because its a node version change it looks like it installs anyway.

Regardless, the current setup will break the installation process if users are not on node version 4. Changing the version would fix it but its not ideal. \_(ツ)_/¯

PR #251, opened 2016-02-12
Use 'indexOf' strategy instead of 'in'

dok:
Problem: the usage of [in](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/in) was incorrect. It was being used like so:

```
var trees = ["redwood", "bay", "cedar", "oak", "maple"];
"bay" in trees
```

The following will return false, as usage of `in` in an array must be an index value, not a key. Only in objects can you use it like so. In this case, if `this.element` is an array, like `this.element = ['p', 'ul', 'ol']`, then the usage of `element.getName() in this.element` was returning false.

I wrote the fix to move away from using `in`, and replaced it with using the `indexOf` strategy instead.

### DIFF
--- a/core/style.js
+++ b/core/style.js
@@ -401,7 +401,7 @@ CKEDITOR.STYLE_OBJECT = 3;
 				name = element.getName();
 
 			// If the element name is the same as the style name.
-			if ( typeof this.element == 'string' ? name == this.element : name in this.element ) {
+			if ( typeof this.element == 'string' ? name == this.element : this.element.indexOf(name) !== -1 ) {
 				// If no attributes are defined in the element.
 				if ( !fullMatch && !element.hasAttributes() )
 					return true;


Reinmar:
Hi! Thanks for your PR and apologies for the late reply.

Unfortuantely, there are two reasons why this PR is invalid:

1, `this.element` is meant to be an object. I can't find a documentation for that but you can see it e.g. here: https://github.com/ckeditor/ckeditor-dev/pull/251/commits/ebcc318fbd9b6eda586fba90974a48f1e2ff1ffd#diff-a3629c9cb3a3e1d14a5d6a64ba3a6288R176. If I'm mistaken and `this.element` is set to an array somewhere, then we should check why.
2. This code won't work in IE8, because `indexOf` is available since IE9.

PR #252, opened 2016-02-12
Fix backspacing into empty lists

caffodian:
As discovered in https://dev.ckeditor.com/ticket/14395, empty lists cause the list backspace key handler to error out (because of a null element case.)  This causes it to fall back to random default behaviour depending on the browser.

This change makes it so that backspacing into a childless list will instead create a list item in that list to put the backspaced content into.


jacekbogdanski:

PR #253, opened 2016-02-20
[Plugin][ColorButton] Fix color name displayed

egeloen:
Hey!

I maintain the Symfony2 integration of CKEditor and someone reports me the custom color label are not working in egeloen/IvoryCKEditorBundle#204

Basically, something like explain in the doc is not working: http://docs.ckeditor.com/#!/api/CKEDITOR.config-cfg-fileTools_defaultFileName#cfg-colorButton_colors

After debugging it, I notice the plugin fallbacks on the `colorCode` if the color trans is not found but IMO, if someone provides a explicit color label, it should be used directly and the current color code should only be involved if the label is not provided at all.

For the testing part, I'm not familiar with it and so, I was not able to add one... If someone can point me the way to go I can add them :)

### DIFF
--- a/plugins/colorbutton/plugin.js
+++ b/plugins/colorbutton/plugin.js
@@ -177,15 +177,21 @@ CKEDITOR.plugins.add( 'colorbutton', {
 
 				var parts = colors[ i ].split( '/' ),
 					colorName = parts[ 0 ],
-					colorCode = parts[ 1 ] || colorName;
+					colorCode = parts[ 1 ] || colorName,
+					colorLabel;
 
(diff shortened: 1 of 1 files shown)


mlewand:
Hi! Sorry for late reply we've had quite a lot of work lately, and the PR queue got long. Finally we're here - looking at your PR it looks reasonable. There is a need for unit tests but we'll take it from here and add them.

I'll merge your PR into a branch where the tests will be added. You can keep track of the issue in #2271 - I think it should get into next minor release.

Thanks once again!

PR #254, opened 2016-03-03
fixes issue #8584: Support Content Security Policy

caridy:
The solutions that other popular libraries and frameworks are adopting to comply with CSP are:

1) provide a hook consumers to bypass the indirect evaluation routine (doesn't apply to this case)

2) provide a less performant solution that can be used under certain conditions.

I have implemented the corresponding forking logic on `CKEDITOR.template` to detect whether or not indirect eval is available, and use an alternative `output` function that is less performant but works well with CSP.

It is very similar to what Angular does for their expression language, and it is a solution that can be ported back to any previous version of CKEditor that relies on the same CKEDITOR.template mechanism.  This solution can be backported to 4.x (I wasn't sure about the branching strategy on this repo)

By forcing to use the new less-performant solution (setting the `isIndirectEvalSupported` flag manually) we have validated that all tests are passing, but someone will have to do a little bit of work here to make sure that every template test runs with and without the flag.

Backpointer: https://dev.ckeditor.com/ticket/8584
SF Backpointer: @W-2958085

### DIFF
--- a/core/template.js
+++ b/core/template.js
@@ -27,7 +27,6 @@
 	 * @param {String} source The template source.
 	 */
 	CKEDITOR.template = function( source ) {
-		// Builds an optimized function body for the output() method, focused on performance.
 		// For example, if we have this "source":
 		//	'<div style="{style}">{editorName}</div>'
 		// ... the resulting function body will be (apart from the "buffer" handling):
@@ -37,18 +36,21 @@
 		if ( cache[ source ] )
 			this.output = cache[ source ];
 		else {
-			var fn = source
-				// Escape chars like slash "\" or single quote "'".
-				.replace( reEscapableChars, '\\$1' )
-				.replace( reNewLine, '\\n' )
-				.replace( reCarriageReturn, '\\r' )
-				// Inject the template keys replacement.
-				.replace( rePlaceholder, function( m, key ) {
-					return "',data['" + key + "']==undefined?'{" + key + "}':data['" + key + "'],'";
-				} );
-
-			fn = "return buffer?buffer.push('" + fn + "'):['" + fn + "'].join('');";
-			this.output = cache[ source ] = Function( 'data', 'buffer', fn );
+			// less performant solution when CSP is disabling indirect evaluation
+			cache[ source ] = function( data, buffer ) {
+				var m,
+					src = source,
+					chunks = [];
+				while ( ( m = rePlaceholder.exec( src ) ) ) {
+					var k = m[ 1 ];
+					chunks.push( src.substr( 0, m.index ), data[ k ] === undefined ? m[ 0 ] : data[ k ] );
+					src = src.substr( m.index + m[ 0 ].length );
+					rePlaceholder.lastIndex = 0;
+				}
+				chunks.push( src ); // ... the rest
+				return buffer ? buffer.push.apply( buffer, chunks ) : chunks.join( '' );
+			};
+			this.output = cache[ source ];
 		}
 	};
 } )();


core/template.js
```diff
@@ -13,10 +13,18 @@
 		rePlaceholder = /{([^}]+)}/g,
 		reEscapableChars = /([\\'])/g,
 		reNewLine = /\n/g,
-		reCarriageReturn = /\r/g;
+		reCarriageReturn = /\r/g,
+		isIndirectEvalSupported = true;
+
+	try {
+		Function( '1' )();
```
caridy: this is enough to detect whether or not CSP is enforcing `unsafe-eval`.

core/template.js
```diff
@@ -13,10 +13,18 @@
 		rePlaceholder = /{([^}]+)}/g,
 		reEscapableChars = /([\\'])/g,
 		reNewLine = /\n/g,
-		reCarriageReturn = /\r/g;
+		reCarriageReturn = /\r/g,
+		isIndirectEvalSupported = true;
```
caridy: we should probably expose this flag somehow to tests can run with and without the flag.

ralfbecker:
Does that fix both cases mentioned in https://dev.ckeditor.com/ticket/8584#comment:11

The first (eval to keep a closure compiler directive) seems rather trivial, but the second (inline javascript in markup, eg. onclick on buttons) seems the big issue and not affected by your pull request, as far as I can tell.

Ralf

caridy:
@ralfbecker this PR is about indirect eval, which is the hard part IMO. the inline javascript can be replaced with the corresponding DOM APIs (e.g.: addEventListener).

Reinmar:
The indirect eval is the simple part. The inline JavaScript listeners are spread around tens (if not hundreds) of places and the way how the DOM is created makes it hard to use `addEventListener()`. Unfortunately, what made sense 5+ year ago is now a pain in the :horse:.

caridy:
These are two separate vectors in CSP, `unsafe-eval` (for indirect evaluation) and `unsafe-inline` (for inline scripts). While inline listeners fit into `unsafe-inline` rule, I understand that removing them will require some extensive work across the source code, but it is a very mechanical process, and certainly doable to comply with CSP.

Reinmar:
Then I guess @mlewand will consider accepting this PR (he's on vacations right now).

caridy:
@Reinmar yes. In many cases, people will use CSP rules with `unsafe-inline` flag, not ideal, but it is hard to remove all inline scripts. I think merging this PR is a step in the right direction. We now need a brave soul to remove all inline `onclick`s.

mlewand:
Hi @caridy, thanks for the PR!

There are couple things that needs to be addressed before adding it to CKE4:

1. `isIndirectEvalSupported` var should be moved to [`CKEDITOR.env`](http://docs.ckeditor.com/#!/api/CKEDITOR.env). Let's name it `CKEDITOR.env.evalSupported`.
1. New templating logic needs to be tested. In this PR it's not tested at all. Actually we have couple of [tests for template](https://github.com/ckeditor/ckeditor-dev/blob/cce5ef6e8583e25392e3910b808d63f493f1d0ad/tests/core/template.js), so we could use could modify tests object, in a way that it creates tests both for both variants of `evalSupported`. Feel free to add any tests that you find to be necessary to existing test suite.

core/template.js
```diff
@@ -37,18 +44,36 @@
...
+					// Inject the template keys replacement.
+					.replace( rePlaceholder, function( m, key ) {
+						return "',data['" + key + "']==undefined?'{" + key + "}':data['" + key + "'],'";
+					} );
+				var fn = "return buffer?buffer.push('" + src + "'):['" + src + "'].join('');";
+				cache[ source ] = Function( 'data', 'buffer', fn );
+			} else {
+				// less performant solution when CSP is disabling indirect evaluation
+				cache[ source ] = function( data, buffer ) {
+					var m,
```
mlewand: Please, use meaningful variable names - `m`, `k` does not tell anything.

Reinmar:
Some performance tests could be interesting. Maybe the regexp based solution will not be significantly slower and then it could totally replace the previous implementation.

Also, the regexp based solution seems to be overcomplicated. Can't it be coded like this https://github.com/ckeditor/ckeditor5-core/blob/master/src/locale.js#L57?

mlewand:
Actually, yes - plain RegExp if doable would be better here. Generally RegExp should be much quicker than any JavaScript manipulation, as long you cache compiled RegExp itself.

Any thoughts on this @caridy?

caridy:
@Reinmar 

> Some performance tests could be interesting. Maybe the regexp based solution will not be significantly slower and then it could totally replace the previous implementation.

I agree with you on this one. Even though I don't have enough information about how is this used internally by CKE, I suspect that the editor doesn't do intensive DOM operations with templates. In the case of big frameworks like Angular, their expression language has to be computed on every DOM diffing process, which is very intense, not the case here AFAICT.

> Also, the regexp based solution seems to be overcomplicated. Can't it be coded like this https://github.com/ckeditor/ckeditor5-core/blob/master/src/locale.js#L57?

If we decide that regex solution is good enough, I can squash the complexity of that, and make it simpler without forking the logic while reusing the existing tests.

@mlewand

> Actually, yes - plain RegExp if doable would be better here. Generally RegExp should be much quicker than any JavaScript manipulation, as long you cache compiled RegExp itself.

yes, same as above. if the decision is to go with the regex (which seems simpler, safer and more robust without a huge perf degradation, probably a perf gain as you mentioned), then yes, I can make the changes.

> isIndirectEvalSupported var should be moved to CKEDITOR.env. Let's name it CKEDITOR.env.evalSupported.

this is only needed if we keep the fork in the logic.

> New templating logic needs to be tested. In this PR it's not tested at all. Actually we have couple of tests for template, so we could use could modify tests object, in a way that it creates tests both for both variants of evalSupported. Feel free to add any tests that you find to be necessary to existing test suite.

As I mentioned in the description of the PR, I have tested with the same tests by forcing the flag, but if we decide to go with the regex solution, there is no issue since we will be using the same tests.

In resume, you guys should make a call about using the regex vs forking logic, then we can clear up all the obstacles to get this merged.

mlewand:
> I agree with you on this one. Even though I don't have enough information about how is this used internally by CKE, I suspect that the editor doesn't do intensive DOM operations with templates. In the case of big frameworks like Angular, their expression language has to be computed on every DOM diffing process, which is very intense, not the case here AFAICT.

Well, we use it a lot building icons, buttons, parts of UI. It's not processed as extensively as in Angular (based on your information), but still we use it a lot so it's important to keep good performance.

> this is only needed if we keep the fork in the logic.

If new proposal will have similar, better performance, then ofc we could use it without forking. However if current solution will be more performant then unfortunately we'll need to have these 2 branches in code.

All right, so first prepare forked code, driven by `CKEDITOR.env.evalSupported` (without adapting tests yet) and check performance of these two. Based on this we'll discuss on what to do next. Thanks for the input @caridy!

You've made me curious about the result. :)

caridy:
good, I will make the changes tonight.

ralfbecker:
One approach we used when we adapted CSP and forbid inline scripts was to add an event-handler via jQuery after, which get's triggered before the inline code, parses some information out of it, calls the required callback and terminates the event, before inline code get called and CSP violation would be triggered. As you parse the inline-code, you can make sure no arbitrary code get executed.
Looking at current stable CKEditor you could make the buttons work with something like:

```
jQuery('a.cke_button').on('click','a[href^="javascript:"]',function(ev){
	// check onclick attribute, eg.: CKEDITOR.tools.callFunction(6,this);
	var matches = this.attr('onclick').match(/^CKEDITOR.tools.callFunction\((\d+),this\);/)
	if (matches) {
		CKEDITOR.tools.callFunction(parseInt(matches[1]), this);
		ev.stopPropagation();	// do NOT execute regular event, as it will violate CSP, when handler does NOT return false
		return false;	// IE11 seems to require this, ev.stopPropagation() does NOT stop link from being executed
	}
	// everything not matching will trigger CSP violation, as now inline onclick runs
}
```

So you could use something like the above code, if user set attribute to run with CSP enabled.

Ralf

caridy:
@Reinmar @mlewand I have created a micro-benchmark of the template method, here are the results:

1. Create a new template, cache it, and call it with the same dataset:
http://jsperf.com/ckeditor-templates/4

2. Create a new template, and call it with the same dataset:
http://jsperf.com/ckeditor-templates-without-cache

Results are consistent with our initial assessment. In the case of the cached template (perf 1), we are talking about millions operations per second, and obviously, the function is 2x-3x faster, but this is not significant due to the number of operations IMO.

As for the other option (perf 2), if you have to create the template function every time, and then call it, the regex is faster, pretty much 2x-3x faster, and we are talking about half million operations per second.

All in all, I think it is going to be just fine to go with the regex approach.

caridy:
@Reinmar @mlewand any update on the perf analysis?

mlewand:
@caridy thanks for this analysis! Did you try to change alternative template building method to use regexp only as suggested in [comment 193732733](https://github.com/ckeditor/ckeditor-dev/pull/254#issuecomment-193732733) and [comment 193777360](https://github.com/ckeditor/ckeditor-dev/pull/254#issuecomment-193777360)? Assuming that replace regexp would be compiled only once, it might present much better performance.

That being said at the end we'll need also to do deeper profiling, as jsperf tests are not reliable as results might be distorted by browser optimization behind the scene, but no need to worry about this at this point yet.

core/template.js
rpicard: Just want to mention that testing like this could end up cluttering the console with CSP violation reports.

caridy: @rpicard this is a one time operation for detection. as far as I can tell, there is no other way to detect CSP `unsafe-eval` rule. Ideally, we settle on the idea to use only the regex solution going forward, instead of a forked logic that requires detection.

caridy:
@Reinmar, @mlewand I have updated the PR with the regexp output, it passes all the tests, and we can now easily compare the perf characteristics of this branch vs master, hopefully you guys have some perf infrastructure that can be run against different branches.

mlewand:
@caridy That's great! We'll try to have a look on it next week. We'll keep you updated, thanks again!

caridy:
any update on this?

caridy:
More updates about CSP:

We finally got 4.x working under CSP, it is far from ideal, but it works, we use multiple tricks but hopefully we can get the original code to adjust accordingly.

Here are the tricks we are using (just for references):

* MutationObserver to observer changes in the editor's container, and whenever a new element is added into the DOM, the element should be inspected, and any inline event should be transformed into a real DOM event.
* DOM events generated by the previous step have to be evaluated somehow (we use a secure iframe for that).
* domobject.js is responsible for creating new iframes (for plugins popups, etc), we do some tricks to guarantee the initialization of those new documents when document.close() is called.

Here are some of the pain points:

* aside from the tons of places where "onclick", "onfocus", "onblur", "onkeydown", "onkeypress", "onkeyup", "onmousedown", "onmouseout", "onmouseover", "onmouseup" and "onchange" are used when producing a markup (which I hope we can fix in 5.x), the biggest problem is how plugins are creating new iframes, and writing inline scripts in those iframes. If `"unsafe-inline"` is not in place, then that initialization will fail. The curious thing about this is that document.open/write/close is sync, in seems that there is not really a need to listen for "DOMContentLoaded" inside the iframe, and instead that initialization can be carry on right after the call to `.close()`

mlewand:
@caridy take a look at [t/8584b](https://github.com/cksource/ckeditor-dev/tree/t/8584b) this is a simplified regexp replace that I was referring to. It's doing the same thing with lesser lines of code and is simpler.
The only difference there is that it pushes different amount of chunks onto `buffer` array (if present). However we're OK with doing so, since [the doc says](http://docs.ckeditor.com/#!/api/CKEDITOR.template-method-output):
> The number of entries appended to the array is unknown.
So it's not breaking the API.


> MutationObserver to observer changes in the editor's container, and whenever a new element is added into the DOM, the element should be inspected, and any inline event should be transformed into a real DOM event.

That's a clever solution! Out of curiosity, would you mind sharing this part of code? We're just curious to se how you got around this. :)

@Comandeer could you do couple perf test with this approach?

mlewand:
@caridy take a look at [t/8584b](https://github.com/cksource/ckeditor-dev/tree/t/8584b) this is a simplified regexp replace that I was referring to. It's doing the same thing with lesser lines of code and is simpler.
The only difference there is that it pushes different amount of chunks onto `buffer` array (if present). However we're OK with doing so, since [the doc says](http://docs.ckeditor.com/#!/api/CKEDITOR.template-method-output):
> The number of entries appended to the array is unknown.
So it's not breaking the API.


> MutationObserver to observer changes in the editor's container, and whenever a new element is added into the DOM, the element should be inspected, and any inline event should be transformed into a real DOM event.

That's a clever solution! Out of curiosity, would you mind sharing this part of code? We're just curious to se how you got around this. :)

@Comandeer could you do couple perf test with this approach?

Comandeer:
Benchmark for approach presented in `t/8584b` is available at http://bzdety.comandeer.pl/benchmark (it runs very long, so you can just leave your browser running and go do something else).

That approach is much slower than our current implementation when cache is enabled and much faster when cache is disabled. I also prepared its [optimized version](http://bzdety.comandeer.pl/benchmark/template-t-8584b-opt.js), which uses string concatenation instead of arrays and it's faster in Chrome and Firefox (however I didn't test it in IE – in older IEs it could be slower than arrays).

caridy:
@Comandeer thanks for stepping up on this issue. The real question here is: when the cache is enable (which I suspect it is the majority of prod configuration), does the slower version make the cut? Is it good enough? IMO it might be orders of magnitude slower, but still not impacting the end user if that operation is not carry on very frequently.

caridy:
any update on this?

mlewand:
Sorry @caridy again we've been in a busy period once again. CSP got some traction recently, so we'd wish to revisit it soon. I'll set a review as soon as we'll be able to. Sorry for trouble, and thanks for your patience!

Comandeer:
I've created new pull request, #332, based on the above discussion and the optimised version of the template parser. Thx @caridy for your input and proposing the original solution, on which the optimised version is based upon.

ralfbecker:
Does CK Editor 4.7.* now supports CSP in full (no more unsafe-eval AND unsafe-inline) or just no unsafe-eval, but no unsafe-inline is still postponed to 5?

mlewand:
@ralfbecker sorry for replying so late - we added here support for `unsafe-eval`. However CKE4 will still fail to comply with `unsafe-inline` directive, so here CKE5 is your best bet.

The problem is that our old legacy UI parts rely heavily on `onxyz` HTML attributes, and it would be a tremendous effort to change that.

ralfbecker:
@mlewand We figures so much out our self, thought CKE5 is still to far from what we need feature-wise :(

mlewand:
@ralfbecker I hear you.

Well we had an idea to as an workaround to dynamically replace our `on-xyz` attributes (that matches `return CKEDITOR.tools.callFunction(.*);`) after the editor is loaded and then remove these attrs and register them as a proper listeners. But then the problem are that many other components register this kind of listeners at a runtime (dialog for instance).

If you have some time you could give it a try as a R&D to see how it works.

Talking about CKE5 it's really getting the shape quickly. I know that guys are now just polishing the API (and they're actually finishing it currently) - so we should see the release soon. From there it's just a question what features you need.

PR #256, opened 2016-03-08
Updated the Persian (fa) language file.

hsamadi:
Fixed some typos. Also replaced some Unicode u+200B characters with u+200C. The u+200C is actually the zero-width non-joiner character that we use in Persian.


mlewand:
Thanks for the PR. Please put your changes to [Transifex](https://www.transifex.net/projects/p/ckeditor/) project to provide translations.

For more information see [http://docs.ckeditor.com/#!/guide/dev_contributing](http://docs.ckeditor.com/#!/guide/dev_contributing).

PR #257, opened 2016-03-11
#14451 - Widget crashes if editable region has a numeric ID

chaluja7:
http://dev.ckeditor.com/ticket/14451

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -5,13 +5,13 @@ CKEditor 4 Changelog
 
 Other Changes:
 
-* [#12707](http://dev.ckeditor.com/ticket/12707): Fixed: Table elements order to meet HTML specification.
+* [#12707](http://dev.ckeditor.com/ticket/12707): Fixed: The order of table elements does not meet the HTML specification.
 
 ## CKEditor 4.5.7
 
 New Features:
 
-* [#14327](http://dev.ckeditor.com/ticket/14327): Added Swiss German localization.
+* [#14327](http://dev.ckeditor.com/ticket/14327): Added Swiss German localization. Thanks to [Miro Grenda](https://twitter.com/mirogrenda)!
 
 Other Changes:
 
@@ -20,8 +20,8 @@ Other Changes:
 	* [#13377](http://dev.ckeditor.com/ticket/13377): [Widget](http://ckeditor.com/addon/widget) plugin issue when typing in Korean.
 	* [#13389](http://dev.ckeditor.com/ticket/13389): [Blink] [`editor.getData()`](http://docs.ckeditor.com/#!/api/CKEDITOR.editor-method-getData) fails when the cursor is next to an `<hr>` tag.
 	* [#13513](http://dev.ckeditor.com/ticket/13513): [Blink, WebKit] [Div Editing Area](http://ckeditor.com/addon/divarea) and [`editor.getData()`](http://docs.ckeditor.com/#!/api/CKEDITOR.editor-method-getData) throw an error when an image is the only data in the editor.
-* [#13884](http://dev.ckeditor.com/ticket/13884): Fixed: Copy/paste table in Firefox results in just first cell being pasted.
-* [#14234](http://dev.ckeditor.com/ticket/14234): Fixed: URL input field is not marked as required in the [Embed](http://ckeditor.com/addon/embed) dialog.
+* [#13884](http://dev.ckeditor.com/ticket/13884): [Firefox] Fixed: Copying and pasting a table results in just the first cell being pasted.
+* [#14234](http://dev.ckeditor.com/ticket/14234): Fixed: URL input field is not marked as required in the [Media Embed](http://ckeditor.com/addon/embed) dialog.
 
 ## CKEditor 4.5.6
 
--- a/core/dom/element.js
+++ b/core/dom/element.js
@@ -2038,7 +2038,7 @@ CKEDITOR.dom.element.clearMarkers = function( database, element, removeFromDatab
 	}
 
 	function getContextualizedSelector( element, selector ) {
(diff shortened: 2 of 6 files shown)


Reinmar:
Thanks for the PR! Great to see that it has a test.

Unfortunately AFAICS `CSS.escape()` is supported only by latest Firefox and Chrome, so it cannot be used this way. The patch must work on all supported browsers (IE8+ and the rest), so you would need to add a function to `CKEDITOR.tools` which polyfills at least that behaviour that you need to the ids.

chaluja7:
I'm sorry, I should have checked the cross-browser compatibility myself before creating pull request. Anyway thanks for your advice on how to solve this issue.

I could of course write a simple polyfill, that would solve the 'all numeric' ID ('#100') or 'starts with number' ID ('#1aa') issue. But there are much more cases, that needs to be escaped in order to become valid CSS selectors (https://mathiasbynens.be/notes/css-escapes). So here comes question, if more robust solution shouldn't be take into consideration. I found already working polyfill under the MIT license (https://github.com/mathiasbynens/CSS.escape) which looks good but I'm not sure if this can be used here with respect to licensing policy etc.

Thank you so much for advice!

Reinmar:
I'll leave the decision to @mlewand, but I always have doubts whether adding 80LOC just to solve a single edge case is right... Especially that we cannot expose it as a global `CSS.escape()` function because CKEditor should not touch globals. So the developer may need to add it twice.

So, the question I would ask is whether we cannot address only this single case (numeric IDs), because that's the only problem that was reported.

fredck:
I also agree that a balanced solution must be found. Much probably as recommended @Reinmar, implementing `CKEDITOR.tools.escapeCss()` which has feature detection for `CSS.escape()` and fallbacks to a one-liner for the most common (edge) case(s).

chaluja7:
Ok, thanks. I'm on it.

mlewand:
Thanks guys for the input. That's right it makes sense to start with simplified function, that will cover the most generic TC.

`CKEDITOR.tools.escapeCss` is a good candidate for it. If possible it should simply use `CSS.escape`, if not simplified polyfill instead.

Once you have function available in `CKEDITOR.tools` namespace, you can write couple unit tests to it.

**@chaluja7** note that we support IE8 and higher, so you don't have to worry about any quirks with <IE8 browser).

chaluja7:
Ok, I just pushed possible solution based on your comments. Thanks for advices and new review.

Comandeer:
Fix seems to work properly, however there are some issues, mainly with codestyle.

All comments should start with space and capital letter and ends with a dot (just like a sentence), so e.g.:
```javascript
// Invalid input.
```

Brackets should be used even for blocks with one statement, e.g.
```javascript
if ( typeof selector === 'undefined' || selector === null || selector.length === 0 ) {
    return '';
}
```

Your `if`s look too defensive. E.g. in this case it's not necessary to check type of `CSS.escape`, only to check for its existence – so that should be sufficient:

```javascript
if ( CSS && CSS.escape )
```

I'll think about making tests for image2 more generic and move them into `tests/plugins/widget`.

In some places (especially in JSDoc comments) spaces are mixed with tabs.

And at last: there is small error in one of your assertions → https://github.com/ckeditor/ckeditor-dev/pull/257/files#diff-8bdfcee314f07a20ed6c1e45912050c0R691 – the message is not suitable for it.

chaluja7:
Thanks for review. I fixed codestyle as suggested (comments, brackets, tabs vs spaces in JSDoc comments, assertion message). 

Unfortunately the suggested condition:

`if ( CSS && CSS.escape )`

fails on `CSS undefined` in IE so I preserved the original one. I also didn't change the location of tests for image2 - I'm not sure if this is related to this issue.

P.S. I deeply apologize for three commits instead of one - I totally screwed it up.

mlewand:
@chaluja7 Please, don't be so hard on yourself! We're happy to see your contribution.

@Comandeer can I ask you to review new changes and feedback?

Comandeer:
I still don't like the `if`s (yep, I'm quite a grumpy man).

```javascript
if ( typeof selector === 'undefined' || selector === null || selector.length === 0 ) {
```

Can't be it simplified just to:

```javascript
if ( !selector ) {
```

And

```javascript
if ( typeof CSS !== 'undefined' && typeof CSS.escape !== 'undefined' ) {
```

to something like:

```javascript
if ( ( 'CSS' in window ) && CSS.escape ) {
```

(still much simpler than your version ;)).

I see also one codestyle issue that I've missed the previous time: we use one `var` statement to declare all the variables, so:

```javascript
var selector,
	escapedSelector = CKEDITOR.tools.escapeCss( selector );
```

About the tests in `image2.js`: that issue affects all widgets, not only image one. Personally I'll put them in `tests/plugins/widget`. Actually that tests use the same tools that are used inside `widget` plugin tests, so it's just simply moving them to other file. Probably `tests/plugins/widget/nestededitables.js` would be the best option.

With that changes I'll (probably) be happy ;)

fredck:
I find the following more elegant as it doesn't make a reference to `window`:

```
if ( typeof CSS !== 'undefined' && CSS.escape ) {
```

mlewand:
Yea, and I'd, for siplicity sake, go with `if ( window.CSS && CSS.escape ) {` :) KISS. We don't need to deal with non-browser environment here.

That being said, I see that we can take it from here.

@Comandeer please take [#14451](http://dev.ckeditor.com/ticket/14451) issue, and apply necessary fixes there.

@chaluja7 We'll close this PR once the branch is merged. Once again, thank you for the contribution!

mlewand:
Thanks @chaluja7 - your PR has been merged to our master along with http://dev.ckeditor.com/ticket/14451issue! We'll ship it with CKEditor 4.5.10. :tada:

PR #258, opened 2016-03-24
Fixed richcombo focus issue in Chrome

DmitryKrekota:
I have bug with richcombo focus in Chrome. CKEditor richcombo breaks page because of using focus.
Video recording: https://drive.google.com/file/d/0Bxtv10NlSOEbWXNJbkNFVGQyUG8/view
My pull request fixes this Chrome issue.

### DIFF
--- a/plugins/floatpanel/plugin.js
+++ b/plugins/floatpanel/plugin.js
@@ -408,7 +408,9 @@ CKEDITOR.plugins.add( 'floatpanel', {
 						this.focus();
 
 						// Focus the block now.
+						block.element.hide();
 						block.element.focus();
+						block.element.show();
 
 						// #10623, #10951 - restore the viewport's scroll position after focusing list element.
 						if ( CKEDITOR.env.webkit )
--- a/plugins/richcombo/plugin.js
+++ b/plugins/richcombo/plugin.js
@@ -270,7 +270,9 @@ CKEDITOR.plugins.add( 'richcombo', {
 					// The "panelShow" event is fired assinchronously, after the
 					// onShow method call.
 					editor.once( 'panelShow', function() {
+						list.hide();
 						list.focus( !list.multiSelect && me.getValue() );
+						list.show();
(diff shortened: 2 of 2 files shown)


f1ames:
Thanks for the PR!

Looks like very specific case. It may be hard to reproduce and test because it somehow depends on custom code outside the editor. If you could provide a step-by-step instruction on how to reproduce the issue it will be easier for us to analyze and understand where exactly the problem is and what the best solution should be.
Optionally you could also report a bug on [our bugtracker](http://dev.ckeditor.com/) describing steps how to reproduce this issue.

Keep in mind that PR should also contain unit and manual tests (or only manual, if the issue can't be really tested with unit tests).

Thanks again for contribution.

DmitryKrekota:
I created simple demo of this issue: https://drive.google.com/file/d/0Bxtv10NlSOEbdWlsTDZZanBTWjg/view?usp=sharing
How to reproduce:
https://drive.google.com/file/d/0Bxtv10NlSOEbVE94Tkl6dzJSeHM/view

f1ames:
I investigated this issue a little bit further and it seems that this behavior is not the fault of CKEditor but more of how browsers handles specific elements/structure. I modified and simplified your example (which was very helpful by the way) and tested it on Chrome and FF.

It was easier to record a video so here is the link with simplified example and video (along with _video_description.txt_ where I explained what's happening in the video): https://drive.google.com/folderview?id=0B2knXdyRNxzPdFAzaGNwNlMwN1U&usp=sharing.

To sum up the issue (based on provided sample which somehow reflects your application): it seems that the right column even though it's content should not take more space than the window height actually does so window can be actually scrolled - in some specific way (see video I uploaded, it's easy to reproduce on FF even without CKEditor) and adding some additional styling (like _overflow_) can solve this issue.

Hope it helps!

DmitryKrekota:
Thanks for your answer.

PR #260, opened 2016-03-31
Fixed ticked 14558

Duskfall:
Fixed ticket 14558 preventing the widgets to be deleted if the editor is in read only mode

### DIFF
--- a/core/selection.js
+++ b/core/selection.js
@@ -303,6 +303,9 @@
 					range = editor.createRange(),
 					found;
 
+				// if editor is read-only do nothing
+				if(editor.config.readOnly) return false;
+				
 				// If haven't found place for caret on the default side,
 				// try to find it on the other side.
 				if ( !( found = range.moveToClosestEditablePosition( evt.selected, right ) ) )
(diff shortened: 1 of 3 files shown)


f1ames:
Thanks for the PR!

It looks good but there are few modifications needed:

1. [This conditioning](https://github.com/ckeditor/ckeditor-dev/pull/260/files#diff-ae237139107d046c67dc8622d25bace9R307) (`if(editor.config.readOnly)`) works only when _readOnly_ mode is set via editor config prior to initialization. It won't work if _readOnly_ mode is set dynamically (e.g. `CKEDITOR.instances.editor1.setReadOnly( true );`). Proper way to check if editor is in _readOnly_ mode should be `if ( editor.readOnly )`.
2. Keep the code style consistent with rest of the code. Line mentioned above should look like `if ( editor.config.readOnly ) return false;`. The easiest way to do this is to use **.jscsrc** and **.jshintrc** available in the repo (many IDEs/editors have integration out of the box).
3. There is no need for creating separate test suit for this ticket (and duplicating _widgetTestsTools_). Unit tests could be integrated into existing widget tests (`/tests/plugins/widget/`).
4. It will be nice if you could also provide some manual tests (it's best to cover cases of normal and inline editor and also static/dynamic setting of _readOnly_ mode).

Thanks again for the PR and we will be glad to review it further after you provide changes mentioned here.

mlewand:
Closing the PR since there is no feedback from initial contributor. @Duskfall feel free to reopen it as you'll be able to provide the code requested by reviewer.

PR #263, opened 2016-04-06
Add support for FLV videos

AngelFQC:


ywarnier:
:+1: 
At the moment there's no way to add an FLV video, but there's nothing else to be done than to accept the extension in order to add support for them...

mlewand:
Hi @AngelFQC sorry for the late reply.

Thanks for the contribution, however here we're not interested in providing any particular player with this plugin. It's intent is limited to inserting flash files, and only that.

You use case sounds suitable for a separate plugin and this make a lot of sense. You could publish your modified version in our [addons repository](http://ckeditor.com/addons/plugins) so that community might benefit from your work.

Thanks!

PR #264, opened 2016-04-15
Removed an null pointer exception we were observing

vortex-5:
TypeError cannot read property getRanges of null

The above error message was observed from our customers we have in our code base patched the cause of the null pointer and are no longer seeing customer errors.

### DIFF
--- a/plugins/clipboard/plugin.js
+++ b/plugins/clipboard/plugin.js
@@ -1087,8 +1087,11 @@
 				return CKEDITOR.TRISTATE_OFF;
 
 			// Cut, copy - check if the selection is not empty.
-			var sel = editor.getSelection(),
-				ranges = sel.getRanges(),
+			var sel = editor.getSelection();
+			if (!sel)
+				return CKEDITOR.TRISTATE_DISABLED;
+			
+			var	ranges = sel.getRanges(),
 				selectionIsEmpty = sel.getType() == CKEDITOR.SELECTION_NONE || ( ranges.length == 1 && ranges[ 0 ].collapsed );
 
 			return selectionIsEmpty ? CKEDITOR.TRISTATE_DISABLED : CKEDITOR.TRISTATE_OFF;


mlewand:
Could you provide any Test Case for it?

mlewand:
Also note that his code is missing tests, see our [Contributing Code](http://docs.ckeditor.com/#!/guide/dev_contributing_code) guide for more details.

mlewand:
Closing the PR since there is no feedback from initial contributor. @vortex-5 feel free to reopen it as you'll be able to provide the code requested by reviewer.

adrienverge:
I can confirm this bug happens to me too, and my webapp crashes at this exact point (`ranges = sel.getRanges()`).

I tried hard to find the cause and recreate the bug on demand, but couldn't. It happens very randomly, looks like a rare race condition. Unfortunately, without being able to trigger it, I can't write tests for this neither.

@mlewand is there any way to help trigger this bug? If not, is it possible to merge this since it is a pretty obvious PR?

adrienverge:
@mlewand I understand the need for a good coverage and prevent regressions. I'm always for adding tests when fixing a bug.

I spent hours trying to programmatically trigger the bug, but couldn't. It looks like other tried too. The bug happens very sporadically.

All I can say now is that the fix from @vortex-5 **really** stops it from happening. (And while reading the change in JS, it can make sense.)

theNailz:
It seems to happen when we dispose the first CKEditor and initalize another editor (called from a setTimeout with 0 delay). Maybe this helps you pinpoint this issue; I can reproduce it and provide a stack trace but I'd need some pointers on what you'd like to see..

mlewand:
@theNailz acutally the stack trace would be very helpful, it would show us what is the path that leads to this exception.

Please include also evaluated value of `CKEDITOR.revision` and `CKEDITOR.version` so that we can use same CKE version.

theNailz:
I have used a fresh copy of CKEditor from the [Builder](https://ckeditor.com/cke4/builder) to have a non-minified source. (Source: Big N'slow) So these values are placeholders:
```
CKEDITOR.revision: "%REV%"
CKEDITOR.version: "%VERSION%"
```

Stack:
```
stateFromNamedCommand (plugin.js:1233)
setToolbarStates (plugin.js:1218)
(anonymous) (plugin.js:742)
setTimeout (async)
(anonymous) (plugin.js:741)
listenerFirer (event.js:144)
CKEDITOR.event.fire (event.js:290)
(anonymous) (domobject.js:46)
```
`plugin.js`'s full path is `vendor/ckeditor/plugins/clipboard/plugin.js`
`event.js`'s full path is `vendor/ckeditor/core/event.js`
`domobject.js`'s full path is `vendor/ckeditor/core/dom/domobject.js`

PR #265, opened 2016-04-15
Add field to edit Link Display Text

ryanguill:
ticket #7154
Check if more than text items are selected and if so disable the input.
Added a test that is currently failing ("test edit link text") , as well as my changes causing another test to fail ("test edit link on a non-editable inline element")


ryanguill:
This is a continuation of https://github.com/ckeditor/ckeditor-dev/pull/238

The changes resolve our manual tests, but there are issues with the unit tests that I need help resolving.

In the test that I have created, I expect that when the link dialog is opened that I am attempting to edit the existing selected link, but from debugging it seems to be opening a new link dialog as if I am attempting to create a new link. https://github.com/ckeditor/ckeditor-dev/pull/265/commits/bd916aa8010f7bc2fbe93d9bd96cc2a6aba19546#diff-c948505c5dc6c6c0e745e42302538929R172  I am sure this is due to my lack of understanding around how the bot works and its API.

I am unsure what existing test that is now failing ("test edit link on a non-editable inline element") is attempting to test.  If I go to the ckeditor instance on the ckeditor website demo page and insert the element in that test and attempt to edit the link in any way the non-contenteditable section is removed.  But it is certainly a regression that I am introducing, the test passes without my changes.

mlewand:
I did not review it thoroughly, but couple things I can tell already:

1. To your problem:
	> //I expect at this point that it would be as though I selected the link and clicked the link icon, is that correct?
	> //but when debugging it looks as though I am trying to create a new link, not edit the existing link.
	
	Your debugging shown you correctly that in fact you don't have the link selected. :) You could take a look on other tests, e.g. [`test edit link (text selected)`](https://github.com/ryanguill/ckeditor-dev/blob/bd916aa8010f7bc2fbe93d9bd96cc2a6aba19546/tests/plugins/link/link.js#L70-L81), you can see how we handle selection there.
	
	But the best way would be simply to check methods used by `bot.setHtmlWithSelection`, which would led you to [`bender.tools.range.setWithHtml`](https://github.com/ryanguill/ckeditor-dev/blob/t/7154/tests/_benderjs/ckeditor/static/tools.js#L1086-L1107) which docs will explain you what's going on.
	
1. In test "test edit link text" you could also test initial value of the `linkDisplayText` dialog input (it should be the URL).
1. There is no test for a case when the `linkDisplayText` is set to disabled, and `lang.linkTextFromSelection` is used as it's value.

Feel free to rebase your own commits, to remove old comment. Thanks for the contribution so far!

ryanguill:
I thought it best to learn how to write a successful test before writing all tests.

I have resolved the issue with my test, added a test for the initial value of the `linkDisplayText` and added another test for the disabled state and the inputs value from the language file. I have also attempted to add a manual test covering the variants.

I still have the issue with the test [test edit link on a non-editable inline element
](https://github.com/ryanguill/ckeditor-dev/blob/bd916aa8010f7bc2fbe93d9bd96cc2a6aba19546/tests/plugins/link/link.js#L126-L138) that I am unsure of what the issue is.

ryanguill:
Wanted to follow up on this to see if there is anything else I can do / provide to keep this PR progressing.

Tade0:
I ran this test step-by-step and noticed, that the link text field is editable, which means that `<em contenteditable="false">bar</em>` is recognized as editable text and eventually replaced here: https://github.com/ckeditor/ckeditor-dev/pull/265/commits/bd916aa8010f7bc2fbe93d9bd96cc2a6aba19546#diff-ec6fc3b3d4ed2c940ae53858f5d2a2acR886

I think the desired behaviour here would be to disable the link text field if such an element is a child of the link.

ryanguill:
Is there a better way to handle that than just doing a search for `contenteditable="false"` in the content? I searched through the plugins/ directory for an example of how it might be handled somewhere else but wasn't able to find any examples. 

Also, how does the `contenteditable="false"` get onto the element in the editor? trying to reproduce the test manually by going to source and putting that content in there with the `contenteditable` attribute and then clicking source again to go back to the main editor - if you click source again the attribute was stripped...

ghidinelli:
@Tade0, @mlewand - thanks for helping @ryanguill make progress here. We're going to be in crunch time on our end soon and I'd like to see this get committed before we get pulled in another direction. This request sat for 5+ years once, I don't want to have it happen again! :)  I am certain you guys are plenty busy as well - would you have a few moments to help get us to the finish line?

Tade0:
Thank you for your patience.

> Also, how does the contenteditable="false" get onto the element in the editor? trying to reproduce the test manually by going to source and putting that content in there with the contenteditable attribute and then clicking source again to go back to the main editor - if you click source again the attribute was stripped...

That's probably the ACF doing that. If you add `config.allowedContent = true;` then it should not disappear anymore.

ryanguill:
The tests are now passing, but im still not certain that what I have done is the proper way to resolve that problem with the contenteditable=false.  Would be interested in your reviews @Tade0, @mlewand

ghidinelli:
Looks like the tests are passing now @Tade0 - can this be merged in and released in a ~4.5.9 update?

Tade0:
Indeed, @mlewand would you like to take a last look on this PR?

@ryanguill There's no obvious alternative to your method and also I was unable to break anything after your patch, so in my opinion it meets the requirements for a proper way to implement this feature.

mlewand:
I'll do my best to have a look on it this week. Thanks to all for the work so far!

mlewand:
Ok, so I had a chance to do more a review, and here's what i found:

1. We should also retreive text when a ranged selection is made.

	You could utilize [getSelectedText](http://docs.ckeditor.com/#!/api/CKEDITOR.dom.selection-method-getSelectedText) text.
	
1. We definitely need to support ranged range too. The way we see it working is:

	* Collapsed selection works the way you did it, that's totally fine.
	* Ranged selection might even remove existing markup, but then apply a link with inner text. Let's see following TCs:
		
		Let's assume that we're inserting a link with url "xyz" and text value "test", and selection is marked with square brackets.
	
		* Contained elements:
			
			```
			<p>f[oo <strong>bar</strong> ba]z</p>
			```
			
			Goes into:
			
			```
			<p>f<a href="xyz">text</a></p>z</p>
			```
	
		* Multiple elements:
		
			```
			<p>f[oo</p>
			<p>bar</p>
			<p>ba]z</p>
			```
			
			Goes into:
			
			```
			<p>f<a href="xyz">text</a></p>
			<p>z</p>
			```
		
		**important remark:** markup inside selection should get overriden **only** if the text value field changed. Otherwise inner markup should not be removed.


> We're going to be in crunch time on our end soon and I'd like to see this get committed before (...)

We have similar thing going on. :) We also would like to see this feature in CKE4. Feel free to update the branch, as we find some spare time on our end, we also will put some code to adjust this branch to what we are looking for. Hopefully we'll be able to add it in upcoming 4.5.10.

ryanguill:
in the examples you gave, how would I go about adding the closing tag (paragraph in this case) after the link? And is that for certain correct, ie, the opening and closing tags are off in this line: `<p>f<a href="xyz">text</a></p>z</p>`

But even in the second example, how would I go about closing the p tag and opening a new one for z?

ryanguill:
also, is there a built in way to know if the value of the field has been modified, or do I need to write something to track that myself?

ghidinelli:
@mlewand - any chance we could get some feedback on the above Qs?

mlewand:
@ryanguill @ghidinelli sorry for late response, but we have a lot of work recently. We started working on this particular enhancement couple days ago, so that it will be included in upcoming 4.5.11 version.

> also, is there a built in way to know if the value of the field has been modified, or do I need to write something to track that myself?

Unfortunately our dialog API doesn't provide it out of the box and that's something to be tracked by the dev for the time being.

> in the examples you gave, how would I go about adding the closing tag (paragraph in this case) after the link? And is that for certain correct, ie, the opening and closing tags are off in this line: <p>f<a href="xyz">text</a></p>z</p>
> 
> But even in the second example, how would I go about closing the p tag and opening a new one for z?

Actually we've been talking here at CKSource about this aspect, and decided to go slightly different way. Essentially paragraphs will be merged.

With that being said, please hold on with developement, as we decided to dedicate to this issue. You can follow the progess in [#7154](http://dev.ckeditor.com/ticket/7154) issue.

mlewand:
All right, good news: the feature made it's way into upcoming 4.5.11 release! It will be out this week.

Thank you all for the contribution!

PR #267, opened 2016-04-24
Correct changelog method reference

mattleff:
Per https://github.com/ckeditor/ckeditor-dev/commit/8470d94 this is the correct name.


mlewand:
Good catch, thanks! Merged to upcoming major release with 056ee7706e.

PR #269, opened 2016-04-28
Fix #14590

dpidcock:
Only output the extra line break if the tag that was closed is breakAfterClose


mlewand:
This issue has been merged to 4.5.10 released. Thank you for the contribution!

PR #270, opened 2016-04-29
Changed tables to be HTML5 compliant

wporter82:
We have a requirement to use tables and found that css libraries typically will reset the td tag to no padding. In order to make this possible, I had to update the tables plugin to use HTML5 compliant code, ie: styles instead of attributes.

When editing an existing table with this update, it will convert it if table properties are viewed. This should keep from breaking sites that upgrade.


mlewand:
Hi @wporter82, thanks for the contribution.

Could you rebase your code onto the latest major / master branch? This PR contains a bit of noise and we'd like it to be sorted out so that we can review it - thanks!

mlewand:

PR #273, opened 2016-05-04
[#14613] prevent plugin initialization if editor was already destroyed…

brianmay27:
…d (race condition)

### DIFF
--- a/core/editor.js
+++ b/core/editor.js
@@ -591,6 +591,10 @@
 
 			// Load all plugin specific language files in a row.
 			CKEDITOR.scriptLoader.load( languageFiles, function() {
+				//Ensure the editor has not been destroyed in the mean time
+				if (editor.status === 'destroyed') {
+					return;
+				}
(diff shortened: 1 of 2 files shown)


brianmay27:
I tried to figure out how to functional test this. I was able to reproduce this and catch the error with a event listener but could not figure out how to wait a few seconds for the call back to be called. Im happy to finish the test but need a little help :)

brianmay27:
Any updates? We would like to get this in as we have been seeing issues around this race issue and would like to get it merged.

mlewand:
Hi @brianmay27 - sorry for the delay, we've been pretty for a long time. We should be able to review this PR soon.

mlewand:
I see that we need manual / unit tests - but the problem is pretty clear, so info for the reviewer: before bouncing the ticket, please give a try to reproduce it on your own, if no problem with this just create mt, and unit tests - as it should be fairly simple.

PR #275, opened 2016-05-27
When widget doesn't get removed from content

toscanov:
Odd bug I ran across when the content contained partial widget data upon loading.   The system would throw an error due to widgetElement being null.  On this rare occasion this code will remove the widget wrapper allowing the code to function as expected.


f1ames:
Thanks for the PR @toscanov.

This pull request is missing tests, see our [Contributing Code](http://docs.ckeditor.com/#!/guide/dev_contributing_code) guide for more details. Please, provide a manual/unit test for this, so that we can verify the issue.

If you could also provided more information on the issue or steps/scenario how to reproduce it, it will be easier to understand and proceed with this PR for us.

mlewand:

PR #276, opened 2016-06-07
[#14291] Fix selection of `required` checkbox in forms plugin

rzueger:
- according to HTML spec, valid values for the attribute are:
  - no value at all
  - empty string (required="")
  - string `required` (required="required")
- with the current implementation, the checkbox is checked,
  if the attribute contains any value
  -> if there is no value or an empty string, the checkbox
     is not selected, which is wrong
  -> if there is a value other than "required", the checkbox
     is selected, which is wrong
- see ticket #14291

### DIFF
--- a/plugins/forms/dialogs/checkbox.js
+++ b/plugins/forms/dialogs/checkbox.js
@@ -131,7 +131,8 @@ CKEDITOR.dialog.add( 'checkbox', function( editor ) {
 				accessKey: 'Q',
 				value: 'required',
 				setup: function( element ) {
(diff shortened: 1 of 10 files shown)


rzueger:
@ckdevs: Do you plan to include this patch in your next release? Or do you require any changes to the proposed pull request?

uwolfer:
Do you plan to include this bugfix soon?

Tade0:
Hello,

thank you for your patience and for the pull request. There is a status update available at:
http://dev.ckeditor.com/ticket/14291

jacekbogdanski:

PR #278, opened 2016-06-20
IE Performance issues

toscanov:
Seeings delay on typing when using autogrow, each keypress was showing 50-120ms.  Keeping the timeouts from occurring when a new one came in seems to avoid the delay.


Tade0:
Thank you for the PR!

Overall the change is justified, but I see one problem with it: If the user holds down a key(e.g. `enter`) the editor doesn't grow until the user releases the key.

Could you provide some tests for this?

Guide just in case: http://docs.ckeditor.com/#!/guide/dev_contributing_code

mlewand:

PR #279, opened 2016-06-24
#14761 issue - add `startDisabled: 1` in createDef() function

likemusic:
Need if add table-context menu commands to toolbar.


mlewand:
This pull request is missing tests, see our [Contributing Code](http://docs.ckeditor.com/#!/guide/dev_contributing_code) guide for more details.

likemusic:
I think, that I can't write test, because this bug happend when toolbar contain buttons for table-related commands. This buttons provides by my [tabletoolstoolbar plugin](https://github.com/likemusic/tabletoolstoolbar).

Those, the table-related buttons by default must be inactive and be accessible only when the cursor is moved inside table.

mlewand:
Hi @likemusic, thanks for getting back on this. Having a need to add some extra code to the manual test in order to reproduce the issue is fine.

You can see an example [in autogrow manual test](https://github.com/ckeditor/ckeditor-dev/blob/master/tests/plugins/autogrow/manual/14413.html), you'll see that there's added a custom handler for dialog definition.

In the same manner you could create a manual test where you add a required command, button and anything that is necessary to reproduce the issue. What is important though is to do it with bare minimal code required to repro it, just to keep it simple.

If you have more questions just let us know!

likemusic:
@mlewand, thanks for support!
Currently I merge my #patch-1 and #patch-2 branches with the same changes for `table` and `tabletools` plugins.

I create small manual test, with button "Delete table" on the toolbar, that must be inactive on startup, and be available only when cursor inside table. And merge them too.

likemusic:
In [contribution guide](http://docs.ckeditor.com/#!/guide/dev_contributing_code) described where place test for tickets in  http://dev.ckeditor.com/, but not for pull requests. Currenctly I place test in test/tickets/279, but I think that this is most likely wrong.

I need create ticket on [Issues Tracker Site](https://dev.ckeditor.com/newticket) before make pull-request (as described in [docs](http://docs.ckeditor.com/#!/guide/dev_issues_tracker)  and move test to them number? All right?

likemusic:
And I think that its must be auto-test, not manual. All right?

mlewand:
> I need create ticket on Issues Tracker Site before make pull-request (as described in docs and move test to them number? All right?

That's absolutely correct, then you'll get a proper number that won't collide with anything in the future.

As for test type, please add one manual test so that anyone can reproduce your issue manually on their own.

As for auto/unit test: seeing that you're just adding an extra property to the command object, it would be fine for you just to add a test case where it checks whether these commands do contain `startDisabled` property equal to `1` - and that's all.

likemusic:
As I see, unfortunely, `toolbar` plugin does not provide api for validating current button state?
If I'm right, мay be would right create issue with this future request?

likemusic:
I mean that validate existing `startDisabled` attribute for command is senselessly afte write this in source code. Therefore I have not yet added it.

f1ames:
Issue link - https://dev.ckeditor.com/ticket/14761.

jacekbogdanski:
Unfortunately, we were unable to review your PR recently. CKEditor 4 codebase rapidly changes due to the active development, so it's unlikely this PR can be still applied without additional adjustments.

PR #280, opened 2016-06-24
add `startDisabled: 1` in createDef() function

likemusic:
Need if add table's context menu commands as toolbar buttons.


mlewand:
It's a duplicate of #279.

PR #281, opened 2016-06-27
handle some <figure> edge cases

liias:
fix error if `<figure class="image">` has no `` or `<a>` element as child


Comandeer:
The fix looks good, but it lacks tests. Please add them in `tests/plugins/image2`. The manual test is optional, but it would be nice to have one.

liias:
Will add a test later on.

For us the problem appeared when user copy-pasted existing ckeditor generated content inside another instance - and probably image was removed for some reason (could have been on our server-side part).

mlewand:

PR #284, opened 2016-07-11
Undo generate error.

devsseb:
If Ctrl-Z pressed, toBeDowncasted is empty and generate an js error.


f1ames:
Thanks for the PR @devsseb!

If you could also provide some description of your use case or steps to reproduce this issue. It will help us to better understand the issue and your approach to fix it.

mlewand:

PR #285, opened 2016-07-19
Colorbutton plugin improvements.

f1ames:
Changes directly related to [14569](http://dev.ckeditor.com/ticket/14569) ticket. This PR changes the default `colorbutton` palette, adds `colorButton_colorsPerRow` config option which makes number of colors per row in color selector configurable. It also changes `Automatic` button html, so the text can be easily centered with CSS.


mlewand:
Issue #14569 has been fixed.

PR #286, opened 2016-07-22
Do not rely on navigator.product to detect UA

KrzysiekJ:
This property has been removed from the web standards. Besides that, generally all
browsers set it to Gecko, so it is meaningless.

Refs #206.


f1ames:
Hi @KrzysiekJ,

PR #287, opened 2016-08-04
improve documentation for menu plugin addMenuItems method - 

jbrekle:
it expects an Object, not an array


AnnaTomanek:
Hi Jonas, you're totally right - thanks for spotting this. This is now merged and will be updated in the online docs with the next minor CKEditor version.

Thank you!

PR #288, opened 2016-08-16
appendBogus function adds a attribute 'data-cke-bogus' 

jbrekle:
I changed the appendBogus function to add a attribute 'data-cke-bogus', that allows to seperate bogus elements from user-edited breaks.

I need this for our companies use-case, where we need to keep the HTML untouched. It messes with change detection and some legacy systems that are upset with tailing br's in a li.
 
I know there is 
http://dev.ckeditor.com/ticket/5767 and 
http://dev.ckeditor.com/ticket/6133
but I am not happy with it. I don't want the bogus br's, but there probably is a good reason for it, but in my case I need to strip them out on save. So I added an special attribute to distiguish them.
What do you think?

### DIFF
--- a/core/dom/element.js
+++ b/core/dom/element.js
@@ -335,6 +335,7 @@ CKEDITOR.dom.element.clearMarkers = function( database, element, removeFromDatab
 				var bogus = this.getDocument().createElement( 'br' );
 
 				CKEDITOR.env.gecko && bogus.setAttribute( 'type', '_moz' );
+				bogus.setAttribute( 'data-cke-bogus', 'true' );
 
 				this.append( bogus );
 			}


jbrekle:
@Tade0 @mlewand can you give some feedback, please?

mlewand:
Sorry, we're having lot of work recently.

Could you provide a little more information on why bogus is problem in your case? What's the reason for working directly on editable?

jbrekle:
I have the requirement to keep HTML as untouched as possible. The HTML is converted to some proprietary RichText format and then processed by other systems. Any changes that are not made by the user and maybe for making it look good in the browser might have unwanted results. 

There is also this "filling empty p's with a br - i prevented this with adding a data-cke-processor=off in the input. 
And I still have an issue with CKEditor trimming trailing and double whitespace...

What do you mean by "working directly on editable"?

mlewand:
Ok, I see. What is the way you get content from editor? Is it simply `editor.getData()`?

> What do you mean by "working directly on editable"?

I've meant working on element returned by `editor.editable()` which is essentially `body` of an `iframe` in case of classic editor, where the (dirty) edited content is.

jbrekle:
It tried both ways.

jbrekle:
For my trimming whitespace issue, I made a stackoverflow question 
http://stackoverflow.com/questions/39493449/ckeditor-trims-whitespace-in-li

mlewand:
Hi again @jbrekle. I did an investigation on what was the reason for not adding an attribute to bogus `br`s.

So the thing is that browsers also put bogus `br`s on their own. Firefox for instance will put br into `body[contenteditable=true]` whet it gets focus (so that the caret might be rendered).

We need to work with bogus `br`s coming from all sources, so also the browser. So this is why we're not using any particular attribute, and we need to keep it that way.

Going back to your problem you can always use [getBogus](http://docs.ckeditor.com/#!/api/CKEDITOR.dom.element-method-getBogus) method to figure out if there's a bogus `br`. Hope it helps!

PR #289, opened 2016-08-22
Added config to inlineAll

roselan:
Adds the option to use a custom config without config.js and CKEDITOR.editorConfig.

### DIFF
--- a/core/creators/inline.js
+++ b/core/creators/inline.js
@@ -117,7 +117,7 @@
 	 * the `contenteditable` attribute set to `true`.
 	 *
 	 */
-	CKEDITOR.inlineAll = function() {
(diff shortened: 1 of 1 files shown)


mlewand:
Hi @roselan it looks like a good idea to allow custom config here, however this pull request is missing tests, see our [Contributing Code](http://docs.ckeditor.com/#!/guide/dev_contributing_code) guide for more details.

Could you please add unit tests checking this enhancement?

roselan:
Thanx, I will as soon as I have the time (and we end up choosing cke)

roselan:
We ended up choosing another solution, as it suits our needs better.

PR #291, opened 2016-09-12
Fix IE 11 error

GabrielDumbrava:
Fixes an error I get when using IE11 in Windows 10, using Emulation. I don't know if it runs in the same situation without being run in emulation.

The error is:
"unable to get property '1' of undefined or null reference"

The agent does not contain the string 'msie' so the returned array is null.

### DIFF
--- a/core/env.js
+++ b/core/env.js
@@ -193,7 +193,12 @@ if ( !CKEDITOR.env ) {
 			if ( edge ) {
 				version = parseFloat( edge[ 1 ] );
 			} else if ( env.quirks || !document.documentMode ) {
-				version = parseFloat( agent.match( /msie (\d+)/ )[ 1 ] );
+				// for IE 11: "mozilla/5.0 (windows nt 6.3; trident/7.0; rv:11.0) like gecko"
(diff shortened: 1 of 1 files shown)


Reinmar:
Hey, thanks for the PR.

Do you use the emulation modes available in the Edge dev tools? They were never supported as they were never accurate. IEs working in those modes were even weirder than normally – something in between real versions.

I guess that the PR may be accepted by the CKE4 core team, but it may happen (hopefully not) that there will be other issues (and the core team won't be interested in working on them I'm afraid).

GabrielDumbrava:
Yeap. I was using the Edge dev tools and I know they are not accurate, but clients use it to demonstrate that the app is not working in IE11, so it might help.

mlewand:
Hi, during all this time we didn't notice any signals for supporting IE in emulation mode other than this PR. Adding a new browser is a huge undertaking for us, because we need to test it out with the entire editor and then test it with each release 🙂 But in case you feel this is important feel free to report a feature request for that kind of support.

PR #292, opened 2016-09-20
- removed percent only checks for size tag

daxborges:


plugins/bbcode/plugin.js
```diff
@@ -137,10 +137,6 @@
 						}
 
 						if ( stylesMap[ part ] ) {
-							// Font size represents percentage.
-							if ( part == 'size' )
```
daxborges: this just always adds % onto whatever the value of the tag would have been

plugins/bbcode/plugin.js
```diff
@@ -660,11 +656,7 @@
 								tagName = 'color';
 								value = CKEDITOR.tools.convertRgbToHex( value );
 							} else if ( ( value = style[ 'font-size' ] ) ) {
-								var percentValue = value.match( /(\d+)%$/ );
```
daxborges: this looks for a digit with a percent and then sets the value and tag name. Instead we just set the tag name and trust the values we setup for our editor.

mlewand:
Hi @daxborges I'd like to ask for more details for this PR. What exactly it's fixing, and what's the use case for it.

Also I can see that this PR is missing tests, see our [Contributing Code](http://docs.ckeditor.com/#!/guide/dev_contributing_code) guide for more details.

daxborges:
@mlewand this was meant to be a PR against our fork of the repo. Apologies.

PR #295, opened 2016-09-26
Add `image2_altRequired` option to `image2` config

andreyfedoseev:
Optionally make the "Alternative Text" field in `image2` plugin required.

### DIFF
--- a/plugins/image2/plugin.js
+++ b/plugins/image2/plugin.js
@@ -1440,7 +1440,7 @@
 						displayTextField.hide();
 					} else {
 						// Make sure that display text is visible, as it might be hidden by image2 integration
-						// before. 
+						// before.
 						displayTextField.show();
 						onShow.apply( this, arguments );
 					}
--- a/plugins/image2/plugin.js
+++ b/plugins/image2/plugin.js
@@ -1697,3 +1697,13 @@ CKEDITOR.config.image2_captionedClass = 'image';
  * @cfg {String[]} [image2_alignClasses=null]
  * @member CKEDITOR.config
  */
+
+/**
+ * Determines whether alternative text is required.
+ *
+ *		config.image2_altRequired = false;
+ *
+ * @since 4.6.0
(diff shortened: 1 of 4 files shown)


plugins/image2/plugin.js
```diff
@@ -1697,3 +1697,13 @@ CKEDITOR.config.image2_captionedClass = 'image';
  * @cfg {String[]} [image2_alignClasses=null]
  * @member CKEDITOR.config
  */
+
+/**
+ * Determines whether alternative text is required.
+ *
+ *		config.image2_altRequired = false;
+ *
+ * @since 4.5
```
mlewand: Please target it for 4.6.0 instead (it's important to use 3 digs, as we're using semver).

plugins/image2/plugin.js
```diff
@@ -1697,3 +1697,13 @@ CKEDITOR.config.image2_captionedClass = 'image';
...
+ * @cfg {Boolean} [image2_prefillDimensions=false]
```
mlewand: Copy'n'pasta left-over here :) instead `image2_prefillDimensions` we want to use `image2_altRequired`.

andreyfedoseev: @mlewand Thank you for the feedback, I'll update the PR shortly.

andreyfedoseev:
@mlewand I updated the pull requests with tests. I also had to fix a couple of unrelated jshint warnings. Please review.

mlewand:
Good job, I've created a ticket [14889](http://dev.ckeditor.com/ticket/14889).

You made a good job with adjusting code style / creating unit tests, so thanks for quality pull request!

Still there were couple of things to be done, like adding manual test, lang entry meta information for our localization contribution or adding lang entries for other langs - but I did it on our own here.

I've also rebased your branch on top of `major` (CKEditor 4.6.0), since we're done with `4.5.x` line.

To sum it up, your contribution will appear in **4.6.0** version, thank you very much!

PR #296, opened 2016-09-29
Doc: Fix `config.toolbar` documentation missing '/'

adrienverge:
Update the documentation to show the usage of `'/'` to produce toolbar lines.

Example:

```js
config.toolbar = [
    [ 'Source', '-', 'Bold', 'Italic' ],
    ['Font', 'FontSize'],
    '/',
    ['NumberedList', 'BulletedList']
];
```

This feature can be seen directly in source: https://github.com/ckeditor/ckeditor-dev/blob/12f0de3/plugins/toolbar/plugin.js#L216-L221

### DIFF
--- a/plugins/toolbar/plugin.js
+++ b/plugins/toolbar/plugin.js
@@ -677,8 +677,9 @@ CKEDITOR.UI_SEPARATOR = 'separator';
 CKEDITOR.config.toolbarLocation = 'top';
 
 /**
- * The toolbox (alias toolbar) definition. It is a toolbar name or an array of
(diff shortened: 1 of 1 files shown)


mlewand:
@adrienverge Thank you for the contribution, however we don't see a need for adding this detail into config.toolbar configuration.

It's already mentioned in couple of places like config,toolbarGroups mentioned in config.toolbar itself. It's also generated by our [CKEditor Toolbar Configurator](http://ckeditor.com/latest/samples/toolbarconfigurator/index.html) which we're promoting for that purpose. So this information is available for any interested people, while w want to keep `config.toolbar` documentation simple.

PR #297, opened 2016-10-17
Support for "bodyClass" and "bodyId" settings in Styles dropdown

aik099:
Closes https://dev.ckeditor.com/ticket/7452

### DIFF
--- a/plugins/panel/plugin.js
+++ b/plugins/panel/plugin.js
@@ -75,7 +75,7 @@
 	var frameDocTpl = CKEDITOR.addTemplate( 'panel-frame-inner', '<!DOCTYPE html>' +
 		'<html class="cke_panel_container {env}" dir="{dir}" lang="{langCode}">' +
 			'<head>{css}</head>' +
-			'<body class="cke_{dir}"' +
+			'<body class="cke_{dir} {bodyClass}" {bodyIdAttr}' +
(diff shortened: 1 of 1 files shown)


aik099:
Any updates?

Tade0:
Thanks for the PR!

Would you like to help us further by including some tests(a manual test is fine in this case).

Here's a guide just in case: http://docs.ckeditor.com/#!/guide/dev_contributing_code

aik099:
I'm new to CKEditor. Is there any manual (I haven't found `CONTRIBUTING.md` file in the repo itself) about how to write tests?

aik099:
What you mean by manual test. Using provided link I've found explanation how to create tests using Bender, but not manual ones.

mlewand:
@aik099 Now I can see it's not exposed well enough, there's a paragraph:

> Some type of features or bugs cannot be tested automatically. In such cases, make sure to create a manual test (see this one for instance).

Which links to https://github.com/ckeditor/ckeditor-dev/tree/master/tests/tickets/12735 as an example. This helps us more quickly to check if the code actually fixes the problem for end user, and is also used by QA again during the final release.

I'll make sure that docs get updated there.

mlewand:
Hi, It's been a while since we last heard from you and the tests were not added. We're closing this PR for now. Still, feel free provide us requested feedback, so that we can reopen this PR.

PR #299, opened 2016-10-25
added <base> to panel to fix 404s on style sheets when baseHref is used

Avenyet:
Added base to panel to fix a problem when using baseHref and the style combo box

Bug is stylesheets will not use the baseHref and cause a 404

This does not contain the IE10 and earlier fix found at line 446 in plugins/wysiwygarea/plugin.js

This has been tested in
Firefox 47
IE11
Chrome 54


Tade0:
Thank you for the PR!

Would you like to include a manual test that would demonstrate the solution?

Here's a guide just in case: http://docs.ckeditor.com/#!/guide/dev_contributing_code

mlewand:

PR #304, opened 2016-11-21
Add support for HOME/END/ARROW keys on fake selection

millerdev:
I've found it really annoying to have keyboard navigation get stalled when a widget is (fake) selected. This adds support for more navigation and selection keys on fake selection.


millerdev:
Rebased on master (also 4.6.1).

jacekbogdanski:

PR #310, opened 2016-12-13
Basic integration with Travis CI

Comandeer:
This branch is based on current master with some "enhancements":

* there is separate Bender configuration for CI (`bender.ci.js`), which tweaks Bender's default especially for `bender run` command (e.g. disabling manual tests in Chrome and choosing Chrome as a default browser).
* the build is run on latest, stable Chrome on Node 4. Unfortunately Bender seems to be too old to run on newer Nodes without refactoring.
* Some tests had issues with paths to the assets (they were hardcoded and not based on `%BASE_PATH%` variable or paths were incorrectly normalised).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,6 +9,7 @@
 
 node_modules/**
 build/**
+coverage/**
 
 dev/builder/release/**
 dev/builder/ckbuilder/**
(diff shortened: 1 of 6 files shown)


f1ames:
Basically, I am quite impressed how little changes were needed for CI integration. I think separate bender config file for CI is also a good idea, because it enables us to add some custom configuration only for CI.

Good job!

PR #313, opened 2016-12-23
Fixed wrong paths in docs to addExternal

arogachev:
Comments don't match code examples.


mlewand:
LGTM, thanks for this catch @arogachev!

PR #315, opened 2017-01-11
Copy bender.ci.js to build folder.

f1ames:
The `bender.ci.js` is removed during build (https://github.com/ckeditor/ckeditor-presets/pull/5) so it needs to be copied to build folder in order to run tests.


mlewand:
Great, I can see now it's working, thanks!

PR #316, opened 2017-01-21
Fixed cloning equals nodes for correct editor.getSelection working (#16813 bug)

DmitryKrekota:
This is my solution to fix the bug: https://dev.ckeditor.com/ticket/16813

### DIFF
--- a/core/dom/range.js
+++ b/core/dom/range.js
@@ -204,10 +204,17 @@ CKEDITOR.dom.range = function( root ) {
 		// This allows us to not think about startNode == endNode case later on.
 		// We do that only when cloning, because in other cases we can safely split this text node
 		// and hence we can easily handle this case as many others.
-		if ( isClone && endNode.type == CKEDITOR.NODE_TEXT && startNode.equals( endNode ) ) {
-			startNode = range.document.createText( startNode.substring( startOffset, endOffset ) );
-			docFrag.append( startNode );
-			return;
+		if ( isClone && endNode.type == CKEDITOR.NODE_TEXT ) {
+			var checkEqualsNode = startNode;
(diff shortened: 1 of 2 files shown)


mlewand:
Thanks for the PR @DmitryKrekota!

DmitryKrekota:
I added test.
There were problems with 'bender server run' command becouse of out of date socket.io dependency in benderjs module. To resolve the problem I put latest version of socket.io to /node_modules/benderjs/node_modules/ instead of old socket.io. It seems that benderjs needs to update dependencies. Maybe it is because of my node version - 7.4.0 :)

mlewand:
@DmitryKrekota Thanks for tests! Let's make sure it gets reviewed.

DmitryKrekota:
@mlewand can you review and marge my PR?

mlewand:
Hi @DmitryKrekota, sorry for the time it took us to get back to you.

I have actually good news, the bug has been fixed in other issue some time ago, and is no longer reproducible with our latest release 🎉 Now for clarity I'll put repro steps and online demos that do not require zip downloads.

## Repro steps

1. Select first word ("foo") with a double click.
1. Click the styles dropdown.
1. Close it with `esc` key.
1. Copy the selected text with `ctrl/cmd+c`.
1. Paste the text with `ctrl/cmd+v`.

### Expected

Only "foo" is being pasted.

### Unexpected

"foo bar" is being pasted.

## Online demos

Here are two samples:

* [CKEditor 4.7.1](https://codepen.io/mlewand/pen/xJqzQq) - this is where the issue was last reproducible
* [CKEditor 4.7.2](https://codepen.io/mlewand/pen/ZjeodZ) - this is where the issue was first fixed

PR #318, opened 2017-02-07
Fire 'change' Event on File Upload (Plugin)

Elidrake:
Per [Documentation](https://docs.cksource.com/ckeditor_api/symbols/CKEDITOR.ui.dialog.uiElement.html#event:change), whenever there is a change to the ui, the event 'change' should be fired.  This is particularly useful for Angular applications, where the event is required after the file upload in order for Angular to know that a change to the text has been made.


Elidrake:
Just realized I was in the incorrect area of the Documentation, and that 'change' is not an official (seemingly) event in the editor.  Feel free to remove this.

PR #320, opened 2017-02-11
Introduced the basicEntityNbspOff configuration option to not to esca…

milossramek:
Hi,
there is a problem related to the config.basicEntities = true option. In our case we want only &gt; and &lt; (we use the < and > characters quite often in math, if they are not escaped, CKeditor or even a browser often deletes them as unknown tags), but do not want the &nbsp; entities - unfortunately, browsers usually enter a lot of &nbsp; entities instead of plain spaces. This is is very annoying. This problem has been discussed already at various places, even with suggested workarounds in a form of a special CKEditor plugin, the goal of which  was to clean up the text in postprocessing. 

In this pull request I propose to add a new configuration option config.basicEntityNbspOff. If true, no nbsps are added, if false, CKeditor behaves as before. Thus, no existing configurations are not broken.

Thank you in advance for considering this change.
Milos

### DIFF
--- a/plugins/entities/plugin.js
+++ b/plugins/entities/plugin.js
@@ -6,6 +6,7 @@
 ( function() {
 	// Basic HTML entities.
 	var htmlbase = 'nbsp,gt,lt,amp';
+	var htmlbasenonbsp = 'gt,lt,amp';
 
 	var entities =
 	// Latin-1 entities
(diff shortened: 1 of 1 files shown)


f1ames:
Hello @milossramek,
This pull request is missing tests, see our [Contributing Code](http://docs.ckeditor.com/#!/guide/dev_contributing_code) guide for more details. Thank you!

jacekbogdanski:
Hi, It's been a while since we last heard from you. We're closing this PR for now. Still, feel free to provide us requested test coverage so that we can reopen it.

PR #322, opened 2017-02-20
Add support for Firefox in Travis builds

mlewand:
This PR adds support for Firefox in our CI builds.

Note that for the time being I've added it into `allow_failures` matrix, which basically means that even though it's tests fail overall build will be marked as success. This is for transistional period, and to test the stability.

I already see [one uploadimage test](https://dev.ckeditor.com/ticket/16861) failing __most__ of the time.

Also note that I did remove `fast_finish` flag on purpose, so that Travis will wait with reporting job as done until Firefox is done (we still care about it).

Also used `addons: firefox "latest"` to make sure it's running latest Firefox version, as by default Travis uses Firefox 38 which is... kida outdated. I suppose it's due to the fact that it's the last version fully compatible with selenium, but that's a different story.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,6 +15,9 @@ matrix:
       env: BUILD=1 BROWSER=Firefox
   fast_finish: true
 
+addons:
+  firefox: "latest"
+
 before_script:
   - 'npm install benderjs-cli -g'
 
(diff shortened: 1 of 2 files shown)


mlewand:
I actually strongly believe there's a lot of sense to get it asap. First off we'll be able to tell if there are any more unstable tests.

Also we'll have FF automatically running which is a great addition, please give it another thought.

PR #325, opened 2017-03-16
Linked an example template file in API docs

mlewand:
As [one of our users points out](http://ckeditor.com/addon/templates#comment-3196671844) template docs could use some improvements.

In terms of API docs one thing missing for me is a better explanation on how the template files should look like, the best example is to point to working code.

### DIFF
--- a/plugins/templates/plugin.js
+++ b/plugins/templates/plugin.js
@@ -74,8 +74,8 @@
  *			'http://www.example.com/user_templates.js'
  *		];
  *
- * For an example template file see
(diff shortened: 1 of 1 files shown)


mlewand:
I've adjusted the link text.

NathanCH:
The documentation could be improved further. It says explicitly "It must match definitions loaded with the templates_files setting." But imminently below, in the documentation, the array for template_files does not match. This makes more sense:

```
config.templates = 'my_templates';
config.templates_files = [
    '/editor_templates/my_templates.js',
    'http://www.example.com/user_templates.js'
];
```

PR #326, opened 2017-03-17
#16906 Fix isArray function in core tools to recognize ES6 iterator protocol…

michaeldimoudis:
… [object Array Iterator]

When using ES6 features (ReactJS/lodash/etc) on your page, the CKEditor core tools `isArray` function does not know how to check the ES Iteration Protocol, as described here: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Iteration_protocols

This is a simple fix by also comparing to `[object Array Iterator]` in the `isArray` function in `tools.js`

### DIFF
--- a/core/tools.js
+++ b/core/tools.js
@@ -270,7 +270,7 @@
 		 * @returns {Boolean} `true` if the object is an Array, otherwise `false`.
 		 */
 		isArray: function( object ) {
-			return Object.prototype.toString.call( object ) == '[object Array]';
+			return Object.prototype.toString.call( object ) == '[object Array]' || Object.prototype.toString.call( object ) == '[object Array Iterator]';
 		},
 
 		/**
--- a/tests/core/tools/array.js
+++ b/tests/core/tools/array.js
@@ -80,6 +80,14 @@
 			assert.isFalse( isArray( { length: 0 } ), 'Case 3' );
 			assert.isFalse( isArray( 'asd' ), 'Case 4' );
 			assert.isTrue( isArray( [ 1, 2 ] ), 'Case 5' );
+
+			var numArrayEntries = [ 1, 5, 7 ].entries();
+			numArrayEntries.toString();           // "[object Array Iterator]"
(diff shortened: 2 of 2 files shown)


michaeldimoudis:
Hi @mlewand thanks for the quick reply. CKEditor was working well for us, until we added some ReactJS code on our page, and the initialization broke. We believe ReactJS and/or lodash, using ES6, is interfering with the initialization of CKEditor, due to an array in CKEditor getting converted to an iterator for some reason. I have added a ticket here with more information: https://dev.ckeditor.com/ticket/16906

The dynamically generated plugins <script> tag becomes malformed. Ie: 
```<script type="text/javascript" src="​https://mydomain/_js/Performance2/ckeditor-full.4.6.2/plugins/dialogui,dialog,about,a11yhelp,dialogadvtab,basicstyles,bidi,blockquote,clipboard,button,panelbutton,panel,floatpanel,colorbutton,colordialog,templates,menu,copyformatting,div,toolbar,enterkey,entities,popup,filebrowser,find,fakeobjects,flash,floatingspace,listblock,richcombo,font,forms,format,horizontalrule,htmlwriter,iframe,wysiwygarea,image,indent,indentblock,indentlist,smiley,justify,menubutton,language,link,list,magicline,maximize,newpage,pagebreak,pastetext,pastefromword,preview,print,removeformat,save,selectall,showblocks,showborders,sourcearea,specialchar,stylescombo,tab,table,undo,wsc/plugin.js?t=v0.01"></script>```

Do you think we should add an `isIterator ` function, and check against that too when building the above script tag?

mlewand:
Hi @michaeldimoudis,

> The dynamically generated plugins <script> tag becomes malformed.

Interesting. Could I ask you to expose minimal setup to reproduce it? Codepen, jsfiddle or a GitHub repo would do.

> Do you think we should add an isIterator function, and check against that too when building the above script tag?

Yes, that would be the correct way to handle it.

mlewand:

PR #327, opened 2017-03-17
Added missing type information

mlewand:
Few config options in template plugin docs are missing var type. This PR fixes it.


Comandeer:
Merged with dc32bb22dae868a2fc927614b46bf90b2eb08435.

PR #328, opened 2017-03-20
Workaround plugin for iOS scrolling bug, #16902.

m-turek:
This is a workaround for this issue: https://dev.ckeditor.com/ticket/16902
Weird bug reproducible only on iOS. It is 'fixed' by removing ranges from selection in `input` event and restoring it shortly after (see comments in the code for more info)

There are no unit test, I tried simulating this bug using `execCommand('insertText')`, however it doesn't reproduce that way. Only manual test is added.


f1ames:
We should see if the initial issue still occurs and get back to this PR if needed.

PR #330, opened 2017-03-23
T/tp 2096: refactored clipboard without dialog

Comandeer:
This PR introduces clipboard plugin with paste dialog replaced by showing notification.


mlewand:
We also need a dedicated message for paste as plain text (ctrl+shift+v), saying that the following paste action will be pasted as plain text.

mlewand:
@f1ames feel free to review whenever you'll have the time to do that.

mlewand:
I also pushed two commits with some cleanup.

mlewand:
I checked the changes. There was only one major missing thing: the paste as plain text was not extracted to pastetext.

I refactored paste command, so it's interface is simpler.

Other than that tere were minor adjustments:
* added meta entries, I guess we have now all we need
* added api docs for pastetext command
* removed leftovers after old lang entries
* adjusted lang messages
* typo fixes

Please review my additions.

Comandeer:
Actually I don't like your change to `pastetext` plugin.

We have three commands connected with pasting:
* `paste` – the main one with the whole pasting logic,
* `pastefromword` – which just exec `paste` command with proper attributes (especially `type`),
* `pastefromtext` – which exec `paste` command, but has its own dedicated logic just to handle notification…

Ideally `pastefromtext` command should be as simple as `pastefromword` one, so no logic for handling showing notification. Maybe something like:

```javascript
exec: function( editor, data ) {
	var lang = editor.lang,
		keyInfo = CKEDITOR.tools.keystrokeToString( lang.common.keyboard, editor.getCommandKeystroke( this ) ),
		showNotification = ( data && typeof data.showNotification !== 'undefined' ) ? data.showNotification :
			!data || !data.from || ( data.from === 'keystrokeHandler' && CKEDITOR.env.ie ),
		msg = lang.pastetext.pasteMsg
			.replace( /%1/, '<kbd aria-label="' + keyInfo.aria + '">' + keyInfo.display + '</kbd>' );

	editor.execCommand( 'paste', {
		type: 'text',
		showNotification: showNotification,
		msg: msg
	} );
}
```

In such way we'll keep whole logic inside `paste` command and keep ability to customise notification content without cumbersome event listeners or `data.description`.

mlewand:
@Comandeer it actually considerably simplifies the process. The problem with former solution was that it introduced multiple stray configuration options, while it should be as simple as saying "I"m going to take care of the notification by myself".

As a result not only API suffered, but it also [caused the message to be overly complex](https://github.com/cksource/ckeditor-dev/blob/69440f9acded4feb6b8ba70ad855793a385ca7dd/plugins/clipboard/lang/en.js#L11). Try to put yourself in a translator shoes. Translators had to anticipate external sentence part at given location. Instead now they have two clear messages, where they hold full control over the message.

As such this change is a major simplification.

Comandeer:
Ok, but with the proposed solution with passing `msg` option we still have two totally separate messages and the logic connected with displaying notification is in one place – therefore `pastetext` command becomes even simpler.

Your solution somehow undermines the sense of having `showNotification` option as you force it to be `false` just to create your own notification handling logic… while you can just pass the real value to the `showNotifcation` option of `paste` command.

mlewand:
> Your solution somehow undermines the sense of having showNotification option as you force it to be false just to create your own notification handling logic… while you can just pass the real value to the showNotifcation option of paste command.

I was considering alternatively passing a string into `showNotification` to allow full message customization but it seemed to be messy. However by changing this property to `data.notification` which is more generic name seems that it could do the trick, so that `notification` could be:
* String - if so, it overrides the default message.
* Bool - if false prevents the notification from showing.

Comandeer:
`notification` seems to be OK. I'll try to implement it then.

Comandeer:
I've added `notification` option to `paste`, `pastetext` and `pastefromword` commands. I've also refactored heavily tests, moving all common things to helpers.

PR #331, opened 2017-03-24
replace arguments.callee with named function expression

IgorRubinovich:
This is a trivial change that prevents the colorbutton plugin from breaking when deployed as part of a browserify bundle. Namely Chrome throws:

    account.html:22179` Uncaught TypeError: 'caller', 'callee', and 'arguments' properties may not be accessed on strict mode functions or the arguments objects for calls to them

Since ES5 when in strict mode the deprecated arguments.callee is explicitly forbidden.

Unless there's a very good reason I can't see from inspecting just this code to keep using arguments.callee this is the recommended way.


Comandeer:
Thx @IgorRubinovich for this fix! It was merged to CKEditor 4 with fb5e3a3e2dc4a34c6f0662d16afd02b0c7b059c2.

PR #332, opened 2017-03-27
Removed eval usage from template

Comandeer:
This PR is based on #254 and fixes part of [ticket #8584](http://dev.ckeditor.com/ticket/8584), replacing template system based on implicit `eval` with a template system based only on regular expressions.

### DIFF
--- a/core/template.js
+++ b/core/template.js
@@ -10,10 +10,7 @@
 
 ( function() {
 	var cache = {},
-		rePlaceholder = /{([^}]+)}/g,
-		reEscapableChars = /([\\'])/g,
-		reNewLine = /\n/g,
-		reCarriageReturn = /\r/g;
+		rePlaceholder = /{([^}]+)}/g;
 
 	/**
 	 * Lightweight template used to build the output string from variables.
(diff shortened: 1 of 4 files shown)


mlewand:
Just a note: don't forget to credit @caridy in the final solution as he did a lot good work in #254.

core/template.js
```diff
@@ -27,28 +24,28 @@
...
-			var fn = source
-				// Escape chars like slash "\" or single quote "'".
-				.replace( reEscapableChars, '\\$1' )
-				.replace( reNewLine, '\\n' )
-				.replace( reCarriageReturn, '\\r' )
-				// Inject the template keys replacement.
-				.replace( rePlaceholder, function( m, key ) {
-					return "',data['" + key + "']==undefined?'{" + key + "}':data['" + key + "'],'";
+			// less performant solution when CSP is disabling indirect evaluation
+			cache[ source ] = function( data, buffer ) {
```
f1ames: The previous implementation created a custom function for each template (something like compiling templates). This one also does this but the function is generic (basically, only `source` is different and the rest of the function is the same). With such approach we could just go with something like:

```
CKEDITOR.template = function( source ) {

	this.source = source;
	
};

CKEDITOR.template.prototype.output = function output( data, buffer ) {

	var output = '',
		source = this.source;

	output += String( source ).replace( rePlaceholder, function( fullMatch, dataKey ) {
		return data[ dataKey ] !== undefined ? data[ dataKey ] : fullMatch;
	} );

	return buffer ? buffer.push( output ) : output;
}
```

Shouldn't be much slower as the `new CKEDITOR.template( source );` is used anyway so the new object is created for each template.

f1ames: Did some profiling against http://tests.ckeditor.dev:1030/#tests/is:unit,group:Core. There are more than 2000 tests so both functions were run probably around few thousand times. Not sure though how much reliable is profiling tool. Anyway (headers may not be visible soe first column is **Self Time** and the second one - **Total Time**):

### Current solution

**CKEDITOR.template**
image


**CKEDITOR.output**
image


### Simplified

**CKEDITOR.template** (not sure why there are two records for each)
image
image

**CKEDITOR.output** (also two records)
image
image

So it seems execution time is very similar for both so maybe it will be better to follow simpler approach? Also looking at those numbers it seems the introduced change (in comparison to previous CSP unsafe one) should not have any negative impact on editor responsibility.

WDYT?

core/template.js
```diff
@@ -27,28 +24,28 @@
 	 * @param {String} source The template source.
 	 */
 	CKEDITOR.template = function( source ) {
-		// Builds an optimized function body for the output() method, focused on performance.
 		// For example, if we have this "source":
```
f1ames: This comment is a little confusing as the resulting function is not exactly as it describes. The resulting function performs the operations described here, but its body is different.

core/template.js
Comandeer: Seems reasonable, I've simplified implementation.

core/template.js
Comandeer: Removed.

core/template.js
```diff
@@ -27,29 +23,16 @@
...
-			var fn = source
-				// Escape chars like slash "\" or single quote "'".
-				.replace( reEscapableChars, '\\$1' )
-				.replace( reNewLine, '\\n' )
-				.replace( reCarriageReturn, '\\r' )
-				// Inject the template keys replacement.
-				.replace( rePlaceholder, function( m, key ) {
-					return "',data['" + key + "']==undefined?'{" + key + "}':data['" + key + "'],'";
-				} );
+	CKEDITOR.template.prototype.output = function( data, buffer ) {
```
f1ames: Missing docs.

core/template.js
```diff
@@ -27,29 +23,16 @@
 	 * @param {String} source The template source.
 	 */
 	CKEDITOR.template = function( source ) {
-		// Builds an optimized function body for the output() method, focused on performance.
-		// For example, if we have this "source":
-		//	'<div style="{style}">{editorName}</div>'
-		// ... the resulting function body will be (apart from the "buffer" handling):
-		//	return [ '<div style="', data['style'] == undefined ? '{style}' : data['style'], '">', data['editorName'] == undefined ? '{editorName}' : data['editorName'], '</div>' ].join('');
+		this.source = String( source );
```
f1ames: Missing docs.

Comandeer: Added

core/template.js
Comandeer: Docs were already there, but in wrong place. I've moved them.

wimleers:
Wow! Great work here :) This also allows me to close https://www.drupal.org/node/2789139 over at Drupal!

PR #333, opened 2017-03-31
T/16938 Extract common PFW tests loop.

f1ames:
Proposed merge commit message:
_Tests: Common Paste from Word test loop extracted as a helper._


tests/plugins/pastefromword/generated/_helpers/createTestSuite.js
```diff
@@ -0,0 +1,73 @@
...
+function createTestSuite( options ) {
+
+	var browsers = options.browsers || [],
+		wordVersions = options.wordVersions || [],
+		tests = options.tests || [],
+		ticketTests = options.ticketTests || [],
+		testData = options.testData || { _should: { ignore: {} } },
+		ignoreAll = options.ignoreAll === true,
+		compareRawData = options.compareRawData === true,
+		customFilters = options.customFilters ? options.customFilters : null;
```
Comandeer: IMO it can be much simplified by using [`CKEDITOR.tools.extend`](http://docs.ckeditor.com/#!/api/CKEDITOR.tools-method-extend) to set default values for undefined options.

tests/plugins/pastefromword/generated/_helpers/createTestSuite.js
```diff
@@ -0,0 +1,73 @@
...
+
+	_prepareTests( testData, tests, browsers, wordVersions, compareRawData, customFilters, false, ignoreAll );
```
Comandeer: And if we're going with simplified handling of `options` object, we can also pass it straight to `_prepareTests` instead of passing absurdly high number of parameters.

tests/plugins/pastefromword/generated/_helpers/createTestSuite.js
```diff
@@ -0,0 +1,73 @@
+/* global createTestCase */
+/* exported createTestSuite */
+
+// Creates a test suite based on the options provided. The test suite then should be passed to `bender.test` to run tests.
```
Comandeer: Why not standard JSDoc comments, similar to our [existing tests](https://github.com/ckeditor/ckeditor-dev/blob/3ee603539f24cbebf0c0c24ec1b3eb48d678570b/tests/plugins/copyformatting/_helpers/tools.js#L64-L72)? Comments from tests don't leak into documentation, so it's safe to use JSDoc in a normal way.

tests/plugins/pastefromword/generated/_helpers/createTestSuite.js
```diff
@@ -0,0 +1,73 @@
...
+// @param {Boolean} areTicketsTests Whenever tests should be generated using `tickets/` directory.
+// @param {Boolean} ignoreAll
+function _prepareTests( testData, tests, browsers, wordVersions, compareRawData, customFilters, areTicketsTests, ignoreAll ) {
+	var testsKeys = CKEDITOR.tools.objectKeys( tests ),
+		wordVersion, testKey, testName, i, k, j;
+
+	if ( testsKeys.length ) {
+		for ( i = 0; i < testsKeys.length; i++ ) {
+			for ( k = 0; k < browsers.length; k++ ) {
+				for ( j = 0; j < wordVersions.length; j++ ) {
```
Comandeer: Shouldn't it be before `browsers` loop? I mean, `j` is usually before `k`… :P

tests/plugins/pastefromword/generated/_helpers/createTestSuite.js
```diff
@@ -0,0 +1,73 @@
...
+		tests = options.tests || [],
+		ticketTests = options.ticketTests || [],
+		testData = options.testData || { _should: { ignore: {} } },
+		ignoreAll = options.ignoreAll === true,
+		compareRawData = options.compareRawData === true,
+		customFilters = options.customFilters ? options.customFilters : null;
+
+	_prepareTests( testData, tests, browsers, wordVersions, compareRawData, customFilters, false, ignoreAll );
+
+	_prepareTests( testData, ticketTests, browsers, wordVersions, compareRawData, customFilters, true, ignoreAll );
```
Comandeer: If I'm seeing correctly, the only place that uses the information if the tests are ticket ones or not is inside `createTestCase` to correctly choose the right path to lookup tests. In that case IMO we could drop `ticketTests` at all and make all tests… well, just `tests`. The info that these tests are ticket ones could be moved into their name.

I'm not seeing any sense in a special treatment of `Tickets` subdirectory, especially when we have also some others (e.g. `No_heuristics`). And this change will simplify test code even more.

tests/plugins/pastefromword/generated/_helpers/createTestSuite.js
```diff
@@ -0,0 +1,73 @@
...
+
+					if ( tests[ testKey ] === true || CKEDITOR.tools.indexOf( tests[ testKey ], wordVersion ) !== -1 ) {
+
+						testName = [ 'test', testKey, wordVersion, browsers[ k ] ].join( ' ' );
+
+						if ( ignoreAll ) {
+							testData._should.ignore[ testName ] = true;
+						}
+
+						testData[ testName ] = createTestCase( testKey, wordVersion, browsers[ k ], areTicketsTests, compareRawData, customFilters );
```
Comandeer: I'm a big fan of `options` object! Maybe we could use this approach also in `createTestCase`? It would be much more readable than 6 parameters.

tests/plugins/pastefromword/generated/_helpers/createTestCase.js
```diff
@@ -1,5 +1,4 @@
-/* global assertWordFilter,Q */
-/* bender-include: _helpers/pfwTools.js */
+/* global assertWordFilter,Q,console */
 /* exported createTestCase */
 
 // @param {boolean} [compareRawData=false] If `true` test case will assert against raw paste's `data.dataValue` rather than
```
Comandeer: If we're adding docs in other places, then we can also add missing docs for `createTestCase` as only one parameter is described.

PR #336, opened 2017-04-06
Docblock for getCss has a typo

knusperpixel:
In line 129 it said " Returns a string will all CSS rules passed" - I thought that might be a typo, so I fixed it.


mlewand:
Thant's right! Thanks for the catch, I had to manually rebase it as we need it in `major` branch. Merged with e88dbddda3.

PR #337, opened 2017-04-11
T/16961 Paste from Excel improvements

f1ames:


plugins/font/plugin.js
```diff
@@ -46,7 +46,8 @@
 						left: function( element ) {
 							return !!element.attributes.size ||
 								!!element.attributes.align ||
-								!!element.attributes.face;
+								!!element.attributes.face ||
+								!!element.styles[ 'font-size' ];
```
f1ames: Looks like I forgot to push this topic further. Created a separate ticket for this http://dev.ckeditor.com/ticket/16977.

f1ames:
> Either I'm missing something obvious or tests/plugins/pastefromword/generated/_fixtures/Table_text_attributes/Mixed/excel2016/datatransfer_firefox.html are exactly the same?

Firefox adds `border-color:windowtext currentcolor windowtext windowtext;` in `td` elements with `colspan`, so the expected html is different in this point. Maybe we could disallow it in ACF for those tests to make it simpler?

tests/plugins/pastefromword/generated/_helpers/pfwTools.js
```diff
@@ -90,14 +90,15 @@
 			} ),
 
 			// Firefox adds quotes around all fonts in font-family.
+			// Safari replaces `&guot;` with `'` in quoted font names.
```
f1ames: I not sure if it is a bug, the specs allows usage of double `"` or single quotes `'` for font names. I can report it if we would like to keep a reference.

You may see [this codepen](https://codepen.io/f1ames/pen/NjxrMN?editors=1010) in both Chrome and Safari - the latter one replaces `"` with `'` in font-family property.

f1ames: The similar (reversed) situation happens in Chrome, where `'Arial'` is replaced (when styles are parsed by the browser) with `"Arial"`.

plugins/font/plugin.js
f1ames: This one is not critical for this issue, so can be fixed later.

f1ames:
Pushed requested fixes:
- Added IE11 - Excel 2016 fixture (we are missing IE11 Excel 2013 fixture due to lack of proper VM)
- Removed background extraction (as `tabletools` PR fixed it - https://github.com/ckeditor/ckeditor-dev/pull/346)
- Merged `datatransfer` fixtures (see https://github.com/ckeditor/ckeditor-dev/pull/337/commits/f5e4a68a0ffb4997f2977dc19981128162533cf6#diff-72d76022afcd3fa31608e03ceb76d6adR15)

plugins/pastefromword/filter/default.js
```diff
@@ -205,7 +205,15 @@
 						delete element.attributes.size;
 					}
 
-					createAttributeStack( element, filter );
+					// Create style stack for td/th > font if only class
+					// and style attributes are present. Such markup is produced by Excel.
+					if ( CKEDITOR.dtd.tr[ element.parent.name ] &&
+						CKEDITOR.tools.arrayCompare( CKEDITOR.tools.objectKeys( element.attributes ), [ 'class', 'style' ] ) ) {
+
+						Style.createStyleStack( element, filter, editor );
```
mlewand: Since you reverted `topmostElementName` usage here, it's no longer needed and can be removed from `createStyleStack` method.

plugins/pastefromword/filter/default.js
```diff
@@ -549,9 +558,12 @@
 		 * @param {CKEDITOR.htmlParser.element} element
 		 * @param {CKEDITOR.htmlParser.filter} filter
 		 * @param {CKEDITOR.editor} editor
+		 * @param {RegExp} [skipStyles] All matching style names will not be extracted to a style stack. Defaults
+		 * to `/margin|text\-align|width|border|padding/i`.
+		 * @param {String} [topmostElementName=span] Element type which is considered as a root element of the stack.
 		 * @member CKEDITOR.plugins.pastefromword.styles
 		 */
-		createStyleStack: function( element, filter, editor ) {
+		createStyleStack: function( element, filter, editor, skipStyles, topmostElementName ) {
```
mlewand: As mentioned before `topmostElementName` can be now removed.

tests/plugins/pastefromword/generated/excel.js
```diff
@@ -0,0 +1,55 @@
...
+/* global createTestSuite,pfwTools */
+
+( function() {
+	'use strict';
+
+	var config = pfwTools.defaultConfig;
+	config.colorButton_normalizeBackground = true;
+	// Firefox adds `border-color:windowtext currentcolor windowtext windowtext;`
+	// for `td` elements with `colspan` attribute.
+	config.disallowedContent = 'td{vertical-align,border-color}';
```
mlewand: Since it adds only `border-color`, what's the reason for adding vertical align too? It is also valuable for us to assert vertical align if possible.

f1ames: The `td{vertical-align}` was disallowed in `pfwTools.defaultConfig` which is used so I just added the `border-color` here. But I see it is not needed so your commit makes perfect sense.

PR #339, opened 2017-04-11
T/16933 Comments in colorbutton plugin pollutes global scope in documentation.

f1ames:


plugins/colorbutton/plugin.js
```diff
@@ -315,8 +315,10 @@ CKEDITOR.plugins.add( 'colorbutton', {
 		/**
 		 * Selects the specified color in the specified panel block.
 		 *
-		 * @param block
-		 * @param color
+		 * @private
+		 * @member CKEDITOR.plugins.colorbutton
```
mlewand: This member doesn't exist. And `selectColor` is not exposed. In the past we used to [comment them with //](https://github.com/cksource/ckeditor-dev/blob/7454f54/core/style.js#L1195-L1197) - so that we have API docs in nice format, and at the same time they're not picked by JSDuck.

But it looks very bad. Instead let's just use `/*` (note no double asterix) for it. It should not be picked by JSDuck.

plugins/colorbutton/plugin.js
```diff
@@ -335,8 +337,10 @@ CKEDITOR.plugins.add( 'colorbutton', {
 		/**
 		 * Converts a CSS color value to an easily comparable form.
 		 *
-		 * @param {string} color
-		 * @returns {string}
+		 * @private
+		 * @member CKEDITOR.plugins.colorbutton
```
mlewand: Same thing as in selectColor.

PR #343, opened 2017-04-13
T/16912 Paste from word incorrectly handles images.

f1ames:
When only image is copied it is the root element sitting inside `DocumentFragment` node. While the parent of the image exists, it does not have any attributes (as `DocumentFragment` simply does not have any attributes by definition).
This is the cause of the issue as the code assumed that `attributes` are always present.

_Merge message proposal:_
Image copied inside the editor is  correctly pasted using `Paste from Word`.


tests/plugins/pastefromword/manual/copypasteimage.md
```diff
@@ -0,0 +1,12 @@
+@bender-tags: tc, 4.6.0
```
mlewand: Please add ticket number `16912` as a tag too. Also correct target version to 4.7.0

tests/plugins/pastefromword/manual/copypasteimage.html
```diff
@@ -0,0 +1,16 @@
+<textarea name="editor1" id="editor1" cols="30" rows="10">
```
mlewand: I'd rename manual test to `pasteimage` too.

tests/plugins/pastefromword/cleanword.js
```diff
@@ -0,0 +1,31 @@
+/* bender-tags: editor,unit,clipboard,pastefromword */
+/* bender-ckeditor-plugins: pastefromword */
+/* bender-include: ../copyformatting/_helpers/tools.js */
```
mlewand: Looks like a copy-n-paste leftover, and should be removed.

tests/plugins/pastefromword/cleanword.js
```diff
@@ -0,0 +1,31 @@
...
+		} );
+	}
+
+	CKEDITOR.scriptLoader.load( CKEDITOR.getUrl( '/plugins/pastefromword/filter/default.js' ), function() {
+		bender.test( {
+			init: function() {
+				this.isCustomDataTypesSupported = CKEDITOR.plugins.clipboard.isCustomDataTypesSupported;
+			},
+
+			'test root image': function() {
```
mlewand: Please, reference 16912 ticket as a comment for this TC.

tests/plugins/pastefromword/cleanword.html
```diff
@@ -0,0 +1,41 @@
+<textarea id="root-image">
```
mlewand: I'd rename this test suite simply to `pasteimage` - shorter name, and still descriptive enough.

plugins/pastefromword/filter/default.js
```diff
@@ -95,7 +95,8 @@
 						}
 					};
 
-					if ( element.parent ) {
+					// If the parent is DocumentFragment it does not have any attributes. #16912
+					if ( element.parent && element.parent.type != CKEDITOR.NODE_DOCUMENT_FRAGMENT ) {
```
mlewand: Fix is good, but I'd go with duck-typing here, so that it gives us even better protection. So I suggest checking `element.parent.attributes`.

plugins/pastefromword/filter/default.js
```diff
@@ -95,7 +95,8 @@
 						}
 					};
 
-					if ( element.parent ) {
+					// If the parent is DocumentFragment it does not have any attributes. #16912
```
mlewand: Issue number should be wrapped with parenthesis.

mlewand:
Merged manually with 8f388d22d1.

PR #344, opened 2017-04-14
T/16968 Paste as plain text doesn't work in Safari.

f1ames:
_Merge message proposal:_
`Paste as plain text` shortcut fixed in Safari.


Comandeer:
LGTM.

PR #346, opened 2017-04-14
Improve background color detection in tabletools plugin

mlewand:
This PR fixes [#16971](https://dev.ckeditor.com/ticket/16971).

Current implementation finds color based on a simple regexp. Some time ago we wrote utility that does a simplified [background parsing](http://docs.ckeditor.com/#!/api/CKEDITOR.tools.style.parse-method-background).

The only thing that bugs me is that it results with a duplicated color property, because it does not remove the color from `background` css prop, while adding a second `background-color`. This is how it was implemented so far in other cases so I decided to make it compatible with how it works today.

### DIFF
--- a/tests/plugins/tabletools/tabletools.js
+++ b/tests/plugins/tabletools/tabletools.js
@@ -192,10 +192,15 @@
 			} );
 		},
 
+		// (#16971)
 		'test background color extraction': function() {
 			var bot = this.editorBot;
 
 			bender.tools.testInputOut( 'background-extraction', function( source, expected ) {
+				if ( CKEDITOR.env.ie && CKEDITOR.env.version === 8 ) {
+					// Just a regular IE quirks.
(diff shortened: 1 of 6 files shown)


tests/plugins/tabletools/tabletools.js
```diff
@@ -192,6 +192,15 @@
 			} );
 		},
 
+		'test background color extraction': function() {
+			var bot = this.editorBot;
+
+			bender.tools.testInputOut( 'background-extraction', function( source, expected ) {
+				bot.setHtmlWithSelection( source );
+				assert.beautified.html( expected, bot.getData( true ) );
```
f1ames: This test fails in IE11 due to different order of rules in `background` style:
screen shot 2017-04-18 at 14 03 58

mlewand:
Pushed fixed tests.

PR #348, opened 2017-04-18
T/16957 [PFW] Styles.inliner.inline does not take selectors specificity into account and duplicates rules.

f1ames:
The fix keeps the order of the css rules when parsing. The rules containing class selectors have higher priority and are applied first.

The styles already existing on the element are not overwritten which means:
- inline styles are preserved
- styles with the highest priority are applied first

The second point above means that the styles inlining applies rules containing class selectors first in a reversed order (so the last on goes first) and then rest of the rules also in reversed order.

Please keep in mind that this solution is a minimalistic (almost) approach to solve this specific issue. There are some cases which are not covered (more complex selectors, rules with multiple selectors, etc), however as Excel or Word does not generate such styles it is sufficient.

### DIFF
--- a/plugins/pastefromword/filter/default.js
+++ b/plugins/pastefromword/filter/default.js
@@ -753,9 +753,9 @@
 			 * Parses content of provided `style` element.
 			 *
 			 * @param {CKEDITOR.dom.element/String} styles `style` element or CSS text
-			 * @returns {Object}
-			 * @returns {Object} return.styles Object containing styles with selectors as keys and styles as values.
-			 * @returns {Array} return.order Array containing selectors in a parse order.
+			 * @returns {Array} Array containing parsed styles. Each item is an object containing two properties:
+			 * 		selector - String representing a CSS selector.
+			 * 		styles - Object containing list of styles (e.g. `{ margin: 0 }`).
 			 * @since 4.7.0
 			 * @private
 			 * @member CKEDITOR.plugins.pastefromword.styles.inliner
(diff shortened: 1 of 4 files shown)


f1ames:
Just would like to point out that styles mentioned by @Comandeer in the previous comment are used in test to check if order of the selectors with the same priority (ofc in this implementation) is preserved when parsing and inlining styles. cc @mlewand

mlewand:
As for specificity we need only a minimal viable product here. So as long as current implementation solves the problem, this is all we need.

I checked a fancy/complex word file with multiple and haven't faced case like this. In case if some version will ever produce any incompatible markup we'll simply fix it in a patch release.

## Structure

I find proposed structure redundant. I feel that it might be simplified by turning returned result into an array, and change a strucutre of entries.

So rather than:

```
.MsoChpDefault
	{mso-style-type:export-only;
	mso-default-props:yes;
	font-family:Calibri;}
p.MsoNormal, li.MsoNormal, div.MsoNormal {
	mso-style-unhide:no;
	mso-style-qformat:yes;
	mso-style-parent:"";
	margin:0cm;
	margin-bottom:.0001pt;
}
=>
{
	"styles": {
		".MsoChpDefault": {
			"font-family": "Calibri"
		},
		"p.MsoNormal, li.MsoNormal, div.MsoNormal": {
			"margin": "0cm 0cm 0.0001pt"
		}
	},
	"order": [ ".MsoChpDefault", "p.MsoNormal, li.MsoNormal, div.MsoNormal" ]
}
```

I'd propose:

```
.MsoChpDefault
	{mso-style-type:export-only;
	mso-default-props:yes;
	font-family:Calibri;}
p.MsoNormal, li.MsoNormal, div.MsoNormal {
	mso-style-unhide:no;
	mso-style-qformat:yes;
	mso-style-parent:"";
	margin:0cm;
	margin-bottom:.0001pt;
}
=>
[
	{
		"selector": ".MsoChpDefault",
		"styles": {
			"font-family": "Calibri"
		}
	},
	{
		"selector": "p.MsoNormal, li.MsoNormal, div.MsoNormal",
		"styles": {
			"margin": "0cm 0cm 0.0001pt"
		}
	}
]
```

Comandeer:
It looks better but isn't it going to complicate the logic for determining correct order of declarations?

mlewand:
I think it's worth the effort, also please extract/encapsulate sorting function into a separate function, so that `parseStyleTags` doesn't get super messy.

f1ames:
Applied changes proposed by @mlewand. The code looks much nicer and I was able to reduce its size a little :)

mlewand:
@Comandeer, please handle the review, thanks!

plugins/pastefromword/filter/default.js
```diff
@@ -824,47 +829,66 @@
...
+						styles = styles.concat( parseStyles( stylesTags.getItem( i ) ) );
 					}
 
-					return stylesObj;
+					styles.sort( getCompareFunction( styles ) );
+
+					return styles;
+				}
+
+				function getCompareFunction( styles ) {
```
Comandeer: I'd move whole sorting logic into dedicated function, `CKEDITOR.plugins.pastefromword.styles.inliner.sort`.

f1ames:
Rebased on latest `major`, on fixture needed some adjustments - https://github.com/ckeditor/ckeditor-dev/pull/348/commits/59c3dc6594892cf5d125bf7669a64e3e53088567.

PR #350, opened 2017-04-20
T/16982 Safari's clipboard API can be trusted

Comandeer:
After enabling Clipboard API in Safari, all clipboard related tests are green and pasting seems to be working as usual.

### DIFF
--- a/plugins/clipboard/plugin.js
+++ b/plugins/clipboard/plugin.js
@@ -1573,9 +1573,12 @@
 
 			// Because of a Firefox bug HTML data are not available in some cases (e.g. paste from Word), in such cases we
 			// need to use the pastebin (#13528, https://bugzilla.mozilla.org/show_bug.cgi?id=1183686).
-			// In newer Safari HTML data also should be available (#16982).
-			if ( ( CKEDITOR.env.gecko || CKEDITOR.env.safari ) &&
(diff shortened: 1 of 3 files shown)


plugins/clipboard/plugin.js
```diff
@@ -1573,11 +1573,13 @@
 
 			// Because of a Firefox bug HTML data are not available in some cases (e.g. paste from Word), in such cases we
 			// need to use the pastebin (#13528, https://bugzilla.mozilla.org/show_bug.cgi?id=1183686).
-			if ( CKEDITOR.env.gecko && ( dataTransfer.getData( 'text/html' ) || dataTransfer.getFilesCount() ) ) {
+			// In newer Safari HTML data also should be available (#16982).
+			if ( ( CKEDITOR.env.gecko || CKEDITOR.env.safari ) &&
```
mlewand: Safari handling is actually not related to Firefox, so let's make it a separate check for readability sake. In addition to that let's check Safari version too. Say sth like this:

```
			// Safari fixed clipboard in 10.1 (https://bugs.webkit.org/show_bug.cgi?id=19893).
			if ( CKEDITOR.env.safari && CKEDITOR.env.version >= 603 ) {
				return true;
			}
```

mlewand:
I have added related docs changes in https://github.com/ckeditor/ckeditor-docs/pull/88.

PR #352, opened 2017-04-25
Tableresize plugin: fire change event

jevanlingen:
Fire the change event after drag is completed. If you would do a save action after every change, the resize action would not trigger a change.


mlewand:
Hi @jevanlingen, your pull request is missing tests, see our [Contributing Code](http://docs.ckeditor.com/#!/guide/dev_contributing_code) guide for more details.

PR #357, opened 2017-05-10
[#13096] Prevent of throwing error with reading path of empty selection.

msamsel:
- Detection when range of selection is empty and prevent of calculating path in such cases.


tests/core/editable/backspaceOrDeleteError.js
```diff
@@ -0,0 +1,22 @@
...
+};
+
+// #13096
+bender.test( {
+    'test deleting text without selection': function() {
+		this.editor.getSelection().removeAllRanges();
+		this.editor.fire( 'key', {
+			domEvent: {
+				getKey: function() {
+					return 8;
```
Comandeer: Let's add test for delete also, for completness.

msamsel:
- Added <kbd>Delete</kbd> key to Unit test.
- Moved test to keystroke.js
- Added manual test
- Moved checking selection outside `if` statement, so it should be more generic.

About <kbd>Enter</kbd>, this is another error, which was should be resolve in this PR @ #359

tests/core/editable/keystrokes/keystrokes.js
```diff
@@ -166,5 +166,34 @@ bender.test( {
...
+		editor.getSelection().removeAllRanges();
+		editor.fire( 'key', {
+			domEvent: {
+				getKey: function() {
+					return DEL;
+				}
+			}
+		} );
+
+		bot.setHtmlWithSelection('<p>^Foo</p>');
```
Comandeer: There are 2 different cases – it'd divide them.

tests/core/editable/keystrokes/delbackspacequirks/manual/noselection.md
```diff
@@ -0,0 +1,14 @@
+@bender-tags: 4.7.0, tc, 13096
+@bender-ui: collapsed
+@bender-ckeditor-plugins: toolbar, wysiwygarea
+
+1. Open browser console.
```
Comandeer: Actually I'm not able to create situation, in which there will be really no-selection in the editor.

msamsel: Hmmm.
I see that currently problem might be reproduced only on Chrome browser. FF, IE and Safari seems to don't have such glitch and caret remain in editor for the whole time.
I precise description to use Chrome browser for testing.

Does it help in this case or should I make some more detail description?

Comandeer: Hm, restricting manual test to only one browser doesn't seem like good solution. Maybe trying to somehow change the test would be better? E.g. add button that removes selection from the editor?

tests/core/editable/keystrokes/delbackspacequirks/manual/noselection.md
```diff
@@ -0,0 +1,15 @@
+@bender-tags: 4.7.0, tc, 13096
```
Comandeer: Now we could update test tag ;)

tests/core/editable/keystrokes/delbackspacequirks/manual/noselection.md
```diff
@@ -0,0 +1,15 @@
...
+@bender-ui: collapsed
+@bender-ckeditor-plugins: toolbar, wysiwygarea
+
+1. Open Chrome browser.
```
Comandeer: As I stated in the comment, restricting this test to one browser is not the best option. The test should be changed to use more generic method.

core/editable.js
```diff
@@ -998,17 +998,23 @@
 					var keyCode = evt.data.domEvent.getKey(),
 						isHandled;
 
+					// #13096
```
Comandeer: Maybe two words of explanation why this change is needed.

core/editable.js
```diff
@@ -1180,8 +1186,14 @@
 						if ( !( key in backspaceOrDelete ) )
 							return;
 
+						// #13096
```
Comandeer: Explanation could be nice also here.

mlewand:
@msamsel @Comandeer One thing that have not been mentioned here is that removing `.DS_Store` is outside of the scope of this ticket and should be done in a separate contribution.

tests/core/editable/keystrokes/delbackspacequirks/manual/noselection.md
msamsel: @Comandeer can you explain how do you see that? I can restrict entire test to one browser but I'm not sure that this is the case here.
Noting else came to my mind about improving this test to be more generic. I'd appreciate a little help or some idea which might help me enchance this part. :)

msamsel: Ohh I see your comment above. Still getting used to github ;)
I'll make change after weekend

msamsel:
Comments and version was updated.
.DS_Store was restored and removing will be requested in another PR.

After weekend I made changes to manual test

tests/core/editable/keystrokes/delbackspacequirks/manual/noselection.md
msamsel: done

core/editable.js
msamsel: done

core/editable.js
msamsel: and done

tests/core/editable/keystrokes/delbackspacequirks/manual/noselection.html
```diff
@@ -0,0 +1,30 @@
...
+	}
+
+	CKEDITOR.replace( 'editor', {
+		height: 300
+	} );
+</script>
+
+<button onclick="start()" >Start.</button>
+<script>
+	function start() {
```
Comandeer: Why so complicated? 😱 

```javascript
function start() {
    var ed = CKEDITOR.instances.editor;
    ed.focus();
    ed.getSelection().removeAllRanges();
}
```
seems to do the trick on Chrome and Firefox, so I suppose that it will also work in other browsers.

tests/core/editable/keystrokes/delbackspacequirks/manual/noselection.md
```diff
@@ -0,0 +1,19 @@
+@bender-tags: 4.7.1, tc, 13096
+@bender-ui: collapsed
+@bender-ckeditor-plugins: toolbar, wysiwygarea
+
+1. Open browser console.
```
Comandeer: If you simplify the above function, then most of these steps could be deleted.

tests/core/editable/keystrokes/delbackspacequirks/manual/noselection.html
msamsel: `Button` get focus after pressing it.
What more I cannot just use `ed.focus()` because in such case selection automatically shows caret at the beginning of line.
I cannot also click into editor to restore focus in editor, because such scenario also show caret.
Idea is to remove ranges without loosing focus on editor.

What just came on my mind is to create some div 20px x 20px, which might simulate similar case with `mouseenter` event. Because `click` event mess up with focus I need to avoid them in this case. What do you think about that?

msamsel: Ok I test it, and it's working as it should. So it simplifies entire test :) I make a new commit with such improvement.

tests/core/editable/keystrokes/delbackspacequirks/manual/noselection.md
msamsel: After change description is now shorter and simpler.

Comandeer:
Added issue for this PR: https://github.com/ckeditor/ckeditor-dev/issues/457

PR #358, opened 2017-05-10
Fixed: range#_find does not return boundary end element

mlewand:
Fix for [#17022](https://dev.ckeditor.com/ticket/17022).

This change doesn't deserve a changelog entry, since this method is marked as a private.


mlewand:
@f1ames Please R this one.

PR #359, opened 2017-05-11
Prevent of throwing an error when is no selection.

### LINKED ISSUE #478: Error in console (Cannot read property 'isContextFor' of null)

Bug

## The issue is already reported:

http://dev.ckeditor.com/ticket/16874

1. Go to http://ckeditor.com/
2. Select Article Editor tab
3. Remove all data
4. Input some string (e.g. "Hello")
5. Select part of the string (e.g. "llo")
6. Click below the text (blinking cursor will disappear)
7. Click "Enter"

### Expected result

No error.

### Actual result

Error in console: Uncaught TypeError: Cannot read property 'isContextFor' of null.

## Other details

* Browser: Google Chrome, v.56.0.2924.87
* CKEditor version: Version: 4.6.2 (Tested on the CKEditor's site)
* Installed CKEditor plugins: Default plugins

beatadelura:
http://dev.ckeditor.com/ticket/16874

---

Closes #478

### DIFF
--- a/plugins/enterkey/plugin.js
+++ b/plugins/enterkey/plugin.js
@@ -521,7 +521,8 @@
 		// Check path block specialities:
 		// 1. Cannot be a un-splittable element, e.g. table caption;
 		var path = editor.elementPath();
-		if ( !path.isContextFor( 'p' ) ) {
+
+		if ( path && !path.isContextFor( 'p' ) ) {
 			mode = CKEDITOR.ENTER_BR;
 			forceMode = 1;
 		}
--- a/tests/plugins/enterkey/manual/noselection.md
+++ b/tests/plugins/enterkey/manual/noselection.md
@@ -1,4 +1,4 @@
-@bender-tags: 4.7.1, 16874
+@bender-tags: 4.7.1, 478
 @bender-ui: collapsed
 @bender-ckeditor-plugins: wysiwygarea, toolbar, enterkey, htmlwriter, sourcearea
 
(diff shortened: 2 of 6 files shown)


tests/plugins/enterkey/manual/noselection.md
```diff
@@ -0,0 +1,18 @@
+@bender-tags: 4.7.1, 16874
```
f1ames: The manual test only covers the issue case on Chrome (there is no selection when pressing enter) while in other browsers the selection exists and cannot be removed/dismissed by clicking inside of the editor.

Maybe it would be possible to add a second scenario for the test which will be covering button removing selection/ranges from the editor to check if `enter` does not produce an error then (similar as in unit test).

The exampe of test with more than one scenario: https://github.com/ckeditor/ckeditor-dev/blob/master/tests/plugins/clipboard/manual/13883.md.

Keep in mind that after pressing button outside of the editor, the editor loses focus - probably clicking should first focus the editor and then remove selection.

plugins/enterkey/plugin.js
```diff
@@ -521,7 +521,8 @@
 		// Check path block specialities:
 		// 1. Cannot be a un-splittable element, e.g. table caption;
 		var path = editor.elementPath();
-		if ( !path.isContextFor( 'p' ) ) {
+
+		if ( path && !path.isContextFor( 'p' ) ) {
```
f1ames: There is similar issue #357 and @Comandeer [mentioned a more generic approach there](https://github.com/ckeditor/ckeditor-dev/pull/357#pullrequestreview-38671902) which may cover all those cases, which is generally a good approach. The cause of the issue is the fact that enter key press is handled while there is no selection (while editable area is focused). This is however a responsibility of the core to handle such situations and `enterkey` plugin by itself also should do some additional checkings as above.

Having said that, I think this check is sufficient for this case.

beatadelura:
@f1ames PTAL

f1ames:
LGTM! I just added some small changes in tests.

f1ames:
Rebased with latest `master`.

PR #364, opened 2017-05-12
[#17020] Unit test for loosing focus in divarea editor

msamsel:
Added unit test for loosing focus in divarea editor. Problem exist in **4.6.2**.
It seems that change introduced in 13446 843720475bc05426f2c147e79f04d4cda1091dce break divarea editor.
Unit test require to have focus in view port of the browser.

### DIFF
--- a/tests/core/focusmanager/lostfocus.js
+++ b/tests/core/focusmanager/lostfocus.js
@@ -6,7 +6,7 @@ bender.editor = true;
 bender.test( {
 	'test lost focus in editor after click': function() {
 		// Test for checking if focusmanager reset ranges in divarea editor (bug introduced in https://dev.ckeditor.com/ticket/13446).
-		// It's necessary to keep focus in browser to get proper result of test. Focus in console or other window will cause test to
+		// It's necessary to keep focus in a browser to get proper result of test. Having focus in a console or other window will cause test to
 		// pass (https://dev.ckeditor.com/ticket/17020).
 
 		// Assert for IE8 returns different selection comparing to other browsers.
@@ -18,7 +18,7 @@ bender.test( {
 
 		var editor = this.editor;
 		editor.focus();
-		bender.tools.setHtmlWithSelection( editor, '<p>[foo] <strong>bar</strong></p>' );
+		bender.tools.selection.setWithHtml( editor, '<p>[foo] <strong>bar</strong></p>' );
 		CKEDITOR.document.getById( 'focusable' ).focus();
 		editor.getSelection().selectElement( editor.editable().findOne( 'strong' ) );
 		wait( function() {
@@ -30,6 +30,5 @@ bender.test( {
 					normalizeSelection: true
 				} );
 		}, 210 );
-
 	}
 } );
--- a/tests/core/focusmanager/lostfocus.js
+++ b/tests/core/focusmanager/lostfocus.js
@@ -8,7 +8,7 @@ bender.test( {
 		// Assert for IE8 returns different selection comparing to other browsers.
 		// IE8: <p>foo [<strong>bar]</strong></p>
 		// Others: <p>foo [<strong>bar</strong>]</p>
-		if ( CKEDITOR.env.ie && CKEDITOR.env.version <= 8 ) {
+		if ( CKEDITOR.env.ie && CKEDITOR.env.version <= 8 || CKEDITOR.env.safari ) {
 			assert.ignore();
 		}
 
--- a/tests/core/focusmanager/lostfocus.js
+++ b/tests/core/focusmanager/lostfocus.js
@@ -1,0 +1,26 @@
+/* bender-tags: editor,unit */
+/* bender-ckeditor-plugins: divarea */
+
+bender.editor = {
+	extraPlugins: 'divarea',
+	allowedContent: true
+};
+
+bender.test({
+	'test lost focus in editor after click': function() {
+		// #17020 - test for checking if focusmanager reset ranges in divarea editor (bug introduced in #13446).
+		// It's necessary to keep focus in browser to get proper result of test. Focus in console or other window will cause test to pass.
+		var editor = this.editor;
+		editor.focus();
+		bender.tools.setHtmlWithSelection( editor, '<p>[foo] <strong>bar</strong></p>' );
+		CKEDITOR.document.getById('focusable').focus();
+		var element = editor.editable().findOne('strong');
+		var selection = editor.getSelection();
+		selection.selectElement(element);
+
+		wait( function() {
+			assert.areSame('Range', editor.window.$.getSelection().type );
+		}, 200);
+
+	}
+});
--- a/tests/core/focusmanager/lostfocus.js
+++ b/tests/core/focusmanager/lostfocus.js
@@ -1,29 +1,26 @@
 /* bender-tags: editor,unit */
 /* bender-ckeditor-plugins: divarea */
 
-bender.editor = {
-	extraPlugins: 'divarea',
-	allowedContent: true
-};
+bender.editor = {};
 
-bender.test({
+bender.test( {
 	'test lost focus in editor after click': function() {
-		// #17020 - test for checking if focusmanager reset ranges in divarea editor (bug introduced in #13446).
+		// Ticket: https://dev.ckeditor.com/ticket/17020
+		// Test for checking if focusmanager reset ranges in divarea editor (bug introduced in https://dev.ckeditor.com/ticket/13446).
 		// It's necessary to keep focus in browser to get proper result of test. Focus in console or other window will cause test to pass.
 		var editor = this.editor;
 		editor.focus();
 		bender.tools.setHtmlWithSelection( editor, '<p>[foo] <strong>bar</strong></p>' );
-		CKEDITOR.document.getById('focusable').focus();
-		var element = editor.editable().findOne('strong');
+		CKEDITOR.document.getById( 'focusable' ).focus();
 		var selection = editor.getSelection();
-		selection.selectElement(element);
+		selection.selectElement( editor.editable().findOne( 'strong' ) );
 		wait( function() {
 			// refresh selection
 			var sel = editor.getSelection();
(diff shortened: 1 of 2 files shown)


msamsel:
First approach used native methods not available in firefox.
Updated PR use CKEDITOR methods to check selection.
Code focus on external element, then make selection in editor. After that check if selection persist in editor.
In 4.6.2 focus manager reset range, what cause selection to disappear.

tests/core/focusmanager/lostfocus.js
```diff
@@ -0,0 +1,29 @@
+/* bender-tags: editor,unit */
+/* bender-ckeditor-plugins: divarea */
+
+bender.editor = {
+	extraPlugins: 'divarea',
```
mlewand: This plugin is already loaded using `bender-ckeditor-plugins` directive.

tests/core/focusmanager/lostfocus.js
```diff
@@ -0,0 +1,29 @@
...
+	allowedContent: true
```
mlewand: I don't think there's need to disable ACF in this unit test.

tests/core/focusmanager/lostfocus.js
```diff
@@ -0,0 +1,29 @@
...
+};
+
+bender.test({
+	'test lost focus in editor after click': function() {
+		// #17020 - test for checking if focusmanager reset ranges in divarea editor (bug introduced in #13446).
```
mlewand: Since https://github.com/ckeditor/ckeditor-dev/pull/470 got in, please update the reference to Trac issue into a full URL.

tests/core/focusmanager/lostfocus.js
```diff
@@ -0,0 +1,29 @@
...
+		// It's necessary to keep focus in browser to get proper result of test. Focus in console or other window will cause test to pass.
+		var editor = this.editor;
+		editor.focus();
+		bender.tools.setHtmlWithSelection( editor, '<p>[foo] <strong>bar</strong></p>' );
+		CKEDITOR.document.getById('focusable').focus();
+		var element = editor.editable().findOne('strong');
```
mlewand: I feel like `element` does not deserve a separate variable in this case. It's used only once, so you can safely inline it into the `selectElement()` call.

tests/core/focusmanager/lostfocus.js
```diff
@@ -0,0 +1,29 @@
...
+		var selection = editor.getSelection();
+		selection.selectElement(element);
+		wait( function() {
+			// refresh selection
+			var sel = editor.getSelection();
+			sel.unlock();
```
mlewand: Is there any reason for `unlock` call?

tests/core/focusmanager/lostfocus.js
```diff
@@ -0,0 +1,29 @@
...
+			sel.reset();
+			assert.isInnerHtmlMatching( '<p>foo [<strong>bar</strong>]@</p>', bender.tools.range.getWithHtml( editor.editable(), sel.getRanges()[0] ) );
```
mlewand: Please, split this line into two, something like that:

```javascript
			assert.isInnerHtmlMatching( '<p>foo [<strong>bar</strong>]@</p>',
				bender.tools.range.getWithHtml( editor.editable(), sel.getRanges()[ 0 ] ) );
```

would be more readable.

msamsel:
I saw 2 options:
- make special assertion dedicated for IE8
- or ignore testing on IE8

As IE8 is quite old browser and bug was reproduced on other newer browsers (Chrome), I've decide to ignore tests on IE8.

Problem appears with assertion methods. I'd like to compare selections in HTML, which are returned wrongly in case of IE8. Example below:
- regular browsers: `<p>foo [<strong>bar</strong>]</p>`
- IE8 browser: `<p>foo [<strong>bar]</strong></p>`

tests/core/focusmanager/lostfocus.js
```diff
@@ -0,0 +1,36 @@
+/* bender-tags: editor,unit */
+/* bender-ckeditor-plugins: divarea */
+
+// Assert for IE8 returns different selection comparing to other browsers.
+// Normal browser <p>foo [<strong>bar</strong>]</p>
+// IE8: <p>foo [<strong>bar]</strong></p>
+if ( CKEDITOR.env.ie && CKEDITOR.env.version <= 8 ) {
+	bender.ignore();
```
f1ames: It should be `assert.ignore()` and should be put inside specific test case.

Using it in the above way would just display info that test was ignored but test cases would run anyway, failing the test suit. The `bender.ignore()` should be used in manual tests.

tests/core/focusmanager/lostfocus.js
```diff
@@ -0,0 +1,36 @@
...
+}
+
+bender.editor = {};
```
f1ames: You can also use `bender.editor = true` here.

tests/core/focusmanager/lostfocus.js
```diff
@@ -0,0 +1,36 @@
...
+bender.test( {
+	'test lost focus in editor after click': function() {
+		// Test for checking if focusmanager reset ranges in divarea editor (bug introduced in https://dev.ckeditor.com/ticket/13446).
+		// It's necessary to keep focus in browser to get proper result of test. Focus in console or other window will cause test to
+		// pass (https://dev.ckeditor.com/ticket/17020).
+		var editor = this.editor;
+		editor.focus();
+		bender.tools.setHtmlWithSelection( editor, '<p>[foo] <strong>bar</strong></p>' );
+		CKEDITOR.document.getById( 'focusable' ).focus();
+		var selection = editor.getSelection();
```
f1ames: You can move `selection` and `sel` declaration to the top, putting it together with `var editor`. Or you can simply use
```
editor.getSelection().selectElement( editor.editable().findOne( 'strong' ) );
editor.getSelection().reset()
```
without declaring those variables as they are not used anywhere else.

tests/core/focusmanager/lostfocus.js
```diff
@@ -0,0 +1,35 @@
+/* bender-tags: editor,unit */
+/* bender-ckeditor-plugins: divarea */
+
+bender.editor = true;
+
+bender.test( {
+	'test lost focus in editor after click': function() {
+		// Test for checking if focusmanager reset ranges in divarea editor (bug introduced in https://dev.ckeditor.com/ticket/13446).
+		// It's necessary to keep focus in browser to get proper result of test. Focus in console or other window will cause test to
```
f1ames: > keep focus in **a** browser

tests/core/focusmanager/lostfocus.js
```diff
@@ -0,0 +1,35 @@
...
+		// Assert for IE8 returns different selection comparing to other browsers.
+		// Normal browser <p>foo [<strong>bar</strong>]</p>
+		// IE8: <p>foo [<strong>bar]</strong></p>
+		if ( CKEDITOR.env.ie && CKEDITOR.env.version <= 8 ) {
+			assert.ignore();
+		}
+
+		var editor = this.editor;
+		editor.focus();
+		bender.tools.setHtmlWithSelection( editor, '<p>[foo] <strong>bar</strong></p>' );
```
f1ames: `bender.tools.setHtmlWithSelection` is deprecated, you should use `bender.tools.selection.setWithHtml` instead.

tests/core/focusmanager/lostfocus.js
```diff
@@ -0,0 +1,35 @@
...
+		wait( function() {
+			// Refresh selection.
+			editor.getSelection().reset();
+			assert.isInnerHtmlMatching( '<p>foo [<strong>bar</strong>]@</p>',
+				bender.tools.selection.getWithHtml( editor ), {
+					compareSelection: true,
+					normalizeSelection: true
+				} );
+		}, 210 );
+
```
f1ames: I think this empty line is not necessary.

msamsel:
Fix things from review and rebase to master.

tests/core/focusmanager/lostfocus.js
```diff
@@ -0,0 +1,34 @@
+/* bender-tags: editor,unit */
+/* bender-ckeditor-plugins: divarea */
+
+bender.editor = true;
+
+bender.test( {
+	'test lost focus in editor after click': function() {
+		// Test for checking if focusmanager reset ranges in divarea editor (bug introduced in https://dev.ckeditor.com/ticket/13446).
+		// It's necessary to keep focus in a browser to get proper result of test. Having focus in a console or other window will cause test to
```
f1ames: This comment is a little confusing:

> Having focus in a console or other window will cause test to pass

Does that mean it will not pass when having focus in a browser? Not sure what is the described behaviour here? Did you mean that test will pass always (even without the fix) when focus is not in the browser?

tests/core/focusmanager/lostfocus.js
```diff
@@ -0,0 +1,34 @@
...
+		// pass (https://dev.ckeditor.com/ticket/17020).
+
+		// Assert for IE8 returns different selection comparing to other browsers.
+		// Normal browser <p>foo [<strong>bar</strong>]</p>
+		// IE8: <p>foo [<strong>bar]</strong></p>
```
f1ames: Well, I know that IE8 is not a normal browser, but maybe something like

```
// Assert for IE8 returns different selection comparing to other browsers.
// IE8: <p>foo [<strong>bar]</strong></p>
// Others: <p>foo [<strong>bar</strong>]</p>

```

tests/core/focusmanager/lostfocus.js
msamsel: I've removed it, as leaving it behind might confuse developers in the future.

msamsel:
I wasn't able to fix it on Safari. It seem that focus automatically influence on selection.  I put ignore statement. Problem was present on Chrome, so test still coverage problematic situation.

mlewand:
It looks OK.

f1ames:
Changelog entry not needed as only unit test was added.

PR #365, opened 2017-05-12
Contributing file.

f1ames:


f1ames:
I think it will be better to put all 3 templates into one PR.

PR #368, opened 2017-05-12
T/17027 Commands event data should be initialized as an empty object

Comandeer:
Currently if command is not passed an object as a second argument, it fall backs to undefined, causing the unability to pass anything from command to afterCommandExec event.

This change is intended to introduce empty object as a default value for command's event data.

### DIFF
--- a/tests/core/command/events.js
+++ b/tests/core/command/events.js
@@ -34,6 +34,28 @@
 
 			this.editor.execCommand( 'mockupCommand' );
 			assert.isTrue( cmdCalled, 'Command should be called' );
+		},
+
+		// #17027
+		'test default event data value': function() {
+			var beforeExecData;
+
+			this.editor.once( 'beforeCommandExec', function( evt ) {
+				beforeExecData = evt.data.commandData;
+
+				assert.isObject( beforeExecData, 'Event data is initialized as an empty object.' );
+			}, null, null, 1 );
+
+			this.editor.once( 'afterCommandExec', function( evt ) {
+				assert.areSame( beforeExecData, evt.data.commandData, 'The same object is passed to afterCommandExec.' );
+			}, null, null, 1 );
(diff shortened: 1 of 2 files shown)


tests/core/command/events.js
```diff
@@ -34,6 +34,28 @@
...
+
+			this.editor.once( 'beforeCommandExec', function( evt ) {
+				beforeExecData = evt.data.commandData;
+
+				assert.isObject( beforeExecData, 'Event data is initialized as an empty object.' );
+			}, null, null, 1 );
+
+			this.editor.once( 'afterCommandExec', function( evt ) {
+				assert.areSame( beforeExecData, evt.data.commandData, 'The same object is passed to afterCommandExec.' );
+			}, null, null, 1 );
```
mlewand: No need to set the priority.

tests/core/command/events.js
```diff
@@ -34,6 +34,28 @@
...
+
+			this.editor.addCommand( 'dataFlow', {
+				exec: function() {
```
mlewand: Since we're testing `beforeExecData` in `beforeCommandExec` and `afterCommandExec` it would make sense to test it in exec too for the completeness.

tests/core/command/events.js
```diff
@@ -34,6 +34,28 @@
...
+		},
+
+		// #17027
+		'test default event data value': function() {
+			var beforeExecData;
+
+			this.editor.once( 'beforeCommandExec', function( evt ) {
+				beforeExecData = evt.data.commandData;
+
+				assert.isObject( beforeExecData, 'Event data is initialized as an empty object.' );
```
mlewand: Don't put dots at the end of assertion msg.

mlewand:
I added these minor changes while reviewing, so R+.

PR #371, opened 2017-05-15
Failing pagebreak tests in WebKit

Comandeer:
It seems that Blink inherits WebKit behavior connected with `page-break-*` and `break-*` properties. In this case enabling Blink-only fix also in WebKit seems to do the trick.

### DIFF
--- a/tests/plugins/pagebreak/pagebreak.js
+++ b/tests/plugins/pagebreak/pagebreak.js
@@ -44,8 +44,8 @@
 
 				var element = elements.getItem( 0 );
 
-				// #14605
-				if ( CKEDITOR.env.chrome ) {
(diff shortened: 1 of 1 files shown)


f1ames:
@Comandeer Which Safari version you tested on?
I have _Safari Version 10.0.1 (11602.2.14.0.7) on OS X El Capitan 10.11.6_ and test passes on `major` and fails after above changes. I assume it might be a newer version issue?

Comandeer:
I have Safari 10.1 (12603.1.30.0.34)

f1ames:
Confirmed the issue on Safari _Version 10.1 (11603.1.30.0.34)_.

PR #372, opened 2017-05-16
[#16808] Justify respect ACF rules

### LINKED ISSUE #455: The Justify plugin doesn't obey the custom Advanced Content Filter rules

Bug

http://dev.ckeditor.com/ticket/16808

1. Make sure the Justify plugin is available.
2. Configure the editor to use the following custom ACF rules:
```
allowedContent: 'p ul{text-align};li'
```
3. Load the editor. The Justify tool bar buttons are available, which is expected because we have the "p ul{text-align}" rule.
4. Verify that the custom ACF rules are active using the JavaScript console:
5 .Create a list
6. Place the caret on the second item. The Justify buttons are still enabled. Click on the "Center" button from the tool bar.

### Expected result

Justify plugin should prevent of applying justification on disallow elements. If ACF forbid to apply text-align style on given element, then plugin should be disabled on such element.

### Actual result

The Justify plugin centers the second list item, thus generating invalid content from the point of view of the ACF.

msamsel:
PR which disable justification option for elements which are not allowed by ACF rules.

close #455


tests/plugins/justify/manual/justify.md
```diff
@@ -0,0 +1,23 @@
+@bender-tags: 4.7.0, tc, 16808, justify
```
Comandeer: We can already change version tag… ;)

tests/plugins/justify/manual/justify.md
```diff
@@ -0,0 +1,23 @@
...
+
+----
+3. Move caret to list item `<li>` (e.g. word "One").
+
+**Expected:** Justification option should be disabled and you should not be allowed to justify list item.
+
+----
+**Unexpected:** Justification is possible to change when caret is inside list item.
+
+**Unexpected:** Justification buttons are not visible in editor.
```
Comandeer: Maybe, instead of writing "Unexpected" twice, just create list with unexpected things? IMO more readable.

tests/plugins/justify/manual/justify.md
msamsel: done

tests/plugins/justify/manual/justify.md
msamsel: Good point. I made such change.

msamsel:
- I updated fix, so now your test case scenario works properly (only `p` is aligned).
- I also update manual test to check for such situation.
- I made rebase to current master, because there was problems with unit tests. Now it's working fine.

plugins/justify/plugin.js
```diff
@@ -150,6 +149,12 @@
 					if ( block.isReadOnly() )
 						continue;
 
+					// #16808 Check if style or class might be applied to currently processed element
+					var tag = block.getName(),
+						isAllowedTextAlign, isAllowedCssClass;
+					isAllowedTextAlign = editor.filter.check( tag + '{text-align}' ) || false;
+					isAllowedCssClass = editor.filter.check( tag + '(' + cssClassName + ')' ) || false;
```
Comandeer: Can't we early return from function if both `isAllowedTextAlign` and `isAllowedCssClass` are false?

Also I'd check if default value is needed as `editor.filter.check` is supposed to return boolean.

Mind also the fact that below you use `editor.activeFilter` instead of `editor.filter`. Probably it would be a better idea to reuse it here.

plugins/justify/plugin.js
```diff
@@ -150,6 +149,12 @@
 					if ( block.isReadOnly() )
 						continue;
 
+					// #16808 Check if style or class might be applied to currently processed element
```
Comandeer: We have slight different convention for such comments:

```javascript
// Check if style or class might be applied to currently processed element (#16808).
```

msamsel: I wasn't aware of such convention. :(
I'll make proper changes to that.

tests/plugins/justify/justify.js
```diff
@@ -252,5 +252,66 @@ bender.test(
...
+				enterMode: CKEDITOR.ENTER_DIV
+			}
+		}, function( bot ) {
+			var editor = bot.editor;
+			bot.setHtmlWithSelection( '<div>Foo</div><ul><li>on^e</li><li>two</li><li>three</li></ul>' );
+			tc.assertCommandState( 0,0,0,0, editor );
+		});
+	},
+
+	'test alignment on disabled elements br mode': function() {
```
msamsel: I add one test where is selected non-collapsed text over paragraphs and list. It should cover this case.

tests/plugins/justify/manual/justify.html
```diff
@@ -0,0 +1,14 @@
+<div id='editor'>
```
msamsel: I make name `justifyrespectacf` it should be now more specific name of the test.

plugins/justify/plugin.js
msamsel: Corrected.

plugins/justify/plugin.js
msamsel: I made a condition to continue loop if both condition are false. You have right, default values are not necessary `filter.check` return bool.  I updated also `check` method to use `activeFilter`.

msamsel:
During test I found also that other test failed on IE8 and IE11. It seems that some older test do not allowed on inline styles. I modified condition. I don't know why test doesn't fail on Chrome, if you think that, this case should be investigate more give me an info.

Comandeer:
I've created issue for this PR: https://github.com/ckeditor/ckeditor-dev/issues/455

tests/plugins/justify/justify.js
```diff
@@ -252,5 +252,87 @@ bender.test(
 		bot.setHtmlWithSelection( '[<p><span contenteditable="false">foo</span></p>]' );
 		bot.execCommand( 'justifyright' );
 		assert.areSame( '<p style="text-align:right;"><span contenteditable="false">foo</span></p>', bot.getData( true ) );
+	},
+
+	'test alignment on disabled elements paragraph': function() {
```
Comandeer: Hmm… Looking through original issue, I've discovered that I completely forgot about `disallowContent` rule. Proposed fix should be working with it without any issues, but we could add unit test for it – just in case.

msamsel: I add unit test for `disallowedContent`.
I also create new manual test with `disallowedContent` just to be sure :)

msamsel:
About failing IE8 and `image2`. It seems that IE8 after making widget focus had selection in `body` element not in `paragraph`. This cause refresh and disabling commands.
I make less restrict condition and allow on calling justification on `body`.

plugins/justify/plugin.js
```diff
@@ -178,8 +188,18 @@
...
 			var firstBlock = path.block || path.blockLimit;
-
-			this.setState( firstBlock.getName() != 'body' && getAlignment( firstBlock, this.editor.config.useComputedState ) == this.value ? CKEDITOR.TRISTATE_ON : CKEDITOR.TRISTATE_OFF );
+			var name = firstBlock.getName();
+
+			// Check if we can give to element proper class, style or current path is body (#16808).
+			// IE8, after focus and refresh return path with 'body'
+			if ( ( this.cssClassName && editor.activeFilter.check( name+'('+this.cssClassName+')' ) ) ||
+				editor.activeFilter.check( name+'{text-align}' ) ||
+				name == 'body' )
```
Comandeer: This should work as long as the classic, iframe based editor is used. However in div-based editor this check will be always `false`. Therefore I'd check if the `firstBlock` is not the same node as editable.

msamsel: @Comandeer there are OR statements so there won't be always `false`. If there will be allowed element which can have style or class then condition will be `true`.
Do I understand you correctly that I should change 3rd condition from `name == 'body'` to `firstElement == editor.editable()`?
If not then please explain a little bit better what do . you mean by:
> Therefore I'd check if the `firstBlock` is not the same node as editable.

msamsel:
@Comandeer I hope that understand you correctly.
I add unit test based on inline editor (which is div-type), and my solution still work with it.
If I understand you wrongly, then please describe it a little bit wider how you see it?

plugins/justify/plugin.js
Comandeer: Yep, I mean `firstElement.equals( editor.editable() )`.

plugins/justify/plugin.js
```diff
@@ -178,8 +188,17 @@
...
 			var firstBlock = path.block || path.blockLimit;
-
-			this.setState( firstBlock.getName() != 'body' && getAlignment( firstBlock, this.editor.config.useComputedState ) == this.value ? CKEDITOR.TRISTATE_ON : CKEDITOR.TRISTATE_OFF );
+			var name = firstBlock.getName();
+
+			// Check if we can give to element proper class, style or current path is body (#455).
+			// IE8, after focus and refresh return path with 'body'
+			if ( ( this.cssClassName && editor.activeFilter.check( name + '(' + this.cssClassName + ')' ) ) ||
+				editor.activeFilter.check( name + '{text-align}' ) ||
+				name == editor.editable().getName() ) {
```
Comandeer: Almost good… but what about such case:

```html
<div>Te^xt</div>
```

in div-based editor with following ACF: `div; p{text-align}`? In such case buttons won't be disabled, yet they won't do anything.

Therefore the safest approach would be simply check if `firstBlock` is not `editor.editable()`.

msamsel: Ok this example showed me what you have on your mind.
Generally 3rd condition was left to preserve possibility to justify in ENTER_BR mode.

Right now I restrict this option to BR mode only ;)
Because of that I have to modify one test condition.
To have coherent visualisation between different editor types and enter modes, button will be turned off when selection will be in `editor.editable()`.

msamsel:
I remake image justify test, by surround it with `<p>` tag. Now IE8 doesn't fail.

plugins/justify/plugin.js
```diff
@@ -178,8 +188,18 @@
 
 		refresh: function( editor, path ) {
 			var firstBlock = path.block || path.blockLimit;
+			var name = firstBlock.getName();
+
+			// Check if we can give to element proper class, style or current path is body (#455).
+			// IE8, after focus and refresh return path with 'body'
+			if ( editor.config.enterMode === CKEDITOR.ENTER_BR && firstBlock === editor.editable() ) {
```
Comandeer: This condition will probably never be true. Mind the fact that path will contain _new_ instances of `CKEDITOR.dom.element` for every fetched element. Therefore there is very little chance that `editor.editable` would return one of them. To check the equality of two nodes, use [`CKEDITOR.dom.element.prototype.equals`](http://docs.ckeditor.com/#!/api/CKEDITOR.dom.domObject-method-equals).

More over this change introduces the strange behavior in IE8 in `justifyrespectacf` test. When editable is selected, justify buttons are disabled. After clicking inside paragraph, the buttons are still disabled and I can't change the alignment of the paragraph. Honestly I liked the previous version more – there was only the issue with checking name of the block, not the block itself.

Comandeer: To sum up: I think that it'd be better to just enable justify if the whole editable is selected

msamsel:
Ok so I revert it to looks like previously, but instead of comparing `node.names` now is used `equal` method.

msamsel:
I've made update for that.

plugins/justify/plugin.js
```diff
@@ -178,8 +188,19 @@
...
 			var firstBlock = path.block || path.blockLimit;
+			var name = firstBlock.getName();
+
+			// Check if we can give to element proper class, style or current path is body (#455).
+			// IE8, after focus and refresh return path with 'body'
+			// path.elements.length === 1: prevent of situation: body > ul, where selected is ul by elements path
+			if ( ( this.cssClassName && editor.activeFilter.check( name + '(' + this.cssClassName + ')' ) ) ||
+					editor.activeFilter.check( name + '{text-align}' ) ||
+					( firstBlock.equals( editor.editable() ) && path.elements.length === 1 ) ) {
+				this.setState( name != 'body' && getAlignment( firstBlock, this.editor.config.useComputedState ) == this.value ? CKEDITOR.TRISTATE_ON : CKEDITOR.TRISTATE_OFF );
```
Comandeer: `name != 'body'` seems like not needed anymore.

PR #373, opened 2017-05-17
T/tp2322 Various docs fixes

Comandeer:
This PR fixes following issues with docs:

* things leaking into global scope,
* links to non-existent classes and methods,
* incorrect or duplicated variables/parameters names.

### DIFF
--- a/plugins/clipboard/plugin.js
+++ b/plugins/clipboard/plugin.js
@@ -2773,7 +2773,7 @@
  * Duration of the notification displayed after pasting was blocked by the browser.
  *
  * @since 4.7.0
- * @cfg {Number} [clipboard_defaultContentType=10000]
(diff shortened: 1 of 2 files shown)


plugins/tabletools/plugin.js
```diff
@@ -1896,6 +1896,7 @@
 	 *
 	 * @since 4.7.0
 	 * @cfg {Boolean} [tableImprovements=true]
+	 * @member CKEDITOR.config
```
mlewand: Please revert this change as it will be deprecated by a soon to be merged table selection plugin extraction.

mlewand: Just to avoid conflicts.

Comandeer:
Ok, reverted change.

PR #375, opened 2017-05-17
[#9780] Fix JS error in IE10

msamsel:
Fix which generated throwing an error in IE10 and below.
In case of IE `range.parentElement()` was undefined and there was reference to `ownerDocument` which throws an error.

Unit test wasn't possible to make. Problem was with reproducing selection in a similar way as IE has. There was necessary to make selection with `type` other than `None` to have test which check second condition in `if` statement. Selection made by regular cursor had `Text` type, so error was visible. Bender tools create selection with `None` type only :(.

### DIFF
--- a/core/selection.js
+++ b/core/selection.js
@@ -995,7 +995,8 @@
 					range = sel.createRange();
 
 				// The selection range is reported on host, but actually it should applies to the content doc.
-				if ( sel.type != 'None' && ( range.parentElement() || {}).ownerDocument == doc.$ )
+				// #9780 added check if range.parentElement() exists
+				if ( sel.type != 'None' && range.parentElement() && range.parentElement().ownerDocument == doc.$ )
 					range.select();
 			}
 
--- a/core/selection.js
+++ b/core/selection.js
@@ -995,7 +995,7 @@
 					range = sel.createRange();
 
 				// The selection range is reported on host, but actually it should applies to the content doc.
-				// #9780 added check if range.parentElement() exists
+				// The parentElement may be null for read only mode in IE10 and below (http://dev.ckeditor.com/ticket/9780).
 				if ( sel.type != 'None' && range.parentElement() && range.parentElement().ownerDocument == doc.$ )
 					range.select();
 			}
--- a/core/selection.js
+++ b/core/selection.js
@@ -995,7 +995,7 @@
 					range = sel.createRange();
 
 				// The selection range is reported on host, but actually it should applies to the content doc.
-				if ( sel.type != 'None' && range.parentElement().ownerDocument == doc.$ )
+				if ( sel.type != 'None' && ( range.parentElement() || {}).ownerDocument == doc.$ )
 					range.select();
 			}
 
--- a/tests/tickets/16820/manual/ie10hrline.html
+++ b/tests/tickets/16820/manual/ie10hrline.html
@@ -1,0 +1,14 @@
+<div id="editor">
+	<p>Foo</p>
+</div>
+<script>
+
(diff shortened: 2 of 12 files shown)


core/selection.js
```diff
@@ -1168,7 +1168,7 @@
 					range = sel.createRange();
 
 				// The selection range is reported on host, but actually it should applies to the content doc.
-				if ( sel.type != 'None' && range.parentElement().ownerDocument == doc.$ )
+				if ( sel.type != 'None' && ( range.parentElement() || {}).ownerDocument == doc.$ )
```
f1ames: We could also check if `range.parentElement()` exists, not sure which one is more readable. Please remember about proper formatting like `( range.parentElement() || {} )` (missing space before closing parenthesis).

msamsel: Ok so i change it also a little bit:
```
if ( sel.type != 'None' && range.parentElement() && range.parentElement().ownerDocument == doc.$ )
```
So if 2nd part of condition failed then 3rd one won't be evaluated.

msamsel:
Manual tests was moved to `tickets`
I added manual tests for 16820.
Also modified a little bit the condition to have more adequate syntax in this case.
Test was restricted to IE only in version below 11.

core/selection.js
```diff
@@ -1168,7 +1168,8 @@
 					range = sel.createRange();
 
 				// The selection range is reported on host, but actually it should applies to the content doc.
-				if ( sel.type != 'None' && range.parentElement().ownerDocument == doc.$ )
+				// #9780 added check if range.parentElement() exists
```
f1ames: Remember to put ticket number at the end of the comment like:
> The `parentElement` may be null for read only mode in IE10 and below. (#9780)

Also try to describe the case why this change was introduced (like above).

tests/tickets/16820/manual/ie10hrline.html
```diff
@@ -0,0 +1,14 @@
+<div id="editor">
```
f1ames: Since this tests are already in dedicated folder `tests/16820/` there is no need for such specific naming. You may simply name files `1.html`, `1.md` (unless the issue covers more cases which should be clearly distinguishable).

tests/tickets/16820/manual/ie10hrline.md
```diff
@@ -0,0 +1,11 @@
+@bender-tags: selection, tc, 4.7.1, 16820
+@bender-ckeditor-plugins: wysiwygarea,toolbar,horizontalrule,newpage
```
f1ames: I think `newpage` dependency is not needed. You may just initialize the test with empty editor which will also simplify the test to 2 steps :)

tests/tickets/9780/manual/ie10readonly.html
```diff
@@ -0,0 +1,11 @@
+<div id="editor"></div>
```
f1ames: I would also update this test name to `1.html` and `1.md` as mentioned in another comment.

tests/tickets/16820/manual/ie10hrline.html
```diff
@@ -0,0 +1,14 @@
...
+	<p>Foo</p>
+</div>
```
f1ames: I think the more readable formatting will be (newline separating editor html and js "logic"):

```
<div id="editor">
	<p>Foo</p>
</div>

<script>
	if ( !( CKEDITOR.env.ie && CKEDITOR.env.version < 11 ) ) {
        bender.ignore();
    }

	CKEDITOR.replace('editor', {
		height: 300,
		enterMode: CKEDITOR.ENTER_BR
	} );
</script>
```

PR #376, opened 2017-05-18
Force Travis to use new version of Chrome

Comandeer:
Using new apt plugin for Travis should do the trick.


mlewand:
It clearly uses `chrome58` so that looks nice. Nice one!

PR #377, opened 2017-05-19
Remove config.tableImprovements

mlewand:
Since we extracted table selection into a separate plugin, we can now safely remove this config variable.

Now the proper way to disable table selection is simply by removing tableselection plugin.

### DIFF
--- a/core/selection.js
+++ b/core/selection.js
@@ -1981,7 +1981,7 @@
 			}
 
 			// Handle special case - fake selection of table cells.
-			if ( editor && editor.config.tableImprovements && isTableSelection( ranges ) && !isSelectingTable ) {
+			if ( editor && editor.plugins.tableselection && isTableSelection( ranges ) && !isSelectingTable ) {
 				performFakeTableSelection.call( this, ranges );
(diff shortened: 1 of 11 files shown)


core/selection.js
```diff
@@ -1981,7 +1981,7 @@
 			}
 
 			// Handle special case - fake selection of table cells.
-			if ( editor && editor.config.tableImprovements && isTableSelection( ranges ) && !isSelectingTable ) {
+			if ( editor && editor.plugins.tableselection && isTableSelection( ranges ) && !isSelectingTable ) {
```
Comandeer: This conditional is leaking. Using element path in IE 8 causes the call of `performFakeTableSelection`. It's caused by the fact that `editor.plugins.tableselection` is always true (it's an object set **before** plugin's init and not removed).

mlewand: Oh, yes, the problem here is that the plugin will get loaded despite table selection not being supported on a given environment. We need additional check to `CKEDITOR.plugins.tableselection.isSupportedEnvironment`.

PR #378, opened 2017-05-19
Fixed tests affected by tabletools in the built version

mlewand:
Some simply were failing because of ACF transformations, the other was failing because tabletools was adding an extra CSS file. Also added extra safety check, which was failing in one of the tests.

Fixes tests:
* http://tests.ckeditor.dev:10470/tests/plugins/wysiwygarea/contentscss
* http://tests.ckeditor.dev:10470/tests/plugins/tableselection/integrations/clipboard/pastenested
* http://tests.ckeditor.dev:1030/tests/plugins/tableselection/integrations/core/style

### DIFF
--- a/plugins/tableselection/plugin.js
+++ b/plugins/tableselection/plugin.js
@@ -761,7 +761,7 @@
 							targetNode = targetNode.getPreviousSourceNode( false, CKEDITOR.NODE_ELEMENT, boundaryTable );
 						}
 
-						if ( !targetNode && !endNode.is( 'table' ) && endNode.getNext() ) {
+						if ( !targetNode && endNode && endNode.is && !endNode.is( 'table' ) && endNode.getNext() ) {
 							// Special case: say we were removing the first row, so there are no more tds before, check if there's a cell after removed row.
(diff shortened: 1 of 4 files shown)


plugins/tableselection/plugin.js
```diff
@@ -761,7 +761,7 @@
 							targetNode = targetNode.getPreviousSourceNode( false, CKEDITOR.NODE_ELEMENT, boundaryTable );
 						}
 
-						if ( !targetNode && !endNode.is( 'table' ) && endNode.getNext() ) {
+						if ( !targetNode && endNode && endNode.is && !endNode.is( 'table' ) && endNode.getNext() ) {
```
Comandeer: I don't understand why this change is added in this PR.

mlewand: When I ran all the tests, this fragment was failing for one TC (don't remember which one exactly), the thing was that endNode was a textNode - so it did not have `is` method. I decided to add it as a just extra safety.

Comandeer: Ok, fair enough ;)

PR #381, opened 2017-05-19
T/14629 IE11/Edge highlighted text wrapped in span tag

### LINKED ISSUE #473: Additional styles added to style attribute after adding/removing styles

Bug

## Issue is already reported

https://dev.ckeditor.com/ticket/14629

1. Update find_highlight config option to: `config.find_highlight = { element : 'span', styles : { 'background-color' : '#004', 'color' : '#fff', 'border' : '1px solid #004' } };`
2. Open any CKEditor sample and type some text, e.g. test.
3. Open Find & Replace dialog.
4. Click 'Find' button to search for a word, e.g. test.
5. Click 'Cancel' button.
6. Switch to 'Source' view.

### Expected result

`test` word shouldn't be wrapped with span tag: `<span style="border-image: none;">test</span>`

### Actual result

`test` word has been wrapped with span tag: `<span style="border-image: none;">test</span>`

## Other details

* Browser: IE11, Edge
* CKEditor version: 4.7.0

The same issue is reproducible using `CKEDITOR.style` and `removeFromRange` method: (https://jsfiddle.net/gzg23t1v/)

beatadelura:
https://dev.ckeditor.com/ticket/14629

Closes #473.

---

Instead of adding inline styles to an element, we are adding a class with those styles.

### DIFF
--- a/plugins/find/dialogs/find.js
+++ b/plugins/find/dialogs/find.js
@@ -67,8 +67,14 @@
 		};
 
 		var highlightAttributes = CKEDITOR.tools.extend( {}, editor.config.find_highlight.attributes, { 'class': 'cke_find_highlight' } );
+
+		if ( highlightAttributes[ 'class' ] != 'cke_find_highlight' ) {
+			highlightAttributes[ 'class' ] = 'cke_find_highlight'.concat( ' ' + highlightAttributes[ 'class' ] );
+		}
+
 		var highlightStyle = new CKEDITOR.style(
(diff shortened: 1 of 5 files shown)


mlewand:
For the records, the issue is caused by bug in IE/Edge available at https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/12263408. Microsoft won't fix this bug, we decided not to patch it too https://github.com/ckeditor/ckeditor-dev/issues/473#issuecomment-308044752.

PR #386, opened 2017-05-20
Add filtering to paste handler in tableselection

Comandeer:
Paste handler in tableselection plugin uses temporary container to handle pasted data. However it allows to paste unsafe HTML inside it. This PR adds appropriate filter.

### DIFF
--- a/plugins/tableselection/plugin.js
+++ b/plugins/tableselection/plugin.js
@@ -515,7 +515,9 @@
 
 		// Pasted value must be filtered using dataProcessor to strip all unsafe code
 		// before inserting it into temporary container.
-		tmpContainer.setHtml( dataProcessor.toHtml( evt.data.dataValue ) );
+		tmpContainer.setHtml( dataProcessor.toHtml( evt.data.dataValue ), {
+			fixForBody: false
+		} );
 		pastedTable = tmpContainer.findOne( 'table' );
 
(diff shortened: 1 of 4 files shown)


plugins/tableselection/plugin.js
```diff
@@ -508,7 +509,13 @@
...
 		}
 
-		tmpContainer.setHtml( evt.data.dataValue );
+		if ( !dataProcessor ) {
+			dataProcessor = new CKEDITOR.htmlDataProcessor( editor );
+		}
+
+		// Pasted value must be filtered using dataProcessor to strip all unsafe code
+		// before inserting it into temporary container.
+		tmpContainer.setHtml( dataProcessor.toHtml( evt.data.dataValue ) );
```
mlewand: Use options argument with `fixForBody` so that no extra blocks are added, like so:
```
		tmpContainer.setHtml( dataProcessor.toHtml( evt.data.dataValue, {
			fixForBody: false
		} ) );
```

tests/plugins/tableselection/integrations/clipboard/xss.js
```diff
@@ -0,0 +1,44 @@
+/* bender-tags: editor,unit */
+/* bender-ckeditor-plugins: tableselection, image2 */
```
mlewand: Rather than relying on `image2` (which will wrap, and potentially modify img listeners), I think it make sense to use it without image2 plugin but adding a config:

			config: {
				allowedContent: 'img[*]'
			}

To each editor instance.

PR #390, opened 2017-05-22
Failing tests for tableselection

Comandeer:
Added ignore for tests in environments that use pastebin,


tests/plugins/tableselection/_helpers/tableselection.js
```diff
@@ -43,13 +43,14 @@
 		 * the current environment is not supported.
 		 *
 		 * @param {Object} testSuite
+		 * @param {Function} check
```
mlewand: Missing docs.

tests/plugins/tableselection/_helpers/tableselection.js
```diff
@@ -43,13 +43,14 @@
...
 		 */
-		ignoreUnsupportedEnvironment: function( testSuite ) {
+		ignoreUnsupportedEnvironment: function( testSuite, check ) {
 			testSuite._should = testSuite._should || {};
 			testSuite._should.ignore = testSuite._should.ignore || {};
 
 			for ( var key in testSuite ) {
-				if ( !this.isSupportedEnvironment ) {
+				if ( ( check && !check() ) || !this.isSupportedEnvironment ) {
```
mlewand: This does not need to be a function, as it's a static value.

tests/plugins/tableselection/_helpers/tableselection.js
mlewand: `check` is also an optional param

PR #392, opened 2017-05-22
[2399] improvements for manual test description

msamsel:
Improvements for manual tests description.

### DIFF
--- a/tests/plugins/clipboard/manual/pastecommands.md
+++ b/tests/plugins/clipboard/manual/pastecommands.md
@@ -3,7 +3,7 @@
 @bender-ckeditor-plugins: clipboard, pastetext, pastefromword, wysiwygarea, toolbar, undo, elementspath, clipboard, floatingspace, sourcearea
 
 
-Have fun with clipboard plugins!
+Have fun with clipboard plugins! Use buttons in editor!
 
 **Things to check:**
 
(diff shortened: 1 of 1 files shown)


tests/plugins/clipboard/manual/pastecommands.md
```diff
@@ -13,3 +13,11 @@ Have fun with clipboard plugins!
 * In IE: `Ctrl+Shift+V` displays security dialog. If user doesn't allow accessing the clipboard, notification will be shown. In the other case, content will be pasted as plain text.
 * Pressing "Paste as plain text" button, then "Paste" button and pasting afterwards should paste HTML content, not plain text. The same goes with "Paste from Word" button and resetting PFW mode.
 
+**Step by step:**
+1. Use buttons in editor to paste.
+1. After each paste reset editor.
+1. Copy helper text.
+1. Paste test using all buttons without access to clipboard.
```
mlewand: Shouldn't it be "Paste text" rather than "Paste test"?

tests/plugins/clipboard/manual/pastecommands.md
```diff
@@ -13,3 +13,11 @@ Have fun with clipboard plugins!
...
+1. Paste text with shortcut `Ctrl+Shift+V`.
+1. Use `Ctrl+V` check if text is paste properly.
+1. Use switch buttons "Paste as plaintext" with other pastes, check if last selected past option is valid for `Ctrl+V` shortcut.
```
mlewand: This could be simplified, especially the first part is iffy. How about sth like "**Use different paste buttons subsequently**, make sure that after pressing `Ctrl+V` the last pasting filter is used."

tests/plugins/clipboard/manual/pastecommands.md
```diff
@@ -3,7 +3,7 @@
 @bender-ckeditor-plugins: clipboard, pastetext, pastefromword, wysiwygarea, toolbar, undo, elementspath, clipboard, floatingspace, sourcearea
 
 
-Have fun with clipboard plugins!
+Have fun with clipboard plugins! Use buttons in editor!
```
mlewand: The change is not precise enough, it should say at lest "clipboard buttons" as there are many other buttons. In that case first sequence is no longer needed.

PR #396, opened 2017-05-23
[2402] Fixed condition for XSS manual test.

msamsel:
Change to condition for bender test to not be executed on IE and iOS.


mlewand:
Duplicate of #395.

PR #398, opened 2017-05-23
Detect meta tag generated by Microsoft Office apps.

f1ames:


plugins/pastefromword/plugin.js
```diff
@@ -66,9 +66,11 @@
 					// dataValue should be used.
 					mswordHtml = dataTransferHtml || data.dataValue,
 					pfwEvtData = { dataValue: mswordHtml },
-					wordRegexp = /(class=\"?Mso|style=(?:\"|\')[^\"]*?\bmso\-|w:WordDocument|<o:\w+>|<\/font>)/;
+					officeMetaRegexp = /<meta\s*name=("|')?generator("|')?\s*content=("|')?microsoft/gi,
```
mlewand: Let's use unnamed groups, just like one line below (so that it is consistenet) e.g. `<meta\s*name=(?:"|')?generator(?:"|')?\s*content=(?:"|')?microsoft`

tests/plugins/pastefromword/generated/meta_generator.js
```diff
@@ -0,0 +1,34 @@
+/* bender-tags: clipboard,pastefromword */
```
mlewand: Not a best idea to put a full additional test suite for a single case, instead we could put it into `tests/plugins/pastefromword/generated/excel.js`. It can be a separate TC, but let's avoid spamming too much suites.

mlewand:
Merged without a changelog, as it's more part of a [#16961](https://dev.ckeditor.com/ticket/16961).

f1ames:
@mlewand Have you checked if it solves http://dev.ckeditor.com/ticket/17049 ?

mlewand:
@f1ames yes it does, I'll close it - thanks for reminding.

PR #399, opened 2017-05-23
Corrections in clipboard/manual/buttons test

mlewand:
Now Firefox and Safari (desktop only) also supports clipboard api allowing for copy and cut.


mlewand:
Desc updated.

PR #406, opened 2017-05-29
#17052 Error on scroll while selecting nested table

### LINKED ISSUE #493: Selection started from a nested table causes an error in a browser while scrolling down

Bug

https://dev.ckeditor.com/ticket/17052

1. Create a 5x5 table (editor should become scrollable).
2. In the first cell make 2x2 nested table with width set to 15.
3. Fill every cell with letter "s".
4. Start selection from nested table, then move it out of the table and drag it to scroll down the editor.

### Expected result

Selection should stay in a nested table without returning any errors in console.

### Actual result

There is an error in a console. ​https://gist.github.com/Mgsy/b5c95910313a7cb70e87beae921ac74f

## Other details

* CKEditor version: 4.7.0

Comandeer:
Issue was caused by the fact that mouse handler is fired also on `mousemove` and scrolling selects cell outside the nested table. Simple check if the newly selected cell is still inside the original table should do the trick.

Closes #493.

### DIFF
--- a/tests/plugins/tableselection/manual/nestedtablescroll.md
+++ b/tests/plugins/tableselection/manual/nestedtablescroll.md
@@ -8,6 +8,8 @@
 2. Select cells inside nested table.
 3. Without releasing mouse button, move pointer to the bottom edge of the editor to cause scrolling down to the end of the editor.
 
+**NOTE:** This issue could not be reproducible in IE and Edge due to https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/12115046/
+
 **Expected result:**
 
 * Selection is contained inside nested table.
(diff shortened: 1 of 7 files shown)


tests/plugins/tableselection/manual/nestedtablescroll.md
```diff
@@ -0,0 +1,18 @@
+@bender-ui: collapsed
+@bender-tags: tc, 17052, gh465, 4.7.1
+@bender-ckeditor-plugins: wysiwygarea, toolbar, tableselection
+
+**Procedure:**
+
+1. Open console.
+2. Select cells inside nested table.
+3. Without releasing mouse button, move pointer to the bottom edge of the editor to cause scrolling down to the end of the editor.
```
mlewand: Add a note that for Edge/IE it acts differently, as it does not scroll.

Comandeer:
Ok, I've prepared unit test and added note about Edge. I've also changed references to ticket number.

PR #411, opened 2017-05-30
Updates to the image plugin to improve handling of percent-sized images

logiclrd:
We have a use of the `image` plugin that frequently inserts images with percentage widths. I've found the behaviour of the dialog to not entirely capture the way images with percentage widths fundamentally work. This PR addresses the issues I bumped into in the following ways:

* If a percentage is input into either `txtWidth` or `txtHeight`, the lock ratio is turned _on_ and the percentage value is copied to the other control.
* Setting a percentage height on an image in a container that doesn't have a defined height is a no-op, equivalent to `height: auto`, so the control will e.g. no longer emit `height: 50%;` when the width/height are set to `50%`.
* The preview area *does* have a fixed height, which resulted in the image appearing distorted, different from how it appears in the document after insertion. This code computes the appropriate height for the preview when the width is set to a percentage.
* The dialog has been updated to detect and properly re-load percentage width/height values when being re-opened on a previously-inserted image.

The code formatting in this codebase is very different from what I am accustomed to -- I have done my best to keep my newly-written code following the same style. Let me know if there are any style changes needed.

### DIFF
--- a/plugins/image/dialogs/image.js
+++ b/plugins/image/dialogs/image.js
@@ -12,6 +12,7 @@
 				CLEANUP = 8,
 				regexGetSize = /^\s*(\d+)((px)|\%)?\s*$/i,
 				regexGetSizeOrEmpty = /(^\s*(\d+)((px)|\%)?\s*$)|^$/i,
+				regexPercentSize = /^\s*\d+\%\s*$/i,
 				pxLengthRegex = /^\d+px$/;
 
 			var onSizeChange = function() {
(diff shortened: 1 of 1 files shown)


logiclrd:
It occurs to me that another approach might be to blank out the `txtHeight` box instead of having it copy the percentage value from `txtWidth`. Either way, the semantics would be the same. However, it has some other complexities -- for instance, right now with this PR, once you are in "percentage mode", you can edit either `txtWidth` or `txtHeight` and the locked aspect will simply copy the value from one to the other. If `txtHeight` is made blank instead, then what should the behaviour be if the user types a percentage into `txtHeight`? What if there is already a percentage in `txtWidth` and the user starts typing in `txtHeight`? On the surface, I think I like the aesthetic of having `txtHeight` blank more, but I think the implications of it add complexity to the code needed to drive it.

f1ames:
Hi @logiclrd,

PR #434, opened 2017-05-31
Add support for align (left, center, right tags)

KminekMatej:

New feature

yes, or at least, they will contai, wait for it

### This PR contains

- [x] Unit tests

## What changes did you make?

Added support for div tags having style text-align

### DIFF
--- a/plugins/bbcode/plugin.js
+++ b/plugins/bbcode/plugin.js
@@ -26,10 +26,10 @@
 		}
 	} );
 
-	var bbcodeMap = { b: 'strong', u: 'u', i: 'em', color: 'span', size: 'span', quote: 'blockquote', code: 'code', url: 'a', email: 'span', img: 'span', '*': 'li', list: 'ol' },
+	var bbcodeMap = { b: 'strong', u: 'u', i: 'em', color: 'span', size: 'span', left: 'div', right: 'div', center: 'div', quote: 'blockquote', code: 'code', url: 'a', email: 'span', img: 'span', '*': 'li', list: 'ol' },
 		convertMap = { strong: 'b', b: 'b', u: 'u', em: 'i', i: 'i', code: 'code', li: '*' },
 		tagnameMap = { strong: 'b', em: 'i', u: 'u', li: '*', ul: 'list', ol: 'list', code: 'code', a: 'link', img: 'img', blockquote: 'quote' },
-		stylesMap = { color: 'color', size: 'font-size' },
+		stylesMap = { color: 'color', size: 'font-size', left: 'text-align', center: 'text-align', right: 'text-align' },
 		attributesMap = { url: 'href', email: 'mailhref', quote: 'cite', list: 'listType' };
 
 	// List of block-like tags.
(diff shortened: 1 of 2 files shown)


KminekMatej:
I found some requests to make bbCode plugin handle text aligning. Since i needed it or my project, why not share it with others. I guess tags [left],[center] and [right] are pretty common in bbcode specs.
No tests run, I expect the reviewer would run the tests? (Dont have bender, dont really want to install it.. thx!)

plugins/bbcode/plugin.js
```diff
@@ -138,8 +138,8 @@
 
 						if ( stylesMap[ part ] ) {
 							// Font size represents percentage.
-							if ( part == 'size' )
-								optionPart += '%';
+//							if ( part == 'size' )
```
mlewand: This code should not be commented out, it's required by the font tag.

plugins/bbcode/plugin.js
```diff
@@ -157,6 +157,12 @@
 					// as "span" with an attribute marker.
 					if ( part == 'email' || part == 'img' )
 						attribs.bbcode = part;
+                                            
+                                        //special handling of left, right and center tags
```
mlewand: Code style is not consistent with rest of the code. Comment should start uppercased, there should be a space etc.

plugins/bbcode/plugin.js
```diff
@@ -157,6 +157,12 @@
...
+                                        if(part == 'left' || part == 'right' || part == 'center'){
+                                                styles[ stylesMap[ part ] ] = part;
```
mlewand: This code could be simplified and handled within the `optionPart` `if`, a couple lines above.

plugins/bbcode/plugin.js
```diff
@@ -660,13 +666,18 @@
...
 							} else if ( ( value = style[ 'font-size' ] ) ) {
+                                                                tagName = 'size';
 								var percentValue = value.match( /(\d+)%$/ );
 								if ( percentValue ) {
 									value = percentValue[ 1 ];
-									tagName = 'size';
 								}
 							}
-						} else if ( tagName == 'ol' || tagName == 'ul' ) {
+						} else if ( tagName == 'div' ) {
```
mlewand: Inconsistent spacing (also appears in few other places).

plugins/bbcode/plugin.js
```diff
@@ -660,13 +666,18 @@
 								tagName = 'color';
 								value = CKEDITOR.tools.convertRgbToHex( value );
 							} else if ( ( value = style[ 'font-size' ] ) ) {
+                                                                tagName = 'size';
 								var percentValue = value.match( /(\d+)%$/ );
 								if ( percentValue ) {
 									value = percentValue[ 1 ];
-									tagName = 'size';
```
mlewand: I don't see a reason to move this assignment, in fact moving it will lead to quirky cases, where tag is inserted, but no value is going to be assigned.

mlewand:
I'll merge it to `t/2248` branch, where this PR will get rebased onto latest `major` branch (upcoming 4.11.0 release) and there I will add my changes on top of that.

PR #467, opened 2017-06-05
T/16908 Allow to paste an image from clipboard to Chrome on macOS

### LINKED ISSUE #505: Allow to paste an image from clipboard to Chrome on macOS

Feature

## The issue is already reported:

http://dev.ckeditor.com/ticket/16908

## Details:

What Chrome gives, when I paste a file is DataTransfer (a.k.a. ClipboardData) object with an empty "files" property, but with 2 "items": first with a filename and the second with a file. It means it is possible to get a file from the second "item", on paste.

Unfortunately, when the upload plugin was implemented there was only one item on Chrome and only the first item is handled (see ​https://github.com/ckeditor/ckeditor-dev/blob/685fc7474fab779afe76edcf36011d2fb6bafab8/plugins/clipboard/plugin.js#L2545). If this line will be replaced with a loop it should be possible to get the file.

Note that on Windows pasted file is not available at all.

Also, note that now you do not need try-catch block there since items have a "kind" property which let you check if it is a file or a string.

* Browser: Chrome
* OS: macOS

beatadelura:

Bug fix

http://dev.ckeditor.com/ticket/16908

Yes

### This PR contains

Changes in: 
- [X] Unit tests
- [X] Manual tests

## What changes did you make?

Allow converting files to `base64` code for Chrome on macOS.

---

Closes #505


f1ames:
We should revisit the original issue to see if the problem still occurs and what is the best solution (since 2 years get past) and try to solve this one.

PR #494, opened 2017-06-09
Fix edition of empty anchor

### LINKED ISSUE #476: Cannot edit empty anchor tags

Bug

1. Click on the anchor button with no text selected
2. Set the name of the anchor tag, the empty anchor tag is created
3. Attempt to edit the added anchor tag, either via right click or by clicking the anchor button with the anchor tag selected

You can observe this here as well: http://sdk.ckeditor.com/samples/classic.html

### Expected result

You should be able to edit the anchor tag.

### Actual result

You cannot edit the anchor tag.

## Other details

* Browser: Chrome 59
* OS: OS X 10.12.5
* CKEditor version: 4.7 / 4.7.1
* Installed CKEditor plugins: The default "Full" package, nothing else.

msamsel:

Bug fix

yes

### This PR contains

- [x] Unit tests
- [x] Manual tests

## What changes did you make?

Correct condition to check anchor elements to use fake element properties.

close #476

### DIFF
--- a/plugins/link/dialogs/anchor.js
+++ b/plugins/link/dialogs/anchor.js
@@ -27,7 +27,7 @@ CKEDITOR.dialog.add( 'anchor', function( editor ) {
 			return null;
 		}
 
-		if ( element.is( 'a' ) ) {
+		if ( element && element.data( 'cke-real-element-type' ) === 'anchor' ) {
 			return element;
 		}
 
@@ -35,7 +35,7 @@ CKEDITOR.dialog.add( 'anchor', function( editor ) {
 		range.shrink( CKEDITOR.SHRINK_ELEMENT );
 		element = range.getEnclosedNode();
 
-		return element.type === CKEDITOR.NODE_ELEMENT && element.is( 'a' ) && element;
+		return element.type === CKEDITOR.NODE_ELEMENT && element && element.data( 'cke-real-element-type' ) === 'anchor';
 	}
 
 	return {
--- a/plugins/link/dialogs/anchor.js
+++ b/plugins/link/dialogs/anchor.js
@@ -23,19 +23,14 @@ CKEDITOR.dialog.add( 'anchor', function( editor ) {
 		var range = selection.getRanges()[ 0 ],
 			element = selection.getSelectedElement();
 
-		if ( !element ) {
-			return null;
-		}
-
-		if ( element && element.data( 'cke-real-element-type' ) === 'anchor' ) {
-			return element;
-		}
-
 		// In case of table cell selection, we want to shrink selection from td to a element.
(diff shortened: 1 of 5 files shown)


plugins/link/dialogs/anchor.js
```diff
@@ -27,15 +27,15 @@ CKEDITOR.dialog.add( 'anchor', function( editor ) {
...
 			return element;
 		}
 
 		// In case of table cell selection, we want to shrink selection from td to a element.
 		range.shrink( CKEDITOR.SHRINK_ELEMENT );
 		element = range.getEnclosedNode();
 
-		return element.type === CKEDITOR.NODE_ELEMENT && element.is( 'a' ) && element;
+		return element.type === CKEDITOR.NODE_ELEMENT && element && element.data( 'cke-real-element-type' ) === 'anchor';
 	}
```
Comandeer: As this check is actually repeated here, let's move it into private helper, which will:

* check if `element` is set (in this case the `if` from the beginning of the function could be removed);
* check if `element` is an element;
* check if `element` is the `a` element or has appropriate value of `cke-real-element-type`.

This fix also not work in case of anchors in table selection. `getSelectedAnchor` must return an element and after your change it returns `Boolean` for anchors inside table cells. However moving `element` to the end of the `if` is also not an option (as getting `element.type` could fail due to `range.getEnclosedNode` returning `null`). The only viable solution is IMO to create mentioned above function and create normal `if` here, not that abomination (sadly it was created by me…).

I also wonder if the function could be simplified to:
```javascript
function getSelectedAnchor( selection ) {
		var range = selection.getRanges()[ 0 ],
			element = selection.getSelectedElement();

		// In case of table cell selection, we want to shrink selection from td to a element.
		range.shrink( CKEDITOR.SHRINK_ELEMENT );
		element = range.getEnclosedNode();

		if ( element && element.type === CKEDITOR.NODE_ELEMENT &&
			( element.data( 'cke-real-element-type' ) === 'anchor' ) ) {
			return element;
		}
	}
```

but I'm not sure if it won't fetch more elements than it should (e.g. getting anchor even when whole paragraph is selected).

tests/plugins/link/manual/anchorediting.md
```diff
@@ -0,0 +1,16 @@
+@bender-tags: link, tc, 4.7.1, 476
```
Comandeer: Please replace `tc` tag with `bug` one.

plugins/link/dialogs/anchor.js
msamsel: It seems to be working solution. I just extend the if statement to also accept non-fake `a` elements as it was made at the origin solution.

PR #497, opened 2017-06-09
Remove double slash from tableselection.css path

mflorea:

Bug fix

this.path ends with a slash so the final path to tableselection.css contains a double slash. This can prevent the stylesheet from being loaded in some cases (e.g. when it is loaded from within a WebJar and not directly from the file system). Other plugins that use this.path (see a11yhelp, find, image2, link, liststyle, magicline, specialchar, table) don't prefix the path fragments appended to this.path with a slash.


Comandeer:
@mflorea thanks for contributing!

mlewand:
Related issue #582.

PR #527, opened 2017-06-19
T/515 Prevent of throwing an error on mouse move

### LINKED ISSUE #515: Mouse actions on CKEditor scrollbar throws 'p.getAscendant is not a function'
## Bug

Mouse events on CKEditor scrollbar results in error 'p.getAscendant is not a function'

### Steps to reproduce
1. Make editor scrollable
2. Mouse over scrollbar
3. Profit.

### Expected result
No errors

### Actual result
Errors

## Other details

Clicking the scrollbar also causes an exception.

* Version Chrome 59.0.3071.86 (Official Build) (64-bit)
* OS: Win 10
* CKEditor version: 4.7.0
* Installed CKEditor plugins: Official site http://ckeditor.com/

### LINKED ISSUE #528: Click on scrollbar throws error `node.getAscendant is not a function`

Bug

1. Make editor scrollable
3. Click on scrollbar

### Expected result

No error in the console.

### Actual result

An error in the console:

`Uncaught TypeError: node.getAscendant is not a function`

## Other details

An error comes from here: https://github.com/ckeditor/ckeditor-dev/blob/master/plugins/tableselection/plugin.js#L281

There is similar issue #515 

* Browser: Chrome 59
* OS: MacOS
* CKEditor version: 4.7.0

beatadelura:

Bug fix

Yes

### This PR contains

- [X] Unit tests
- [X] Manual tests

## What changes did you make?

Prevent of throwing an error in the console if target is undefined.

---

Closes #515
Closes #528


tests/plugins/tableselection/manual/mouseover.md
```diff
@@ -0,0 +1,20 @@
+@bender-ui: collapsed
```
mlewand: File name rather than `mouseover.md` should also mention scroll, so `scrollmouseover.md`?

tests/plugins/tableselection/manual/mouseover.md
```diff
@@ -0,0 +1,20 @@
...
+@bender-tags: tc, bug, 4.7.1, 515
+@bender-ckeditor-plugins: wysiwygarea, toolbar, sourcearea, tableselection
+
+**Procedure:**
+
+1. Open developer console.
+2. Focus on editor.
+3. Scroll down/up.
```
mlewand: Can 3 and 4 be combined? Write that the editor should be scrolled using a mouse, so this will automatically force mousing over.

plugins/tableselection/plugin.js
```diff
@@ -262,11 +262,15 @@
...
-			table = target && target.getAscendant( 'table', true ),
 			tableElements = { table: 1, thead: 1, tbody: 1, tfoot: 1, tr: 1, td: 1, th: 1 },
 			canClear;
 
+		// Prevent of throwing error in console if target is undefined (#515).
+		if ( target.$.tagName !== undefined ) {
+			var cell = target && target.getAscendant( { td: 1, th: 1 }, true ),
+			table = target && target.getAscendant( 'table', true );
+		}
+
```
mlewand: You can check if object is wrapped on HTMLElement either by checking if [getName method](http://docs.ckeditor.com/#!/api/CKEDITOR.dom.element-method-getName) exists or by checking [type property](http://docs.ckeditor.com/#!/api/CKEDITOR.dom.element-property-type).

Here it would be simpler to leave the original logic untouched, but just to do an "early return", by adding

```
if( !evt.data.getTarget().getName ) {
	return;
}
```

At the beginning of the function.

That will be safer way of fixing it. So please revert the changes in this file, and just do an early return.

tests/plugins/tableselection/mouseover.js
```diff
@@ -0,0 +1,43 @@
...
+				'<p>Test</p><p>Test</p><p>Test</p><p>Test</p>' +
+				'<p>Test</p><p>Test</p><p>Test</p><p>Test</p>', function() {
+
+				editable.$.scrollTop = 150;
+				editor.document.fire( 'mousemove', new CKEDITOR.dom.event( {
+					target: editable.getDocument()
+				} ) );
+			} );
+
+			assert.pass();
```
mlewand: In that construction the test would **always pass** despite having exception thrown.

You should put `assert.pass();` after the `editor.document.fire` call.

tests/plugins/tableselection/mouseover.js
```diff
@@ -0,0 +1,43 @@
...
+			config: {
+				height: 300
+			}
+		}
+	};
+
+	var tests = {
+		// #515
+		'test mouseover on scrollbar': function( editor ) {
+			var bot = bender.editors.scrollable,
```
mlewand: Bot is a second argument passed to test case, so you should just use argument.

tests/plugins/tableselection/mouseover.js
```diff
@@ -0,0 +1,43 @@
...
+			creator: 'replace',
+			config: {
+				height: 300
+			}
+		}
+	};
+
+	var tests = {
+		// #515
+		'test mouseover on scrollbar': function( editor ) {
```
mlewand: Bot is passed as an  argument here, see comment 2 lines below.

tests/plugins/tableselection/mouseover.js
```diff
@@ -0,0 +1,43 @@
+/* bender-tags: tableselection */
+/* bender-ckeditor-plugins: tableselection */
+/* bender-include: _helpers/tableselection.js */
+
+( function() {
+	'use strict';
+
+	bender.editors = {
+		scrollable: {
```
mlewand: Please rename this property to `classic` and add a `divarea` editor too, see https://github.com/ckeditor/ckeditor-dev/blob/b8d9a84/tests/core/editable/readonly.js#L36-L41

tests/plugins/tableselection/tableselection.html
```diff
@@ -51,5 +51,4 @@
 	</table>
 </div>
 
-
```
mlewand: This change is not needed and should  be  reverted.

plugins/tableselection/plugin.js
beatadelura: Right, it's a better solution :) Thanks!

beatadelura:
@mlewand PTAL

tests/plugins/tableselection/scrollmouseover.js
```diff
@@ -6,20 +6,21 @@
...
 		}
 	};
 
 	var tests = {
 		// #515
-		'test mouseover on scrollbar': function( editor ) {
-			var bot = bender.editors.scrollable,
-			editable = bot.editable();
+		'test mouseover on scrollbar': function( editor, bot ) {
+			// var bot = bender.editors.classic,
```
mlewand: Don't comment the code, just remove it 😉

tests/plugins/tableselection/scrollmouseover.js
```diff
@@ -6,20 +6,21 @@
 	'use strict';
 
 	bender.editors = {
-		scrollable: {
+		classic: {
```
mlewand: What i meant here is to define two separate editors. One classic (basically same definition as it was before, but different name). And the other one, called `divarea` with `divarea` added to extraPlugins.

tests/plugins/tableselection/manual/scrollmouseover.md
```diff
@@ -6,8 +6,7 @@
 
 1. Open developer console.
 2. Focus on editor.
-3. Scroll down/up.
-4. Mouse over scrollbar.
+3. Use mouse to scroll editor and point the cursor over the scrollbar.
```
mlewand: This sentence is not clear enough. Enough would be some thing "Use mouse to scroll down the editor." - and simple as that. It implies that there will be cursor over the scroll bar.

tests/plugins/tableselection/scrollmouseover.js
beatadelura: Ooops..

mlewand:
LGTM

PR #538, opened 2017-06-21
colorbutton plugin does not allow setting childRule

Jeroenwise:

Other elements like coreStyles_bold and fontSize_style allow setting the childRule. For the colorButton plugin this is not possible because the plugin overwrites the childRule, no matter what.

Unittest is not possible since the color style is directly set via the plugin's dialog.

### This PR contains

- [ ] Unit tests
- [X] Manual tests

## What changes did you make?

Added a condition around the childRule override. The childRule will only be set if it was not defined yet.
Also fixed codestyle for the changed piece of code.


f1ames:
Hi @Jeroenwise,

PR #569, opened 2017-06-27
List all available hotkeys in the Accessibility Instructions dialog

### LINKED ISSUE #456: List hotkeys for all commands in help dialog

Feature request

1. Open a page with CKEditor with the `a11yhelp` plugin.
1. Focus the editor.
1. Press `alt+0`.

### Expected result

Dialog should include all the registered commands, looking up the hotkey using [getCommandKeystroke](http://docs.ckeditor.com/#!/api/CKEDITOR.editor-method-getCommandKeystroke).

Also, if possible, it should attempt to lookup the label from related button.

### Actual result

Currently it's support is limited to [list of hardcoded commands](https://github.com/ckeditor/ckeditor-dev/blob/606ced773bcae95ab734727a3bbefae52f7a67f1/plugins/a11yhelp/dialogs/lang/en.js#L57-L108).

msamsel:
New feature

yes

### This PR contains
- [x] Unit tests
- [x] Manual tests

## What changes did you make?
- Automatically created list with available keystroke in editor. List is used in Accessibility help plugin
- Move description from accessibility plugin to proper plugin which register given keystroke.
- Add new `editor.event` `keystrokeEntry` which allows on modification displayed keystrokes
- Add option to Command Definition to provide `label` and `description` property which later on will be used by accessibility plugin to present it in a table.
- Move method `representKeystroke` to be available in editor.

close #456


AnnaTomanek:
Please make sure that you remember about the following things:
- No translations that are already present can get lost (also applies to translations available in other plugins - e.g. if "Paste as plain text" is translated in the plugin, we should copy it).
- Reuse existing language strings and translations; also if the same language string is used in different places, add it to the core language file.

msamsel:
@AnnaTomanek Generally entire command section, which was predefined in Accessibility Helper plugin now is removed and replaced by automatically generated table, where `Commands label` has proper `keystroke`.
Change introduce possibility, that plugin creator can define inside his plugin some description, which will be automatically used by Accessibility Help plugin.
Earlier such change required modification of `a11yHelp` plugin.

I'm not sure what is proper way in such case. Should those translations remain in files event if those are not used any longer?

f1ames:
This case is a little tricky, as @AnnaTomanek said we would like to make as little changes as possible in language files (especially when translations are already there) especially when just moving language keys/strings around.

However, as @msamsel pointed out the previous solution basically had predefined values which were used in Commands/Shortcuts section in `a11yhelp` dialog. In order to display new shortcuts, the `a11yhelp` plugin needed to be modified. In a proposed solution, shortcut list is generated dynamically based only on information contained in each plugin.
This solution is more flexible and gives an ability (also for community plugins) to _register_ an information about command (label and optional description) which then can be reused to list all commands shortcuts with proper naming and description (e.g. in a help dialog).
It seems like this improvement in flexibility is worth the cost. WDYT @mlewand @AnnaTomanek?

@mlewand I think if we decide to go with such solution it will be good to extract it to a new issue (which will be a new feature introducing possibility for each command to contain/register additional information like command label and description). WDYT?

msamsel:
Generally there wasn't much text to remain. Most of it had to be removed. Commands are now in a table which downloads label for them automatically. It takes values from command label or if there is no command label, then from command ui button label. So most of them should have translations already.

mlewand:
Sure, we need to retain translations whenever possible. Unfortunately in this case it won't be possible, as new translations will be different than the former one - new labels are much shorter/atomic.

Old approach was bad, having redundant text and a knowledge about external plugins in `a11yhelp` plugin.

Luckily it's set for a major release so there's plenty of time for our translation Contributors to pick it up.

msamsel:
I correct problems mention by @mlewand, add milestone `4.8.0` and change base branch to `major`.

msamsel:
There was a lot of changes.
I've remake entire solution to use `getLabel` and `getKeyDescription` instead of event handling. What also affect on `pastetext` plugin which requires those modifications.

About magicline, linters report many issues. Most of them was declaring something after usage. Changes are rearranging code to have most of variables declaration at the beginning, also some functions was invoked inside another which also had to be rearranged.

As far as I remember I've made only significant changes here (https://github.com/ckeditor/ckeditor-dev/blob/master/plugins/magicline/plugin.js#L892-L894)
where I add label and keyDescription to code.

tests/core/command/command.js
```diff
@@ -302,5 +302,106 @@ bender.test( {
 		editor._.elementsPath.onClick( 0 );
 
 		wait();
+	},
+
+	// #456
+	'test basic behavor of getLabel': function() {
```
f1ames: **behavior** (missing `i`)

core/commanddefinition.js
```diff
@@ -160,3 +160,46 @@
...
+ * @method getKeyDescription
+ * @param {CKEDITOR.editor} editor The editor instance.
+ * @returns {String} command's description
+ */
+
+/**
+ * Defines command label. This property is used by different plugins to display command information (by e.g. Accessibility Help).
+ * For geting command's label please use {@link CKEDITOR.commandDefinition#getLabel}, instead of directly reading `label` property.
+ *
+ * @since 4.8.0
```
msamsel: I have doubts about it. If those will be marked as `@private`, then some kind of 'setter' might be required. Users might not see those variables, or be afraid of modifying them. What do you think about that?

core/command.js
```diff
@@ -116,6 +116,42 @@ CKEDITOR.command = function( editor, commandDefinition ) {
 		return allowed = editor.activeFilter.checkFeature( this );
 	};
 
+	/**
+	 * Providing this method will overwrite {@link CKEDITOR.commandDefinition#getLabel}, what allows on better customization
```
msamsel: I rearrange it, I hope it will be ok ;D

tests/plugins/a11yhelp/a11yhelp.js
```diff
@@ -0,0 +1,146 @@
...
+
+
+		'test representKeystroke': function() {
+			this._createEditor( {}, function() {
+				var editor = this.editor;
+				assert.areSame( '<kbd><kbd>Alt</kbd>+<kbd>0</kbd></kbd>', CKEDITOR.plugins.a11yhelp.representKeystroke( editor, CKEDITOR.ALT + 48 ) );
+			} );
+		},
+
+		'test keystrokeEntry event': function() {
```
f1ames: There are still some leftovers of `keystrokeEntry` event in this file and also in `tests/plugins/a11yhelp/manual/commandstable` manual test (which fails now).

core/command.js
```diff
@@ -116,6 +116,49 @@ CKEDITOR.command = function( editor, commandDefinition ) {
...
+	 * @since 4.8.0
+	 * @returns {String} Command's label
+	 */
+	this.getLabel = function() {
+		if ( typeof commandDefinition.getLabel == 'function' ) {
+			return commandDefinition.getLabel.call( this, editor );
+		} else if ( commandDefinition.label ) {
+			return commandDefinition.label;
+		} else if ( this.uiItems && this.uiItems.length && this.uiItems[ 0 ].label ) {
+			return this.uiItems[ 0 ].label;
```
f1ames: What was the reason for this fallback, was it used anywhere before?

core/commanddefinition.js
```diff
@@ -162,12 +162,8 @@
  */
 
 /**
- * Provides command's label string. Label is define with few fallbacks in specific order described below:
- *
- * 1. If {@link CKEDITOR.command#getLabel} method exists, it is used to provide label,
- * 2. If {@link CKEDITOR.commandDefinition#label} atribute exists, it is used to provide label.
- * 3. If {@link CKEDITOR.command#uiItems} is non empty, then label of first array's item is used as label
- * 4. Empty string.
+ * Providing this method will overwrite {@link CKEDITOR.command#getLabel}, what allows on better customization
```
f1ames: My previous comment might be misleading a little, sorry for that. You moved the description to the right place as I suggested, but still:

> Providing this method will overwrite {@link CKEDITOR.command#getLabel}

Providing this method will not overwrite `CKEDITOR.command#getLabel`. The provided method will be just used by `CKEDITOR.command#getLabel` to return a label. I suggest to update the description to something like:

> If present, this method will be used to provide command's label. It allows customizing label for a specific command.

Same for `getKeyDescription` description.

core/commanddefinition.js
f1ames: Makes sense 👍

core/command.js
msamsel: This kind of fallback allows on reusing already created labels for buttons without modifying all other plugins.
So steps are to use `getLabel` method if exist, then `label` property, and then button's label if exist. So we use translations for button's label such as **bold**, *italic*, etc, without waiting for them being translate and just use what we have already.

plugins/a11yhelp/plugin.js
```diff
@@ -22,30 +22,148 @@
...
+		 * @param {CKEDITOR.editor} editor Editor instance.
+		 * @param {Number} keystroke Number which represent command keystroke.
+		 * @returns {String} HTML code with key names wrappeed with `<kbd></kbd>` tags.
+		 */
+		representKeystroke: function( editor, keystroke ) {
+			var lang = editor.lang.a11yhelp,
+				coreLang = editor.lang.common.keyboard;
+
+			// CharCode <-> KeyChar.
+			var keyMap = {
```
f1ames: Maybe this map could be defined as `CKEDITOR.plugins.a11yhelp` property? Now, every time the method is called, the map is defined again.

plugins/a11yhelp/styles/a11yhelp.css
```diff
@@ -0,0 +1,50 @@
...
+	padding: 0.8em;
+	font-weight: bold;
+}
+
+.cke_accessibility_legend td {
+	border: 0.1em solid black;
+	padding: 0.8em;
+}
+
+.cke_accessibility_legend tbody > tr:nth-child(odd) {
```
f1ames: This is not supported in IE8 😞  Maybe we could use `.odd` class when generating the commands table?

dev/langtool/meta/ckeditor.core/meta.txt
```diff
@@ -69,7 +69,7 @@ common.invalidCssLength = Error message displayed to the user when the specified
 common.invalidHtmlLength = Error message displayed to the user when the specified length value is not a valid HTML measurement unit.
 common.invalidInlineStyle = Error message displayed to the user when the specified style value is not a valid CSS declaration.
 common.cssLengthTooltip = Tooltip displayed for the Width and Height fields in the Table Properties dialog window explaining the correct format of these length values.
-common.unavailable =
+common.unavailable =
```
f1ames: Not sure why this lang entry get here?

plugins/a11yhelp/plugin.js
```diff
@@ -60,21 +63,27 @@
...
 		 * @since 4.8.0
-		 * @param {CKEDITOR.editor} editor Editor instance.
-		 * @param {Number} keystroke Number which represent command keystroke.
-		 * @returns {String} HTML code with key names wrapped with `<kbd></kbd>` tags.
+		 * @private
+		 * @param {CKEDITOR.editor} editor Editor instance
+		 * @returns {Object}
 		 */
-		representKeystroke: function( editor, keystroke ) {
+		_renderKeyMap: function( editor ) {
```
f1ames: I think we can simplify creating `keyMap` even more, just creating singleton like method:

```
_renderKeyMap: ( function() {
    var keyMapCache;
    return function( editor ) {
        if ( !keyMapCache ) {
            // Here goes the body of the current function. Created keyMap should be assigned to keyMapCache variable.
        }
        return keyMapCache;
    }
} )();
```

And then we can completely drop `_keyMap` property and just use `var keyMap = this._renderKeyMap( editor );` in other places.

---

Also `renderKeyMap` is somehow misleading name, this functions does not render anything it creates/generates key map object so it can be e.g. `[create/generate/get]keyMap`.

dev/langtool/meta/ckeditor.core/meta.txt
msamsel: Me neither. ;) I just use `dev/langtool/langtool.sh` and it was somehow affected by this.

msamsel: BTW it was only added space at the end of lien if I see correct comparing to repository: https://github.com/ckeditor/ckeditor-dev/blob/master/dev/langtool/meta/ckeditor.core/meta.txt#L72

So I'm not sure what should with that :) any ideas?

msamsel:
Ok this hidden comment is corrected also ;)

dev/langtool/meta/ckeditor.core/meta.txt
f1ames: Ok, we decided to left it as it is.

f1ames:
@mlewand, please check if your previous Change Requests were fulfilled or if you have any more comments.

tests/core/command/command.js
```diff
@@ -302,5 +302,106 @@ bender.test( {
...
+			name: 'getLabel1',
+			config: {
+				on: {
+					// This allows on setting up nice enviornment before asserts.
+					'pluginsLoaded': function( evt ) {
+						var editor = evt.editor;
+
+						editor.addCommand( 'custom1', {
+							exec: function() {
+								return;
```
mlewand: Unnecessarily expressive function. Use just `function() {}`. This is repeated few times in this file.

tests/core/command/command.js
```diff
@@ -302,5 +302,106 @@ bender.test( {
...
 		wait();
+	},
+
+	// #456
+	'test basic behaviour of getLabel': function() {
+		bender.editorBot.create( {
+			name: 'getLabel1',
+			config: {
+				on: {
+					// This allows on setting up nice enviornment before asserts.
```
mlewand: Typo, should be "environment".

tests/core/command/command.js
```diff
@@ -302,5 +302,106 @@ bender.test( {
...
+		} );
+	},
+
+	// #456
+	'test basic behaviour of getDescription': function() {
+		bender.editorBot.create( {
+			name: 'getKeystrokeDescription1',
+			config: {
+				on: {
+					// This allows on setting up nice enviornment before asserts.
```
mlewand: Typo, should be "environment".

msamsel:
Current changes should be merged after #830. Without it both keystrokes <kbd>Alt</kbd>+<kbd>0</kbd> and <kbd>Ctrl/Cmd</kbd>+<kbd>/</kbd> will open first tab. With changes from #830, 2nd shortcut will open keystroke list.

msamsel:
Note: because of changes in #830, it have to be remember to change `tab` occurrence into `tabId`.

tests/plugins/a11yhelp/keystrokes.js
```diff
@@ -0,0 +1,230 @@
+/* bender-tags: editor,unit */
+/* bender-ckeditor-plugins: a11yhelp,dialog*/
+
+( function() {
+	'use strict';
+
+	bender.editor = {};
```
mlewand: You're actually not using this line. This line creates a default editor, but you're creating a custom instance each time in `_createEditor` method. So it can be removed.

plugins/a11yhelp/styles/a11yhelp.css
```diff
@@ -0,0 +1,97 @@
...
+.cke_accessibility_legend dd {
+	margin-top: 6px;
+	margin-bottom: 10px;
+	line-height: 19px;
+}
+
+.cke_accessibility_legend table {
+	border-collapse: collapse;
+	border: 1px solid #d9dce0;
+	margin: 10px !important;
```
mlewand: We should not use `!important`. Please adjust selectors to achieve this effect.

plugins/a11yhelp/plugin.js
```diff
@@ -20,32 +20,190 @@
...
+		 * 		"<kbd><kbd>Alt</kbd>+<kbd>0</kbd></kbd>"
+		 *
+		 * @since 4.8.0
+		 * @param {CKEDITOR.editor} editor Editor instance.
+		 * @param {Number} keystroke Number which represent command keystroke.
+		 * @param {String} [wrappingTag=kbd] Wrapping tag. It's inner string of tag: e.g. 'kbd' string in case of `<kbd></kbd>` tags pair.
+		 * Providing empty string will remove wrapping tags from returned result.
+		 * @returns {String} HTML code with key names wrapped with wrapping tag by default it is `<kbd></kbd>`.
+		 */
+		representKeystroke: function( editor, keystroke, wrappingTag ) {
```
mlewand: Please, make it a private property by adding `@private` tag. Technically this method does not belong to this plugin, but it's fine to leave it here as long as it's only used by a11yhelp plugin.

plugins/a11yhelp/plugin.js
```diff
@@ -20,32 +20,190 @@
...
+		/**
+		 * Return html which represent keystroke. Result is wrapped with `<kbd></kbd>` tags or other defined by the user.
+		 *
+		 * 		CKEDITOR.plugins.a11yhelp.representKeystroke( CKEDITOR.instances.editor, 4456496 );
+		 * 		"<kbd><kbd>Alt</kbd>+<kbd>0</kbd></kbd>"
+		 *
+		 * @since 4.8.0
+		 * @param {CKEDITOR.editor} editor Editor instance.
+		 * @param {Number} keystroke Number which represent command keystroke.
+		 * @param {String} [wrappingTag=kbd] Wrapping tag. It's inner string of tag: e.g. 'kbd' string in case of `<kbd></kbd>` tags pair.
```
mlewand: Default string value ('kbd') should be wrapped in quotes (single quotes).

plugins/a11yhelp/plugin.js
```diff
@@ -20,32 +20,190 @@
...
+	CKEDITOR.plugins.a11yhelp = {
+		/**
+		 * Method creates key map object used by {@link CKEDITOR.plugins.a11yhelp#representKeystroke} to provide proper names for keys.
+		 *
+		 * @since 4.8.0
+		 * @private
+		 * @param {CKEDITOR.editor} editor Editor instance
+		 * @returns {Object}
+		 */
+		_createKeyMap: ( function() {
```
mlewand: Private methods should go to the end of a type/class.

tests/plugins/a11yhelp/manual/commandstable.html
```diff
@@ -0,0 +1,44 @@
...
+</textarea>
+
+
+<script>
+	if ( bender.tools.env.mobile ) {
+		bender.ignore();
+	}
+	CKEDITOR.replace( 'editor1' );
+
+	var editor = CKEDITOR.replace( 'editor2');
```
mlewand: Missing space after `'editor2'`.

tests/plugins/a11yhelp/manual/commandstable.md
```diff
@@ -0,0 +1,17 @@
...
+@bender-ui: collapsed
+@bender-ckeditor-plugins: wysiwygarea,a11yhelp
+
+**Perform below steps in both editors.**
+
+1. Focus editor.
+1. Press `Ctrl/Cmd + /`.
+
+**Expected:**
+* First editor - Table should contain at least 2 keystrokes `Alt + 0` amd `Ctrl/Cmd + Forward Slash`.
```
mlewand: Typo "amd" --> "and".

plugins/a11yhelp/plugin.js
```diff
@@ -20,32 +20,190 @@
...
+					tab: 'keystrokes'
+				} ) );
 
-			editor.setKeystroke( CKEDITOR.ALT + 48 /*0*/, 'a11yHelp' );
-			CKEDITOR.dialog.add( commandName, this.path + 'dialogs/a11yhelp.js' );
+				editor.setKeystroke( CKEDITOR.ALT + 48 /*0*/, 'a11yHelp' );
+				CKEDITOR.dialog.add( commandName, plugin.path + 'dialogs/a11yhelp.js' );
+
+				editor.setKeystroke( CKEDITOR.CTRL + 191 /*/*/, 'keystrokesTab' );
+				CKEDITOR.dialog.add( 'keystrokesTab', plugin.path + 'dialogs/a11yhelp.js' );
```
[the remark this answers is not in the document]
msamsel: I didn't know that. I've base on find plugin:
https://github.com/ckeditor/ckeditor-dev/blob/2249f1a3172f98d62cc404f2b1ab4c9018d427dc/plugins/find/plugin.js#L35-L36
Which opens on different tab when is 'find' or 'replace' used.

plugins/a11yhelp/plugin.js
msamsel: As far as I remember it supposed to be available for everybody. To give user possibility to make nicely wrapped keystroke if he overwrite keystroky description.

mlewand: It's not a valid place for this method, if anything it should be placed in general purpose tools. But I don't think we need to care about this now.

If someone will be interested in reusing this logic, he or she will figure check this private method, and file a proper request if needed.

msamsel: Ok this explains the situation. I'll change it to private then.

plugins/a11yhelp/plugin.js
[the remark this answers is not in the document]
mlewand: It's a slightly different thing, as in fact it does define two separate dialogs: https://github.com/ckeditor/ckeditor-dev/blob/2249f1a3172f98d62cc404f2b1ab4c9018d427dc/plugins/find/dialogs/find.js#L817-L823

And it does so because... someone was lazy enough not to do what you made in #830. 🙂 

Actually having #830 in place we can refactor it, I will create a proper ticket for that in a minute.

mlewand:
Pushed one commit with some code style adjustments.

mlewand:
This feature should utilize features already available in tools, there's a [keystrokeToString](https://docs.ckeditor.com/#!/api/CKEDITOR.tools-method-keystrokeToString) method that is perfect candidate.

Currently `representKeystroke` and `_createKeyMap` combo duplicates it.

However we can't use `keystrokeToString` at it's current shape, as it returns a string like "CTRL + X". While it might be tempting to just split it using "+" char, this is not the right thing to do. Instead we should have an underlying method that parsers keycode, and return array of keys in a keystroke, so that it can be joined for presentation.

I have created an issue #856 for that - issue #856 will block this pull request.

PR #599, opened 2017-07-03
Fix for remaining fake selection in advance tab of table properties

### LINKED ISSUE #579: Unwanted "cke_table-faked-selection-table" class in table properties dialog box

Bug

1. Select some cells in the table. For example you can use table from this page: http://sdk.ckeditor.com/samples/fullpreset.html
2. Go to the 2nd tab in the "Table properties" dialog box

### Expected result

In the "CSS classes" input field you should see only actual classes, not the virtual ones. In this particular case the field should be empty

### Actual result

In the "CSS classes" input field you'll see "cke_table-faked-selection-table". It can confuse some users

## Other details

* Browser: Firefox
* OS: Windows
* CKEditor version: 4.7.0 at least
* Installed CKEditor plugins: advanced tab, table selection

2017-06-29 14_05_28-ckeditor sdk samples full preset

msamsel:

Bug fix

yes
### This PR contains

- [x] Unit tests
- [x] Manual tests

## What changes did you make?
Move classes to dataset attributes

close #579

### DIFF
--- a/plugins/tableselection/plugin.js
+++ b/plugins/tableselection/plugin.js
@@ -9,6 +9,7 @@
 	var fakeSelectedClass = 'cke_table-faked-selection',
 		fakeSelectedTableClass = fakeSelectedClass + '-table',
 		fakeSelectedEditorClass = fakeSelectedClass + '-editor',
+		fakeSelectedTableDatasetClass = 'cke-table-faked-selection-table',
 		fakeSelection = { active: false },
 		tabletools,
 		getSelectedCells,
@@ -122,7 +123,7 @@
 		}
 
 		if ( selectedCells.count() > 0 ) {
-			selectedCells.getItem( 0 ).getAscendant( 'table' ).removeClass( fakeSelectedTableClass );
+			selectedCells.getItem( 0 ).getAscendant( 'table' ).data( fakeSelectedTableDatasetClass, false );
 		}
 
 		editor.fire( 'unlockSnapshot' );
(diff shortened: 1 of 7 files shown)


tests/plugins/table/manual/fakeselectionadvancetab.md
```diff
@@ -0,0 +1,11 @@
+@bender-tags: 4.7.2, tc, bug, 579
```
f1ames: `tc` tag is not longer used.

plugins/tableselection/plugin.js
```diff
@@ -251,7 +252,7 @@
 
 		if ( cells.length > 0 ) {
 			editor.editable().addClass( fakeSelectedEditorClass );
-			cells[ 0 ].getAscendant( 'table' ).addClass( fakeSelectedTableClass );
+			cells[ 0 ].getAscendant( 'table' ).data( fakeSelectedTableDatasetClass, fakeSelectedTableClass );
```
msamsel: I used empty string. Setting `true` just parse it to string, so it reflects with result of attribute: `data-cke-table-faked-selection-table="true"`.
Unfortunately method `data` is combination of setter and getter, so I cannot omit 2nd attribute when I want to use it as a setter.
Current approach seems to be most convenient. Another option would be design some simple general purpose data attribute, like `data-cke-class` and assign class values, but it might be confusing.

f1ames: Makes sense 👍 .
It is similar behaviour to how `Boolean attributes` works (e.g. `input[checked]`). Here we also have only two values, `selected` / `not selected` so it is a proper approach IMHO.

tests/plugins/table/fakeselectionadvancetab.js
```diff
@@ -0,0 +1,28 @@
...
+			}
+
+			var bot = this.editorBot;
+
+			// Add table with fake selection.
+			bot.setHtmlWithSelection( '<table border="1">[<tr><td>Cell1</td><td>Cell2</td></tr>]</table>' );
+
+			bot.dialog( 'tableProperties', function( dialog ) {
+				assert.areSame( '', dialog.getValueOf( 'advanced', 'advCSSClasses' ) );
+				dialog.hide();
```
f1ames: There is no need to hide dialog here, especially that it happens after assert so does not have any effect on the test result. This line can be simply removed.

PR #615, opened 2017-07-06
Font options are not toggleable any longer

### LINKED ISSUE #584: Fonts setting should not be toggling options

Feature request

http://dev.ckeditor.com/ticket/16865

1. Change the font size for the first time it works correctly.
2. But if select the same font size again, it will revert back to the original font size. The same for font style.
3. This behavior doesn't conform to other main editors, and could bring problems. 

### Expected result

Please prevent revert font back to original font if select same font again.

### Actual result

Resetting the font to original font if select font again.

## Additional Information

To provide a possibility to remove the font all the way. It was specified in [the comment below](https://github.com/ckeditor/ckeditor-dev/issues/584#issuecomment-313076114).

msamsel:
New feature

yes

### This PR contains
- [x] Unit tests
- [x] Manual tests

## What changes did you make?
- Disable possibility to toggle option on drop down list of: **font size**, **font family name**, **paragraph format**
- Add `(Default)` option to **font size** and **font family** name, which remove formating from selection
- Add new language entry with `Default` to `lang.common` file
- Fix one unit test which click twice into this same menu option and start to fail after change

close #584

### DIFF
--- a/plugins/font/plugin.js
+++ b/plugins/font/plugin.js
@@ -5,7 +5,6 @@
 
 ( function() {
 	function addCombo( editor, comboName, styleType, lang, entries, defaultLabel, styleDefinition, order ) {
-		// debugger;
 		var config = editor.config,
 			style = new CKEDITOR.style( styleDefinition );
 
@@ -94,7 +93,7 @@
 
 				this.startGroup( lang.panelTitle );
 
-				// Add (default) option as first element on the list.
+				// Add `(Default)` item as first element on the drop down list.
 				this.add( defaultValue, defaultText, defaultText );
 
 				for ( var i = 0; i < names.length; i++ ) {
--- a/plugins/font/plugin.js
+++ b/plugins/font/plugin.js
@@ -93,7 +93,7 @@
 
 				this.startGroup( lang.panelTitle );
 
-				// Add `(Default)` item as a first element on the drop down list.
+				// Add `(Default)` item as a first element on the drop-down list.
 				this.add( this.defaultValue, defaultText, defaultText );
 
 				for ( var i = 0; i < names.length; i++ ) {
--- a/plugins/font/plugin.js
+++ b/plugins/font/plugin.js
@@ -4,7 +4,7 @@
  */
 
 ( function() {
-	function addCombo( editor, comboName, styleType, lang, entries, defaultLabel, styleDefinition, order ) {
+	function addCombo( editor, comboName, styleType, lang, entries, defaultLabel, styleDefinition, order, twoStateList, clearFontAvailable ) {
 		var config = editor.config,
 			style = new CKEDITOR.style( styleDefinition );
 
@@ -89,9 +89,13 @@
 			init: function() {
 				this.startGroup( lang.panelTitle );
 
+				// Adding (clear) option to list.
+				if ( clearFontAvailable ) {
+					this.add( 'cke-clear', '<span>(clear)</span>', '(clear)' );
+				}
+
 				for ( var i = 0; i < names.length; i++ ) {
 					var name = names[ i ];
-
 					// Add the tag entry to the panel list.
 					this.add( name, styles[ name ].buildPreview(), name );
 				}
@@ -106,7 +110,7 @@
 
 				// When applying one style over another, first remove the previous one (http://dev.ckeditor.com/ticket/12403).
 				// NOTE: This is only a temporary fix. It will be moved to the styles system (http://dev.ckeditor.com/ticket/12687).
-				if ( previousValue && value != previousValue ) {
+				if ( previousValue && ( value != previousValue || value === 'cke-clear' ) ) {
 					var previousStyle = styles[ previousValue ],
 						range = editor.getSelection().getRanges()[ 0 ];
 
@@ -157,7 +161,9 @@
 					}
 				}
 
-				editor[ previousValue == value ? 'removeStyle' : 'applyStyle' ]( style );
+				if ( value !== 'cke-clear' ) {
+					editor[ twoStateList && previousValue == value ? 'removeStyle' : 'applyStyle' ]( style );
+				}
 
 				editor.fire( 'saveSnapshot' );
 			},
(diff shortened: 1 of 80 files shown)


plugins/font/plugin.js
```diff
@@ -87,11 +87,17 @@
...
 			init: function() {
+				var name,
+					defaultValue = 'cke-default',
+					defaultText = '(' + editor.lang.common.optionDefault + ')';
+
 				this.startGroup( lang.panelTitle );
 
-				for ( var i = 0; i < names.length; i++ ) {
-					var name = names[ i ];
+				// Add `(Default)` item as first element on the drop down list.
```
f1ames: > // Add `(Default)` item as **a** first element on the drop down list.

plugins/font/plugin.js
```diff
@@ -102,27 +108,35 @@
...
+					style = styles[ value ],
+					previousStyle,
+					range,
+					path,
+					matching,
+					startBoundary,
+					endBoundary,
+					node,
+					bm,
+					defaultValue = 'cke-default';
```
f1ames: Same default value is already defined in an `init` method, maybe it could be extracted somewhere to be more "global"? Especially that it is used also in other places (tests).

plugins/format/plugin.js
```diff
@@ -64,10 +64,10 @@ CKEDITOR.plugins.add( 'format', {
 				editor.focus();
 				editor.fire( 'saveSnapshot' );
 
-				var style = styles[ value ],
-					elementPath = editor.elementPath();
+				var style = styles[ value ];
 
-				editor[ style.checkActive( elementPath, editor ) ? 'removeStyle' : 'applyStyle' ]( style );
+				// Always aply style, do not allow on toggle it by clicking on list item (#584).
```
f1ames: > // Always **apply** style, do not allow on toggle it by clicking on list item (#584).

plugins/format/plugin.js
```diff
@@ -64,10 +64,10 @@ CKEDITOR.plugins.add( 'format', {
...
+				editor.applyStyle( style );
```
f1ames: By default all format tags are block elements so reapplying them to partial selection doesn't do anything. However, if you change the config like:

```
config.format_p = { element: 'span', attributes: { 'class': 'normalPara' } };
```

You can apply it to partial selection despite the fact it was already applied (so it creates nested tags). This is an edge case IMHO but still we should handle this. You could just check if style is active (as before `style.checkActive( elementPath, editor )`) and apply it only if it is not already there.

tests/plugins/font/font.js
```diff
@@ -170,6 +172,43 @@
 				'<p>x<span style="font-size:12px"><em>foo</em></span><em><span style="font-size:24px">bar</span></em>x@</p>' );
 		},
 
+		// #584
```
f1ames: Is there any test checking if second click (reapplying) of the same font/size doesn't change the content?

tests/plugins/font/manual/preservestyle.md
```diff
@@ -0,0 +1,16 @@
+@bender-tags: 4.8.0, feature, 584
+@bender-ui: collapsed
+@bender-ckeditor-plugins: wysiwygarea, toolbar, font, elementspath, sourcearea
+
+1. Select text
```
f1ames: Each list item should end with `.`. `Select some text` sounds a little better.

tests/plugins/font/font.js
f1ames: I see there is on in a `format` plugin. Still those are separate plugin with different code so it should be tested for both.

msamsel:
I hope it is fixed now :)

plugins/format/plugin.js
```diff
@@ -67,7 +67,10 @@ CKEDITOR.plugins.add( 'format', {
 				var style = styles[ value ],
 					elementPath = editor.elementPath();
 
-				editor[ style.checkActive( elementPath, editor ) ? 'removeStyle' : 'applyStyle' ]( style );
+				// Always apply style, do not allow on toggle it by clicking on list item (#584).
```
f1ames: > Always apply style, do not allow **to** toggle it by clicking on **corresponding** list item (#584).

?

tests/plugins/font/font.js
```diff
@@ -35,8 +48,10 @@
 
 				this.wait( function() {
 					// Click again to exit the style.
+					// Since 4.8.0 2nd click into this same menu item do not unselect it.
```
f1ames: > Since 4.8.0, 2nd click **on the** same menu item does not unselect it.

tests/plugins/font/font.js
```diff
@@ -35,8 +48,10 @@
...
+					// It is required to click into '(Default)' option to reset style (#584).
```
f1ames: > It is required to click **on the** '(Default)' option to reset style (#584).

tests/plugins/format/plugin.js
```diff
@@ -25,5 +25,24 @@ bender.test( {
...
+			combo = editor.ui.get( name );
+
+		bot.setHtmlWithSelection( '<h1>f^oo</h1>' );
+
+		// Apply the same style format to element already styled.
+		assert.areSame( CKEDITOR.TRISTATE_OFF, combo.getState(), 'check state OFF' );
+		bot.combo( name, function( combo ) {
+			assert.areSame( CKEDITOR.TRISTATE_ON, combo.getState(), 'check state ON when opened' );
+			combo.onClick( 'h1' );
+			assert.areSame( '<h1>f^oo</h1>', bot.htmlWithSelection(), 'applied h1 block style' );
```
f1ames: > applied h1 block style

`h1` was not applied here as it was already there. The test (and assert) check if `h1` is not removed when applied second time, so maybe _preserved h1 block style_ or something similar?

msamsel:
It seems that sometimes IE8 on some system can have strange behaviour. I removed checking if the format panel is opened and left only checking styles if it's properly applied.

PR #642, opened 2017-07-12
`cloneContents` properly return document fragment

### LINKED ISSUE #426: CloneContents selects whole element when selection starts at the start of that element

Bug

http://dev.ckeditor.com/ticket/17030

1. Put attached sample into ```ckeditor/samples``` folder
2. Open browser console 
3. Select e.g. ```[sample te]xt```. It is important that selection starts at the beginning of the element. If you select e.g. ```sam[ple te]xt``` it will work as expected
4. Inspect documentFragment in console

### Expected result

Text should contain ```sample te```

### Actual result

Text contains ```sample text```

## Other details

Problem can be reproduced from CKEditor 4.5.0 in all browsers.

msamsel:
bug fix

yep!

### This PR contains

- [x] Unit tests
- [x] Manual tests

## What changes did you make?

Modify `if` statement to accept situation when selection start and ends in this same element, but `startNode` has type of `element` where `endNode` has type of `text`

close #426

### DIFF
--- a/tests/core/dom/range/clonecontents.js
+++ b/tests/core/dom/range/clonecontents.js
@@ -538,7 +538,7 @@
 			assert.isInnerHtmlMatching( '<p>Foo bar</p>', docFrag.getHtml(), 'Cloned HTML' );
 		},
 
-		// #426
+		// Variety of edge test cases with selection range in text and elements nodes inside one element and multiple ones (#426).
 		'test cloneContents - inner selection1': function() {
 			this.assertHtmlFragment( this.editors.classic, '<p>foo <strong>[bar} baz</strong> foo</p>', 'bar' );
 		},
--- a/tests/core/dom/range/clonecontents.js
+++ b/tests/core/dom/range/clonecontents.js
@@ -233,7 +233,6 @@
 
 			root.setHtml( 'foo<b>bar</b>' );
 			doc.getBody().append( root );
-
 			var startContainer = root.getFirst(),
 				endContainer = root.getLast().getFirst();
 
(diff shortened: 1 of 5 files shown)


f1ames:
I think I run the unit tests with some local changes so I have to recheck it.

f1ames:
I rechecked test, Edge/IE11 is fine, however Safari still fails (updated the comment).

f1ames:
I also run those added unit test without the fix in a code and there is only on test failing on Chrome, FF, Edge, IE9 - IE11 which is `tests/core/dom/range/clonecontents test cloneContents - inner selection1`. Maybe we could reduce the amount of tests or find more failing cases (if there is more than one).

None test is failing on IE8 - you should probably check if the issue is reproducible there, if not we may skip this browser.

tests/core/dom/range/clonecontents.js
```diff
@@ -526,6 +536,52 @@
 
 			// See: execContentsAction in range.js.
 			assert.isInnerHtmlMatching( '<p>Foo bar</p>', docFrag.getHtml(), 'Cloned HTML' );
+		},
+
+		// Variety of edge test cases with selection range in text and elements nodes inside one element and multiple ones (#426).
+		'test cloneContents - inner selection1': function() {
+			this.assertHtmlFragment( this.editors.classic, '<p>foo <strong>[bar} baz</strong> foo</p>', 'bar' );
```
msamsel: tests was added at the end.

msamsel:
Because IE8 and Safari handling a little bit differently with selection than other browser, so I have to ignore part of the tests on those browsers. Generally they kept selection inside specific node, what would be repetition of earlier tests.

PR #653, opened 2017-07-17
Fix error which appear with `columnDelete` command when table contained colspans

### LINKED ISSUE #577: Removing table columns throw an error

Bug

### Case 1
1. Open http://sdk.ckeditor.com/samples/table.html
2. Select whole 2013 section inside the table (see screenshot below)
3. Right click inside the selection and choose "Column"→"Delete Columns"

Example of selected columns

#### Expected result

Columns are deleted.

#### Actual result

Columns are not deleted and the error is thrown:

> ckeditor.js:66 Uncaught TypeError: Cannot read property 'ownerDocument' of undefined

### Case 2
1. Open http://sdk.ckeditor.com/samples/table.html
2. Select 2013 section without the header inside the table (see screenshot below)
3. Right click inside the selection and choose "Column"→"Delete Columns"

Example of selected columns

#### Expected result

Columns are deleted.

#### Actual result

Columns are deleted, but the error is thrown:

> ckeditor.js:66 Uncaught TypeError: Cannot read property 'ownerDocument' of undefined

Additionally selection is put inside incorrect cell (see screenshot below).

Incorrect position of cursor

## Other details

* CKEditor version: 4.7.0

msamsel:
bug fix

yes

### This PR contains
- [x] Unit tests
- [x] Manual tests

## What changes did you make?
- fix bug which forbid to delete columns when those were under colspans.
- fix bug where only last column was removed when colspan cell was selected.
- fix bug where cursor position might won't be returned, what generat the console error.
- add some brackets `{}` to if statements.

close #577

### DIFF
--- a/plugins/tabletools/plugin.js
+++ b/plugins/tabletools/plugin.js
@@ -304,24 +304,32 @@
 
 	function deleteColumns( selection ) {
 		function processSelection( selection ) {
-			// If selection leakt to next td/th cell, then move it back to previous cell.
-			var ranges = selection.getRanges();
+			// If selection leak to next td/th cell, then preserve it in previous cell.
+
+			var ranges,
+				range,
+				endNode,
+				endNodeName,
+				previous;
+
+			ranges = selection.getRanges();
 			if ( ranges.length !== 1 ) {
 				return selection;
 			}
 
-			var range = ranges[0];
-			if ( range.endOffset !== 0 ) {
+			range = ranges[0];
+			if ( range.collapsed || range.endOffset !== 0 ) {
 				return selection;
 			}
 
-			var endNode = range.endContainer;
-			var endNodeName = endNode.getName().toLowerCase();
+			endNode = range.endContainer;
+			endNodeName = endNode.getName().toLowerCase();
 			if ( !( endNodeName === 'td' || endNodeName === 'th' ) ) {
 				return selection;
 			}
+
 			// Get previous td/th element or the last from previous row.
-			var previous = endNode.getPrevious();
+			previous = endNode.getPrevious();
 			if ( !previous ) {
 				previous = endNode.getParent().getPrevious().getLast();
 			}
@@ -339,8 +347,9 @@
 			return range.select();
 		}
 
-		// Problem occures only on webkit in case of native selection (#577
-		if ( CKEDITOR.env.webkit && !selection.isFake && selection.getNative().type === 'Range' ) {
+		// Problem occures only on webkit in case of native selection (#577).
+		// Upstream: https://bugs.webkit.org/show_bug.cgi?id=175131, https://bugs.chromium.org/p/chromium/issues/detail?id=752091
+		if ( CKEDITOR.env.webkit && !selection.isFake ) {
 			selection = processSelection( selection );
 		}
 
--- a/plugins/tabletools/plugin.js
+++ b/plugins/tabletools/plugin.js
@@ -303,7 +303,6 @@
 	}
 
 	function deleteColumns( selection ) {
-		// debugger;
 		var ranges = selection.getRanges(),
 			cells = getSelectedCells( selection ),
 			firstCell = cells[ 0 ],
@@ -321,7 +320,7 @@
 				// #577
 				// Map might contain multiple times this same element, because of existings collspan.
 				// We don't want to overwrite startIndex in such situation and take first one.
-				if ( !startColIndex && map[ i ][ j ] == firstCell.$ ) {
+				if ( startColIndex === undefined && map[ i ][ j ] == firstCell.$ ) {
 					startColIndex = j;
 				}
 				if ( map[ i ][ j ] == lastCell.$ ) {
--- a/plugins/tabletools/plugin.js
+++ b/plugins/tabletools/plugin.js
@@ -303,6 +303,7 @@
 	}
 
 	function deleteColumns( selection ) {
+		// debugger;
 		var ranges = selection.getRanges(),
 			cells = getSelectedCells( selection ),
 			firstCell = cells[ 0 ],
@@ -317,10 +318,15 @@
 		// Figure out selected cells' column indices.
 		for ( var i = 0, rows = map.length; i < rows; i++ ) {
 			for ( var j = 0, cols = map[ i ].length; j < cols; j++ ) {
-				if ( map[ i ][ j ] == firstCell.$ )
+				// #577
+				// Map might contain multiple times this same element, because of existings collspan.
+				// We don't want to overwrite startIndex in such situation and take first one.
+				if ( !startColIndex && map[ i ][ j ] == firstCell.$ ) {
 					startColIndex = j;
-				if ( map[ i ][ j ] == lastCell.$ )
+				}
+				if ( map[ i ][ j ] == lastCell.$ ) {
 					endColIndex = j;
+				}
 			}
 		}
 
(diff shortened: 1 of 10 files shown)


plugins/tabletools/plugin.js
```diff
@@ -325,27 +330,34 @@
...
 
-		var firstRowCells = table.$.rows[ 0 ] && table.$.rows[ 0 ].cells;
-
 		// Where to put the cursor after columns been deleted?
 		// 1. Into next cell of the first row if any;
 		// 2. Into previous cell of the first row if any;
 		// 3. Into table's parent element;
-		var cursorPosition = new CKEDITOR.dom.element( firstRowCells[ startColIndex ] || ( startColIndex ? firstRowCells[ startColIndex - 1 ] : table.$.parentNode ) );
+		var cursorPosition;
+		if ( map[ 0 ].length - 1 > endColIndex && map[ 0 ][ endColIndex + 1 ].cellIndex !== -1 ) {
```
Comandeer: Hmm… I don't get the logic behind this condition. It seems to work, but I don't understand why :D Could elaborate?

tests/plugins/tabletools/columndeletion.js
```diff
@@ -0,0 +1,74 @@
+/* bender-tags: editor,unit */
```
Comandeer: Tests connected with `tableselection` plugin should be inserted into `tableselection/integrations` directory. It allows to easily find all issues related to integrations between `tableselection` and other plugins.

tests/plugins/tabletools/manual/columndeletionerror.md
```diff
@@ -0,0 +1,14 @@
+@bender-tags: 4.7.2, bug, 577
```
Comandeer: Tests connected with `tableselection` plugin should be inserted into `tableselection/integrations` directory. It allows to easily find all issues related to integrations between `tableselection` and other plugins.

tests/plugins/tabletools/manual/columndeletionerror.md
```diff
@@ -0,0 +1,14 @@
...
+@bender-ui: collapsed
+@bender-ckeditor-plugins: wysiwygarea, toolbar, table, tabletools, tableselection, contextmenu, undo
+
+1. Open browser console.
+1. Select some cells somwhere under merged header.
+1. Selection should use `tableselection` plugin.
+1. Right click to open cntentxt menu.
```
Comandeer: Please fix typos in test description.

tests/plugins/tabletools/columndeletion.js
```diff
@@ -0,0 +1,74 @@
...
+		// #577
+		'test remove first column': function() {
+			// tests requires tableselection
+			if ( ( CKEDITOR.env.ie && CKEDITOR.env.version < 11 ) || bender.tools.env.mobile ) {
+				assert.ignore();
+			}
+			this.doTest( 'table-1', 'columnDelete' );
+		},
+
+		'test remove 2 last columns by single row selection': function() {
```
Comandeer: Please add ticket references before every test connected with it.

plugins/tabletools/plugin.js
```diff
@@ -325,27 +330,34 @@
...
 		}
 
-		var firstRowCells = table.$.rows[ 0 ] && table.$.rows[ 0 ].cells;
-
 		// Where to put the cursor after columns been deleted?
 		// 1. Into next cell of the first row if any;
 		// 2. Into previous cell of the first row if any;
 		// 3. Into table's parent element;
-		var cursorPosition = new CKEDITOR.dom.element( firstRowCells[ startColIndex ] || ( startColIndex ? firstRowCells[ startColIndex - 1 ] : table.$.parentNode ) );
+		var cursorPosition;
```
Comandeer: Actually I wonder if it could be moved to `tableselection` plugin using its internal `customizeTableCommand` function.

msamsel: Generally `tabletools` doesn't depend on `tableselection` so we cannot assume that `tableselection` will be loaded.

plugins/tabletools/plugin.js
msamsel: First part of condition `map[ 0 ].length - 1 > endColIndex` check if there was some columns on the right after deleted part.
- `map[ 0 ].length - 1` is the last index of possible column value
- `endColIndex` is last index of deleted column.

Second part of condition `map[ 0 ][ endColIndex + 1 ].cellIndex !== -1` checks, if cell (`endColIndex + 1`) after deleted column, where we want to put cursor, is still in table. Theoretically it should be true, if first part is true ;)

tests/plugins/tabletools/columndeletion.js
msamsel: I'll prepare two sets of test with and without tableselection plugin to cover more cases.

plugins/tabletools/plugin.js
msamsel: I removed second part it seems to be confusing.

tests/plugins/tabletools/manual/columndeletionerror.md
msamsel: it's done. ;)

msamsel:
@Comandeer 
It's actually browser problem. I've report it to them:
- https://bugs.webkit.org/show_bug.cgi?id=175131
- https://bugs.chromium.org/p/chromium/issues/detail?id=752091

Browser natively, after opening context menu, highlight clicked word. In this case it cause enlarging selection to next cell if we're at the end of cell.

Do you think that I should prepare some custom solution to fix in CKEditor?

Comandeer:
There are already some workarounds for context menu in our `tableselection` plugin, so it won't be any extraordinary situation. Such temporary workaround could be deleted when upstream issues will be fixed.

plugins/tabletools/plugin.js
```diff
@@ -296,6 +296,47 @@
...
+					return selection;
+				}
+			}
+
+			range.setEndAt( previous, CKEDITOR.POSITION_BEFORE_END );
+			return range.select();
+		}
+
+		// Problem occures only on webkit in case of native selection (#577
+		if ( CKEDITOR.env.webkit && !selection.isFake && selection.getNative().type === 'Range' ) {
```
Comandeer: Is the last part of the statement really needed? Probably it could be replaced with checking if range is non-collapsed.

plugins/tabletools/plugin.js
```diff
@@ -296,6 +296,47 @@
...
+				if ( !previous ) {
+					return selection;
+				}
+			}
+
+			range.setEndAt( previous, CKEDITOR.POSITION_BEFORE_END );
+			return range.select();
+		}
+
+		// Problem occures only on webkit in case of native selection (#577
```
Comandeer: It'll be nice to link also upstream issue here.

tests/plugins/tabletools/manual/columndeletionerror.md
```diff
@@ -0,0 +1,24 @@
+@bender-tags: 4.7.2, bug, 577
+@bender-ui: collapsed
+@bender-ckeditor-plugins: wysiwygarea, toolbar, table, tabletools, contextmenu, undo
+
+----
+1. Open browser console.
+1. Select some cells somewhere under merged header.
+1. Selection should use native selection.
+1. Right click to open context menu.
+1. Select Column -> Delete Column.
```
Comandeer: It's "Delete Column**s**", not "~Column~".

msamsel:
@Comandeer mentioned behaviour seems to be not related to this bug.
* You can reproduce your steps but instead of right-click use shortcut to open context menu `Shift + F10`. You will see that applying undo after that will restore proper collapsed selection.
* You can modify your steps and after opening context menu close it with ESC button. You will see that selection is extend to both cells even without using column delete.

Entire situation has happens because of using `right-click` not because of using `column delete`. And that's why such case should be fixed separately not in a scope of this task.

PR #680, opened 2017-07-26
T/605 Preserve trailing space for inline widget

### LINKED ISSUE #605: Inline widget does not preserve trailing space

Bug

1. Create a widget plugin for an inline element (eg: `span`)
2. Load some data in CKEditor with a trailing space inside the element to be converted to a widget (eg: `<span>lorem </span>ipsum` )

see JSFiddle here: https://jsfiddle.net/cf5d457L/5/

### Expected result

The following text to be displayed: `lorem ipsum`

### Actual result

The following text is displayed (notice the missing space): `loremipsum`

## Other details

* Browser: all
* OS: all
* CKEditor version: latest (4.7.0)
* Installed CKEditor plugins: the Full preset together with all other plugins created by CKSource

beatadelura:

Bug fix

Yes

### This PR contains

- [X] Unit tests
- [X] Manual tests

## What changes did you make?

I'm replacing initial and trailing space with `&nbsp`.

---

Closes #605


plugins/widget/plugin.js
```diff
@@ -2177,6 +2177,16 @@
 							element.attributes[ 'data-cke-widget-data' ] = encodeURIComponent( JSON.stringify( data ) );
 							element.attributes[ 'data-cke-widget-upcasted' ] = 1;
 
+							// Preserve initial and trailing space by replacing white space with &nbsp; (#605).
```
f1ames: I am not sure if this method is the right place for such transformations. As it should be placed in `toHtml` listener as it is now, I think it should be put a little further (e.g. https://github.com/ckeditor/ckeditor-dev/pull/680/files#diff-d932030cd4e22ec8b0e30a362269b7e8R2806). The `createUpcastIterator` has different responsibility so it does not fit here.

tests/plugins/widget/widgetwithspaces.js
```diff
@@ -0,0 +1,66 @@
...
+		config: {
+			allowedContent: true
+		}
+	};
+
+	// 605
+	bender.test( {
+		'test trailing space': function() {
+			var editor = this.editor;
+			editor.widgets.add( 'spanwidget', {
```
beatadelura: Good idea 🌞

tests/plugins/widget/widgetwithspaces.js
```diff
@@ -0,0 +1,66 @@
...
+					}
+				},
+				upcast: function( element ) {
+					return element.name == 'span';
+				}
+			} );
+
+			editor.setData( '<span>lorem </span>ipsum' );
+
+			setTimeout( function() {
```
beatadelura: Sounds good! 👍

plugins/widget/plugin.js
```diff
@@ -2701,6 +2733,24 @@
 				var attrs = element.attributes,
 					widget, widgetElement;
 
+				// Reset initial and trailing space by replacing &nbsp; with white space (#605).
+				if ( 'data-cke-widget-white-space' in attrs ) {
```
f1ames: Same here, a more generic solution is needed. Maybe when replacing spaces you could tag an element in which space was replaced with something similar to `data-cke-widget-white-space` so then it is easier to find it in this code (assuming that there can be many nested elements)? Additionally, you may also mark widgets (same as now) so you don't have to iterate over all of them.

tests/plugins/widget/manual/widgetwithspaces.html
```diff
@@ -0,0 +1,24 @@
+<div id="editor"></div>
+
+<script>
+	CKEDITOR.replace( 'editor', {
+        extraAllowedContent: 'span',
+	    extraPlugins: 'spanwidget'
```
f1ames: There are spaces mixed with tabs here (that's why GH renders it like):
image

Please keep the spacing consistent.

tests/plugins/widget/manual/widgetwithspaces.md
```diff
@@ -0,0 +1,16 @@
+@bender-tags: 4.7.2, bug, 605
+@bender-ui: collapsed
+@bender-ckeditor-plugins: widget, wysiwygarea, toolbar, sourcearea
+
+**Scenario:**
+1. Click `Source` button.
+2. Input `lorem<span> ipsum&nbsp;dolor sit </span>amet` into source area.
```
f1ames: Maybe **Paste `lorem...` into source area** ?

tests/plugins/widget/manual/widgetwithspaces.md
```diff
@@ -0,0 +1,16 @@
...
+3. Click `Source` button.
+
+**Expected result:**
+
+The text displayed should equal: `lorem ipsum dolor sit amet` and `Source` data should be `<p>lorem<span> ipsum&nbsp;dolor sit </span>amet</p>`.
+
+**Unexpected result:**
+
+There are a missing spaces in a text: `loremipsum dolor sitamet`.
```
f1ames: > There are missing spaces in text in `WYSIWYG` mode: `loremipsum dolor sitamet`.

?

tests/plugins/widget/widgetwithspaces.js
```diff
@@ -0,0 +1,84 @@
...
+( function() {
+	'use-strict';
+
+	bender.editor = {
+		config: {
+			allowedContent: true
+		}
+	};
+
+	// #605
```
f1ames: If all tests covers `#605`, I think it will be enough to mark it here only.

tests/plugins/widget/manual/widgetwithspaces.md
beatadelura: 👻

tests/plugins/widget/manual/widgetwithspaces.md
beatadelura: 👍

tests/plugins/widget/widgetwithspaces.js
beatadelura: 👍

tests/plugins/widget/manual/widgetwithspaces.html
beatadelura: 👍

plugins/widget/plugin.js
```diff
@@ -2053,9 +2053,62 @@
...
+				firstTextNode.parent.attributes[ 'data-cke-white-space-first' ] = 1;
+				firstTextNode.value = firstTextNode.value.replace( /^\s/g, '&nbsp;' );
+			}
+
+			// Check whether the value of the last text node contains white space at the end and replace it with &nbsp;.
+			if ( lastTextNode.value && lastTextNode.value.match( /\s$/g ) ) {
+				lastTextNode.parent.attributes[ 'data-cke-white-space-last' ] = 1;
+				lastTextNode.value = lastTextNode.value.replace( /\s$/g, '&nbsp;' );
+			}
+		}
```
f1ames: This whole code block contains some redundant parts (checking for spaces in text node), I suggest to modify it as follows:

First `if` checks only if it operates on widget element:
`if ( el.attributes[ 'data-cke-widget-data' ] ) {`

`firstTextNode` and `lastTextNode` declaration is inside the above `if` (as this are used only here) with a third variable like `spacesReplaced`:
```
if ( el.attributes[ 'data-cke-widget-data' ] ) {
    var firstTextNode = getFirstTextNode( el ),
        lastTextNode = getLastTextNode( el ),
        spacesReplaced = false;
```

In each if where space is replaced you should set `spacesReplaced` to true:
```
if ( firstTextNode && firstTextNode.value && firstTextNode.value.match( /^\s/g ) ) { // Remember to check firstTextNode as previous if was modified.
    firstTextNode.parent.attributes[ 'data-cke-white-space-first' ] = 1;
    firstTextNode.value = firstTextNode.value.replace( /^\s/g, '&nbsp;' );
    spacesReplaced = true;
}
```

Finally, add `data-attribute` to widget element:
```
if ( spacesReplaced ) {
    el.attributes[ 'data-cke-widget-white-space' ] = 1;
}
```

This makes this code longer (3 lines more), but should be more readable and does not repeat any checks. WDYT?

plugins/widget/plugin.js
```diff
@@ -2053,9 +2053,62 @@
...
+
 		var parent = el.parent;
-		if ( parent.type == CKEDITOR.NODE_ELEMENT && parent.attributes[ 'data-cke-widget-wrapper' ] )
+
+		if ( parent.type == CKEDITOR.NODE_ELEMENT && parent.attributes[ 'data-cke-widget-wrapper' ] ) {
 			parent.replaceWith( el );
+		}
+	}
+
+	// Get first text node (#605).
```
f1ames: I think ticket number is not needed here as this function is used by the code which already contains proper description. Same for the second function below.

plugins/widget/plugin.js
```diff
@@ -2053,9 +2053,62 @@
...
+	function getFirstTextNode( el ) {
+		var firstIteration = true,
```
f1ames: You can use [`find`](https://docs.ckeditor.com/#!/api/CKEDITOR.htmlParser.element) method in both `getFirstTextNode` and `getLastTextNode` functions. `Find` returns array of `CKEDITOR.htmlParser.node`, so to find first text node:

```
return el.find( function( node ) {
    return node.type === 3;
}, true ).shift();
```

And to find last:
```
return el.find( function( node ) {
    return node.type === 3;
}, true ).pop();
```

plugins/widget/plugin.js
```diff
@@ -2701,6 +2754,22 @@
 				var attrs = element.attributes,
 					widget, widgetElement;
 
+				// Reset initial and trailing space by replacing &nbsp; with white space (#605).
+				if ( 'data-cke-widget-white-space' in attrs ) {
+					var firstTextNode = getFirstTextNode( element ),
```
f1ames: The main point of adding 'data-cke-white-space-first|last' data attributes to specific elements was to enable the easy way to find them later without a need to iterate over whole widget structure. However, to find them you still need to use `find` method same as in `getFirstTextNode`/`getLastTextNode` so it doesn't make a difference. Lets leave it as it is.

plugins/widget/plugin.js
```diff
@@ -2701,6 +2754,22 @@
...
+						lastTextNode = getLastTextNode( element );
+
+					// Check whether the value of the text node contains &nbsp; at the beginning and replace it with white space.
+					if ( firstTextNode.parent.attributes[ 'data-cke-white-space-first' ] && firstTextNode.value.match( /^&nbsp;/g ) ) {
```
f1ames: Is there any reason why `firstTextNode.value.match( /^&nbsp;/g )` is checked here? If the element has `data-cke-white-space-first` attribute it already means that space was replaced there, so the second check seems redundant.

plugins/widget/plugin.js
```diff
@@ -2053,9 +2053,62 @@
 	//
 	// @param {CKEDITOR.htmlParser.element} el
 	function cleanUpWidgetElement( el ) {
+		// Preserve initial and trailing space by replacing white space with &nbsp; (#605).
+		var firstTextNode = getFirstTextNode( el ),
+			lastTextNode = getLastTextNode( el );
+
+		if ( firstTextNode || lastTextNode && el.attributes[ 'data-cke-widget-data' ] ) {
```
beatadelura: It's true. My mistake :)

plugins/widget/plugin.js
beatadelura: Looks much better 👍

plugins/widget/plugin.js
beatadelura: 👍

plugins/widget/plugin.js
beatadelura: I agree with you 👍

tests/plugins/widget/manual/widgetwithspaces.html
```diff
@@ -0,0 +1,25 @@
...
+		extraAllowedContent: 'span',
+		extraPlugins: 'spanwidget',
+		allowedContent: true
+	} );
+
+	CKEDITOR.plugins.add( 'spanwidget', {
+		requires: 'widget',
+		init: function ( editor )	{
+			editor.widgets.add( 'spanwidget', {
+				editables: {
```
f1ames: Only block widgets support `editables`. Even if you define it like that, the inline widget or its part will not be editable. You can remove the `editables` definition as it does not change anything here (you can see by running this test and trying to edit the `span` element inside widget).

tests/plugins/widget/manual/widgetwithspaces.md
```diff
@@ -0,0 +1,16 @@
...
+@bender-ckeditor-plugins: widget, wysiwygarea, toolbar, sourcearea
+
+**Scenario:**
+1. Click `Source` button.
+2. Paste `lorem<span> ipsum&nbsp;<strong>dolor sit </strong></span>amet` into source area.
+3. Click `Source` button.
+
+**Expected result:**
+
+The text displayed should equal: `lorem ipsum dolor sit amet` and `Source` data should be `<p>lorem<span> ipsum&nbsp;<strong>dolor sit </strong></span>amet`.
```
f1ames: > The displayed text is equal to ... and `Source` data to ...

tests/plugins/widget/manual/widgetwithspaces.md
```diff
@@ -0,0 +1,16 @@
...
+
+**Unexpected result:**
+
+The text displayed in editor equal: `loremipsum dolor sitamet`.
```
f1ames: > The text displayed in the editor is equal to ...

tests/plugins/widget/widgetwithspaces.js
```diff
@@ -0,0 +1,85 @@
...
+	bender.test( {
+
+		init: function() {
+			this.editor.widgets.add( 'spanwidget', {
+				editables: {
```
f1ames: Same here, `editables` not needed.

tests/plugins/widget/manual/widgetwithspaces.md
```diff
@@ -0,0 +1,16 @@
+@bender-tags: 4.8.0, bug, 605
+@bender-ui: collapsed
+@bender-ckeditor-plugins: widget, wysiwygarea, toolbar, sourcearea
+
+**Scenario:**
+1. Click `Source` button.
+2. Paste `lorem<span> ipsum&nbsp;<strong>dolor sit </strong></span>amet` into source area.
```
f1ames: Can we wrap `lorem<span> ipsum&nbsp;<strong>dolor sit </strong></span>amet` inside `textarea` or `input`? On Edge it is impossible to select without content around which makes executing this test very annoying ;)

plugins/widget/plugin.js
```diff
@@ -2054,8 +2064,62 @@
...
+
+			if ( spacesReplaced ) {
+				el.attributes[ 'data-cke-widget-white-space' ] = 1;
+			}
+		}
+	}
+
+	// It returns first text node from the element.
+	//
+	// @param {CKEDITOR.htmlParser.element} el
```
f1ames: `@returns` param missing.

plugins/widget/plugin.js
```diff
@@ -2054,8 +2064,62 @@
...
+	function getFirstTextNode( el ) {
+		return el.find( function( node ) {
+			return node.type === 3;
+		}, true ).shift();
+	}
+
+
+	// It returns last text node from the element.
+	//
+	// @param {CKEDITOR.htmlParser.element} el
```
f1ames: `@returns` param missing.

tests/plugins/widget/manual/widgetwithspaces.md
beatadelura: Of course, we can :)

plugins/widget/plugin.js
beatadelura: Done 👍

plugins/widget/plugin.js
beatadelura: Done 👍

PR #708, opened 2017-08-01
T/638 iframe/anchor is not visible after copy/paste from editor in other tab

### LINKED ISSUE #638: iframe is not visible after Copy pasting content with iframe in google chrome
Reporting a bug

create some content with a iframe in it, open another editor instance try to copy and paste the content with iframe.

It doesnt display the iframe what we pasted but if we switch from edit to source mode and back to edit mode the iframe comes up in the editor.

Below are the screenshots of my test

image

After pasting into a new editor instance

edit mode (Note: here after pasting the content, iframe is not visible)

image

source mode (Note: it has the iframe tag in the source)

image

### Expected result

when copy paste contents with iframes, iframes should be visible in the editor  

### Actual result

iframes are visible on the UI once switch to source mode and then back to edit mode

## Other details

* Browser: Googel Chrome
* OS: windows10
* CKEditor version: 4.7.1
* Installed CKEditor plugins: OOB

beatadelura:

Bug fix

Yes

### This PR contains

- [X] Unit tests
- [X] Manual tests

## What changes did you make?

Checking whether `dataTransfer` comes from the external resource and if the value contains attribute related to the fake element, original data is set.

---

Closes #638


plugins/iframe/plugin.js
```diff
@@ -70,7 +70,6 @@
 		afterInit: function( editor ) {
 			var dataProcessor = editor.dataProcessor,
 				dataFilter = dataProcessor && dataProcessor.dataFilter;
-
```
Comandeer: It seems like unnecessary, accidental change

beatadelura:
The cases introduced by @Comandeer works now but I found a bug which has been reported here: #823.

tests/plugins/clipboard/pastefakeelement.js
```diff
@@ -0,0 +1,89 @@
+/* bender-tags: editor */
```
Comandeer: It'd be nice to have mentioned breakable cases tested here.

plugins/clipboard/plugin.js
```diff
@@ -363,6 +363,25 @@
...
+				// External paste and pasteFilter exists and filtering isn't disabled.
+				if ( evt.data.dataTransfer.getTransferType( editor ) === CKEDITOR.DATA_TRANSFER_EXTERNAL && editor.pasteFilter && !evt.data.dontFilter ) {
+					var dataTransferHtml = CKEDITOR.plugins.clipboard.isCustomDataTypesSupported ?
+							evt.data.dataTransfer.getData( 'text/html', true ) : null,
+						regex = /<img.+?data-cke-realelement.+?>/g,
+						dataValue = evt.data.dataValue;
+
+					if ( dataTransferHtml && dataValue.match( regex ) ) {
+						for ( var i = 0; i < dataValue.match( regex ).length; i++ ) {
+							dataValue = dataValue.replace( dataValue.match( regex )[ i ], dataTransferHtml.match( regex )[ i ] );
```
Comandeer: It looks like it could be optimized by moving result of `dataValue.match( regex )` into a variable.

plugins/clipboard/plugin.js
```diff
@@ -363,6 +363,25 @@
 					}, 0 );
 				}
 			}, null, null, 1000 );
+
+			// Instert fake element from external data transfer (#638).
+			editor.on( 'paste', function( evt ) {
```
Comandeer: I'm wondering if it shouldn't be moved into `fakeobjects` plugin as without it there is no way to insert such element into the editor. In that case tests also could be moved.

tests/plugins/clipboard/pastefakeelement.js
Comandeer: There is also one more issue with unit tests: they pass even without the fix – as whole pasting is treated as from the same page/tab.

tests/plugins/fakeobjects/manual/pastefakeelement.md
```diff
@@ -0,0 +1,33 @@
+@bender-tags: 4.8.0, bug, clipboard, 638
+@bender-ui: collapsed
+@bender-ckeditor-plugins: wysiwygarea, toolbar, undo, basicstyles, iframe, clipboard, sourcearea, link
+
+**Scenario 1:**
+
+1. Open this page in new tab.
+2. Click `Source` in editor and paste:
+`<p>Lorem ipsum</p><p><iframe frameborder="0" height="100" src="https://example.com" width="100"></iframe></p><p>cke-real-element-type</p>`
```
Comandeer: This step could be omitted and instead this HTML could be already inserted into the editor.

tests/plugins/fakeobjects/manual/pastefakeelementwithimage.md
```diff
@@ -0,0 +1,17 @@
+@bender-tags: 4.8.0, bug, clipboard, 638
+@bender-ui: collapsed
+@bender-ckeditor-plugins: wysiwygarea, toolbar, undo, basicstyles, iframe, clipboard, sourcearea, link, image2
+
+**Scenario:**
+
+1. Open this page in new tab.
+2. Click `Source` in editor and paste:
```
Comandeer: This step could be omitted and instead this HTML could be already inserted into the editor.

beatadelura:
> In pastefakeelementwithimage manual test iframe after paste is slightly more narrow than the original, copied one.

@Comandeer I reported a separate issue https://github.com/ckeditor/ckeditor-dev/issues/823.

beatadelura:
> Unit tests are still passing even without the fix. Probably mocking dataTransfer.getTransferType to always return CKEDITOR.DATA_TRANSFER_EXTERNAL could do the trick.

@Comandeer I checked, ` CKEDITOR.DATA_TRANSFER_EXTERNAL` in each test returns 3. I'm not sure which browser did you checked, but this bug occurs in Chrome only. Other browsers work fine. So except Chrome tests will pass even without the fix. I added [proper condition](https://github.com/ckeditor/ckeditor-dev/pull/708/files#diff-6ec9a50df53ca6927c755b355ce7e475R67) to run this block of code on Chrome only.

I did not ignore tests in other browsers because I thought, it's good to check this behavior on them. What do you think?

The problem was only with test `test paste iframe with image2` - this one passed even without the fix. It should work now :)

f1ames:
This is a minor issue on which we won't be working for some time, let's close this PR and get back to it if necessary.

PR #734, opened 2017-08-07
[#704] Fix broken widget after using Ctrl+Z in Edge

### LINKED ISSUE #704: Broken content in article editor after using shortcut CTRL+Z

Bug

1. Open Edge browser.
2. Go to http://ckeditor.com/ and choose `Article Editor`.
3. Put caret at the end of the text.
4. Press `Enter`.
5. Use `CTRL+Z` shortcut to undo.

### Expected result

The layout looks the same as before use `CTRL+Z` shortcut.

### Actual result

The layout is broken.

## Other details
* Support ticket number: 32334
* Browser: Microsoft Edge 38.14393.1066.0, Microsoft EdgeHTML 14.14393
* OS: Windows 10

If you use the F12 developer tools to change the User Agent String to Chrome the site works as expected.

Here is additional information about the Edge UA String: https://msdn.microsoft.com/library/hh869301(v=vs.85).aspx

msamsel:

Bug fix

yes

### This PR contains

- [x] Unit tests
- [x] Manual tests

## What changes did you make?

* Modify condition which remove unnecessary `div` in Edge.
* Modify condition to work only in Edge up to 14.xxx. (Edge 15 doesn't add `div`).
* Modify tests to respect new condition related to Edge version.

Close #704

### DIFF
--- a/plugins/wysiwygarea/plugin.js
+++ b/plugins/wysiwygarea/plugin.js
@@ -165,7 +165,7 @@
 		if ( CKEDITOR.env.ie && !CKEDITOR.env.edge && editor.enterMode != CKEDITOR.ENTER_P ) {
 			removeSuperfluousElement( 'p' );
 		}
-		// From version Edge 15 additional `div` are not added to the editor.
+		// Starting from Edge 15 additional `div` is not added to the editor.
 		else if ( CKEDITOR.env.edge && editor.enterMode != CKEDITOR.ENTER_DIV && CKEDITOR.env.version < 15 ) {
 			removeSuperfluousElement( 'div' );
 		}
--- a/plugins/wysiwygarea/plugin.js
+++ b/plugins/wysiwygarea/plugin.js
@@ -297,7 +297,7 @@
 				var elements = doc.getElementsByTag( tagName );
 				if ( lockRetain ) {
 					if ( elements.count() == 1 && !elements.getItem( 0 ).getCustomData( 'retain' ) &&
-						!elements.getItem( 0 ).hasAttribute( 'data-cke-temp' ) ) {
+						CKEDITOR.tools.isEmpty( elements.getItem( 0 ).getAttributes() ) ) {
 						elements.getItem( 0 ).remove( 1 );
 					}
 					lockRetain = false;
--- a/plugins/wysiwygarea/plugin.js
+++ b/plugins/wysiwygarea/plugin.js
@@ -161,10 +161,12 @@
 			doc.getDocumentElement().addClass( doc.$.compatMode );
 		}
 
-		// Prevent IE/Edge from leaving a new paragraph/div after deleting all contents in body. (http://dev.ckeditor.com/ticket/6966, http://dev.ckeditor.com/ticket/13142)
+		// Prevent IE/Edge from leaving a new paragraph/div after deleting all contents in body (http://dev.ckeditor.com/ticket/6966, http://dev.ckeditor.com/ticket/13142).
 		if ( CKEDITOR.env.ie && !CKEDITOR.env.edge && editor.enterMode != CKEDITOR.ENTER_P ) {
 			removeSuperfluousElement( 'p' );
-		} else if ( CKEDITOR.env.edge && editor.enterMode != CKEDITOR.ENTER_DIV && CKEDITOR.env.version < 15 ) {
+		}
+		// From version Edge 15 additional `div` are not added to the editor.
+		else if ( CKEDITOR.env.edge && editor.enterMode != CKEDITOR.ENTER_DIV && CKEDITOR.env.version < 15 ) {
 			removeSuperfluousElement( 'div' );
 		}
 
--- a/plugins/wysiwygarea/plugin.js
+++ b/plugins/wysiwygarea/plugin.js
@@ -166,7 +166,7 @@
 			removeSuperfluousElement( 'p' );
 		}
 		// Starting from Edge 15 additional `div` is not added to the editor.
-		else if ( CKEDITOR.env.edge && editor.enterMode != CKEDITOR.ENTER_DIV && CKEDITOR.env.version < 15 ) {
+		else if ( CKEDITOR.env.edge && CKEDITOR.env.version < 15 && editor.enterMode != CKEDITOR.ENTER_DIV  ) {
 			removeSuperfluousElement( 'div' );
 		}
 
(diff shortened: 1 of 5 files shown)


plugins/wysiwygarea/plugin.js
```diff
@@ -297,7 +297,7 @@
 				var elements = doc.getElementsByTag( tagName );
 				if ( lockRetain ) {
 					if ( elements.count() == 1 && !elements.getItem( 0 ).getCustomData( 'retain' ) &&
-						!elements.getItem( 0 ).hasAttribute( 'data-cke-temp' ) ) {
+						CKEDITOR.tools.isEmpty( elements.getItem( 0 ).getAttributes() ) ) {
```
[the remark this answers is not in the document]
msamsel: Entire function `removeSuperfluousElement` was design to remove additional `div` which appear in Edge, when all text is removed from editor. This div never had any attribute. Unfortunately author hadn't predict that not only those extra `divs` might appear, but also widgets might be construct based on `div`. And in case of `undo` command entire widget might be redraw so we lost `retain` customData attached to `div`.
This change detects more carefully only those extra divs from Edge, so it becomes more specific. Now divs connected to widgets won't be removed.

tests/plugins/widget/integration/wysiwygarea/manual/preventofremovingdiv.html
```diff
@@ -0,0 +1,19 @@
...
+	Combine the flour, sugar and cocoa powder in a medium bowl. In another small bowl, whisk together the butter and eggs. Stir the two mixtures until just combined.
+	Bake the brownies for 25 to 35 minutes. Remove from the oven and let it cool for 5 minutes.
+	</p>
+</textarea>
+<pre id="output" style="background-color: lightgreen;"></pre>
+<script>
+	var editor = CKEDITOR.replace( 'editor1', {
+		height: '600px'
+	} );
+	setInterval( function() {
```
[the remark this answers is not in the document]
msamsel: Unfortunately `change` event is not working as it should. :( It's execute before `keyup` listener removes div from `wysiwygarea`. I've used `keyup` listener instead.

tests/plugins/widget/integration/wysiwygarea/superflousdoesntremovedivwrapper.js
```diff
@@ -0,0 +1,57 @@
...
+				var editor = bot.editor;
+
+				editor.editable().fire( 'keydown', new CKEDITOR.dom.event( {
+					keyCode: 90,
+					ctrlKey: true,
+					shiftKey: false
+				} ) );
+
+				var snap = editor.getSnapshot();
+				editor.loadSnapshot( snap );
```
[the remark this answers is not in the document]
msamsel: Yes.
`keydown` and `keyup` don't trigger undo command, which triggers re-creating widget (there even is no history to revert it). 
So this simulation works as follows:
1. Simulate `keydown` to register listeners on `div` by superflous function.
2. Reload snapshot, to simulate <kbd>Ctrl</kbd>+<kbd>Z</kbd>, when previous snapshot is restored.
3. Reloading snapshot automatically triggers re-creating widget. Here will be destroyed wrapping div for widget, and `retain` attribute will be lost (as it happens in Edge).
4. Simulate `keyup` to execute proper listener from superflous function.

Re-loading snapshot might be replaced with some kind of triggering undo event, but it would require much more work. First to make some history, and then revert it. So I extract only important part from entire undo process, which is reloading snapshot, what cause the issue in case of using shortcut. This allows on proper simulation behaviour on Edge and proper fail of test in case of Edge environment without bug fix.

plugins/wysiwygarea/plugin.js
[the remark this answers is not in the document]
f1ames: Makes sense 👍

tests/plugins/widget/integration/wysiwygarea/manual/preventofremovingdiv.md
```diff
@@ -0,0 +1,11 @@
+@bender-ui: collapsed
+@bender-tags: bug, 4.7.2, 704
+@bender-ckeditor-plugins: image2, wysiwygarea, undo, sourcearea
+
+----
+1. Put caret in the text on the right and type some text.
+1. Use `Ctrl + Z` or `Cmd + Z` shortcut to make undo few times.
+
+**Expected:** Text returns to previous form. Below editor will be written `div` on green background.
```
f1ames: Maybe:

> Text returns to previous form. Below **the** editor, you should see `div` on green background.

tests/plugins/widget/integration/wysiwygarea/manual/preventofremovingdiv.md
```diff
@@ -0,0 +1,11 @@
...
+
+**Unexpected:** Text moves below the image. Below editor will be written `figure` instead of `div`.
```
f1ames: And:

> Text moves below the image. Below editor, it will be written `figure` instead of `div` on the green background.

tests/plugins/widget/integration/wysiwygarea/superflousdoesntremovedivwrapper.js
```diff
@@ -0,0 +1,58 @@
...
+
+	bender.editor = {
+		config: {
+			allowedContent: true
+		}
+	};
+
+	bender.test( {
+		// #704
+		'test keeping widget wrapper in editor when superflous elemnts are checked': function() {
```
f1ames: Typo:

> test keeping widget wrapper in editor when superfl**u**ous elem**e**nts are checked

tests/plugins/widget/integration/wysiwygarea/manual/preventofremovingdiv.html
```diff
@@ -0,0 +1,22 @@
+<textarea id="editor1" name="editor1">
```
f1ames: This test fails on IE8 due to `addEventListener`, if it should be run only for Edge add a proper `bender.ignore` or fix attaching listener if it should be run also for IE8.

msamsel: I added both things. I know it wasn't necessary, but this not harm anything. ;)

msamsel:
I've made fixes and rebase branch to major. Also switch target branch on github to major.

Issue #678, opened 2017-07-26
Removing toolbar disable possibility to execute commands


msamsel:
## Are you reporting a feature or a bug?

bug

## [Check if the issue is already reported](http://docs.ckeditor.com/#!/guide/dev_issues_readme-section-avoid-duplicates)

*Put all reference links here…*

## Provide detailed reproduction steps (if any)
You can use this example: https://codepen.io/msamsel/pen/gxpOmy?editors=1010

1. Remove toolbar.
1. Register plugin with some command. Here is used `link` plugin.
1. Open console, switch to proper iframe context.
1. Check if command is registered: CKEDITOR.instances.editor.getCommand( 'link' );
1. Try to run command from console: CKEDITOR.instances.editor.execCommand( 'link' );

Alternatively you can also check for `keystrokes` which should also run command here it will be:
 <kbd>Ctrl</kbd>/<kbd>Cmd</kbd>+<kbd>L</kbd>

### Expected result

Dialog with possibility for adding link will appear.

### Actual result

Nothing happen.

## Other details

* CKEditor version: 4.7.1
* Installed CKEditor plugins: link, wysiwygarea, 
* **Disabled** plugins: toolbar

mlewand:
Since #654 just became dependent on this issue, let's solve this issue ASAP. Also due to fact that it has been a PITA for a long, long time. 🙂

dimaip:
We at Neos CMS trigger commands via a custom toolbar (rendered in React) and `execCommand`, without adding any native CKE buttons. For now we have to use `extraAllowedContent` to work around the issue.

msamsel:
@dimaip I've extract the problem to separate issue. You can take look here: #1103. As @jswiderski mention it might be a little bit different problem.

jonnott:
Why have this issue, and #654, both been removed from 4.9.0 milestone? @mlewand @msamsel

PR #744, opened 2017-08-08
Fix dependency between toolbar and clipboard

### LINKED ISSUE #654: Remove dependency on toolbar from clipboard

Task

## Explanation

Currently clipboard plugin depends on two other plugins: notification and toolbar. However it seems that toolbar-specific functions are used only to disallow dropping files on it. Therefore dropping it as a dependency seems like a relatively simple task.

It's kinda-sorta followup of #491.

msamsel:

Bug fix.

yes

### This PR contains

- [x] Unit tests
- [x] Manual tests

## What changes did you make?

* Remove dependancy on toolbar in clipboard plugin
* Improvement in Link and Table plugin to register proper ACF rules even when toolbar is not present. This is required to pass tests (tableselection, autoemebed).

Close #654

### DIFF
--- a/plugins/clipboard/plugin.js
+++ b/plugins/clipboard/plugin.js
@@ -116,7 +116,7 @@
 ( function() {
 	// Register the plugin.
 	CKEDITOR.plugins.add( 'clipboard', {
-		requires: 'notification,toolbar',
+		requires: 'notification',
 		// jscs:disable maximumLineLength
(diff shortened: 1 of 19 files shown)


plugins/link/plugin.js
```diff
@@ -49,20 +49,25 @@
```
Comandeer: It doesn't seem to be part of the fix for #654. If #654 does not make sense without #678, it should be blocked on it.

What's more, if #678 is resolved, these changes will become obsolete.

msamsel:
I haven't thought in such way. I removed commits which change `link` and `table` plugin, but I had to modify `autoembed` and `tableselection` tests instead.

plugins/clipboard/plugin.js
```diff
@@ -1306,8 +1306,13 @@
 			// -------------- DRAGOVER TOP & BOTTOM --------------
 
 			// Not allowing dragging on toolbar and bottom (http://dev.ckeditor.com/ticket/12613).
-			clipboard.preventDefaultDropOnElement( top );
-			clipboard.preventDefaultDropOnElement( bottom );
+			// Remove dependacy on toolbar (#654).
```
Comandeer: Please fix typo here and in other places (it's "dependency", not "dependancy").

mlewand:
@Comandeer actually you're right. Removing this dep would break many dependent plugin, unaware of this long standing issue. Given that I'm all for blocking this issue until #678 is done.

PR #745, opened 2017-08-08
Rich combos reflects whole selection

### LINKED ISSUE #525: Dropdowns reflect only the beginning / start of selection.

Feature request

1. Create 3 paragraphs with 3 different fonts e.g. 
```
<p><span style="font-family:Comic Sans MS,cursive">test1</span></p>
<p><span style="font-family:Georgia,serif">test2</span></p>
<p><span style="font-family:Lucida Sans Unicode,Lucida Grande,sans-serif">test3</span></p>
```
2. Select all content

### Expected result

Font dropdown doesn't show any font.

### Actual result

Font dropdown shows font from first paragraph.

## Other details

* Browser: all
* OS: all
* CKEditor version: 4.0+
* Installed CKEditor plugins: 

**NOTE:** This is how selection and styles work in the editor. Dropdowns always reflect beginning of selection. Perhaps here we could check if selection contains something more than one element and more than one font and if it does, simply not showing any font. Same goes for font-size, format and styles.

Comandeer:

New feature

### This PR contains

- [x] Unit tests
- [x] Manual tests

## What changes did you make?

I've introduced `CKEDITOR.plugins.richcombo` namespace, which currently holds only one method – `createSelectionListener`. It contains common logic for creating `selectionChange` listeners for rich combos (Font, Font Size, Styles and Format). These listeners ensures that when determining the displayed value of rich combo state from the whole, not only the beginning, of selection is considered.

There are however some issues with selecting new text, starting from the same element as previous selection. Such selection does not refresh the state of rich combo.

I'm also not 100% sure about `createSelectionListener`. It simplifies a logic and put it into one place, but at the same time I feel it somehow illogical. Probably it's a place that could be further improved.

closes #525.

### DIFF
--- a/plugins/richcombo/plugin.js
+++ b/plugins/richcombo/plugin.js
@@ -114,7 +114,9 @@ CKEDITOR.plugins.add( 'richcombo', {
 			 * to this button.
 			 */
 			render: function( editor, output ) {
-				var env = CKEDITOR.env;
+				var env = CKEDITOR.env,
+					instance,
+					selLocked;
 
 				var id = 'cke_' + this.id;
 				var clickFn = CKEDITOR.tools.addFunction( function( el ) {
@@ -127,7 +129,7 @@ CKEDITOR.plugins.add( 'richcombo', {
 				}, this );
 
 				var combo = this;
-				var instance = {
+				instance = {
 					id: id,
 					combo: this,
 					focus: function() {
@@ -209,7 +211,7 @@ CKEDITOR.plugins.add( 'richcombo', {
 					instance.onfocus && instance.onfocus();
 				} );
 
-				var selLocked = 0;
+				selLocked = 0;
 
 				// For clean up
 				instance.keyDownFn = keyDownFn;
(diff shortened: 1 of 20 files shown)


plugins/richcombo/plugin.js
```diff
@@ -416,4 +418,96 @@ CKEDITOR.plugins.add( 'richcombo', {
...
+					}
+
+					function runHandler( node ) {
+						if ( node.getCustomData( 'processed_richcombo' ) ) {
+							return;
+						}
+
+						CKEDITOR.dom.element.setMarker( markers, node, 'processed_richcombo', true );
+
+						nodeHandler( node, getNewValue );
```
f1ames: Would it be possible to gather all nodes (as an Array/NodeList), pass it at once to a _callback_ and then check if all elements which passes `checkElementMatch` have the same styling? Is the performance (less loops) or order of processing nodes the reason for using above approach with running _callback_ for each node?

Comandeer:
This PR needs a proper refactor and more testing.

PR #750, opened 2017-08-09
Corrections to table selection pasting improvements PR

mlewand:

Bug fix

- [x] Unit tests
- [x] Manual tests (unit tests inherited from intiial PR)

 ## What changes did you make?

Since paste listener was huge, and very difficult to understand I had to split the logic into smaller pieces, so that it's easier to work with it.

The biggest change is encapsulated `TableSelection` type in tableselection plugin, so that there's no `firstCell`, `lastRow` etc reassigning all over the place. Instead new types take care of that. It also features methods related to the manipulation on the table selection itself.

Also `insertRow` and `insertColumn` in tabletools now returns added DOM elements.

### DIFF
--- a/plugins/tableselection/plugin.js
+++ b/plugins/tableselection/plugin.js
@@ -612,6 +612,69 @@
 			} );
 
 			return tmpContainer.getChildCount() > 1 ? null : tmpContainer.findOne( 'table' );
+		},
+
+		// Performs an actual paste into selectedTableMap based on content in pastedTableMap.
+		pasteTable: function( selectedTable, tableSel, selectedTableMap, pastedTableMap, startIndex ) {
+			var cellToReplace,
+				markers = {},
+				currentRow,
+				prevCell,
+				cellToPaste,
+				i,
+				j;
+
+			// And now paste!
+			for ( i = 0; i < pastedTableMap.length; i++ ) {
+				currentRow = new CKEDITOR.dom.element( selectedTable.$.rows[ tableSel.rows.first.$.rowIndex + i ] );
+
+				for ( j = 0; j < pastedTableMap[ i ].length; j++ ) {
+					cellToPaste = new CKEDITOR.dom.element( pastedTableMap[ i ][ j ] );
+
+					if ( selectedTableMap[ i ] && selectedTableMap[ i ][ j ] ) {
+						cellToReplace = new CKEDITOR.dom.element( selectedTableMap[ i ][ j ] );
+					} else {
+						cellToReplace = null;
+					}
+
+					// Only try to paste cells that aren't already pasted (it can occur if the pasted cell
+					// has [colspan] or [rowspan]).
+					if ( cellToPaste && !cellToPaste.getCustomData( 'processed' ) ) {
+						// If the cell to being replaced has [colspan], it could have been already
+						// replaced. In that case, it won't have parent.
+						if ( cellToReplace && cellToReplace.getParent() ) {
+							cellToPaste.replace( cellToReplace );
+						} else if ( j === 0 || pastedTableMap[ i ][ j - 1 ] ) {
+							if ( j !== 0 ) {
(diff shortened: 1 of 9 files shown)


plugins/tableselection/plugin.js
```diff
@@ -538,57 +754,43 @@
...
-			selectCellContents( selectedCells[ 0 ] );
+		// In case of mixed content or non table content just insert it to a first cell, erasing the others.
+		if ( !pastedTable ) {
+			selectCellContents( tableSel.cells.first );
 
 			// Due to limitations of our undo manager, in case of mixed content
 			// cells must be emptied after pasting (#520).
 			editor.once( 'afterPaste', function() {
-				emptyCells( selectedCells.slice( 1 ), true );
+				emptyCells( tableSel.cells.all.slice( 1 ), true );
```
Comandeer: Maybe we could move `emptyCells` into `TableSelection.prototype`?

plugins/tabletools/plugin.js
```diff
@@ -146,6 +147,8 @@
 		}
 
 		insertBefore ? newRow.insertBefore( row ) : newRow.insertAfter( row );
+
+		return newRow;
```
Comandeer: There are no tests for these new return values. It would be nice to add them.

plugins/tableselection/plugin.js
```diff
@@ -443,37 +459,159 @@
...
+			selectedCells = selectedCells.concat( insertColumn( selectedCells ) );
+		}
+
+		this.setSelectedCells( selectedCells );
+	};
+
+	// Sorts given arr according to DOM position.
+	//
+	// @param {CKEDITOR.dom.node[]} arr
+	TableSelection.prototype._arraySortByDOMOrder = function( arr ) {
```
mlewand: It looks like I have lost it's usage during cleanups. It was introduced specifically for `insertColumn` call, as it [just appends created `td` elements into `selectedCells`](https://github.com/ckeditor/ckeditor-dev/blob/501b8b2/plugins/tableselection/plugin.js#L586), so this will append `n` columns from each row.

However I think it makes a good idea to put it in `_reset`. It's not a super performance-intensive function so that's going to be fine.

Comandeer: Yes, I meant `setSelectedCells`.

plugins/tableselection/plugin.js
```diff
@@ -538,57 +754,43 @@
...
 
-		// Handle mixed content (if the table is not the only child in the tmpContainer, we
-		// are probably dealing with mixed content). We handle also non-table content here.
-		// We just collapse selection to the first selected cell and pass non-table/mixed
-		// content to other paste handlers (#520).
-		if ( tmpContainer.getChildCount() > 1 || !pastedTable ) {
-			evt.data.dataValue = tmpContainer.getHtml();
-
-			selectCellContents( selectedCells[ 0 ] );
+		// In case of mixed content or non table content just insert it to a first cell, erasing the others.
```
mlewand: Well this comments goes a bit outside of the scope of what table selection plugin is doing. I'll update it accordingly.

plugins/tableselection/plugin.js
mlewand: Did it although emptyCells no longer features undo logic (it's been inlined in host code) as it's only used in that one single place.

mlewand:
The bug is caused by [insertColumn](https://github.com/ckeditor/ckeditor-dev/blob/37583ef/plugins/tableselection/plugin.js#L590) as it returns cells in the whole column (as one might expect), while this logic assumes that only selected cells would be returned.

See a following selection: 
image

In this case it adds to the `selectedCell` also highlighted "dummy cell".

I'll fix this bug shortly.

mlewand:
Fixed the bug caught above, in addition I played a bit with `rowspan` and added an improvement in 2a6f7ecee51a81f95f338eb83805a4c7b9a4f8d7. I was able to improve it, but it's not a priority to fix it completely (extracted to #754).

PR #757, opened 2017-08-10
T/719 Image could be resized in read-only mode

### LINKED ISSUE #719: Image could be resized in read-only mode

Bug

1. Open https://jsfiddle.net/eLzo16jn/
2. Try to resize image.

### Expected result

Nothing happens

### Actual result

Image is resized, changing HTML source.

## Other details

* CKEditor version: reproducible at least from 4.5.0
* Installed CKEditor plugins: image2

beatadelura:

Bug fix

Yes

### This PR contains

- [X] Unit tests
- [X] Manual tests

## What changes did you make?

A new condition has been added to prevent initializing resizer.

---

Closes #719

### DIFF
--- a/plugins/image2/plugin.js
+++ b/plugins/image2/plugin.js
@@ -428,8 +428,10 @@
 
 				// Setup dynamic image resizing with mouse.
 				// Don't initialize resizer when dimensions are disallowed (http://dev.ckeditor.com/ticket/11004).
-				if ( editor.filter.checkFeature( this.features.dimension ) && editor.config.image2_disableResizer !== true )
+				// Don't initialize resizer when config.readOnly is set to true (#719).
+				if ( editor.filter.checkFeature( this.features.dimension ) && editor.config.image2_disableResizer !== true && editor.config.readOnly != true ) {
 					setupResizer( this );
+				}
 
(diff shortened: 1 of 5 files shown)


plugins/image2/plugin.js
```diff
@@ -428,8 +428,10 @@
 
 				// Setup dynamic image resizing with mouse.
 				// Don't initialize resizer when dimensions are disallowed (http://dev.ckeditor.com/ticket/11004).
-				if ( editor.filter.checkFeature( this.features.dimension ) && editor.config.image2_disableResizer !== true )
+				// Don't initialize resizer when config.readOnly is set to true (#719).
+				if ( editor.filter.checkFeature( this.features.dimension ) && editor.config.image2_disableResizer !== true && editor.config.readOnly != true ) {
```
Comandeer: What if the editor is in read-only mode because someone uses `editor.setReadOnly` method (before initializing widgets), not a configuration option?

Such possibility introduces even more problematic question: what if read-only mode would change during using an editor? I honestly don't know how to fix such issue in the current version of editor as we don't have something like "refresh" method for widgets, which can listen to any changes in the editor and react accordingly.

PR #819, opened 2017-08-22
CKEDITOR.tools.getMouseButton

### LINKED ISSUE #810: Introduce CKEDITOR.tools.getMouseButton

Task

Currently `tableselection` plugin and `copyformatting` plugin have their own independent implementations of detecting left mouse button (to be honest version in `tableselection` is just copied from `copyformatting` and slightly simplified). The very same logic would eventually become a part of widget plugin (#711, #746). Probably detecting mouse button could be used also in other places in the editor.

Therefore I propose to unify the implementation of this function by moving it into `CKEDITOR.tools` namespace. It could be further broaden to support also detecting right mouse button.

Comandeer:

Task

### This PR contains

- [x] Unit tests
- [x] Manual tests

## What changes did you make?

I've created `CKEDITOR.tools.getMouseButton` method. It detects if the action was fired by left, middle or right mouse button.

The main issue was the fact that IE 8 has different values for `event.button` property, so to unify them I've introduced three constants:

* `CKEDITOR.MOUSEBUTTON_LEFT`
* `CKEDITOR.MOUSEBUTTON_MIDDLE`
* `CKEDITOR.MOUSEBUTTON_RIGHT`

Alternative approach could use `event.buttons`, which is compatible with IE8's `event.button`.

closes #810


tests/core/tools.js
```diff
@@ -773,6 +789,32 @@
 
 			// Check standard selector.
 			assert.areSame( escapedSelector, 'aaa', 'standard selector' );
+		},
+
+		// #810
+		'test getMouseButton': function() {
+			if ( CKEDITOR.env.ie && CKEDITOR.env.version < 9 ) {
```
mlewand: Let's simplify this, rather than doing two tests for ie8 and non ie8, please add a variable `isIe8` and assert would look like:

```js
generateMouseButtonAsserts( [
	[ CKEDITOR.MOUSEBUTTON_LEFT, isIe8 ? 1 : CKEDITOR.MOUSEBUTTON_LEFT ],
	[ CKEDITOR.MOUSEBUTTON_MIDDLE, isIe8 ? 4 : CKEDITOR.MOUSEBUTTON_MIDDLE ],
	[ CKEDITOR.MOUSEBUTTON_RIGHT, isIe8 ? 2 : CKEDITOR.MOUSEBUTTON_RIGHT ]
] );
```

mlewand: And since we'll have one test, I think it will be beneficial to move `generateMouseButtonAsserts` here, so that it does not distract others.

core/tools.js
```diff
@@ -1463,6 +1463,36 @@
 		},
 
 		/**
+		 * Detects, which mouse button generated given DOM event
+		 *
+		 * @since 4.7.3
+		 * @param {CKEDITOR.dom.event} evt DOM event.
+		 * @returns {Number}
```
mlewand: You made a lie here! 🙀 It can also return a boolean. That has to be documented. 😉

tests/core/tools/manual/getmousebutton.html
```diff
@@ -0,0 +1,34 @@
+<body>
```
mlewand: We don't need a `body` tag here, can be safely removed..

Comandeer:
>  I'm wondering would it be better if we put constants, and function in a dedicated tools namespace.

That was also my thought. However we have all constants in global namespace and nothing (nearly nothing?) elsewhere. IMO it would be quite strange to introduce new namespace just for 3 new constants.

core/tools.js
Comandeer: Yup, I'm a liar :( But I've fixed it.

core/tools.js
```diff
@@ -1980,7 +2010,35 @@
...
 
+	/**
+	 * Left mouse button.
+	 *
+	 * @since 4.7.3
+	 * @readonly
+	 * @property {Number} [=0]
+	 * @member CKEDITOR
+	 */
+	CKEDITOR.MOUSEBUTTON_LEFT = 0;
```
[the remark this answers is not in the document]
Comandeer: Done 👍

tests/core/tools.js
Comandeer: Done 👍

tests/core/tools/manual/getmousebutton.html
Comandeer: Done 👍

PR #822, opened 2017-08-23
save support in source mode

### LINKED ISSUE #817: Save enabled only in design mode

Bug

https://dev.ckeditor.com/ticket/8354

1. Open editor and switch to source mode
2. Try saving using editor toolbar button

### Expected result

Saving with toolbar button should be possible in wysiwyg and source mode.

### Actual result
Saving with toolbar button is only possible in wysiwyg mode.

## Other details

* Browser: Any
* OS: Any
* CKEditor version: 3.0
* Installed CKEditor plugins: save

Initially this plugin was supposed to work in wysiwyg and source: 
https://dev.ckeditor.com/ticket/2836#comment:14
https://dev.ckeditor.com/attachment/ticket/2836/2836_6.patch
but it has most likely been made to work in wysiwyg only by mistake:
https://dev.ckeditor.com/ticket/8354
https://dev.ckeditor.com/changeset/3285

Except for this "won't fix" ticket https://dev.ckeditor.com/ticket/10902, I haven't found any problems related to it.

-----
Setting both modes to be dependent on form existence should fix this problem:
`command.modes = { wysiwyg: !!( editor.element.$.form ), source: !!( editor.element.$.form ) };`

wojtekw92:

bug fix

### This PR contains

- [X] Unit tests
- [x] Manual tests

## What changes did you make?

Added support for save button is source mode

closes #817

### DIFF
--- a/plugins/save/plugin.js
+++ b/plugins/save/plugin.js
@@ -10,6 +10,7 @@
 ( function() {
 	var saveCmd = {
 		readOnly: 1,
+		modes: { wysiwyg: 1,source: 1 },
 
 		exec: function( editor ) {
 			if ( editor.fire( 'save' ) ) {
@@ -44,9 +45,8 @@
 			if ( editor.elementMode != CKEDITOR.ELEMENT_MODE_REPLACE )
 				return;
 
-			saveCmd.startDisabled = !( editor.element.$.form );
 			var command = editor.addCommand( pluginName, saveCmd );
-			command.modes = { wysiwyg: 1,source: 1 };
+			command.startDisabled = !( editor.element.$.form );
 
 			editor.ui.addButton && editor.ui.addButton( 'Save', {
 				label: editor.lang.save.toolbar,
--- a/plugins/save/plugin.js
+++ b/plugins/save/plugin.js
@@ -44,8 +44,9 @@
 			if ( editor.elementMode != CKEDITOR.ELEMENT_MODE_REPLACE )
 				return;
 
+			saveCmd.startDisabled = !( editor.element.$.form );
 			var command = editor.addCommand( pluginName, saveCmd );
-			command.modes = { wysiwyg: !!( editor.element.$.form ), source: !!( editor.element.$.form ) };
+			command.modes = { wysiwyg: 1,source: 1 };
 
 			editor.ui.addButton && editor.ui.addButton( 'Save', {
 				label: editor.lang.save.toolbar,
--- a/plugins/save/plugin.js
+++ b/plugins/save/plugin.js
@@ -45,7 +45,7 @@
 				return;
 
 			var command = editor.addCommand( pluginName, saveCmd );
-			command.modes = { wysiwyg: !!( editor.element.$.form ) };
+			command.modes = { wysiwyg: !!( editor.element.$.form ), source: !!( editor.element.$.form ) };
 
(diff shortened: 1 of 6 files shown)


plugins/save/plugin.js
```diff
@@ -45,7 +45,7 @@
 				return;
 
 			var command = editor.addCommand( pluginName, saveCmd );
-			command.modes = { wysiwyg: !!( editor.element.$.form ) };
+			command.modes = { wysiwyg: !!( editor.element.$.form ), source: !!( editor.element.$.form ) };
```
Comandeer: If at last we got to fix it, let's do it properly! I don't feel that `command.modes` is a proper place to do any of such checks (apparently [not only me](https://dev.ckeditor.com/ticket/8354#comment:1)). Modes should be just enabled and check should be moved e.g. to [`command.startDisabled`](https://docs.ckeditor.com/#!/api/CKEDITOR.commandDefinition-property-startDisabled).

tests/plugins/save/save.js
```diff
@@ -1,9 +1,31 @@
 /* bender-tags: editor */
-/* bender-ckeditor-plugins: toolbar,save,wysiwygarea */
+/* bender-ckeditor-plugins: toolbar,save,wysiwygarea,sourcearea */
 
 bender.test( {
-	'test save event': function() {
-		var editor = CKEDITOR.replace( 'editor' ),
+	'test save event in WYSIWYG mode': function() {
```
Comandeer: As code for tests is nearly identical, I'd think about creating some function to generate them or just to move common logic into it.

plugins/save/plugin.js
```diff
@@ -44,8 +44,9 @@
 			if ( editor.elementMode != CKEDITOR.ELEMENT_MODE_REPLACE )
 				return;
 
+			saveCmd.startDisabled = !( editor.element.$.form );
```
Comandeer: This property should be added to `command`, not `saveCmd`. `saveCmd` is a template to create commands in all editors and `command` is an instance of that command in particular editor. We shouldn't modify the template, but the instance.

tests/plugins/save/save.js
```diff
@@ -1,26 +1,35 @@
...
-		editor.on( 'instanceReady', function() {
-			editor.execCommand( 'save' );
+	editor.on( 'instanceReady', function() {
+		editor.execCommand( 'save' );
 
-			setTimeout( function() {
-				resume( function() {
-					assert.areSame( 1, count, 'save was fired once' );
-				} );
+		setTimeout( function() {
```
Comandeer: I wonder if `afterCommandExec` could be used here instead of `setTimeout`.

plugins/save/plugin.js
```diff
@@ -44,8 +44,9 @@
...
 			var command = editor.addCommand( pluginName, saveCmd );
-			command.modes = { wysiwyg: !!( editor.element.$.form ) };
+			command.modes = { wysiwyg: 1,source: 1 };
```
Comandeer: This property OTOH should be added to template (`saveCmd`) as it's shared across all commands created using that template.

tests/plugins/save/save.js
wojtekw92: unfortunately no, save plugin doesn't fire `afterCommandExec`

PR #832, opened 2017-08-28
[#830] Add possibility to define default tab in dialog

### LINKED ISSUE #830: Extend dialogCommand definition with parameter to select active tab

task

### Expected result

https://docs.ckeditor.com/#!/api/CKEDITOR.dialogCommand-method-constructor
Will gain option to have additional attribute in command definition (`[ext]`) which will set up starting tab of dialog.

This issue blocks #456 and has to be merged before #456.

msamsel:

task request

yes
### This PR contains

- [x] Unit tests
- [x] Manual tests

## What changes did you make?
- provide possibility to define tab which should be open as default one after loading dialog.
- `tests/plugins/dialog/plugin.js` can seems to have a lot of changes. I've wrapped entire test suite with IIFE and add `'use strict;'`. That's why there was added extra tabs in entire file. 

Close #830

### DIFF
--- a/plugins/dialog/plugin.js
+++ b/plugins/dialog/plugin.js
@@ -862,8 +862,10 @@ CKEDITOR.DIALOG_STATE_BUSY = 2;
 			// Reset all inputs back to their default value.
 			this.reset();
 
-			// Select the first tab by default.
-			this.selectPage( this.definition.contents[ 0 ].id );
+			// Select provided tab or first one if there is nothing provided.
+			if ( this._.currentTabId === null ) {
+				this.selectPage( this.definition.contents[ 0 ].id );
+			}
 
 			// Set z-index.
 			if ( CKEDITOR.dialog._.currentZIndex === null )
@@ -2993,6 +2995,9 @@ CKEDITOR.DIALOG_STATE_BUSY = 2;
 	 * @param {String} dialogName The name of the dialog to open when executing
 	 * this command.
 	 * @param {Object} [ext] Additional command definition's properties.
+	 * You can provide additional property (`tab`) to ext object if you wish to open dialog on specific tab.
+	 * 		// Open dialog on 'keystroke' tab
+	 *		editor.addCommand( 'keystroke', new CKEDITOR.dalogCommand( 'a11yHelp', { tab: 'keystroke' } ) );
 	 */
 	CKEDITOR.dialogCommand = function( dialogName, ext ) {
 		this.dialogName = dialogName;
@@ -3001,7 +3006,13 @@ CKEDITOR.DIALOG_STATE_BUSY = 2;
 
 	CKEDITOR.dialogCommand.prototype = {
 		exec: function( editor ) {
-			editor.openDialog( this.dialogName );
+			var that = this;
+			editor.openDialog( this.dialogName, function( dialog ) {
+				// Select different tab if it's provided (#830).
+				if ( that && that.tab ) {
+					dialog.selectPage( that.tab );
+				}
+			} );
 		},
 
 		// Dialog commands just open a dialog ui, thus require no undo logic,
--- a/plugins/dialog/plugin.js
+++ b/plugins/dialog/plugin.js
@@ -862,7 +862,7 @@ CKEDITOR.DIALOG_STATE_BUSY = 2;
 			// Reset all inputs back to their default value.
 			this.reset();
 
-			// Select provided tab or first one if there is nothing provided.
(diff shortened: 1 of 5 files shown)


plugins/dialog/plugin.js
```diff
@@ -862,8 +862,10 @@ CKEDITOR.DIALOG_STATE_BUSY = 2;
 			// Reset all inputs back to their default value.
 			this.reset();
 
-			// Select the first tab by default.
-			this.selectPage( this.definition.contents[ 0 ].id );
+			// Select provided tab or first one if there is nothing provided.
```
f1ames: It selects first tab if no tab is already selected. The comment is a little misleading.

plugins/dialog/plugin.js
```diff
@@ -2993,6 +2995,9 @@ CKEDITOR.DIALOG_STATE_BUSY = 2;
 	 * @param {String} dialogName The name of the dialog to open when executing
 	 * this command.
 	 * @param {Object} [ext] Additional command definition's properties.
+	 * You can provide additional property (`tab`) to ext object if you wish to open dialog on specific tab.
```
f1ames: You can add `tab` docs like:

```
@param {String} [ext.tab] Tab id which should be active when dialog is opened, e.g.
    here goes the code sample...
```

I also suggest changing `ext.tab` to `ext.tabId`.

tests/plugins/dialog/manual/opennonfirsttabasdefault.md
```diff
@@ -0,0 +1,11 @@
+@bender-tags: 830, feature, 4.7.3
+@bender-ui: collapsed
+@bender-ckeditor-plugins: wysiwygarea, toolbar, dialog
+
+----
+
+1. Press button below editor with name `Press me`.
```
f1ames: Maybe:
> Press `Open Dialog` button below editor.

tests/plugins/dialog/manual/opennonfirsttabasdefault.md
```diff
@@ -0,0 +1,11 @@
...
+
+**Expected:** Dialog shows up, where `Tab two` will be displayed as default.
```
f1ames: You could shorten the description to something like:
> `Tab two` is active/shown when dialog opens.

plugins/dialog/plugin.js
```diff
@@ -3001,7 +3006,13 @@ CKEDITOR.DIALOG_STATE_BUSY = 2;
 
 	CKEDITOR.dialogCommand.prototype = {
 		exec: function( editor ) {
-			editor.openDialog( this.dialogName );
+			var that = this;
+			editor.openDialog( this.dialogName, function( dialog ) {
+				// Select different tab if it's provided (#830).
+				if ( that && that.tab ) {
```
f1ames: Is `that.tab` available only here (or can it be changed between `exec` command is executed and this line)? If not instead of using `var that = this`, you may already assign the `tab` value there.

tests/plugins/dialog/plugin.js
```diff
@@ -1,347 +1,407 @@
...
+			editor.execCommand( 'testDialog5' );
+
+			wait();
+		},
 
-		wait();
-	}
-} );
+		// #830
+		'test dialog opens non first tab as default': function() {
```
f1ames: Maybe "test dialog opens tab defined in tabId as default"?

tests/plugins/dialog/manual/opennonfirsttabasdefault.html
```diff
@@ -0,0 +1,49 @@
...
+			}
+		} );
+
+		editor.addCommand( 'testDialog', new CKEDITOR.dialogCommand( 'testDialog', {
+			tabId: 'tab2'
+		} ) );
+
+	} );
+
+	editor.on( 'dialogShow', function() { console.log( 'Show' ) } );
```
f1ames: Is this `console.log` needed here? If yes, please include checking console output as a scenario step, if not simple remove it.

tests/plugins/dialog/plugin.js
msamsel: good point it has more sense 👍

tests/plugins/dialog/manual/opennonfirsttabasdefault.html
msamsel: I removed it.
It left after my checking if `dialogShow` event is execute inside openDialog method.

PR #833, opened 2017-08-28
removed nodejs dependency in build.sh script

### LINKED ISSUE #815: Pick CKEditor version from package.json without using node

Task

> This means that we now need to have Node.js installed in order to build the CKEditor. This is a pain, especially if you build the CKEditor on a Continuous Integration server and you need to ask your infrastructure team to install yet another tool on all the CI agents. I would rather use grep, or perl, or an environment variable, or parse the JSON using Java, rather than adding a dependency on Node.js for such a small thing.

As noted by our user it would be better if we don't rely on `node` to pick the version as it adds extra dependency to deal with.

I think that grep would be a nice alternative. Since the script is in bash, we can assume it has the regular linux env binaries around.

## Other details

Node-based detection was initially introduced in #572.

wojtekw92:

New feature

### This PR contains

- [ ] Unit tests
- [ ] Manual tests

## What changes did you make?

now build.sh script use only bash tools to get version from package.json

closes #815

### DIFF
--- a/dev/builder/build.sh
+++ b/dev/builder/build.sh
@@ -57,7 +57,7 @@ echo "Starting CKBuilder..."
 
 JAVA_ARGS=${ARGS// -t / } # Remove -t from args.
 
-VERSION=$(cat ./../../package.json | grep \"version\"\: | awk -F: '{ print $2}' | sed 's/[\",]//g' | tr -d '[[:space:]]')
+VERSION=$(grep '"version":' ./../../package.json | sed $'s/[\t\",: ]//g; s/version//g' | tr -d '[[:space:]]')
 REVISION=$(git rev-parse --verify --short HEAD)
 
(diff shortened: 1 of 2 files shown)


dev/builder/build.sh
```diff
@@ -57,7 +57,7 @@ echo "Starting CKBuilder..."
 
 JAVA_ARGS=${ARGS// -t / } # Remove -t from args.
 
-VERSION=`node -pe "require('./../../package.json').version"`
+VERSION=$(cat ./../../package.json | grep \"version\"\: | awk -F: '{ print $2}' | sed 's/[\",]//g' | tr -d '[[:space:]]')
```
f1ames: This works fine, but maybe it will be simpler and more readable to use something like:

`grep '"version":' package.json | sed $'s/[\t\",: ]//g; s/version//g'`

it greps the version line and then replaces/removes all unnecessary characters and `version` string which leaves only a version number.

PR #868, opened 2017-09-05
Typing into selected table

### LINKED ISSUE #867: Typing into selected table throws an error

Bug

1. Open https://sdk.ckeditor.com/samples/table.html
2. Select whole table using elements path.
3. Type any letter.

### Expected result

Table is replaced with typed text.

### Actual result

Table is replaced with typed text, but at the same moment error is thrown:

> Uncaught TypeError: Cannot read property 'type' of null

## Other details

* CKEditor version: 4.7.0+

Comandeer:

Bug fix

### This PR contains

- [x] Unit tests
- [x] Manual tests

## What changes did you make?

Former implementation assumed that it would always got ranges with cells, therefore it didn't work with selected table element. Adding a little bit checks should do the trick.

Additionaly function that would remove content from table cells would also do it with table element, leaving it empty inside the editor. After this change whole table will be deleted. 

Closes #867.


mlewand:
Added details that I was missing. Thanks for a quick turnaround.

PR #890, opened 2017-09-11
Fixing bug which cleared clipboard when collapsed selection was copied

### LINKED ISSUE #869: CTRL + C without selected text clears the clipboard buffer
Bug

*Put all reference links here…*

1. Select and copy text from another application.  
2. Paste the selected content into a text editor to make sure it was copied.
3. Place cursor in CKEditor without selecting any text.
4. Use CTRL+C to copy
5. Paste the "copied" text to a text editor.

### Expected result
The copy buffer should still have the copied text that existed in it before the steps above.  This is consistent with other applications where hitting CTRL + C when text isn't selected does not modify the active contents of the clipboard.   CTRL + C without selected text is handled this way because it only occurs when a user does it accidentally.

This is reproducible on the CKEditor demo page.
 
### Actual result
The copy buffer is cleared.

## Other details

* Browser:  Chrome/Firefox
* OS: Windows 10
* CKEditor version:  4.7.2

msamsel:

Bug fix

yes

### This PR contains

- [x] Unit tests
- [X] Manual tests

## What changes did you make?

- activate copying or cutting only when selection is non-collapsed

_**Note:** Because browse doesn't allow on free access to clipboard, there is no unit test which would simulate properly this case._

Close #869

### DIFF
--- a/plugins/clipboard/plugin.js
+++ b/plugins/clipboard/plugin.js
@@ -510,12 +510,13 @@
 
 			if ( CKEDITOR.plugins.clipboard.isCustomCopyCutSupported ) {
 				var initOnCopyCut = function( evt ) {
-					// Make copy/cut only when we have non-collapsed selection (#869).
-					// If user tries to cut in read-only editor, we must prevent default action (https://dev.ckeditor.com/ticket/13872).
-					var collapsedSelection = editor.getSelection().isCollapsed(),
-						readOnlyCut = editor.readOnly && evt.name == 'cut';
+					// There shouldn't be anythong to copy/cut when selection is collapsed (#869).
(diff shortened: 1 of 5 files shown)


plugins/clipboard/plugin.js
```diff
@@ -509,7 +509,12 @@
 
 			if ( CKEDITOR.plugins.clipboard.isCustomCopyCutSupported ) {
 				var initOnCopyCut = function( evt ) {
-					// If user tries to cut in read-only editor, we must prevent default action. (http://dev.ckeditor.com/ticket/13872)
+					// There shouldn't be anythong to copy/cut when selection is collapsed (#869).
```
Comandeer: Typo: "anythong"

Issue #910, opened 2017-09-14
Adapt 'scrollTop' to changes in Chrome 61+


msamsel:
## Are you reporting a feature request or a bug?

bug

## [Check if the issue is already reported](http://docs.ckeditor.com/#!/guide/dev_issues_readme-section-avoid-duplicates)

https://www.chromestatus.com/feature/6386758136627200
https://bugs.chromium.org/p/chromium/issues/detail?id=157855
https://developer.mozilla.org/en-US/docs/Web/API/document/scrollingElement

## Provide detailed reproduction steps (if any)

Chrome disallow on setting scrollTop directly on document (`document.scrollTop`).  Right now we should check our code and correct behaviour of using scrollTop

### Expected result
scrollTop will work properly in Chrome 61+

### Actual result
potential troubles with scrolling.

## Other details

* Browser: Chrome 61+
* OS: any
* CKEditor version: 4.7.3
* Installed CKEditor plugins: …

PR #922, opened 2017-09-15
[#910][#895] Fix problem with using scrollTop on document in Chrome 61+

### LINKED ISSUE #895: "test scroll editable and focus" in tests/core/editable/misc fails in Chrome 61+

Other: failing unit test

None

Run http://tests.ckeditor.dev:1030/tests/core/editable/misc tests.

### Expected result

All tests passing.

### Actual result

The last test in test suite (**test scroll editable and focus**) fails, starting from Chrome 61.

## Other details

* Browser: Chrome 61+
* OS: All (checked on OS X and Windows)
* CKEditor version: 4.7.2

msamsel:

Bug fix

yes

### This PR contains

- [x] Unit tests
- [x] Manual tests

## What changes did you make?

- Add bunch of new methods to API, which set and get scroll of element, document and editable.
- Add method to get scrollingElement from document.
- Change most of the methods where `scrollTop` from document was read by using native methods.
- remove failing test, which is not needed any longer. Test reproduce this part of the code: https://github.com/ckeditor/ckeditor-dev/blob/dee99e283ce259f8e16f1092e3d7613a5e45a499/core/editable.js#L92-L98 This code now look different, so test is no longer needed.

Close #895 #910

### DIFF
--- a/core/dom/document.js
+++ b/core/dom/document.js
@@ -248,6 +248,15 @@ CKEDITOR.tools.extend( CKEDITOR.dom.document.prototype, {
 	},
 
 	/**
+	 * Gets the DOM scrolling element for this document.
+	 *
+	 * @since 4.8.0
+	 * @returns {CKEDITOR.dom.element} The scrolling element or `null` if element doesn't exists.
+	 */
+	getScrollingElement: function() {
+		return this.$.scrollingElement ? new CKEDITOR.dom.element( this.$.scrollingElement ) : null;
+	},
+	/**
 	 * Defines the document content through `document.write`. Note that the
 	 * previous document content will be lost (cleaned).
 	 *
--- a/core/dom/document.js
+++ b/core/dom/document.js
@@ -248,13 +248,14 @@ CKEDITOR.tools.extend( CKEDITOR.dom.document.prototype, {
 	},
 
 	/**
-	 * Gets the DOM scrolling element for this document.
+	 * Gets the DOM scrolling element for this document. Scrolling element allows for using `scrollTop` and `scrollLeft` on entire document.
 	 *
 	 * @since 4.8.0
-	 * @returns {CKEDITOR.dom.element} The scrolling element or `null` if element doesn't exists.
+	 * @returns {CKEDITOR.dom.element} Scrolling element.
 	 */
 	getScrollingElement: function() {
-		return this.$.scrollingElement ? new CKEDITOR.dom.element( this.$.scrollingElement ) : null;
+		var scrollingElement = this.$.scrollingElement || this.$.documentElement || this.$.body;
+		return new CKEDITOR.dom.element( scrollingElement );
 	},
 	/**
 	 * Defines the document content through `document.write`. Note that the
--- a/core/dom/document.js
+++ b/core/dom/document.js
@@ -250,7 +250,7 @@ CKEDITOR.tools.extend( CKEDITOR.dom.document.prototype, {
 	/**
 	 * Gets the DOM scrolling element for this document. Scrolling element allows for using `scrollTop` and `scrollLeft` on entire document.
 	 *
-	 * @since 4.8.0
+	 * @since 4.8.1
 	 * @returns {CKEDITOR.dom.element} Scrolling element.
 	 */
 	getScrollingElement: function() {
--- a/core/dom/document.js
+++ b/core/dom/document.js
@@ -250,13 +250,43 @@ CKEDITOR.tools.extend( CKEDITOR.dom.document.prototype, {
 	/**
 	 * Gets the DOM scrolling element for this document. Scrolling element allows for using `scrollTop` and `scrollLeft` on entire document.
 	 *
-	 * @since 4.8.1
+	 * @since 4.10.0
 	 * @returns {CKEDITOR.dom.element} Scrolling element.
 	 */
 	getScrollingElement: function() {
 		var scrollingElement = this.$.scrollingElement || this.$.documentElement || this.$.body;
 		return new CKEDITOR.dom.element( scrollingElement );
 	},
+	/**
+	 * Returns scroll position of the document.
+	 *
+	 * @since 4.10.0
+	 * @returns {Object} Object representing vertical and horizontal scroll position of the document.
+	 * @returns {Number} return.scrollTop vertical scroll position.
+	 * @returns {Number} return.scrollLeft horizontal scroll position.
+	 */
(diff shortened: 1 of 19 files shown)


core/dom/element.js
```diff
@@ -2095,6 +2083,69 @@ CKEDITOR.dom.element.clearMarkers = function( database, element, removeFromDatab
...
+		},
+
+		/**
+		 * Returns scroll position of the element itself.
+		 *
+		 * @since 4.8.0
+		 * @param {Boolean} [onlyScrollTop] If set to `true` returns only scrollTop value
+		 * @returns {Object} Object with keys: `scrollTop` and `scrollLeft`
+		 */
+		'getScroll': function( onlyScrollTop ) {
```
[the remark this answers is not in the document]
msamsel: Probably I've made it automatically after writing some unit tests.

tests/core/editable/scrolls.js
```diff
@@ -0,0 +1,73 @@
+( function() {
+	'use strict';
+
+	bender.editors = {
+		divarea: {
```
f1ames: Inline editor should be added here too.

tests/core/dom/element/manual/scrolldocument.md
```diff
@@ -0,0 +1,7 @@
+@bender-tags: 4.8.0, feature
```
f1ames: Isn't it more like a bug fix? The original issue #910 is a bug type.

tests/core/dom/element/manual/scrollelement.md
```diff
@@ -0,0 +1,7 @@
+@bender-tags: 4.8.0, feature
+@bender-ui: collapsed
+@bender-ckeditor-plugins: toolbar, divarea, undo, sourcearea, basicstyles, div
+
+----
+
+Press bottom buttons and check if editor is scroleld properly. Compare scroll position with values displayed above editor.
```
f1ames: Typo in scrolled (now is scro**lel**d).

tests/core/editable/scrolls.js
msamsel: I'm not sure about it. Is it possible to generate inline editor with scrolling editable inside our unit test? Or it require prepare special test suite?

Comandeer: Aren't appropriate CSS styles (`overflow: auto` + small height) enough?

msamsel:
I add inline editor and update tags. I also ignore iOS from testing, because of upstream issue: https://bugs.webkit.org/show_bug.cgi?id=172854

tests/core/dom/document.js
```diff
@@ -142,6 +134,20 @@ bender.test( appendDomObjectTests(
...
+		},
+
+		// (#910)
+		'test getScrollingElement': function() {
+			var scrollingElement = CKEDITOR.document.getScrollingElement(),
+				nativeEl = document.scrollingElement || document.documentElement || document.body;
+
+			if ( !( CKEDITOR.env.ie && CKEDITOR.env.version < 9 ) ) {
+				// IE8 doesn't support isEqualNode
+				assert.isTrue( scrollingElement.$.isEqualNode( nativeEl ) );
```
Comandeer: It should be rather `===` than `isEqualNode`. We are not looking for equal nodes, but checking if nodes are **the same**. [Example on MDN](https://developer.mozilla.org/en-US/docs/Web/API/Node/isEqualNode#Example) clearly shows this difference. Introducing this change would also allow us to test it in old IEs.

tests/core/dom/element/scrolls.js
```diff
@@ -0,0 +1,118 @@
...
+			assert.areSame( expected.x, actual.scrollLeft );
+			assert.areSame( expected.y, actual.scrollTop );
+		},
+
+		'test getDocumentScrollPosition': function() {
+			var element = document.getElementById( 'small' ),
+				ckEl = new CKEDITOR.dom.element( element ),
+				doc,
+				docEl;
+			// Reset position with native methods
```
Comandeer: Resetting could be moved into separate helper function.

tests/core/dom/element/scrolls.js
```diff
@@ -0,0 +1,118 @@
...
+			docEl.scrollLeft = 20;
+			this._assertBothScrolls( { x: 20, y: 20 }, ckEl.getDocumentScrollPosition() );
+
+			// document scroll, element no-scroll
+			ckEl.$.scrollTop = 0;
+			ckEl.$.scrollLeft = 0;
+			this._assertBothScrolls( { x: 20, y: 20 }, ckEl.getDocumentScrollPosition() );
+		},
+
+		'test getScrollPosition': function() {
```
Comandeer: This test is very similar to the previous one and it seems that only the checked element is different. Maybe it'd be better to create factory for such tests?

Comandeer:
There is also missing changelog entry.

Comandeer:
Rebased to latest `major`.

core/dom/document.js
```diff
@@ -247,6 +247,46 @@ CKEDITOR.tools.extend( CKEDITOR.dom.document.prototype, {
...
+	 */
+	getScrollingElement: function() {
+		var scrollingElement = this.$.scrollingElement || this.$.documentElement || this.$.body;
+		return new CKEDITOR.dom.element( scrollingElement );
+	},
+	/**
+	 * Returns scroll position of the document.
+	 *
+	 * @since 4.10.0
+	 * @returns {Object} Object representing vertical and horizontal scroll position of the document.
```
Comandeer: If such object is returned in many places, I wonder about adding some virtual type for it (`CKEDITOR.dom.scrollPosition?`) – similar to `CKEDITOR.dom.rect`

core/dom/element.js
```diff
@@ -2112,6 +2101,55 @@ CKEDITOR.dom.element.clearMarkers = function( database, element, removeFromDatab
...
+
+		/**
+		 * Returns scroll position of the element's document.
+		 *
+		 * @since 4.10.0
+		 * @returns {Object} Object representing vertical and horizontal scroll position of the document.
+		 * @returns {Number} return.scrollTop vertical scroll position.
+		 * @returns {Number} return.scrollLeft horizontal scroll position.
+		 */
+		getDocumentScrollPosition: function() {
```
Comandeer: I wonder if this (and `setDocumentScrollPosition`) is longer needed after adding `get/setScrollPosition` to `CKEDITOR.dom.document` itself.

PR #927, opened 2017-09-18
[#801] Fix elementspath plugin to work properly with Firefox 55+

### LINKED ISSUE #801: The `elementspath/divarea` unit tests fails on FF 55, when browser window focused

Bug

No

Two unit tests in http://tests.ckeditor.dev:1030/tests/plugins/elementspath/divarea fails if browser window is focused, see below:

aug-18-2017 12-47-38

### Expected result

Tests passing regarding the focus.

### Actual result

Tests fails.

## Other details

* Browser: FF 55 (works on all other browsers and FF < 55)
* OS: All
* CKEditor version: 4.7.2
* Installed CKEditor plugins: `divarea,elementspath`

msamsel:

Bug fix

PR fix tests

### This PR contains

- [ ] Unit tests
- [ ] Manual tests

## What changes did you make?
- change in elementspath plugin to use `evt.data.path` instead of `editor.elementPath()`

Close #801

### DIFF
--- a/plugins/elementspath/plugin.js
+++ b/plugins/elementspath/plugin.js
@@ -153,7 +153,7 @@
 				return true;
 			} );
 
-		editor.on( 'selectionChange', function() {
+		editor.on( 'selectionChange', function( evt ) {
 			var html = [],
 				elementsList = elementsPath.list = [],
 				namesList = [],
@@ -161,7 +161,7 @@
 				isContentEditable = true,
 
 				// Use elementPath to consider children of editable only (http://dev.ckeditor.com/ticket/11124).
-				elementsChain = editor.elementPath().elements,
+				elementsChain = evt.data.path.elements,
 				name;
 
 			// Starts iteration from body element, skipping html.
--- a/plugins/elementspath/plugin.js
+++ b/plugins/elementspath/plugin.js
@@ -161,6 +161,7 @@
 				isContentEditable = true,
 
 				// Use elementPath to consider children of editable only (http://dev.ckeditor.com/ticket/11124).
+				// Use elementPath provided in evt, which is accurate in all cases (#801).
 				elementsChain = evt.data.path.elements,
 				name;
 
--- a/plugins/elementspath/plugin.js
+++ b/plugins/elementspath/plugin.js
@@ -161,7 +161,7 @@
 				isContentEditable = true,
 
 				// Use elementPath to consider children of editable only (http://dev.ckeditor.com/ticket/11124).
-				// Use elementPath provided in evt, which is accurate in all cases (#801).
+				// Use elementPath from event (instead of editor.elementPath()), which is accurate in all cases (#801).
 				elementsChain = evt.data.path.elements,
 				name;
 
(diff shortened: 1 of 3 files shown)


tests/plugins/elementspath/manual/elementspath.html
```diff
@@ -0,0 +1,82 @@
...
+		<figure class="image">
+			<img src="/tests/_assets/lena.jpg" alt="avatar" />
+			<figcaption>Caption</figcaption>
+		</figure>
+
+</div>
+
+<script>
+	CKEDITOR.replace( 'editor1', {
+		extraPlugins: 'image2',
```
f1ames: The `image2` plugin is not needed here as it is already in `bender-ckeditor-plugins`.

tests/plugins/elementspath/manual/elementspath.html
```diff
@@ -0,0 +1,82 @@
...
+		allowedContent: true
+	} );
+	CKEDITOR.replace( 'editor2', {
+		extraPlugins: 'divarea,image2',
```
f1ames: Same as above.

tests/plugins/elementspath/manual/elementspath.html
```diff
@@ -0,0 +1,82 @@
+<textarea id="editor1">
+	<p>Hello world. <strong>Bold text and <em>italic with additional <u>underline</u></em></strong>.</p>
+	<p>Hello world. <strong contenteditable="false">Bold text and <em contenteditable="true">italic with additional <u>underline</u></em></strong>.</p>
```
f1ames: This line looks identical as the previous line in the editor, however it has an inline widget. I would change the text so it will be more clear that there is a widget.

tests/plugins/elementspath/manual/elementspath.html
```diff
@@ -0,0 +1,82 @@
...
+	<ul>
+		<li>One</li>
+		<li>Two<ol>
+			<li><strong>STRONG: </strong>1111</li>
+			<li><em>ITALIC: </em>2222</li>
+		</ol></li>
+	</ul>
+	<div><div><div><div><div><div><div><div><div><div><div>Divception</div></div></div></div></div></div></div></div></div></div></div>
```
f1ames: Nice idea, but it is hard to evaluate if path is correct without looking at the source. You may slightly decrese the number of divs (like 5) and change the text to e.g.

> Divception - when clicked elements path will show 5 divs.

tests/plugins/elementspath/manual/elementspath.html
f1ames: > italic with additional underline inside inline widget

tests/plugins/elementspath/manual/elementspath.html
```diff
@@ -0,0 +1,82 @@
+<textarea id="editor1">
```
f1ames: Instead of duplicating the same content in both editors you may put part of it with some elements (e.g. image, styling, widget) in the first one and rest (e.g. tables, list) in the second one. It is always less monotonic to test different editor with different contents (ofc not applicable in all cases, but here it will be ok).

plugins/elementspath/plugin.js
```diff
@@ -153,15 +153,16 @@
...
+		editor.on( 'selectionChange', function( evt ) {
 			var html = [],
 				elementsList = elementsPath.list = [],
 				namesList = [],
 				filters = elementsPath.filters,
 				isContentEditable = true,
 
 				// Use elementPath to consider children of editable only (http://dev.ckeditor.com/ticket/11124).
-				elementsChain = editor.elementPath().elements,
+				// Use elementPath provided in evt, which is accurate in all cases (#801).
```
f1ames: > Use elementPath from event (instead of `editor.elementPath()`), which is...

f1ames:
Reported one issue not related to this fix #1113.

Issue #662, opened 2017-07-19
Pasting images from Word


Comandeer:
## Are you reporting a feature or a bug?

Feature request

## Provide detailed reproduction steps (if any)

1. Open attached Word document.
2. Select all its content.
3. Paste into CKEditor.

### Expected result

Image is pasted correctly.

### Actual result

Image is not pasted and console shows error:

> Not allowed to load local resource: `file:///[redacted]/clip_image002.png`

There should be a sensible way to detect such images and try to embed them anyway, converting to Blob (see also #505). However I'm not sure how to omit limitation with ability to get access to local files… Maybe try to extract images from RTF representation of pasted Word content?

## Other details

* Browser: Chrome, Firefox
* OS: macOS
* CKEditor version: 4.7.1
* Installed CKEditor plugins: pastefromword

## Attachments

* [Document1.docx](https://github.com/ckeditor/ckeditor-dev/files/1159137/Document1.docx)

mlewand:
Yes, image blobs are extractable from RTF clipboard.

bahuma20:
Works in Internet Explorer

mlewand:
That's correct AFAIR it does work only in IE11 (as it inlines all the images natively). This feature has been reverted in MS Edge though.

mlewand:
Added information on browser compatibility.

fikoborquez:
I have this issue pasting images with text in Chrome but not in Firefox. Can anybody give me some help?

JURALOVE:
Thanks. This hile was my solution ;)

PR #951, opened 2017-09-21
Hexstring converter to base64

msamsel:

New feature

### This PR contains

- [x] Unit tests
- [ ] Manual tests

## What changes did you make?

- Add tools for converting hexstrings to base64.

Required by #662

### DIFF
--- a/core/tools.js
+++ b/core/tools.js
@@ -1522,7 +1522,7 @@
 		 *
 		 * @since 4.8.0
 		 * @param {String} hexString Contains input string which represent bytes, e.g. `"08A11D8ADA2B"`.
-		 * @returns {Array} Bytes stored in a form of Integer numbers, e.g. `[ 8, 161, 29, 138, 218, 43 ]`.
+		 * @returns {Number[]} Bytes stored in a form of Integer numbers, e.g. `[ 8, 161, 29, 138, 218, 43 ]`.
 		 */
 		convertHexStringToBytes: function( hexString ) {
 			var bytesArray = [],
@@ -1536,10 +1536,10 @@
 		},
 
 		/**
-		 * Convert bytes array into base64 code.
+		 * Convert bytes array into a base64 encoded string.
 		 *
 		 * @since 4.8.0
-		 * @param {Array} bytesArray An array which stores 1 byte in each cell as Integer number.
+		 * @param {Number[]} bytesArray An array which stores 1 byte in each cell as Integer number.
 		 * @returns {String} Base64 string which represents input bytes.
 		 */
 		convertBytesToBase64: function( bytesArray ) {
--- a/core/tools.js
+++ b/core/tools.js
@@ -1556,8 +1556,8 @@
 		 * Decimal: `2 10 4 29 34 45 40 43` → base64: `CKEditor`.
 		 *
 		 * @since 4.8.0
-		 * @param bytesArray Array which stores 1 byte in each cell as Integer number.
-		 * @returns Base64 string which represents input bytes.
+		 * @param {Array} bytesArray An array which stores 1 byte in each cell as Integer number.
+		 * @returns {String} Base64 string which represents input bytes.
 		 */
 		convertBytesToBase64: function( bytesArray ) {
 			var base64characters = 'ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+/',
--- a/core/tools.js
+++ b/core/tools.js
@@ -1518,6 +1518,71 @@
 		},
 
 		/**
+		 * Converts hexstring to base64 coode.
+		 *
+		 * Hexstring contains 8bit numbers, where base64 use 6bit to code.
+		 * We need to process 3 * 8bit chunks of hexstring into 4 * 6bit characters of base64.
+		 * Algorithm:
+		 * 1. Take 3 * 8bit.
+		 * 2. If there is less than 3 byte, fill it with zeros.
+		 * 3. Transform 3 * 8bit into 4 * 6bit numbers.
+		 * 4. Translate those numbers to proper character related to base64.
+		 * 5. If there was added extra zeros bytes fill them with '=' sign.
+		 *
+		 * Example:
+		 * 	hex: 08A11D8ADA2B -> binary: 0000 1000 1010 0001 0001 1101 1000 1010 1101 1010 0010 1011
+		 * 	| <- shows where base64 will cut bits
+		 * 	binary: 0000 10|00 1010| 0001 00|01 1101| 1000 10|10 1101| 1010 00|10 1011
+		 * 	This gives indexes in base64character table
+		 * 	decimal: 2 10 4 29 34 45 40 43 -> base64: "CKEditor"
+		 *
+		 * @since 4.8.0
+		 * @param {String} hexstring string contained only hexadecimal values, e.g. "0AB7"
+		 * @returns {String} base64 string
+		 */
+		hexstring2base64: function( hexstring ) {
+			var binaryArray = [],
+				base64characters = 'ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+/',
+				base64string = '',
+				binArrLength = hexstring.length / 2,
+				i;
+
+
+			for ( i = 0; i < binArrLength; i++ ) {
+				binaryArray.push( parseInt( hexstring.substr( i * 2, 2 ), 16 ) );
+			}
+
+			for ( i = 0; i < binArrLength; i += 3 ) {
+				var array3 = binaryArray.slice( i, i + 3 ),
+					array3length = array3.length,
+					array4 = [],
+					j;
+
+				if ( array3length < 3 ) {
+					for ( j = array3length; j < 3; j++ ) {
+						array3[ j ] = 0;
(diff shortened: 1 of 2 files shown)


core/tools.js
```diff
@@ -1518,6 +1518,71 @@
...
+		 * 	| <- shows where base64 will cut bits
+		 * 	binary: 0000 10|00 1010| 0001 00|01 1101| 1000 10|10 1101| 1010 00|10 1011
+		 * 	This gives indexes in base64character table
+		 * 	decimal: 2 10 4 29 34 45 40 43 -> base64: "CKEditor"
+		 *
+		 * @since 4.8.0
+		 * @param {String} hexstring string contained only hexadecimal values, e.g. "0AB7"
+		 * @returns {String} base64 string
+		 */
+		hexstring2base64: function( hexstring ) {
```
mlewand: When picking a name, check the surrounding names. Any other convert methods uses `convertXyzToXyz` so you're missing convert prefix and it should be "to" rather than "2".

core/tools.js
```diff
@@ -1518,6 +1518,71 @@
 		},
 
 		/**
+		 * Converts hexstring to base64 coode.
```
mlewand: This markdown does not get converted at all:
image
Take the time and set up the ckeditor-docs builder, and preview any docs changes before pushing the code.

mlewand: Reason for the above is probably no new line before the ordered list.

core/tools.js
mlewand: Use consistent letter casing: `hexString` rather than `hexstring`.

core/tools.js
```diff
@@ -1518,6 +1518,71 @@
...
+		 *
+		 * Example:
+		 * 	hex: 08A11D8ADA2B -> binary: 0000 1000 1010 0001 0001 1101 1000 1010 1101 1010 0010 1011
+		 * 	| <- shows where base64 will cut bits
+		 * 	binary: 0000 10|00 1010| 0001 00|01 1101| 1000 10|10 1101| 1010 00|10 1011
+		 * 	This gives indexes in base64character table
+		 * 	decimal: 2 10 4 29 34 45 40 43 -> base64: "CKEditor"
+		 *
+		 * @since 4.8.0
+		 * @param {String} hexstring string contained only hexadecimal values, e.g. "0AB7"
```
mlewand: `@param` and `@returns` contain full sentences. So start with a capital letter and end with a dot.

core/tools.js
```diff
@@ -1518,6 +1518,71 @@
...
+		 * We need to process 3 * 8bit chunks of hexstring into 4 * 6bit characters of base64.
+		 * Algorithm:
+		 * 1. Take 3 * 8bit.
+		 * 2. If there is less than 3 byte, fill it with zeros.
+		 * 3. Transform 3 * 8bit into 4 * 6bit numbers.
+		 * 4. Translate those numbers to proper character related to base64.
+		 * 5. If there was added extra zeros bytes fill them with '=' sign.
+		 *
+		 * Example:
+		 * 	hex: 08A11D8ADA2B -> binary: 0000 1000 1010 0001 0001 1101 1000 1010 1101 1010 0010 1011
```
mlewand: Don't be afraid to wrap code-related parts with back ticks, like so: `08A11D8ADA2B`.

tests/core/tools.js
```diff
@@ -775,6 +775,44 @@
...
+					hex: 'FFFFFF',
+					base64: '////'
+				},
+				{
+					hex: 'fc0fc0',
+					base64: '/A/A'
+				},
+				{
+					hex: '08A11D8ADA2B',
+					base64: 'CKEditor'
```
mlewand: ❤️

mlewand:
@msamsel and please, don't put these dots in a pull request title :D

core/tools.js
```diff
@@ -1518,6 +1518,76 @@
...
+		 * * `|` (pipe) shows where base64 will cut bits during transform.
+		 * * Binary: `0000 10|00 1010| 0001 00|01 1101| 1000 10|10 1101| 1010 00|10 1011`.
+		 * * Now we have 6bit numbers (decimal values below), which are translated to indexes in `base64characters` array.
+		 * * Decimal: `2 10 4 29 34 45 40 43` -> base64: `CKEditor`.
+		 *
+		 * @since 4.8.0
+		 * @param {String} hexString String contained only hexadecimal values, e.g. `0AB7`.
+		 * @returns {String} Base64 string represnet input value.
+		 */
+		convertHexstringToBase64: function( hexString ) {
```
mlewand: There's one more thing, I thought that I have included it in original review, but turns out I forgot to paste it in review summary.

Rather than adding `convertHexstringToBase64` which does the job for your particular purpose, you should do two separate function:

* `convertHexStringToBytes`
* `convertBytesToBase64`

That way this code will be more reusable, and can be used by a developer that looks only to use bytes to base64 conversion, etc.

core/tools.js
```diff
@@ -1518,6 +1518,87 @@
...
+		 * Example:
+		 *
+		 * * Bytes Array: [ 8, 161, 29, 138, 218, 43 ] -> binary: `0000 1000 1010 0001 0001 1101 1000 1010 1101 1010 0010 1011`.
+		 * * Binary: `0000 10|00 1010| 0001 00|01 1101| 1000 10|10 1101| 1010 00|10 1011` ← `|` (pipe) shows where base64 will cut bits during transformation.
+		 * * Now we have 6bit numbers (written in decimal values), which are translated to indexes in `base64characters` array.<br />
+		 * Decimal: `2 10 4 29 34 45 40 43` → base64: `CKEditor`.
+		 *
+		 * @since 4.8.0
+		 * @param bytesArray Array which stores 1 byte in each cell as Integer number.
+		 * @returns Base64 string which represents input bytes.
```
f1ames: There is no such type as `bytesArray` or `Base64`, it should probably be `Array` and `String`. You can include more information (that it is array of bytes / string in base64 format) in the `@param`/`@returns` description. Also remember about brackets when defining type - `{Array}`.

core/tools.js
```diff
@@ -1518,6 +1518,87 @@
...
+			for ( i = 0; i < bytesArrayLength; i++ ) {
+				bytesArray.push( parseInt( hexString.substr( i * 2, 2 ), 16 ) );
+			}
+			return bytesArray;
+		},
+
+		/**
+		 * Convert bytes array into base64 code.
+		 *
+		 * Bytes are `8bit` numbers, where base64 use `6bit` to store data. That's why we process 3 Bytes into 4 characters representing base64.
```
mlewand: Is this information in any way needed for the person using this API? I think not - this is an implementation detail of base64 and are not relevant for the person looking for this method.

Same goes for the "Algorithm" part. **It's great that you have added it**, but you did simply in a wrong place.

You don't want to put all of this information in our API docs, as there should be just summary. Instead it should be added as a inline comment insde the function, so that only programmers who work on this internals know it.

So I'll ask you to just do that, and we'll keep this great value of having these comments. 🙂

tests/core/tools.js
```diff
@@ -775,6 +775,88 @@
...
+					hex: 'fc0fc0',
+					bytes: [ 252, 15, 192 ]
+				},
+				{
+					hex: '08A11D8ADA2B',
+					bytes: [ 8, 161, 29, 138, 218, 43 ]
+				}
+			];
+			CKEDITOR.tools.array.forEach( testCases, function( test ) {
+				assert.isTrue( CKEDITOR.tools.arrayCompare( test.bytes, CKEDITOR.tools.convertHexStringToBytes( test.hex ) ) );
```
mlewand: There are dedicated assertions for array assertion (`arrayAssert`), see `arrayAssert.itemsAreEqual`.

core/tools.js
```diff
@@ -1518,6 +1518,87 @@
...
+		 * Example:
+		 *
+		 * * Bytes Array: [ 8, 161, 29, 138, 218, 43 ] -> binary: `0000 1000 1010 0001 0001 1101 1000 1010 1101 1010 0010 1011`.
+		 * * Binary: `0000 10|00 1010| 0001 00|01 1101| 1000 10|10 1101| 1010 00|10 1011` ← `|` (pipe) shows where base64 will cut bits during transformation.
+		 * * Now we have 6bit numbers (written in decimal values), which are translated to indexes in `base64characters` array.<br />
+		 * Decimal: `2 10 4 29 34 45 40 43` → base64: `CKEditor`.
+		 *
+		 * @since 4.8.0
+		 * @param {Array} bytesArray An array which stores 1 byte in each cell as Integer number.
+		 * @returns {String} Base64 string which represents input bytes.
```
AnnaTomanek: Base64-encoded string.

core/tools.js
AnnaTomanek: Agreed, first of all - kudos for explaining this in details, great job.

However, I'd recommend moving the detailed explanation of the algorithm to (internal) comments and leaving the public API docs more straightforward/basic. A nice example how we usually do it is the Magicline plugin with excellent comments by @oleq: https://github.com/ckeditor/ckeditor-dev/blob/major/plugins/magicline/plugin.js#L851

oleq: @AnnaTomanek 😊

AnnaTomanek: Credit where it's due! Learn from the master, guys! :)

PR #952, opened 2017-09-21
Initial implementation of commands in Easy Image

Comandeer:

New feature

### This PR contains

- [x] Unit tests
- [x] Manual tests

## What changes did you make?

Initial implementation of "Full Image", "Side image" and "Change alternative text" commands for Easy Image.

### DIFF
--- a/plugins/easyimage/plugin.js
+++ b/plugins/easyimage/plugin.js
@@ -7,20 +7,12 @@
 	'use strict';
 
 	function addCommands( editor ) {
-		var commandsToRefresh = [ 'easyimageFull', 'easyimageSide' ];
-
 		function isSideImage( widget ) {
 			return widget.element.hasClass( 'easyimage-side' );
 		}
 
 		function isFullImage( widget ) {
 			return !isSideImage( widget );
-		}
-
-		function refreshCommands( editor, commands ) {
-			CKEDITOR.tools.array.forEach( commands, function( command ) {
-				editor.getCommand( command ).refresh( editor );
-			} );
 		}
 
 		function createCommandRefresh( enableCheck ) {
@@ -45,9 +37,9 @@
 
 					exec( widget );
 
-					// We have to manually refresh commands as refresh seems
+					// We have to manually force refresh commands as refresh seems
 					// to be executed prior to exec.
-					refreshCommands( editor, commandsToRefresh );
+					editor.forceNextSelectionCheck();
 				},
 
 				refresh: createCommandRefresh( refreshCheck )
--- a/plugins/easyimage/plugin.js
+++ b/plugins/easyimage/plugin.js
@@ -21,7 +21,7 @@
 		lang: 'en',
 
 		onLoad: function() {
-			CKEDITOR.dialog.add( 'eiAltText', this.path + 'dialogs/eiAltText.js' );
+			CKEDITOR.dialog.add( 'easyimageAlt', this.path + 'dialogs/easyimagealt.js' );
 		},
 
 		init: function( editor ) {
@@ -50,7 +50,7 @@
 				}
 			} );
 
-			editor.addCommand( 'easyimageAlt', new CKEDITOR.dialogCommand( 'eiAltText' ) );
+			editor.addCommand( 'easyimageAlt', new CKEDITOR.dialogCommand( 'easyimageAlt' ) );
 
 			editor.addMenuGroup( 'easyimage' );
 			editor.addMenuItems( {
--- a/plugins/easyimage/plugin.js
+++ b/plugins/easyimage/plugin.js
@@ -10,7 +10,7 @@
 
 	function addCommands( editor ) {
 		function isSideImage( widget ) {
-			return widget.element.hasClass( 'easyimage-side' );
+			return widget.element.hasClass( editor.config.easyimage_sideClass );
 		}
 
 		function isFullImage( widget ) {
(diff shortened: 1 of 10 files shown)


plugins/easyimage/plugin.js
```diff
@@ -6,6 +6,102 @@
...
+			if ( !stylesLoaded ) {
+				CKEDITOR.document.appendStyleSheet( this.path + 'styles/easyimage.css' );
+				stylesLoaded = true;
+			}
+
+			if ( editor.addContentsCss ) {
+				editor.addContentsCss( this.path + 'styles/easyimage.css' );
+			}
+
+			editor.addCommand( 'eiFullImage', {
```
mlewand: Follow the convention `<pluginName>CommandName`, so `easyimageFull`?

plugins/easyimage/plugin.js
```diff
@@ -6,6 +6,102 @@
 ( function() {
 	'use strict';
 
+	function isSideImage( widget ) {
+		return widget.element.hasClass( 'easyimage--side' );
```
mlewand: We're not using double dash in CSS class names. How about just `side-image`?

Comandeer: Maybe `.easyimage-side`? It'll contain plugin name then.

mlewand: Sounds good to me.

Comandeer:
I've moved logic connected with managing commands's state to command's `refresh` method. However it seems that refreshing is done before executing command, requiring to manually refresh them after `exec`.

plugins/easyimage/plugin.js
```diff
@@ -6,6 +6,141 @@
...
+			contextSensitive: true,
+
+			exec: function( editor ) {
+				var widget = editor.widgets.focused;
+
+				exec( widget );
+
+				// We have to manually refresh commands as refresh seems
+				// to be executed prior to exec.
+				refreshCommands( editor, commandsToRefresh );
```
mlewand: Refreshing elements path will be enough to trigger it.

Just call `editor.forceNextSelectionCheck();` and it should be good enough. It's simpler, has less LOC, but make sure to leave the comment as it will be helpful.

Actually I see that many existing plugins are doing that this way.

plugins/easyimage/plugin.js
```diff
@@ -6,6 +6,141 @@
...
+				return isSideImage( widget );
+			} ) );
+
+			editor.addCommand( 'easyimageAlt', new CKEDITOR.dialogCommand( 'easyimageAlt', {
+				startDisabled: true,
+				contextSensitive: true,
+				refresh: createCommandRefresh()
+			} ) );
+
+			editor.addMenuGroup( 'easyimage' );
```
mlewand: Can you extract context-menu definition logic into a separate method? To improve `init()`'s readability?

plugins/easyimage/plugin.js
```diff
@@ -6,6 +6,141 @@
...
+			if ( !stylesLoaded ) {
+				CKEDITOR.document.appendStyleSheet( this.path + 'styles/easyimage.css' );
+				stylesLoaded = true;
+			}
+
+			if ( editor.addContentsCss ) {
+				editor.addContentsCss( this.path + 'styles/easyimage.css' );
+			}
+
+			editor.addCommand( 'easyimageFull', createCommand( function( widget ) {
```
mlewand: Can you extract command definition logic into a separate method? To improve `init()`'s readability? Also this function could encapsulate methods like `refreshCommands`, `createCommandRefresh`, `createCommand`.

plugins/easyimage/plugin.js
```diff
@@ -6,6 +6,141 @@
...
+
+				easyimageAlt: {
+					label: editor.lang.easyimage.commands.altText,
+					command: 'easyimageAlt',
+					group: 'easyimage',
+					order: 3
+				}
+			} );
+
+			// Overwrite some image2 defaults
```
mlewand: Missing tailing dot.

plugins/easyimage/plugin.js
```diff
@@ -6,6 +6,141 @@
 ( function() {
 	'use strict';
 
+	var commandsToRefresh = [ 'easyimageFull', 'easyimageSide' ];
+
+	function isSideImage( widget ) {
+		return widget.element.hasClass( 'easyimage-side' );
```
mlewand: Let's extract class names into constants, so that it is not duplicated all over the place.

plugins/easyimage/plugin.js
```diff
@@ -6,6 +6,141 @@
...
+		return !isSideImage( widget );
+	}
+
+	function refreshCommands( editor, commands ) {
+		CKEDITOR.tools.array.forEach( commands, function( command ) {
+			editor.getCommand( command ).refresh( editor );
+		} );
+	}
+
+	function createCommandRefresh( check ) {
```
mlewand: How about `activeCheck` / `enabledCheck` or anything along these lines? The reason why I'm asking is `check` says almost nothing about the purpose of variable.

mlewand:
Other than things reported above, the base seems to be good enough to work on it (the only exception is that we need to extract `image2`, but we want to make it in other ticket).

plugins/easyimage/plugin.js
Comandeer: It should be configurable, so maybe just extract it to config variables?

mlewand: Make sense we can use configurable classes from the very beginning.

PR #957, opened 2017-09-25
Inline toolbar plugin: toolbar view

wojtekw92:

 New feature

### This PR contains

- [X] Unit tests
- [X] Manual tests

## What changes did you make?

inline toolbar plugin.


plugins/inlinetoolbar/plugin.js
```diff
@@ -0,0 +1,285 @@
...
+ */
+
+( function() {
+	'use strict';
+
+	CKEDITOR.plugins.add( 'inlinetoolbar', {
+		requires: 'balloonpanel',
+		init: function( editor ) {
+			CKEDITOR.ui.inlineToolbar.prototype = CKEDITOR.tools.extend( CKEDITOR.ui.balloonPanel.prototype );
+			CKEDITOR.ui.inlineToolbar.prototype.build = function() {
```
mlewand: Why, copy and pasting stuff? Keep it DRY. You want to reuse original method + apply any adjustments before / after it was called.

plugins/inlinetoolbar/plugin.js
```diff
@@ -0,0 +1,285 @@
...
+				// Activates listeners on panel show.
+				// All listeners will be deactivated on panel hide.
+				this.on( 'show', this.activateShowListeners, this );
+
+				// Deactivate all listeners on panel hide.
+				this.on( 'hide', this.deactivateShowListeners, this );
+			};
+
+			CKEDITOR.ui.inlineToolbar.prototype.attach = ( function() {
+				var winGlobal, frame, editable, isInline;
```
mlewand: Same thing as above, why copy'n'paste?

tests/plugins/inlinetoolbar/positioning.js
```diff
@@ -0,0 +1,92 @@
...
+	bender.editor = {
+		name: 'editor1',
+		config: {
+			allowedContent: true,
+			height: 300,
+			width: 300
+		}
+	};
+
+	var inlineToolbarConfig = {
```
mlewand: There's no benefit in extracting this variable, it's used only in one place.

tests/plugins/inlinetoolbar/positioning.js
```diff
@@ -0,0 +1,92 @@
...
+			width: 100,
+			height: 200
+		},
+		spy = sinon.spy,
+		tests = {
+			// There's no need to call editor.window.scrollTo, because balloon logic initialy considers only
```
mlewand: Typo `initialy` -> `initially`

tests/plugins/inlinetoolbar/positioning.js
```diff
@@ -0,0 +1,92 @@
+/* bender-tags: a11ychecker,inlinetoolbar */
```
mlewand: This is a test unrelated to a11ychecker.

tests/plugins/inlinetoolbar/positioning.js
```diff
@@ -0,0 +1,92 @@
...
+			setUp: function() {
+				this.inlineToolbar = new CKEDITOR.ui.inlineToolbar( this.editor, inlineToolbarConfig );
+
+				this.markerElement = this.editor.editable().findOne( '#marker' );
+
+				// Make sure that global window is scrolled down to 0,0.
+				window.scrollTo( 0, 0 );
+
+				if ( !this._getFrameMethodReplaced ) {
+					// The problem is also window.getFrame().getClientRect() as it retursn different results from dashboard and directly.
```
mlewand: Typo retursn -> returns

tests/plugins/inlinetoolbar/positioning.js
```diff
@@ -0,0 +1,92 @@
+/* bender-tags: a11ychecker,inlinetoolbar */
+/* bender-ckeditor-plugins: toolbar,link,inlinetoolbar */
+/* bender-include: ../balloonpanel/_helpers/tools.js */
+/* global balloonTestsTools */
+
+( function() {
+	'use strict';
+
+	CKEDITOR.dom.window.prototype.getViewPaneSize = function() {
```
mlewand: Never overwrite any external method without restoring it at the end.

tests/plugins/inlinetoolbar/positioning.js
```diff
@@ -0,0 +1,92 @@
...
+		return {
+			width: 1183,
+			height: 1818
+		};
+	};
+
+	bender.editor = {
+		name: 'editor1',
+		config: {
+			allowedContent: true,
```
mlewand: I don't think there's a strong reason to disable ACF.

tests/plugins/inlinetoolbar/positioning.js
mlewand: Whole body of this if instruction could be replaced just with `sinon.stub( this.editor.window, 'getFrame', { height: 300, width: 300, left: 1, bottom: 643, right: 301, top: 343 } );`.

It's much more readable.

tests/plugins/inlinetoolbar/positioning.js
```diff
@@ -0,0 +1,92 @@
...
+				}
+			},
+
+			tearDown: function() {
+				this.inlineToolbar.destroy();
+				this.inlineToolbar = null;
+			},
+
+
+			'test classic - out of view - bottom center': function() {
```
mlewand: Include also tests for `divarea` editor.

Comandeer:
I don’t feel that having `attach` method, which allow to attach it to particular element is a right approach. I think that `attach` method should be a private one.

I imagine that to use such toolbar I have to:

* create an instance of it and attach it to the editor (something like `editor.createInlineToolbar` or `editor.addInlineToolbar`)
* define buttons in toolbar
* define contexts, so when the toolbar would be visible (e.g. `img` would mean that  the toolbar will be visible after selecting image in the editor).

Whole logic connected with actual displaying the toolbar could be even hidden from the end user, as defining context should be good enough (look at our commands).

WDYT?

mlewand:
Thanks @Comandeer, you're right that proposal makes a lot of sense. As a result I think that:

* All the `attach` alike methods should stay only in `InlineToolbarView` type.
* Let's create a separate TP issue for the toolbar context API.

Meaning that we won't deal with the context API in this PR yet. Does that make sense?

plugins/inlinetoolbar/plugin.js
```diff
@@ -0,0 +1,92 @@
...
+					label: 'test inlinetoolbar',
+					command: 'testInlineToolbar',
+					toolbar: 'insert'
+				} );
+			}
+			temporaryCode();
+		}
+	} );
+
+	CKEDITOR.ui.inlineToolbarView = function( editor, definition ) {
```
Comandeer: It seems odd to define constructor after defining its whole prototype. I'd move const

Comandeer: It seems odd to define constructor after defining its whole prototype. I'd move the constructor to the top.

plugins/inlinetoolbar/plugin.js
```diff
@@ -0,0 +1,92 @@
...
+				this._detachListiners();
+			};
+
+			CKEDITOR.ui.inlineToolbarView.prototype.create = function( element ) {
+				this.attach( element );
+
+				var that = this,
+					editable = this.editor.editable();
+				this._detachListiners();
+				this.listeners = {
```
Comandeer: I don't like how currently listeners are working. If `attachListener` and `on` methods return object with `removeListener` method, then `listeners` could be just an array of such objects, which would simplify `_detachListeners` implementation to something like:

```javascript
CKEDITOR.tools.forEach( this.listeners, function( listener ) {
    listener.removeListener();
} );
```

It'll also make whole listener code more universal and flexible.

plugins/inlinetoolbar/plugin.js
```diff
@@ -0,0 +1,92 @@
...
+				CKEDITOR.ui.balloonPanel.prototype.destroy.call( this );
+				this._detachListiners();
+			};
+
+			CKEDITOR.ui.inlineToolbarView.prototype.create = function( element ) {
+				this.attach( element );
+
+				var that = this,
+					editable = this.editor.editable();
+				this._detachListiners();
```
Comandeer: Typo, should be `_detachListeners`.

plugins/inlinetoolbar/plugin.js
```diff
@@ -0,0 +1,92 @@
...
+				this.editor.on( 'resize', this.listeners.resize );
+				this.listeners.scrollEvent = editable.attachListener( editable.getDocument(), 'scroll', this.listeners.scroll );
+			};
+
+			CKEDITOR.ui.inlineToolbarView.prototype.detach = function() {
+				this._detachListiners();
+				this.hide();
+			};
+
+			function temporaryCode() {
```
Comandeer: I'd move this code to the manual test/sample.

tests/plugins/inlinetoolbar/positioning.js
```diff
@@ -0,0 +1,78 @@
+/* bender-tags: inlinetoolbar */
+/* bender-ckeditor-plugins: toolbar,link,inlinetoolbar */
+/* bender-include: ../balloonpanel/_helpers/tools.js */
+/* global balloonTestsTools */
+
+( function() {
+	'use strict';
+
+	var restore = CKEDITOR.dom.window.prototype.getViewPaneSize;
+	CKEDITOR.dom.window.prototype.getViewPaneSize = function() {
```
Comandeer: Can't it be done in before callback using `sinon.stub` and restored in after callback?

plugins/inlinetoolbar/plugin.js
```diff
@@ -0,0 +1,114 @@
...
+				this.listeners.push( this.editor.on( 'resize', function() {
+					that.attach( element, false );
+				} ) );
+				this.listeners.push( editable.attachListener( editable.getDocument(), 'scroll', function() {
+					that.attach( element, false );
+				} ) );
+			};
+
+			/**
+			 * Hides the inline toolbar, detaches listners and moves the focus back to the editable.
```
mlewand: typo listners -> listeners

plugins/inlinetoolbar/plugin.js
```diff
@@ -0,0 +1,114 @@
...
+			 * all associated event listeners.
+			 */
+			CKEDITOR.ui.inlineToolbarView.prototype.destroy = function() {
+				CKEDITOR.ui.balloonPanel.prototype.destroy.call( this );
+				this._detachListeners();
+			};
+
+			/**
+			 * Places the inline toolbar next to a specified element so the tip of the toolbar's triangle
+			 * touches that element. Once the toolbar is attached it gains focus and attach DOM change listiners.
```
mlewand: Typo listiners -> listeners.

mlewand: Hmm I'm wondering on this part:

> Once the toolbar is attached it gains focus and attach DOM change listeners.

I think this should be implementation detail that should not be exposed to the here in API docs. If anything, it should be mentioned just in the `inlineToolbarView` type docs.

plugins/inlinetoolbar/plugin.js
```diff
@@ -0,0 +1,114 @@
...
+	CKEDITOR.inlineToolbar = function( editor ) {
+		this.inlineToolbar = new CKEDITOR.ui.inlineToolbarView( editor );
+	};
+
+	CKEDITOR.plugins.add( 'inlinetoolbar', {
+		requires: 'balloonpanel',
+		init: function() {
+			CKEDITOR.ui.inlineToolbarView.prototype = CKEDITOR.tools.extend( {}, CKEDITOR.ui.balloonPanel.prototype );
+			/**
+			 * build inline toolbar DOM representation.
```
mlewand: Sentence should start with a capitalized letter. Also one extra line of space between the comment and previous line should be kept.

tests/plugins/inlinetoolbar/positioning.js
f1ames: As @Comandeer suggested, please use `sinon.stub` and move it to `setUp` and `tearDown` methods. As we don't have `after` method, and only `tearDown` one (like `afterEach`) it makes sense to keep it in `setUp` (like `beforeEach`) and `tearDown` methods, so the stub should be created in `setUp` and then restored in `tearDown`.

tests/plugins/inlinetoolbar/positioning.js
```diff
@@ -0,0 +1,78 @@
...
+					height: 200
+				} );
+
+				this.markerElement = this.editor.editable().findOne( '#marker' );
+
+				// Make sure that global window is scrolled down to 0,0.
+				window.scrollTo( 0, 0 );
+
+				if ( !this._getFrameMethodReplaced ) {
+					sinon.stub( this.editor.window, 'getFrame', { height: 300, width: 300, left: 1, bottom: 643, right: 301, top: 343 } );
```
f1ames: Same here, you can just create stub in `setUp` and restore in `tearDown` - no flag is needed then.

tests/plugins/inlinetoolbar/positioning.js
```diff
@@ -0,0 +1,78 @@
...
+			tearDown: function() {
+				this.inlineToolbar.destroy();
+				this.inlineToolbar = null;
+			},
+
+
+			'test classic - out of view - bottom center': function() {
+				// Position acts as if the editor viewport was at position x: 260, 0.
+				this.markerElement.getClientRect = sinon.stub().returns( { height: 15, width: 25, left: 129, bottom: 416.34375, right: 154, top: 401.34375 } );
+				var moveSpy = spy( this.inlineToolbar, 'move' );
```
f1ames: Same here, setting `moveSpy` can also be extracted to `setUp`/`tearDown`.

plugins/inlinetoolbar/plugin.js
```diff
@@ -0,0 +1,145 @@
...
+		/**
+	 * A class that represents a inline toolbar capable of presenting defined
+	 * menu buttons.
+	 *
+	 * @class
+	 * @since 4.8
+	 * @param {CKEDITOR.editor} editor The editor instance for which the panel is created.
+	 */
+	CKEDITOR.inlineToolbar = function( editor ) {
+		this.inlineToolbar = new CKEDITOR.ui.inlineToolbarView( editor );
```
f1ames: I think it should be `this.inlineToolbarView = ...`. Also missing docs here.

plugins/inlinetoolbar/plugin.js
```diff
@@ -0,0 +1,145 @@
...
+			 * @method attach
+			 * @param {CKEDITOR.dom.element} element The element to which the panel is attached.
+			 */
+			CKEDITOR.ui.inlineToolbarView.prototype.create = function( element ) {
+				this.attach( element );
+
+				var that = this,
+					editable = this.editor.editable();
+				this._detachListeners();
+				this.listeners = [];
```
f1ames: This line is not needed here as it is already in `_detachListeners`.

wojtekw92:
@f1ames as mentioned earlier `CKEDITOR.inlineToolbar` is subject of another task from user story. This PR is only about positioning and and `inlineToolbarView`.

plugins/inlinetoolbar/plugin.js
```diff
@@ -0,0 +1,145 @@
...
+	};
+
+	CKEDITOR.plugins.add( 'inlinetoolbar', {
+		requires: 'balloonpanel',
+		init: function() {
+			CKEDITOR.ui.inlineToolbarView.prototype = CKEDITOR.tools.extend( {}, CKEDITOR.ui.balloonPanel.prototype );
+			/**
+			 * Build inline toolbar DOM representation.
+			 */
+			CKEDITOR.ui.inlineToolbarView.prototype.build = function() {
```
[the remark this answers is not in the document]
wojtekw92: Here no but next PR is using this to rebuild button structure. So after all we will overwrite this.

plugins/inlinetoolbar/plugin.js
```diff
@@ -0,0 +1,145 @@
...
+				} ) );
+				this.listeners.push( editable.attachListener( editable.getDocument(), 'scroll', function() {
+					that.attach( element, false );
+				} ) );
+			};
+
+			/**
+			 * Hides the inline toolbar, detaches listeners and moves the focus back to the editable.
+			 */
+			CKEDITOR.ui.inlineToolbarView.prototype.detach = function() {
```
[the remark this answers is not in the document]
wojtekw92: It will be probably moved to `CKEDITOR.inlineToolbar`

plugins/inlinetoolbar/plugin.js
```diff
@@ -0,0 +1,145 @@
...
+					CKEDITOR.tools.array.forEach( this.listeners, function( listener ) {
+						listener.removeListener();
+					} );
+					this.listeners = [];
+				}
+			};
+
+			/**
+			 * Destroys the inline toolbar by removing it from DOM and purging
+			 * all associated event listeners.
```
mlewand: Again, docs copy pasta from the base class. Instead use `@inheritdoc`.

f1ames:
@wojtekw92 Ok, makes sense, didn't catch that one. So as `inlineToolbarView` is only subject of this PR I suggest removing `CKEDITOR.inlineToolbar` entirely from this PR as it will be covered in another task. This way we could make it clearer and contain only code which is related to this issue.

plugins/inlinetoolbar/plugin.js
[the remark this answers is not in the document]
f1ames: See my comment below then -> https://github.com/ckeditor/ckeditor-dev/pull/957#issuecomment-334168780.

plugins/inlinetoolbar/plugin.js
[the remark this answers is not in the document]
f1ames: Ok, I see. It might be hard to code the same functionality in 3 separated PRs :) While this code is correct, and my proposal was only a simplification it might stay as it is.

mlewand:
There's an issue with branching in this PR. It's using `t/933` branch targetted directly to `major`, while it should use `t/933-2955` targetted at `t/933`. `t/933` should be a branch where all the R+ PR are merged to, we obviosuly don't want to merge incomplete code to the `major` branch.

Now we can't change the source branch of this PR (and I don't want to lose the history), so the correct feature branch (`t/933`) will be set after this PR is closed. For now it will be targetted to a temp branch `t/933-2955`.

tests/plugins/inlinetoolbar/positioning.js
wojtekw92: It was the shortest way to make `<span id="marker">bom</span>` available

tests/plugins/inlinetoolbar/positioning.js
wojtekw92: After consultation with @Comandeer, about problems in this solution we decided to use the solution from `ballonpanel test`

f1ames:
Just wanted to mention that there is one issue regarding first balloonpanel position/appearance - https://dev.ckeditor.com/ticket/16622. It is visible in some inlinetoolbar manual test too (e.g. `tests/plugins/inlinetoolbar/manual/inlinetoolbar`, `http://tests.ckeditor.dev:1030/tests/plugins/inlinetoolbar/manual/kama`) and could be addressed. WDYT?

mlewand:
> Just wanted to mention that there is one issue regarding first balloonpanel position/appearance - https://dev.ckeditor.com/ticket/16622.

This is another issue that should not be tracked as a part of this ticket. I'll move it to GH.

wojtekw92:
@mlewand `positioning.divarea.js` is testing inlinetoolbar in divareaa editor. There are two tests: `test classic - out of view - bottom center` and `test classic - out of view - hcenter top`.
Both tests assume that inline toolbar is attached to the element that is out of view, below or above current view. According to this we have to assume that:
* if we have our `marker` above we will have `inlinetoolbar` sticked to the top edge of editor with arrow facing upwards.
* if we have our `marker` below the situation should be opposite and we will have inlinetoolbar sticked to the bottom edge of editor with arrow facing down.

The way that is wortking:
* At the beginning we are creating our `window` with size 
```
{
    width: 1183,
    height: 1818
},
```
* After that we are creating instance of our inlineToolbarView with size `100px x 200px`.
* next think that we have to do is to define our `getClientRect` for window that is moved `343px` below top
* after that we can start our test cases where we are setting up client rect for our `marker` element below or above our `window` and checking if inlineToolbar position is proper.

mlewand:
What you wrote above makes a perfect sense to me, and this is what we need. However this is not what this test case checks. Below is a screenshot of position for `test divaera - out of view - bottom center`:

test divaera - out of view - bottom center screen

This does not align in the correct position, it's in the middle of nowhere.

Second case `test divaera - out of view - hcenter top`, same thing:

test divaera - out of view - hcenter top screen

Despite incorrect positions the test passes.

f1ames:
> Just wanted to mention that there is one issue regarding first balloonpanel position/appearance - https://dev.ckeditor.com/ticket/16622.

> This is another issue that should not be tracked as a part of this ticket. I'll move it to GH.

@mlewand Yes, I agree. Just wanted to point out that this issue exists and maybe we could try to solve it while working on `inlinetoolbar`as it is already visible during tests.

tests/plugins/inlinetoolbar/positioning.js
```diff
@@ -0,0 +1,82 @@
...
+				height: 200,
+				width: 600
+			}
+		}
+	};
+
+	var tests = {
+
+		'test divaera - out of view - bottom center': function( editor ) {
+			var inlineToolbar = new CKEDITOR.ui.inlineToolbarView( editor, {
```
mlewand: Code style: subsequent `var`s should be declared/initialized in a single `var`, like so:

```js
var inlineToolbar = new CKEDITOR.ui.inlineToolbarView( editor, {
		width: 100,
		height: 200
	} ),
	markerElement = editor.editable().findOne( '#marker' ),
	frame = editor.editable().isInline() ? editor.editable().getClientRect() : editor.window.getFrame().getClientRect(),
	elementFrame = markerElement.getClientRect();
```

Also `inlineToobarRect` should be joined in it.

tests/plugins/inlinetoolbar/positioning.js
```diff
@@ -0,0 +1,82 @@
...
+			markerElement.getParent().getNext().scrollIntoView( true );
+			inlineToolbar.create( markerElement );
+			var inlineToobarRect = inlineToolbar.parts.panel.getClientRect();
+			assert.areEqual( frame.left + elementFrame.left + elementFrame.width / 2 - 50, inlineToobarRect.left, 'left align' );
+			assert.areEqual( frame.top, inlineToobarRect.top - inlineToolbar.triangleHeight, 'top align' );
+			inlineToolbar.destroy();
+			inlineToolbar = null;
+		}
+	};
+	tests = bender.tools.createTestsForEditors( CKEDITOR.tools.objectKeys( bender.editors ), tests );
```
mlewand: Please make one line of spacing between tests, and the close of object literal.

tests/plugins/inlinetoolbar/positioning.js
```diff
@@ -0,0 +1,82 @@
...
+				extraPlugins: 'divarea',
+				extraAllowedContent: 'span[id]',
+				height: 200,
+				width: 600
+			}
+		}
+	};
+
+	var tests = {
+
```
mlewand: This (spacing) line is not needed.

tests/plugins/inlinetoolbar/manual/inlinetoolbar.html
```diff
@@ -0,0 +1,58 @@
+<style>
+	body {
```
mlewand: Add a reference to proper ticket for this CSS trick.

mlewand:
Let's wait for @f1ames review. After that I'll merge the changes.

PR #958, opened 2017-09-25
Image extraction methods for PFW Images plugin

msamsel:

New feature

### This PR contains

- [x] Unit tests
- [ ] Manual tests

## What changes did you make?
- 2 new methods to extract images:
  - first to get images from RTF
  - second to extract img tags from html string.


tests/plugins/pastefromwordimage/helpers.js
```diff
@@ -0,0 +1,84 @@
...
+		'test extract images from rtf': function() {
+			var testCase,
+				result,
+				i,
+				length;
+
+			while ( testCase = fixtures.shift() ) {
+				result = CKEDITOR.plugins.pastefromwordimage.extractImagesFromRtf( testCase.rtf );
+				if ( result ) {
+					for ( i = 0, length = result.length; i < length; i++ ) {
```
mlewand: It's not really worth to extract this `length` variable.

tests/plugins/pastefromwordimage/helpers.js
```diff
@@ -0,0 +1,84 @@
...
+				{
+					htmlString: '<p>There is no img</p>',
+					result: []
+				}
+			];
+
+			while ( item = test.shift() ) {
+				result = CKEDITOR.plugins.pastefromwordimage.extractImgTagsFromHtmlString( item.htmlString );
+				assert.areSame( item.result.length, result.length );
+				for ( i = 0, length = result.length; i < length; i++ ) {
```
mlewand: It's not really worth to extract this `length` variable. It decreases the readability.

plugins/pastefromwordimage/plugin.js
```diff
@@ -0,0 +1,96 @@
...
+		 * @returns {Object} return.Object Single image found in `rtfClipboard`.
+		 * @returns {String/null} return.Object.hex Hexadecimal string of image embeded in rtf clipboard.
+		 * @returns {String/null} return.Object.type String represent type of image, allowed values: 'image/png', 'image/jpeg' or `null`
+		 */
+		extractImagesFromRtf: function( rtfClipboard ) {
+			var images = [],
+				rePictureOrShape = /((?:\{\\\*\\shppict[\s\S]+?{\\\*\\blipuid\s+[0-9a-f]+\}\s?)|(?:\{\\shp[\s\S]+?\{\\\*\\svb\s?))([0-9a-f\s]+)\}\}/g,
+				rePictureHeader = /\{\\\*\\shppict[\s\S]+?{\\\*\\blipuid\s+[0-9a-f]+\}\s?/,
+				reShapeHeader = /\{\\shp[\s\S]+?\{\\\*\\svb\s?/,
+				wholeImage,
```
mlewand: It's not a single image, it's about multiple images.

plugins/pastefromwordimage/plugin.js
```diff
@@ -0,0 +1,96 @@
...
+		 * @private
+		 * @since 4.8.0
+		 * @param {String} rtfClipboard Data obtained from rtf clipboard.
+		 * @returns {Array} Contains array of objects with images.
+		 * @returns {Object} return.Object Single image found in `rtfClipboard`.
+		 * @returns {String/null} return.Object.hex Hexadecimal string of image embeded in rtf clipboard.
+		 * @returns {String/null} return.Object.type String represent type of image, allowed values: 'image/png', 'image/jpeg' or `null`
+		 */
+		extractImagesFromRtf: function( rtfClipboard ) {
+			var images = [],
```
mlewand: Would it be worth to rename this variable to `ret`? Reason why I'm asking for this is that we have `wholeImage` in other places, which sounds similar and can be mistaken. At the same time this variable will emphasize it's purpose.

plugins/pastefromwordimage/plugin.js
```diff
@@ -0,0 +1,96 @@
...
+				reShapeHeader = /\{\\shp[\s\S]+?\{\\\*\\svb\s?/,
+				wholeImage,
+				imageType;
+
+			wholeImage = rtfClipboard.match( rePictureOrShape );
+			if ( !wholeImage ) {
+				return;
+			}
+
+			for ( var i = 0, len = wholeImage.length; i < len; i++ ) {
```
mlewand: There's very little if any benefit from extracting len into a var, while it decreases the readability.

plugins/pastefromwordimage/plugin.js
```diff
@@ -0,0 +1,96 @@
...
+		 * @since 4.8.0
+		 * @param {String} rtfClipboard Data obtained from rtf clipboard.
+		 * @returns {Array} Contains array of objects with images.
+		 * @returns {Object} return.Object Single image found in `rtfClipboard`.
+		 * @returns {String/null} return.Object.hex Hexadecimal string of image embeded in rtf clipboard.
+		 * @returns {String/null} return.Object.type String represent type of image, allowed values: 'image/png', 'image/jpeg' or `null`
+		 */
+		extractImagesFromRtf: function( rtfClipboard ) {
+			var images = [],
+				rePictureOrShape = /((?:\{\\\*\\shppict[\s\S]+?{\\\*\\blipuid\s+[0-9a-f]+\}\s?)|(?:\{\\shp[\s\S]+?\{\\\*\\svb\s?))([0-9a-f\s]+)\}\}/g,
```
mlewand: You could dynamically build this regexp, based on two former. JS has a ways to build regexp from string, that way you'll avoid copy-pasta.

plugins/pastefromwordimage/plugin.js
```diff
@@ -0,0 +1,99 @@
...
+		 *
+		 * @private
+		 * @since 4.8.0
+		 * @param {String} rtfClipboard Data obtained from RTF clipboard.
+		 * @returns {Array} Contains array of objects with images or empty array if there weren't a match.
+		 * @returns {Object} return.Object Single image found in `rtfClipboard`.
+		 * @returns {String/null} return.Object.hex Hexadecimal string of image embedded in RTF clipboard.
+		 * @returns {String/null} return.Object.type String represent type of image, allowed values: 'image/png', 'image/jpeg' or `null`
+		 */
+		extractImagesFromRtf: function( rtfClipboard ) {
```
mlewand: `rtfClipboard` could be renamed to `rtfContent` it doesn't have to be anything related to clipboard. It might be just simply a plain RTF file and that's fine too.

msamsel:
While testing there was detect situation, when images are not properly extract from the RTF. This requires tunning a little bit better RegExp, which process RTF. Detected problems:
- images with changed wrap around setting
- images with hyperlinks
- images under MacOS word

plugins/pastefromwordimage/plugin.js
```diff
@@ -0,0 +1,99 @@
...
+	 */
+	CKEDITOR.plugins.pastefromwordimage = {
+		/**
+		 * Methods parse RTF clipboard to find embedded images.
+		 *
+		 * @private
+		 * @since 4.8.0
+		 * @param {String} rtfClipboard Data obtained from RTF clipboard.
+		 * @returns {Array} Contains array of objects with images or empty array if there weren't a match.
+		 * @returns {Object} return.Object Single image found in `rtfClipboard`.
```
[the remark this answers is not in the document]
msamsel: Hmmm it seems to work quite well and docs are build without any error. I can't find information how to represent return value in such case (array of objects with properties description). Do you have any suggestions how it might be described? Or should I just remove all information what is returned by this method?
screenshot-localhost-2017-10-02-13-51-17-800

plugins/pastefromwordimage/plugin.js
```diff
@@ -0,0 +1,98 @@
...
+		 * @returns {Object} return.Object Single image found in `rtfContent`.
+		 * @returns {String/null} return.Object.hex Hexadecimal string of image embedded in RTF clipboard.
+		 * @returns {String/null} return.Object.type String represent type of image, allowed values: 'image/png', 'image/jpeg' or `null`
+		 */
+		extractImagesFromRtf: function( rtfContent ) {
+			var ret = [],
+				rePictureHeader = CKEDITOR.env.mac ? /\{\\pict[\s\S]+?\\bliptag\-?\d+(\\blipupi\d+?)?\s?/ : /\{\\pict[\s\S]+?\{\\\*\\blipuid[\s0-9a-f]+?\}\s?/,
+				reShapeHeader = /\{\\shp[\s\S]+?\{\\\*\\svb\s?/,
+				rePictureOrShape = new RegExp( '(?:(' + rePictureHeader.source + ')|(' + reShapeHeader.source + '))([0-9a-f\\s]+)\\}', 'g' ),
+				unwantedType = CKEDITOR.env.mac ? '\\macpict' : '\\wmetafile',
```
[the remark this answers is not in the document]
mlewand: It would actually solve the need to use this hack: https://github.com/ckeditor/ckeditor-dev/pull/958/files#diff-086a650d277ef8af879ae71f10875ae1R62

tests/plugins/pastefromwordimage/helpers.js
```diff
@@ -0,0 +1,112 @@
...
+				actualResult,
+				testCase,
+				// Test description:
+				// 1. Single Image
+				// 2. Mixed: image online, image offline, word shape
+				// 3. No images in RTF
+				// 4. Image with hyperlink
+				// 5. Image with changed wrapping option
+				testCases = [
+				{
```
mlewand: Wrong indentation, it should be intended by one tab more (following elements too), just like it is in the "test extract images hexstrings from rtf for MacOS" TC.

tests/plugins/pastefromwordimage/helpers.js
```diff
@@ -0,0 +1,112 @@
...
+				},
+				{
+					rtf: CKEDITOR.ajax.load( './_fixtures/helpers5.rtf' + '?' + this.cacheBuster() ),
+					result: JSON.parse( CKEDITOR.ajax.load( './_fixtures/helpers5.json' + '?' + this.cacheBuster() ) )
+				}
+			];
+			CKEDITOR.env.mac = false;
+			while ( testCase = testCases.shift() ) {
+				actualResult = CKEDITOR.plugins.pastefromwordimage.extractImagesFromRtf( testCase.rtf );
+				assert.areSame( testCase.result.length, actualResult.length );
```
mlewand: There's no information which case (file) failed, if it fails. Just modify one expectation so that it fails, and run the tests - are you able to guess which one is? 😉

tests/plugins/pastefromwordimage/helpers.js
```diff
@@ -0,0 +1,112 @@
...
+			return ( new Date() ).getTime().toString() + Math.floor( Math.random() * 1000000 ).toString();
+		},
+
+		'test img tags extraction from html string': function() {
+			var sourceLength = 5,
+				i;
+
+			for ( i = 0; i < sourceLength; i++ ) {
+				bender.tools.testInputOut( 'source' + i, function( input, output ) {
+					var expectedResult = JSON.parse( output.replace( /<!--/, '' ).replace( /-->/, '' ) ),
```
mlewand: I like the creative use of `testInputOutput` to use it for array comparison using JSON. However why are you wrapping expected results with a HTML comments?

Keep it simple.

tests/plugins/pastefromwordimage/helpers.js
mlewand: And here we come to resource loading. Please change it to load resources asynchronously (test of will have to work asynchronously too, so it will use `wait` and `resume`).

Still `CKEDITOR.ajax.load` can be used for async loading.

mlewand:
Ok I think we're closer to concluding this one. I have pushed 2 commits with docs adjustments.

plugins/pastefromwordimage/plugin.js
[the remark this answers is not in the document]
msamsel: I've removed this logic but still hack needs to remain, because of this: https://github.com/ckeditor/ckeditor-dev/pull/958/files/9eab02612c88318ed129b7697d8e98c4ab584188#diff-d3d4bfb15d2eb9b8bcdfda799d8f109dR35
MacOS and Windows had 2 different rtf structures, which are recognised with 2 different patterns. So if we want to have working tests (for MacOS and Windows) running under single system, then env variable must be overwritten before test start.

tests/plugins/pastefromwordimage/helpers.js
msamsel: I was pretty sure that incorrect HTML markup might break something in bender, but it seems to work fine. Most probably I tested it not on `textarea`s but on regular `div`s. :/

PR #976, opened 2017-09-28
Paste replacement

msamsel:

New feature

yes

### This PR contains

- [x] Unit tests
- [x] Manual tests

## What changes did you make?

- add plugin which embed images while pasting from Word

Related to #662


dev/pastefromwordimage/getclipboard.html
```diff
@@ -0,0 +1,97 @@
...
+		<div style="float: right; width: 49%">
+			<p>Raw RTF Data Received:</p>
+			<textarea data-name="input.rtf" id="rawRtf" readonly="readonly"></textarea>
+			<button id="rtfData">Save RTF Data</button>
+		</div>
+	</div>
+	<div style="width: 100%; float: left;">
+		<p>After Paste Processing:</p>
+		<textarea id="output" readonly="readonly"></textarea>
+	</div>
```
f1ames: Maybe you could also add a container which will show all images pasted to the CKEditor (in the form already transformed by CKEditor), something like `Extracted Images`? If it's relatively easy to do (like 0,5h) you could take a look at it. From the other hand you can see pasted images inside editor instance, so not sure if it is needed. WDYT?

tests/plugins/pastefromwordimage/manual/pastewordwithimage.md
```diff
@@ -0,0 +1,16 @@
+@bender-tags: bug, 4.8.0, 662, pastefromwordimage
+@bender-ui: collapsed
+@bender-ckeditor-plugins: wysiwygarea, toolbar, undo, pastefromwordimage, sourcearea, elementspath, newpage
```
f1ames: I would add `resize` plugin so you can enlarge editor to see whole pasted content.

tests/plugins/pastefromwordimage/manual/pastewordwithimage.html
```diff
@@ -0,0 +1,4 @@
+<textarea id="editor" cols="30" rows="10"></textarea>
+<script>
```
f1ames: This test should be ignored in browsers not supporting Clipboard API IMHO.

f1ames: Alternatively you may describe what is the expected behaviour for such browsers.

msamsel:
1. I redesigned PFW helpers to have possibility for using it in PFW Image tests, but in future most probably it will have to be separate again. Because there will be required to process files outside of PFW event, so logic which make assertion will have to change.
https://github.com/ckeditor/ckeditor-dev/blob/128516a52485505176189743962eaf97d52aa067/tests/plugins/pastefromwordimage/generated/_helpers/createTestCase.js#L40-L115
2. I rename tests to include "Image" word at the end.
3. Failing example is caused by wrong processing of Shapes by PFW. I've created an upstream issue with it #1069.

dev/pastefromwordimage/getclipboard.html
msamsel: On this stage of implementation it might be problematic. I could process RTF clipboard, but it might contains also Word Shapes, which are ignored for processing while pasting. That's why preparing similar solution here might be repeating functionality of the plugin, which seems to be not a good thing.
It should be easier, when further changes related to file transfer will be implemented:
https://github.com/ckeditor/ckeditor-dev/blob/128516a52485505176189743962eaf97d52aa067/plugins/pastefromwordimage/plugin.js#L60-L63
There is nice event when will be exposed processed URL and image which is going to be embed. With that implementation it should like 0,5h of work :)

f1ames: @msamsel Ok, so let's leave it as it is for now.

plugins/pastefromwordimage/plugin.js
```diff
@@ -8,9 +8,55 @@
 
 	CKEDITOR.plugins.add( 'pastefromwordimage', {
 		requires: 'pastefromword',
-		init: function() {}
+		init: function( editor ) {
+			if ( CKEDITOR.env.ie || CKEDITOR.env.iOS ) {
```
f1ames: Was there any particular reason why it was changed from `!CKEDITOR.plugins.clipboard.isCustomDataTypesSupported` to `CKEDITOR.env.ie || CKEDITOR.env.iOS`?

If it is about iOS you may always go with `!CKEDITOR.plugins.clipboard.isCustomDataTypesSupported || CKEDITOR.env.iOS` which is more flexible (as `CKEDITOR.plugins.clipboard.isCustomDataTypesSupported` may change at some point and then it will automatically work here too).

tests/plugins/pastefromwordimage/manual/pastewordwithimage.html
```diff
@@ -1,4 +1,7 @@
 <textarea id="editor" cols="30" rows="10"></textarea>
 <script>
+	if ( !CKEDITOR.plugins.clipboard.isCustomDataTypesSupported ) {
```
f1ames: This code should be called at least after [`pluginsLoaded`](https://docs.ckeditor.com/ckeditor4/docs/#!/api/CKEDITOR.editor-event-pluginsLoaded) event as at this point it throws an error
image

because `CKEDITOR.plugins.clipboard` is not initialized yet.

f1ames:
> but in future most probably it will have to be separate again. Because there will be required to process files outside of PFW event, so logic which make assertion will have to change.

@msamsel Ok, I hope we will be able to extract the different logic/flows to some separate helper and still keep it DRY, without duplicating too much code.

plugins/pastefromwordimage/plugin.js
msamsel: I've change wrong file :( 
It supposed to be switched in manual test here: https://github.com/ckeditor/ckeditor-dev/blob/aa75e0874a372dd89168314eaae2162c248a1ff9/tests/plugins/pastefromwordimage/manual/pastewordwithimage.html#L3
To avoid problem with loading plugins. As manual test are usually fired only while releasing, then such change should be sufficient there.

msamsel:
@f1ames linking images problem was reported here: #995

msamsel:
>>but in future most probably it will have to be separate again. Because there will be required to process files outside of PFW event, so logic which make assertion will have to change.

> Ok, I hope we will be able to extract the different logic/flows to some separate helper and still keep it DRY, without duplicating too much code.

What I think now it might be possible to prepare alternative assert file, similar to current `assertWordFilter.js`, but dedicated for assert with images.

f1ames:
> What I think now it might be possible to prepare alternative assert file, similar to current assertWordFilter.js, but dedicated for assert with images.

@msamsel I think it might make a lot of sense for this case. We will see after starting working on it 👍

tests/plugins/pastefromwordimage/manual/pastewordwithimage.md
```diff
@@ -0,0 +1,18 @@
...
+@bender-ckeditor-plugins: wysiwygarea, toolbar, undo, pastefromwordimage, sourcearea, elementspath, newpage, resize
+
+----
+
+1. Open document and copy whole content.
+1. Clear editor content.
+1. Paste copied content into CKEditor.
+
+### Perform above steps for all files:
+  * [Offline img](../generated/_fixtures/SimpleImages/SimpleImages.docx).
```
f1ames: Links doesn't work anymore as folder names were changed.

tests/plugins/pastefromwordimage/manual/pastewordwithimage.md
```diff
@@ -0,0 +1,18 @@
...
+  * [Online and offline img](../generated/_fixtures/MixedOnline/Mixed_drawings_and_online.docx),
+  * [Document with shape](../generated/_fixtures/MixedOnlineAndShapes/Mixed_drawings_and_online_and_shapes.docx). _Please notice that shape should be ignored while pasting._
+
+**Expected:** Images are pasted to the editor.
```
f1ames: As one of the files contains a shape which is not pasted it should ba also described here, e.g. _Keep in mind that shapes (like last object in`Mixed_drawings_and_online_and_shapes.docx`) will not be pasted_.

msamsel:
> Also some fixtures like e.g `OnlineAndOfflineImage/macos/firefox.html` and `OnlineAndOfflineImage/macos/chrome.htm` are identical so I was thinking about merging them to one (no sense in keeping both identical as then it will generate two identical tests). However, the corresponding RTF fixtures (`OnlineAndOfflineImage/macos/firefox.rtf` and `OnlineAndOfflineImage/macos/chrome.html`) are different so I think that is the main reason there are two (or rather four) separate files.
It wasn't a reason. I just haven't check it and use both available browsers supported RTF under OSX.

About merging it won't possible easily. Currently test generator assume that html and rtf files has the same names, so it won't work with merged file. Such possibility would require changes in test generator.
From what I see RTFs differ with some ID values inside. So those might be removed also. WDYT? If you consider, that those test doesn't provide additional value feel free to remove it.

f1ames:
> About merging it won't possible easily. Currently test generator assume that html and rtf files has the same names, so it won't work with merged file. Such possibility would require changes in test generator.
From what I see RTFs differ with some ID values inside. So those might be removed also. WDYT? If you consider, that those test doesn't provide additional value feel free to remove it.

I see. I was think about it as some kind of optimization, but as the overhead is very little we may leave it as is for now.

PR #977, opened 2017-09-28
CSS styling for inline toolbar

wojtekw92:

### This PR contains

- [ ] Unit tests
- [X] Manual tests

## What changes did you make?

*Give an overview…*

### DIFF
--- a/plugins/inlinetoolbar/plugin.js
+++ b/plugins/inlinetoolbar/plugin.js
@@ -32,24 +32,19 @@
 		this._listeners = [];
 	};
 
-	var stylesLoaded = false;
-
 	CKEDITOR.plugins.add( 'inlinetoolbar', {
 		requires: 'balloonpanel',
+		onLoad: function() {
+			CKEDITOR.document.appendStyleSheet( this.path + 'skins/' + CKEDITOR.skinName + '/inlinetoolbar.css' );
+		},
 		init: function() {
-			if ( !stylesLoaded ) {
-				CKEDITOR.document.appendStyleSheet( this.path + 'skins/' + CKEDITOR.skinName + '/inlinetoolbar.css' );
-				stylesLoaded = true;
-			}
 			CKEDITOR.ui.inlineToolbarView.prototype = CKEDITOR.tools.extend( {}, CKEDITOR.ui.balloonPanel.prototype );
-			if ( CKEDITOR.ui.inlineToolbarView.prototype.templateDefinitions && CKEDITOR.ui.inlineToolbarView.prototype.templateDefinitions.panel.indexOf( 'cke_inlinetoolbar' ) === -1 ) {
-				CKEDITOR.ui.inlineToolbarView.prototype.templateDefinitions.panel = CKEDITOR.ui.inlineToolbarView.prototype.templateDefinitions.panel.replace( 'cke_balloon', 'cke_balloon cke_inlinetoolbar' );
-			}
 			/**
 			 * Build inline toolbar DOM representation.
 			 */
 			CKEDITOR.ui.inlineToolbarView.prototype.build = function() {
 				CKEDITOR.ui.balloonPanel.prototype.build.call( this );
+				this.parts.panel.addClass( 'cke_inlinetoolbar' );
 				this.parts.title.remove();
 				this.parts.close.remove();
 			};
--- a/plugins/inlinetoolbar/plugin.js
+++ b/plugins/inlinetoolbar/plugin.js
@@ -24,9 +24,6 @@
 			triangleHeight: 10
 		} );
 		CKEDITOR.ui.balloonPanel.call( this, editor, definition );
-		this.listeners = [];
-	};
-
 		/**
 		 * Listeners registered by this toolbar view.
 		 *
--- a/plugins/inlinetoolbar/plugin.js
+++ b/plugins/inlinetoolbar/plugin.js
@@ -27,11 +27,16 @@
 		this._listeners = [];
 	};
 
+	var stylesLoaded = false;
+
 	CKEDITOR.plugins.add( 'inlinetoolbar', {
 		requires: 'balloonpanel',
 		init: function() {
+			if ( !stylesLoaded ) {
+				CKEDITOR.document.appendStyleSheet( this.path + 'skins/' + CKEDITOR.skinName + '/inlinetoolbar.css' );
+				stylesLoaded = true;
+			}
 			CKEDITOR.ui.inlineToolbarView.prototype = CKEDITOR.tools.extend( {}, CKEDITOR.ui.balloonPanel.prototype );
-
 			CKEDITOR.ui.inlineToolbarView.prototype.build = function() {
 				CKEDITOR.ui.balloonPanel.prototype.build.call( this );
 				this.parts.title.remove();
--- a/plugins/inlinetoolbar/plugin.js
+++ b/plugins/inlinetoolbar/plugin.js
@@ -18,19 +18,15 @@
 	 * docs for an example definition.
 	 */
 	CKEDITOR.ui.inlineToolbarView = function( editor, definition ) {
-<<<<<<< HEAD
-		CKEDITOR.ui.balloonPanel.call( this, editor, definition );
-=======
-		var defParams = CKEDITOR.tools.extend( definition || {}, {
+		definition = CKEDITOR.tools.extend( definition || {}, {
 			width: 'auto',
 			triangleWidth: 10,
 			triangleHeight: 10
 		} );
-		CKEDITOR.ui.balloonPanel.call( this, editor, defParams );
+		CKEDITOR.ui.balloonPanel.call( this, editor, definition );
 		this.listeners = [];
 	};
 
->>>>>>> review fixes and manual tests
 		/**
 		 * Listeners registered by this toolbar view.
 		 *
@@ -49,13 +45,12 @@
 				stylesLoaded = true;
 			}
 			CKEDITOR.ui.inlineToolbarView.prototype = CKEDITOR.tools.extend( {}, CKEDITOR.ui.balloonPanel.prototype );
-<<<<<<< HEAD
-=======
-			CKEDITOR.ui.inlineToolbarView.prototype.templateDefinitions.panel = CKEDITOR.ui.inlineToolbarView.prototype.templateDefinitions.panel.replace( 'cke_balloon', 'cke_inlinetoolbar' );
+			if (CKEDITOR.ui.inlineToolbarView.prototype.templateDefinitions && CKEDITOR.ui.inlineToolbarView.prototype.templateDefinitions.panel.indexOf( 'cke_inlinetoolbar' ) === -1 ) {
+				CKEDITOR.ui.inlineToolbarView.prototype.templateDefinitions.panel = CKEDITOR.ui.inlineToolbarView.prototype.templateDefinitions.panel.replace( 'cke_balloon', 'cke_balloon cke_inlinetoolbar' );
+			}
 			/**
 			 * Build inline toolbar DOM representation.
 			 */
->>>>>>> review fixes and manual tests
 			CKEDITOR.ui.inlineToolbarView.prototype.build = function() {
 				CKEDITOR.ui.balloonPanel.prototype.build.call( this );
 				this.parts.title.remove();
--- a/plugins/inlinetoolbar/plugin.js
+++ b/plugins/inlinetoolbar/plugin.js
@@ -45,7 +45,7 @@
 				stylesLoaded = true;
 			}
 			CKEDITOR.ui.inlineToolbarView.prototype = CKEDITOR.tools.extend( {}, CKEDITOR.ui.balloonPanel.prototype );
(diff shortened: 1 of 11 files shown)


plugins/inlinetoolbar/skins/kama/inlinetoolbar.css
```diff
@@ -0,0 +1,90 @@
+/*
+Copyright (c) 2003-2017, CKSource - Frederico Knabben. All rights reserved.
+For licensing, see LICENSE.md or http://ckeditor.com/license
+*/
+
+.cke_inlinetoolbar
+{
+	position: absolute;
+	z-index: 10000;
+
```
beatadelura: Extra line here added :)

plugins/inlinetoolbar/skins/kama/inlinetoolbar.css
```diff
@@ -0,0 +1,90 @@
...
+	background: #ffffff;
```
beatadelura: Could be used shorthand here: `#fff`

plugins/inlinetoolbar/skins/kama/inlinetoolbar.css
```diff
@@ -0,0 +1,90 @@
...
+	width: 0;
+	height: 0;
+}
+
+.cke_inlinetoolbar > .cke_balloon_triangle_outer > .cke_balloon_triangle_inner
+{
+	z-index: 0;
+}
+
+.cke_inlinetoolbar > .cke_balloon_triangle_outer
```
beatadelura: It could go to line `33` as `z-index: 0` has been repeated ;)

plugins/inlinetoolbar/skins/kama/inlinetoolbar.css
```diff
@@ -0,0 +1,90 @@
...
+}
+
+/* align: [ hcenter ] */
+.cke_inlinetoolbar > .cke_balloon_triangle_outer.cke_balloon_triangle_align_hcenter
+{
+	left: 50%;
+	margin-left: -10px;
+}
+
+
```
beatadelura: Extra line added.

plugins/inlinetoolbar/skins/moono-lisa/inlinetoolbar.css
```diff
@@ -0,0 +1,90 @@
+/*
+Copyright (c) 2003-2017, CKSource - Frederico Knabben. All rights reserved.
+For licensing, see LICENSE.md or http://ckeditor.com/license
+*/
+
+.cke_inlinetoolbar
+{
+	position: absolute;
+	z-index: 10000;
+
```
beatadelura: Same here, extra line ;)

plugins/inlinetoolbar/skins/kama/inlinetoolbar.css
```diff
@@ -0,0 +1,83 @@
+/*
+Copyright (c) 2003-2017, CKSource - Frederico Knabben. All rights reserved.
+For licensing, see LICENSE.md or http://ckeditor.com/license
+*/
+
+.cke_inlinetoolbar
```
mlewand: I don't like the fact of re-implementing such a generic CSS again in the inline toolbar plugin.

Instead the balloon panel created by this plugin should **add** `cke_inlinetoolbar` class to that outer most element.

Then you'd need to adjust selector from `.cke_inlinetoolbar` to `cke_balloon.cke_inlinetoolbar (...)`.

The main goal is to avoid code redundancy.

plugins/inlinetoolbar/skins/kama/inlinetoolbar.css
```diff
@@ -0,0 +1,83 @@
...
+}
+
+.cke_inlinetoolbar > .cke_balloon_triangle_outer,
+.cke_inlinetoolbar > .cke_balloon_triangle_outer > .cke_balloon_triangle_inner
+{
+	z-index: 0;
+}
+
+/* side: [ bottom, top ] */
+.cke_inlinetoolbar > .cke_balloon_triangle_outer.cke_balloon_triangle_bottom,
```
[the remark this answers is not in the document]
mlewand: Apply the above change to all the CSS files.

plugins/inlinetoolbar/plugin.js
```diff
@@ -18,18 +18,30 @@
 	 * @param {Object} definition An object containing the panel definition.
 	 */
 	CKEDITOR.ui.inlineToolbarView = function( editor, definition ) {
-		CKEDITOR.ui.balloonPanel.call( this, editor, definition );
+		var defParams = CKEDITOR.tools.extend( definition || {}, {
```
mlewand: Instead of adding a new var could you just overwrite the `definition` parameter?

mlewand:
@f1ames I'll ask you to continue the review for this one.

tests/plugins/inlinetoolbar/manual/kama.html
```diff
@@ -0,0 +1,47 @@
+<style>
+	body {
+		margin-left: 0px;
+		padding-left: 400px;
```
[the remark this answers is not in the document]
wojtekw92: I assume that it is a know issue as this problem is solved the same way in `balloonpanel` tests.

tests/plugins/inlinetoolbar/manual/kama.html
```diff
@@ -0,0 +1,61 @@
...
+	}
+</style>
+<textarea id="editor1" cols="10" rows="10">
+	<p>Apollo 11</p>
+	<p>Apollo 11</p>
+	<p>Apollo 11</p>
+	<img src="http://lorempixel.com/400/200/">
```
f1ames: Please use local images e.g. `tests/_assets/` as the remote ones may not be always available (or load slow).

plugins/inlinetoolbar/skins/kama/inlinetoolbar.css
```diff
@@ -0,0 +1,61 @@
...
+}
+
+/* side: [ bottom ] */
+.cke_balloon.cke_inlinetoolbar .cke_balloon_triangle_outer.cke_balloon_triangle_bottom
+{
+	border-width: 10px 10px 0;
+	bottom: -10px;
+}
+
+.cke_balloon.cke_inlinetoolbar .cke_balloon_triangle_outer > .cke_balloon_triangle_inner.cke_balloon_triangle_bottom
```
mlewand: Again (I asked for this before) please relax the CSS selectors. I'm sure this one will be good with `.cke_balloon.cke_inlinetoolbar .cke_balloon_triangle_outer .cke_balloon_triangle_inner.cke_balloon_triangle_bottom`, possibly even with `.cke_balloon.cke_inlinetoolbar .cke_balloon_triangle_inner.cke_balloon_triangle_bottom` (just assuming, don't have the branch checked out for testing).

Hell, looking at original balloonpanel selectors, it could be even possible to simplify it down to `.cke_inlinetoolbar .cke_balloon_triangle_inner.cke_balloon_triangle_bottom`.

mlewand: Talking about specificity, see https://www.smashingmagazine.com/2007/07/css-specificity-things-you-should-know/#specificity-basic-principles in a spare time.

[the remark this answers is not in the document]
wojtekw92: I have not noticed it while I was removing this selectors.

plugins/inlinetoolbar/plugin.js
```diff
@@ -25,10 +30,19 @@
...
+
 	CKEDITOR.plugins.add( 'inlinetoolbar', {
 		requires: 'balloonpanel',
 		init: function() {
+			if ( !stylesLoaded ) {
+				CKEDITOR.document.appendStyleSheet( this.path + 'skins/' + CKEDITOR.skinName + '/inlinetoolbar.css' );
+				stylesLoaded = true;
+			}
 			CKEDITOR.ui.inlineToolbarView.prototype = CKEDITOR.tools.extend( {}, CKEDITOR.ui.balloonPanel.prototype );
+			if ( CKEDITOR.ui.inlineToolbarView.prototype.templateDefinitions && CKEDITOR.ui.inlineToolbarView.prototype.templateDefinitions.panel.indexOf( 'cke_inlinetoolbar' ) === -1 ) {
```
mlewand: This is not an elegant way to add this class.

You could just add `this.parts.panel.addClass( 'cke_inlinetoolbar' );` in `build` method and that will do it.

I have pushed unit test checking if the class is added by the plugin.

tests/plugins/inlinetoolbar/manual/kama.html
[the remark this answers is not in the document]
mlewand: That was not a proper way to handle it, now we can do it the proper way so we need to put a proper reference, so that when looking at the code in couple of months we know exactly what's the reason for this.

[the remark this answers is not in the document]
mlewand: In other words, please report the issue in `ckeditor/ckeditor-dev` and reference it next to these workarounds in tests.

plugins/inlinetoolbar/skins/moono/inlinetoolbar.css
```diff
@@ -0,0 +1,68 @@
...
+}
+
+.cke_balloon.cke_inlinetoolbar .cke_balloon_triangle_inner.cke_balloon_triangle_top,
+.cke_balloon.cke_inlinetoolbar .cke_balloon_triangle_inner.cke_balloon_triangle_bottom
+{
+	border-color: #cfd1cf transparent;
+}
+
+.cke_balloon.cke_inlinetoolbar .cke_balloon_triangle_inner.cke_balloon_triangle_bottom,
+.cke_balloon.cke_inlinetoolbar .cke_balloon_triangle_inner.cke_balloon_triangle_top
```
f1ames: Why isn't it merged with previous item as it affects identical elements (has the same selectors)?

plugins/inlinetoolbar/skins/moono/inlinetoolbar.css
```diff
@@ -0,0 +1,68 @@
+/*
+Copyright (c) 2003-2017, CKSource - Frederico Knabben. All rights reserved.
+For licensing, see LICENSE.md or http://ckeditor.com/license
+*/
+
+.cke_balloon.cke_inlinetoolbar
+{
+	background: linear-gradient(to bottom, #f5f5f5, #cfd1cf);
```
f1ames: I know ATM we are not targeting older browsers probably, but it is good to keep this in mind. So the rule should be same as for standard toolbar (including `filter` for IEs):

image

plugins/inlinetoolbar/skins/moono-lisa/inlinetoolbar.css
```diff
@@ -0,0 +1,65 @@
...
+	border-color: #d1d1d1 transparent;
+}
+
+.cke_balloon.cke_inlinetoolbar .cke_balloon_triangle_inner.cke_balloon_triangle_top,
+.cke_balloon.cke_inlinetoolbar .cke_balloon_triangle_inner.cke_balloon_triangle_bottom
+{
+	border-color: #f8f8f8 transparent;
+}
+
+.cke_balloon.cke_inlinetoolbar .cke_balloon_triangle_inner.cke_balloon_triangle_bottom,
```
f1ames: Can be merged with previous item.

plugins/inlinetoolbar/plugin.js
f1ames: I think it is because`CKEDITOR.ui.inlineToolbarView.prototype.templateDefinitions.panel` is still a string there. So that's why I suggested adding comment there at it is a little misleading. But maybe there is a better way to handle this, WDYT @mlewand ?

mlewand: I mean the solution that I have proposed above is much more readable and elegant.

String replacement looks bad and makes a very imprecise replacement that could as well replace any other part of HTML, say someone will add another class `cke_balloonpanel`  `cke_balloon_custom` or an attribute that will include it.

wojtekw92: I will go along with @mlewand solution with adding class in build method.

wojtekw92:
screen shot 2017-10-16 at 13 17 20
screen shot 2017-10-16 at 13 17 53
I am not able to reproduce issue with moved toolbar in Firefox and Safari @f1ames can you give some reproduce steps?

wojtekw92:
The `IE` click problem is not on our side. After second click IE is sending `click` event

f1ames:
@wojtekw92 I see the issue with top/bottom spacing happens only during bender test run:

oct-16-2017 16-22-11

So it is caused by some specific bender styling and there is no need to work on it now.

f1ames:
> The IE click problem is not on our side. After second click IE is sending click event

I can see this is a native issue (https://codepen.io/f1ames/pen/QqJWzE?editors=1010). Haven't found any similar issues on Edge/IE bugtracker, however as they no longer provide fixes for IE11 (except security ones) there is no use in reporting it IMHO.

f1ames:
@mlewand Please take a look too if your earlier requests were meet.

mlewand:
R+, let's wait with merging till #957 is done.

PR #978, opened 2017-09-28
New widget for Easy Image

Comandeer:

New feature

### This PR contains

- [ ] Unit tests
- [ ] Manual tests

_coming soon_

## What changes did you make?

I've introduced new plugin, `imagebase`, which creates new image widget. This plugin could be used as a base for both `image2` and `easyimage` plugins.

Currently the implementation is very basic and it does not do much beyond just displaying image.

### DIFF
--- a/plugins/easyimage/plugin.js
+++ b/plugins/easyimage/plugin.js
@@ -21,7 +21,7 @@
 			return function( editor ) {
 				var widget = editor.widgets.focused;
 
-				if ( widget && widget.name === 'image' ) {
+				if ( widget && widget.name === 'easyimage' ) {
 					this.setState( ( enableCheck && enableCheck( widget ) ) ? CKEDITOR.TRISTATE_ON : CKEDITOR.TRISTATE_OFF );
 				} else {
 					this.setState( CKEDITOR.TRISTATE_DISABLED );
@@ -94,32 +94,16 @@
 	}
 
 	function registerWidget( editor ) {
-		// Overwrite some image2 defaults.
-		editor.on( 'widgetDefinition', function( evt ) {
-			var oldInit = evt.data.init;
-
-			if ( evt.data.name !== 'image' ) {
-				return;
-			}
-
-			evt.data.allowedContent.figure.classes += ',!' + editor.config.easyimage_class + ',' +
-				editor.config.easyimage_sideClass;
-
-			// Block default image dialog.
-			evt.data.dialog = null;
-
-			evt.data.init = function() {
-				oldInit.call( this );
-
+		CKEDITOR.plugins.imagebase.createWidget( editor, 'easyimage', {
+			basicClass: editor.config.easyimage_class,
+			additionalClasses: [ editor.config.easyimage_sideClass ],
+			init: function() {
 				this.on( 'contextMenu', function( evt ) {
-					// Remove "Image Properties" option.
-					delete evt.data.image;
-
 					evt.data.easyimageFull = editor.getCommand( 'easyimageFull' ).state;
 					evt.data.easyimageSide = editor.getCommand( 'easyimageSide' ).state;
 					evt.data.easyimageAlt = editor.getCommand( 'easyimageAlt' ).state;
 				} );
-			};
+			}
 		} );
 	}
 
@@ -135,7 +119,7 @@
 	}
 
 	CKEDITOR.plugins.add( 'easyimage', {
-		requires: 'image2,contextmenu,dialog',
+		requires: 'imagebase,contextmenu,dialog',
 		lang: 'en',
 
 		onLoad: function() {
--- a/plugins/easyimage/plugin.js
+++ b/plugins/easyimage/plugin.js
@@ -96,7 +96,7 @@
 	function registerWidget( editor ) {
 		var config = editor.config;
 
-		CKEDITOR.plugins.imagebase.createWidget( editor, 'easyimage', {
+		CKEDITOR.plugins.imagebase.addImageWidget( editor, 'easyimage', {
 			allowedContent: {
 				figure: {
 					classes: '!' + config.easyimage_class + ',' + config.easyimage_sideClass
--- a/plugins/easyimage/plugin.js
+++ b/plugins/easyimage/plugin.js
@@ -94,10 +94,28 @@
 	}
 
 	function registerWidget( editor ) {
+		var config = editor.config;
+
 		CKEDITOR.plugins.imagebase.createWidget( editor, 'easyimage', {
-			basicClass: editor.config.easyimage_class,
-			additionalClasses: [ editor.config.easyimage_sideClass ],
-			captionAllowedContent: 'br em strong sub sup u s; a[!href,target]',
+			allowedContent: {
+				figure: {
+					classes: '!' + config.easyimage_class + ',' + config.easyimage_sideClass
+				}
+			},
+
+			requiredContent: 'figure(!' + config.easyimage_class + ')',
+
+			upcast: function( element ) {
+				// http://dev.ckeditor.com/ticket/11110 Don't initialize on pasted fake objects.
+				if ( element.attributes[ 'data-cke-realelement' ] ) {
+					return;
+				}
+
+				if ( element.name === 'figure' && element.hasClass( config.easyimage_class ) ) {
+					return element;
+				}
+			},
+
 			init: function() {
 				this.on( 'contextMenu', function( evt ) {
 					evt.data.easyimageFull = editor.getCommand( 'easyimageFull' ).state;
--- a/plugins/easyimage/plugin.js
+++ b/plugins/easyimage/plugin.js
@@ -97,6 +97,7 @@
 		CKEDITOR.plugins.imagebase.createWidget( editor, 'easyimage', {
 			basicClass: editor.config.easyimage_class,
 			additionalClasses: [ editor.config.easyimage_sideClass ],
+			captionAllowedContent: 'br em strong sub sup u s; a[!href,target]',
 			init: function() {
 				this.on( 'contextMenu', function( evt ) {
 					evt.data.easyimageFull = editor.getCommand( 'easyimageFull' ).state;
--- a/plugins/easyimage/plugin.js
+++ b/plugins/easyimage/plugin.js
@@ -105,14 +105,16 @@
 
 			requiredContent: 'figure(!' + config.easyimage_class + ')',
 
-			upcast: function( element ) {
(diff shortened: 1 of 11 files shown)


plugins/imagebase/lang/en.js
```diff
@@ -0,0 +1,21 @@
+/*
+Copyright (c) 2003-2017, CKSource - Frederico Knabben. All rights reserved.
+For licensing, see LICENSE.md or http://ckeditor.com/license
+*/
+CKEDITOR.plugins.setLang( 'imagebase', 'en', {
```
mlewand: There are multiple unused lang entries.

plugins/imagebase/plugin.js
```diff
@@ -0,0 +1,83 @@
...
+				},
+				figcaption: true
+			},
+			requiredContent: 'figure(!' + options.basicClass + ')',
+
+			editables: {
+				caption: {
+					selector: 'figcaption',
+					pathName: editor.lang.imagebase.pathNameCaption,
+					allowedContent: 'br em strong sub sup u s; a[!href,target]'
```
mlewand: I guess that should be configurable with options.

plugins/imagebase/plugin.js
```diff
@@ -0,0 +1,149 @@
...
+	 *
+	 * @since 4.8.0
+	 * @singleton
+	 * @class CKEDITOR.plugins.imagebase
+	 */
+	CKEDITOR.plugins.imagebase = {
+		/**
+		 * Registers a new widget definition based on passed options.
+		 *
+		 * @since 4.8.0
```
mlewand: Since the parent class already h as a `since` tag I think it's redundant here, as it's the same version.

plugins/imagebase/plugin.js
```diff
@@ -0,0 +1,149 @@
...
+		 * It adds several parts of image and implements the {@link CKEDITOR.plugins.widget.definition#upcast} callback.
+		 * This callback should not be overwritten.
+		 *
+		 * @abstract
+		 * @class CKEDITOR.plugins.imagebase.imageWidgetDefinition
+		 * @mixins CKEDITOR.plugins.widget.definition
+		 */
+		return {
+			/**
+			 * @inheritdoc
```
mlewand: There's no need to explicitly set `@inheritdoc` if you're not modifying docs of inherited object, it will be implicitly inherited.

plugins/imagebase/plugin.js
```diff
@@ -0,0 +1,149 @@
...
+		 * @param {Object} options Widget's options.
+		 * @param {String} options.basicClass Primary class for widget's element, that allows to properly
+		 * identify the widget.
+		 * @param {String[]} options.additionalClasses Additional classes that could be applied to widget's
+		 * element.
+		 * @param {CKEDITOR.filter.allowedContentRules} options.captionAllowedContent Allowed content inside
+		 * image's caption.
+		 * @param {Function} options.init Widget's initialization method.
+		 */
+		createWidget: function( editor, name, options ) {
```
mlewand: Let's stay consistent with `addFooWidget` [that we started in upload plugin](https://github.com/ckeditor/ckeditor-dev/blob/8f1d80f/plugins/uploadwidget/plugin.js#L535). So it should be `addImageWidget`.

mlewand: Current API is very, very rigid and little extensive. Instead I think it should allow you to customize any of the about-to-be-created widget definition.

In that case I can see `createWidgetDefinition` returning a plain/"ideal" definition for image base, and then within `addImageWidget` it would call `CKEDITOR.object.extend` if custom overriding rule was provided.

mlewand: `options` argument should simply allow to customize widget definition. So say it would be a  `definition` argument, an instance of `CKEDITOR.plugins.imagebase.imageWidgetDefinition`.

plugins/imagebase/plugin.js
```diff
@@ -0,0 +1,113 @@
...
+		var defaultTemplate = new CKEDITOR.template(
+			'<figure class="{basicClass}">' +
+				'<img alt="" src="" />' +
+				'<figcaption>{captionPlaceholder}</figcaption>' +
+			'</figure>' ),
+			baseDefinition;
+
+		/**
+		 * This is an abstract class that describes a definition of a basic image widget
+		 * creted by {@link CKEDITOR.plugins.imagebase#addImageWidget} method.
```
mlewand: Typo -> created.

plugins/imagebase/plugin.js
```diff
@@ -0,0 +1,113 @@
...
+						return element;
+					}
+				}
+			}
+		};
+
+		definition = CKEDITOR.tools.object.merge( baseDefinition, definition );
+
+		/**
+		 * Image widget definition overwrites the {@link CKEDITOR.plugins.widget.definition#upcast} property.
```
mlewand: This should be a bit better documented, just briefly what it does? We want to avoid docs like "This is a property that should that should not interest you! Do not touch!" 😉 

How about something simple, like:

`Overrides {@link CKEDITOR.plugins.widget.definition#upcast} property, it's automatically set to enumerate keys of {@link #upcasts}. Avoid changes, unless you know what you're doing.`

?

plugins/imagebase/plugin.js
```diff
@@ -0,0 +1,113 @@
...
+
+			parts: {
+				image: 'img',
+				caption: 'figcaption'
+			},
+
+			upcasts: {
+				figure: function( element ) {
+					// http://dev.ckeditor.com/ticket/11110 Don't initialize on pasted fake objects.
+					if ( element.attributes[ 'data-cke-realelement' ] ) {
```
mlewand: Same thing as in `easyimage.widgetDefinition.figure.upcasts`.

plugins/imagebase/plugin.js
```diff
@@ -0,0 +1,113 @@
...
+						return;
+					}
+
+					if ( element.name === 'figure' ) {
```
mlewand: Well, having #1094 fixed this check is no longer needed, right? So it should be simply a `return element;` function.

plugins/imagebase/plugin.js
```diff
@@ -0,0 +1,113 @@
...
+
+			allowedContent: {
+				img: {
+					attributes: '!src,alt,width,height'
+				},
+				figure: true,
+				figcaption: true
+			},
+
+			requiredContent: 'figure',
```
mlewand: `img[src]` should be required too, as there's pretty much no point in putting figure alone.

plugins/imagebase/plugin.js
```diff
@@ -0,0 +1,113 @@
...
+ * @license Copyright (c) 2003-2017, CKSource - Frederico Knabben. All rights reserved.
+ * For licensing, see LICENSE.md or http://ckeditor.com/license
+ */
+
+( function() {
+	'use strict';
+
+	function createWidgetDefinition( editor, definition ) {
+		var defaultTemplate = new CKEDITOR.template(
+			'<figure class="{basicClass}">' +
```
mlewand: Is `basicClass` defined anywhere or is it just a leftover?

plugins/easyimage/plugin.js
```diff
@@ -94,32 +94,37 @@
...
+		CKEDITOR.plugins.imagebase.addImageWidget( editor, 'easyimage', {
+			allowedContent: {
+				figure: {
+					classes: '!' + config.easyimage_class + ',' + config.easyimage_sideClass
+				}
+			},
 
-			evt.data.allowedContent.figure.classes += ',!' + editor.config.easyimage_class + ',' +
-				editor.config.easyimage_sideClass;
+			requiredContent: 'figure(!' + config.easyimage_class + ')',
```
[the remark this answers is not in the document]
mlewand: Similarly like imagebase, this widget should also require `img[src]` otherwise it has little sense.

This should be covered with small unit test, e.g. in `config.js` suite, just create extra editor with `config.disallowContent='img'`, and ensure that widgets are not upcasted.

tests/plugins/imagebase/imagebase.js
```diff
@@ -0,0 +1,59 @@
...
+
+			CKEDITOR.plugins.imagebase.addImageWidget( editor, 'testWidget', {
+				parts: {
+					loader: 'progress'
+				}
+			} );
+
+			objectAssert.ownsKeys( [ 'testWidget' ], editor.widgets.registered );
+			objectAssert.areDeepEqual( expectedParts, editor.widgets.registered.testWidget.parts );
+
```
mlewand: Redundant empty line, let's just remove it.

tests/plugins/imagebase/imagebase.js
```diff
@@ -0,0 +1,59 @@
...
+		var editor = bot.editor;
+
+		bot.setData( data, function() {
+			var widgets = editor.editable().find( '[data-widget="' + widget + '"]' );
+
+			assert.areSame( 1, widgets.count(), 'Widget is properly upcasted' );
+		} );
+	}
+
+
```
mlewand: Let's remove this small space over here, it's a detail but somehow it hurts. :D

tests/plugins/imagebase/imagebase.js
```diff
@@ -0,0 +1,59 @@
+/* bender-tags: editor,widget */
+/* bender-ckeditor-plugins: imagebase,toolbar */
+
+( function() {
+	'use strict';
+
+	function assertUpcast( bot, data, widget ) {
```
mlewand: You can get rid of it and reuse widget just like I did in 08ef0dcf6a88a99e79b55ef46e361f73c192b844.

plugins/imagebase/plugin.js
Comandeer: It's a leftover from the previous version, it could be safely removed.

tests/plugins/imagebase/imagebase.js
Comandeer: I don't see any similar method in widget tools for that – for now I'd stick to my custom `assertUpcast`.

mlewand: Look at the commit I have referneced:

```js
'test easyimage_class - changed': function() {
	widgetTestsTools.assertWidget( {
		count: 1,
		widgetOffset: 0,
		nameCreated: 'easyimage',
		html: CKEDITOR.document.getById( 'changedClass' ).getHtml(),
		bot: this.editorBots.classCustomized
	} );
},
```

It does check widget count and a widget type, abstracting away how to get widget count etc.

PR #983, opened 2017-09-28
[WIP] Cloud services transpilation

Comandeer:

New feature

## What changes did you make?

I've added transpilation and bundling for `@ckeditor/ckeditor-cloudservices-core` using Grunt + Webpack + Babel, however there are some issues left:

* [ ] Add polyfills for ES6 features (mainly `Promise`)
* [ ] Find a way to make this work in IE 8 (I've already add some ES3 transforms, however `babel-polyfill` is not compatible with IE8, the same goes with `Object.defineProperty` used by Babel to mark modules)
* [ ] Add license comment to the beginning of the bundle (use official `BannerPlugin`?)


dev/tasks/webpack.js
```diff
@@ -0,0 +1,57 @@
...
+							{
+								test: /\.js$/,
+								loader: 'babel-loader',
+								query: {
+									cacheDirectory: true,
+									presets: [
+										[ require( 'babel-preset-env' ), {
+											modules: false,
+											targets: {
+												browsers: [ 'last 2 versions', 'ie >= 8' ]
```
mlewand: We could even further relax this query. We're targeting at Chrome, Firefox, Edge, Safari ATM.

dev/tasks/utils/cs-build-entry.js
```diff
@@ -0,0 +1,16 @@
+// jshint ignore: start
+
+import FileUploader from '@ckeditor/ckeditor-cloudservices-core/src/uploadgateway/fileuploader';
+import UploadGateway from '@ckeditor/ckeditor-cloudservices-core/src/uploadgateway/uploadgateway';
+
+window.CKEDITOR = ( window.CKEDITOR || {} );
+
+window.CKEDITOR.cloudservices = {
```
mlewand: It should be a camelCased property name. I like the idea about putting it statically to `CKEDITOR` namespace. Later, if more plugins uses CS, we could create a `cloudservices` plugin that sets up this namespace.

dev/tasks/utils/cs-build-entry.js
```diff
@@ -0,0 +1,16 @@
+// jshint ignore: start
```
mlewand: This is more of an asset rather than utility. In a sense that it's not bringing value for the grunt task itself, but is used as src for bundling.

I feel like that file should in the end go to the `easyimage` directory. How about `plugins/easyimage/src/buildentry.js`? This directory should be removed by the builder (though `plugins/easyimage/lib/cs.js` should not ofc).

Any thoughts on it @Comandeer @f1ames?

f1ames: It is part of a Webpack task, but indeed `easyimage` specific. Moving to `plugins/easyimage` makes sense.

Couldn't we go a step further and make `webpack` task generic so if the plugin has a build script e.g. `plugins/easyimage/src/build-cs.js` it runs it and outputs the resulted file to some build dir `plugins/easyimage/lib/cs.js`? (well, I'm not sure how hard it is to do with Grunt...)

mlewand: > Couldn't we go a step further and make webpack task generic

I don't think we need it at this point. As soon as we face a case that we have another solution using webpack task, we'll generalize it.

Comandeer:
I've fixed all issues mentioned above, except for API docs, which were moved to dedicated issue.

I've also added license block using Webpack's official `BannerPlugin`.

mlewand:
I just added two minor changes in commits above:

* Linked directly to CS license (it's CS's license that should be used, not CKE4 license).
* Let's look to support 2 recent releases of our target browsers.

Comandeer:
LGTM

mlewand:
I'll leave this issue with R+ until we decide whether we go with CS or `fileTools` for uploading.

PR #1023, opened 2017-10-09
Fix unit test failing in Edge 16

### LINKED ISSUE #945: The 'tests/plugin/indent' breaks a bender test run in Edge 16

Bug (probably)

It is not possible to execute a full test run in Edge 16 (test in Edge 16 preview VM) because `tests/plugin/indent` test breaks the test run:

sep-20-2017 15-48-30

### Expected result

The passing `tests/plugin/indent` unit test should not break the test run.

### Actual result

The passing `tests/plugin/indent` unit test breaks the test run.

As it was tested on preview version there is some possibility that it will get fixed in a stable release, still however it is worth checking on stable when it is available.
Also it might not be CKEditor issue, but a bender one (seems unlikely).

## Other details

* Browser: Edge 16 (`16.16257`)
* OS: Win10 (VM)
* CKEditor version: 4.7.3

f1ames:

Other: fixing tests failing on newest Edge (version 16). Closes #945.

### This PR contains

- [ ] Unit tests
- [ ] Manual tests

Tests are not needed as this PR is fixing tests only.

## What changes did you make?

As stated in #945, the browser behaviour seems quite random. The troublesome test when executed alone or as a first test works fine.  I also tried to check which preceding test may affect it, commenting out one after another. I found a few which when enabled causes the behaviour, however when those tests and only the troublesome one were enabled, again it worked fine (so tests in between where needed to trigger the issue). I also checked if timeout or clearing editor contents in-between test helps, but it doesn't.

It seems there is some issue with many tests operating on lists running in the same editor instance. My proposition for a workaround is to:

* `tests/plugins/indent/indent.js` - run each test in newly created editor instance (as it significantly increases test execution time it will only work this way in Edge, in other browsers it will reuse already created instances).
* `tests/plugins/list/list.js` - instead of one there are two editor instances created. All tests before troublesome one uses first instance and the rest second one.

The above fixes works and all tests now passes in Edge 16, I think it is sufficient for now. However, if we encounter similar behaviour in the future it will definitely require more detailed look.

### DIFF
--- a/tests/plugins/indent/indent.js
+++ b/tests/plugins/indent/indent.js
@@ -3,11 +3,38 @@
 ( function() {
 	'use strict';
 
-	var enterModeAbbr = {};
+	var editorCounter = 0,
+		enterModeAbbr = {},
+		editorsCache = {},
+		editors;
 
 	enterModeAbbr[ CKEDITOR.ENTER_P ] = 'P';
 	enterModeAbbr[ CKEDITOR.ENTER_BR ] = 'BR';
 	enterModeAbbr[ CKEDITOR.ENTER_DIV ] = 'DIV';
+
+	function createEditor( editorName, callback ) {
+		var editorConfig = editors[ editorName ];
+
+		editorCounter++;
+		editorConfig.name = editorConfig.name + editorCounter;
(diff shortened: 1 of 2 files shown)


mlewand:
@Comandeer please have a look at it in a spare time.

PR #1025, opened 2017-10-10
T/1008 Highlight colors in color panel

### LINKED ISSUE #1008: Black and White colors don't get highlighted in Color Button panels

Bug

1. Clear editor contents and type dew words. 
2. Select All and set background color to 'Strong Red'.
3. Select a word and set its text color to 'White'.
4. Click inside the word or select part or all of it.
5. Open the 'Font Color' panel button.

### Expected result

White color is highlighted.

### Actual result

Automatic color (which is 'Black') is highlighted.

## Other details

* Browser: Any
* OS: Any
* CKEditor version: 4.7.0+

**The same issue can be reproduced with Background Color button and Black color.**

beatadelura:

Bug fix

Yes

### This PR contains

- [X] Unit tests
- [X] Manual tests

## What changes did you make?

Colors which wasn't highlighted was written in shorthand hex notation. What I did, I changed it from 3-characters format to 6-characters.

---

Closes #1008

### DIFF
--- a/tests/plugins/colorbutton/manual/highlightedcolor.md
+++ b/tests/plugins/colorbutton/manual/highlightedcolor.md
@@ -1,6 +1,6 @@
 @bender-tags: 4.8.0, bug, 1008
 @bender-ui: collapsed
-@bender-ckeditor-plugins: wysiwygarea, toolbar, colorbutton, colordialog
+@bender-ckeditor-plugins: wysiwygarea, toolbar, colorbutton, colordialog, format
 
 ** Scenario 1: **
 
--- a/tests/plugins/colorbutton/manual/highlightedcolor.md
+++ b/tests/plugins/colorbutton/manual/highlightedcolor.md
@@ -2,7 +2,7 @@
 @bender-ui: collapsed
 @bender-ckeditor-plugins: wysiwygarea, toolbar, colorbutton, colordialog
 
-** Scenario: **
+** Scenario 1: **
 
 1. Select each word or put caret in each word and click `Text Colour`/`Background Colour` button.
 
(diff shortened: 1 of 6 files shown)


beatadelura:
**Just a question:**

With regard to [CKEDITOR.tools.normalizeHex](https://docs.ckeditor.com/ckeditor4/docs/#!/api/CKEDITOR.tools-method-normalizeHex), wouldn't be easier to use `charAt()` method ([fiddle](https://jsfiddle.net/wojwuLsh/))
instead of splitting string into array:
```
var parts = normalizedHexColor.split( '' );
normalizedHexColor = [ parts[ 0 ], parts[ 0 ], parts[ 1 ], parts[ 1 ], parts[ 2 ], parts[ 2 ] ].join( '' );
```

f1ames:
@beatadelura I'm not sure about `CKEDITOR.tools.normalizeHex`, but I assume that it was the matter of preferences here. Both approaches works. If you would like, you can try to benchmark it (using e.g. jsperf) to check which version is faster. But I think it will be super micro optimization here 😄

tests/plugins/colorbutton/manual/highlightedcolor.md
```diff
@@ -0,0 +1,15 @@
+@bender-tags: 4.8.0, bug, 1008
+@bender-ui: collapsed
+@bender-ckeditor-plugins: wysiwygarea, toolbar, colorbutton, colordialog, format
```
f1ames: I think `colordialog` is not needed here. You may also remove `format` plugin (I assume it is to have `h1` in the editor so the text is bigger and more readable - but you can also achieve this using [addCss](https://docs.ckeditor.com/ckeditor4/docs/#!/api/CKEDITOR-method-addCss) and just change the `font-size`).

PR #1055, opened 2017-10-16
#1053 Introduce CKEDITOR.tools.object.merge

### LINKED ISSUE #1053: Introduce CKEDITOR.tools.object.merge function

New feature

We need a function to recursively merge objects.

This function should take 2 params `obj1`, `obj2` and return the object that is a result of merging the two. **Keep in mind that neither `obj1` nor `obj2` should be modified by this function** (add a TC for that).

If the same primitive value is available in `obj1` and `obj2`, the value from `obj2` should  be used.

This function should land in [`CKEDITOR.tools.object`](https://docs.ckeditor.com/ckeditor4/docs/#!/api/CKEDITOR.tools.object) namespace, as a `merge` function.

If possible function may do a feature detection and use `Object.assign`.

 ## Examples

```
{
	one: 1,
	conflicted: 10,
	obj: {
		nested1: 1
	}
}
```

```
{
	two: 2,
	conflicted: 20,
	obj: {
		nested2: 2
	}
}
```

Should result with the following object:

```
{
	one: 1,
	two: 2,
	conflicted: 20,
	obj: {
		nested1: 1,
		nested2: 2
	}
}
```

Comandeer:

New feature

### This PR contains

- [x] Unit tests
- [ ] Manual tests

## What changes did you make?

Introduce `CKEDITOR.tools.object.merge`, which merges two objects.

Closes #1053.

### DIFF
--- a/core/tools.js
+++ b/core/tools.js
@@ -2025,7 +2025,7 @@
 			 *			}
 			 *		}
 			 *
-			 * @param {Object} obj1 A base object.
+			 * @param {Object} obj1 Source object, which will be used to create a new base object.
 			 * @param {Object} obj2 An object, which properties will be merged to the base one.
 			 * @returns {Object} Merged object.
 			 * @member CKEDITOR.tools.object
--- a/core/tools.js
+++ b/core/tools.js
@@ -2001,14 +2001,14 @@
 			 * @returns {Object} Merged object.
 			 * @member CKEDITOR.tools.object
 			 */
-			merge: function merge( obj1, obj2 ) {
-				var copy1 = CKEDITOR.tools.clone( obj1 ),
-					copy2 = CKEDITOR.tools.clone( obj2 ),
-					tools = CKEDITOR.tools;
+			merge: function( obj1, obj2 ) {
+				var tools = CKEDITOR.tools,
+					copy1 = tools.clone( obj1 ),
+					copy2 = tools.clone( obj2 );
 
 				tools.array.forEach( tools.objectKeys( copy2 ), function( key ) {
 					if ( typeof copy2[ key ] === 'object' && typeof copy1[ key ] === 'object' ) {
-						copy1[ key ] = merge( copy1[ key ], copy2[ key ] );
+						copy1[ key ] = tools.object.merge( copy1[ key ], copy2[ key ] );
 					} else {
 						copy1[ key ] = copy2[ key ];
 					}
--- a/core/tools.js
+++ b/core/tools.js
@@ -1996,6 +1996,35 @@
 			/**
 			 * Merges two objects and returns new one.
 			 *
+			 *		var obj1 = {
+			 *				a: 1,
+			 *				conflicted: 10,
+			 *				obj: {
+			 *					c: 1
+			 *				}
+			 *			},
+			 *			obj2 = {
+			 *				b: 2,
+			 *				conflicted: 20,
+			 *				obj: {
+			 *					d: 2
+			 *				}
+			 *			};
+			 *
+			 *		CKEDITOR.tools.object.merge( obj1, obj2 );
+			 *
+			 * This code produces the following object;
+			 *
+			 *		{
+			 *			a: 1,
+			 *			b: 2,
+			 *			conflicted: 20,
+			 *			obj: {
+			 *				c: 1,
+			 *				d: 2
+			 *			}
+			 *		}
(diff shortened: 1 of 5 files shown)


core/tools.js
```diff
@@ -1991,6 +1991,30 @@
...
+			 * @member CKEDITOR.tools.object
+			 */
+			merge: function merge( obj1, obj2 ) {
+				var copy1 = CKEDITOR.tools.clone( obj1 ),
+					copy2 = CKEDITOR.tools.clone( obj2 ),
+					tools = CKEDITOR.tools;
+
+				tools.array.forEach( tools.objectKeys( copy2 ), function( key ) {
+					if ( typeof copy2[ key ] === 'object' && typeof copy1[ key ] === 'object' ) {
+						copy1[ key ] = merge( copy1[ key ], copy2[ key ] );
```
mlewand: Can't we use simply tools.merge? `merge` is defined in a hacky way, very uncommon to our code base.

core/tools.js
```diff
@@ -1991,6 +1991,30 @@
...
+			 * Merges two objects and returns new one.
+			 *
+			 * @param {Object} obj1 A base object.
+			 * @param {Object} obj2 An object, which properties will be merged to the base one.
+			 * @returns {Object} Merged object.
+			 * @member CKEDITOR.tools.object
+			 */
+			merge: function merge( obj1, obj2 ) {
+				var copy1 = CKEDITOR.tools.clone( obj1 ),
+					copy2 = CKEDITOR.tools.clone( obj2 ),
```
mlewand: both members could reuse `tools` that was defined just a line below.

core/tools.js
```diff
@@ -1991,6 +1991,30 @@
 				}
 
 				return null;
+			},
+
+			/**
+			 * Merges two objects and returns new one.
```
mlewand: An example would on how it actually merges would be a brilliant idea.

tests/_benderjs/ckeditor/static/extensions.js
```diff
@@ -135,6 +135,40 @@
...
+		if ( expectedKeys.length != actualKeys.length ) {
+			YUITest.Assert.fail( YUITest.Assert._formatMessage( message,
+				'Object should have ' + expectedKeys.length + ' keys but has ' + actualKeys.length ) );
+		}
+
+		// Then check values.
+		for ( var name in expected ) {
+			if ( expected.hasOwnProperty( name ) ) {
+				if ( expected[ name ] && typeof expected[ name ] === 'object' ) {
+					areDeepEqual( expected[ name ], actual[ name ] );
```
f1ames: Same as https://github.com/ckeditor/ckeditor-dev/pull/1055/files#r144820711, couldn't it be just `bender.objectAssert.areDeepEqual( ... );`

Comandeer:
> do you think the merge function should do some type checking (if passed objects are really an objects, not array, null, etc) to prevent misuse?

There are no such checkings in other `CKEDITOR.tools` method. Method's name IMO is clear enough to allow us to skip them.

core/tools.js
```diff
@@ -1991,6 +1991,59 @@
...
+			 *			b: 2,
+			 *			conflicted: 20,
+			 *			obj: {
+			 *				c: 1,
+			 *				d: 2
+			 *			}
+			 *		}
+			 *
+			 * @param {Object} obj1 A base object.
+			 * @param {Object} obj2 An object, which properties will be merged to the base one.
```
f1ames: > An object, which properties will be merged to the base one.

I think this is a little misleading, as properties are not merged to the base object, but to its copy, so the base object is not modified. And it is also inconsistent with method description which suggest the new object is created (_Merges two objects and returns new one_).

ATM I don't have any clear idea how it could be described. I checked in other libs/docs how it is described there:

* https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/assign
* https://lodash.com/docs/#merge
* http://api.jquery.com/jQuery.merge/
* http://ramdajs.com/docs/#merge

but still not sure. Maybe you could come up with something reasonable?

Comandeer: Maybe something like that?

```
obj1 Source object, which will be used to create a new base object.
```

And then `obj2` description is already good :D WDYT @f1ames?

f1ames: @Comandeer works for me 👍

PR #1052, opened 2017-10-16
Upload widget for Easy Image

Comandeer:
## What is the purpose of this pull request?

New feature

## Does your PR contain necessary tests?

All patches which change the editor code must include tests. You can always read more
on [PR testing](http://docs.ckeditor.com/#!/guide/dev_contributing_code-section-tests),
[how to set the testing environment](http://docs.ckeditor.com/#!/guide/dev_tests) and
[how to create tests](http://docs.ckeditor.com/#!/guide/dev_tests-section-creating-your-own-test)
in the official CKEditor documentation.

### This PR contains

- [x] Unit tests
- [ ] Manual tests

## What changes did you make?

I've introduced upload widget for easy image, based on existing `uploadimage` widget.

Currently information about easy image is passed as additional request parameters to `fileUploadRequest` and `fileLoader`'s property to `fileUploadResponse`. Additionally there is some code duplication between `uploadimage` and current widget (especially whole handling of pasting images, which is just copied).

There is also WIP `[srcset]` support, waiting for dedicated image plugin to be fully functional.

PR #1062, opened 2017-10-17
Cloud Services file loader

mlewand:
New feature

- [x] Unit tests
- [ ] Manual tests - this is strictly an API PR, thus I skipped manual test

Reuses logic proposed in #1052, but extracts [Code Services](https://ckeditor.com/ckeditor-cloud-services) handing into a separate type.

So far it's only a loader type, integration with the [UploadRepository](https://docs.ckeditor.com/ckeditor4/docs/#!/api/CKEDITOR.fileTools.uploadRepository) or UploadWidget will be added in a followup PR.

### DIFF
--- a/plugins/cloudservices/plugin.js
+++ b/plugins/cloudservices/plugin.js
@@ -1,3 +1,8 @@
+/**
+ * @license Copyright (c) 2003-2017, CKSource - Frederico Knabben. All rights reserved.
+ * For licensing, see LICENSE.md or http://ckeditor.com/license
+ */
+
 ( function() {
 	'use strict';
 
--- a/plugins/cloudservices/plugin.js
+++ b/plugins/cloudservices/plugin.js
@@ -22,11 +22,11 @@
 			 * @class CKEDITOR.plugins.cloudservices.cloudServicesLoader
 			 * @extends CKEDITOR.fileTools.fileLoader
 			 */
-			function cloudServicesLoader( editor, fileOrData, fileName ) {
+			function CloudServicesLoader( editor, fileOrData, fileName ) {
 				FileLoader.call( this, editor, fileOrData, fileName );
 			}
 
-			cloudServicesLoader.prototype = CKEDITOR.tools.extend( {}, FileLoader.prototype );
+			CloudServicesLoader.prototype = CKEDITOR.tools.extend( {}, FileLoader.prototype );
 
 			/**
 			 * @inheritdoc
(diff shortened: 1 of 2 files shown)


mlewand:
I see your point. To allow token customization I think the best way would be to allow custom token as a CloudServicesLoader constructor parameter.

To be honest I just made `url` customization possible in order not to have a dummy parameter that says "it's here but don't use it", while the implementation and tests were very simple. I think that we should add token customization as we have a real use case for it down the road, WDYT?

Also I have pushed one small commit changing the casing of local type var in the plugin itself.

mlewand:
Actually... hold that thought. On the other hand it's so simple change that it make sense to think about extensibility at this point.

plugins/cloudservices/plugin.js
```diff
@@ -0,0 +1,127 @@
...
+			 * @param {String} [token] A token used for [Cloud Service](https://ckeditor.com/ckeditor-cloud-services/) request. If
+			 * skipped {@link CKEDITOR.config#cloudServices_token} will be used.
+			 */
+			function CloudServicesLoader( editor, fileOrData, fileName, token ) {
+				FileLoader.call( this, editor, fileOrData, fileName );
+
+				/**
+				 * Custom [Cloud Service](https://ckeditor.com/ckeditor-cloud-services/) token.
+				 *
+				 * @property {String} custom_token
```
Comandeer: Typo in property name.

plugins/cloudservices/plugin.js
```diff
@@ -0,0 +1,127 @@
...
+				 * @member CKEDITOR.plugins.cloudservices.cloudServicesLoader
+				 */
+				this.customToken = token;
```
Comandeer: Maybe name it just `this.token` and move here assigning custom token or default one? Then setting appropriate header in `fileUploadRequest` could be simplify to just setting it to `fileLoader.token`.

mlewand: There's a nice thing about the current implementation that it will work correctly even though config will be changed at runtime.

Officially we do not support changing config at a runtime, but it's nice to keep it whenever possible.

So the way I saw it is that `customToken` would only be set if it should really be overriden, otherwise it should be used from a single source of truth, which is config ATM.

Do you think we should change this?

Comandeer: Keeping ability to change config at runtime sounds reasonable to me. Let's keep it

PR #1065, opened 2017-10-18
Remove console error with not allowed resources for images

msamsel:
Bug fix

### This PR contains

- [ ] Unit tests <-- I have no idea how to check it with unit tests :(
- [x] Manual tests

## What changes did you make?
- fix situation which generates `Not allowed resources` error for inserted images.
- it **does not** prevent of having the same error when shapes are pasted to the editor. 

related to #662


mlewand:
We decided that uploading will be done with the [Upload Image](https://ckeditor.com/cke4/addon/uploadimage) plugin. Thus we no longer need this ingeration next to Paste from Word code responsible for inlining images. Thus closing this PR as expired.

PR #1074, opened 2017-10-23
Fix Arabic typos in common lang file

atefBB:
`need review`


mlewand:
Thank you very much for contribution! We're using [Transifex](https://www.transifex.com/ckeditor/ckeditor/) for managing language strings. Please send your contribution there, so that we can pull it into upcoming release, so that it can be seen for other lang contributors.

In case of further questions just let us know!

Thanks!

atefBB:
thx!

Issue #1068, opened 2017-10-19
Upload widget paste listener ignores changes in the uploadWidgetDefinition


mlewand:
## Are you reporting a feature or a bug?

Bug

 ## Provide detailed reproduction steps (if any)

1. Open https://codepen.io/mlewand/pen/mBvxeM.
1. Open developer console.
1. Drag any jpg file from your desktop, and drop it over the editor's editable.

 ### Expected result

Black square should appear, no exceptions of missing method should be logged in the console.

 ### Actual result

Exception `Uncaught TypeError: d[m] is not a function` is reported.

 ## Other details

Reason why it happens is because `addUploadWidget` function in `widgetupload` plugin uses `def` that was given as an argument for the very first time. This object can not be modified by the `widgetDefinition` listeners.

The solution here is to simply get the definition from `widgetRepository` in `paste` listener.

Reproducible in all browsers.

PR #1079, opened 2017-10-23
Easy image upload refactoring

mlewand:

Other, refactoring of other PR.

### This PR contains

- [x] Unit tests
- [x] Manual tests

## What changes did you make?

This PR adds some fixes to #1052, the general idea was to simplify it and extract cloud-service (backend) stuff to a separate plugin.

There were already quite a few changes, many of which has been already fixed in #1062 and #1066.

Summary for this PR:

* [`uploadUrl`](https://github.com/ckeditor/ckeditor-dev/blob/de017b395220ebaba8136ef7650d54dd035b5411/plugins/easyimage/plugin.js#L137) is now optional, as the plugin uses CloudService loader which automatically picks up URL from the config.
* [`loadMethod`](https://github.com/ckeditor/ckeditor-dev/blob/de017b395220ebaba8136ef7650d54dd035b5411/plugins/easyimage/plugin.js?utf8=%E2%9C%93#L139) is now changed to upload. Since you proposed a more modern approach with object URL - I think it's fine, however we need a followup ticket to keep track of data allocated with it. See TP#3131.
* Removed `fileUploadResponse`, `fileUploadRequest` as they're no longer needed (fixed in #1062, #1066).
* [paste listener](https://github.com/ckeditor/ckeditor-dev/blob/4c7e90d472cc1490df1963c6eb158914e8fefbe4/plugins/easyimage/plugin.js#L206) for `uploadeasyimage` widget uses widget definition at a time of paste event to avoid #1068. Though still I'm not happy that it's a copy pasta of a listener from `imageupload` plugin, we need to combine it. See TP#3132.
* Config options `easyimageUploadUrl`, `easyimage_token` have been renamed to `cloudServices_url`, `cloudServices_token` respectively.
* Corrected [`responseData` structure](https://github.com/ckeditor/ckeditor-dev/blob/4c7e90d472cc1490df1963c6eb158914e8fefbe4/tests/plugins/easyimage/uploadwidget.js?utf8=%E2%9C%93#L61-L66) in unit tests.

There's one issue that latest FF is extremely unstable, that's something that we need to investigate in a followup ticket.

Once this is merged, I'll just give a final check to #1052 and that will be it in terms of uploading.

### DIFF
--- a/tests/plugins/easyimage/uploadwidget.js
+++ b/tests/plugins/easyimage/uploadwidget.js
@@ -4,9 +4,9 @@
 /* bender-include: %BASE_PATH%/plugins/uploadfile/_helpers/waitForImage.js */
 /* global pasteFiles, waitForImage */
 
-'use strict';
-
 ( function() {
+	'use strict';
+
 	var uploadCount, loadAndUploadCount, resumeAfter, tests,
 		IMG_URL = '%BASE_PATH%_assets/logo.png',
 		DATA_IMG = 'data:',
@@ -50,6 +50,8 @@
 
 	CKEDITOR.on( 'instanceCreated', function() {
 		if ( !CKEDITOR.plugins.cloudservices.cloudServicesLoader.restore ) {
+			// Because of #1068 we can't provide loaderType directly. Instead we just replace cloudServicesLoader
+			// with our loader stub that is going to be customized in init method.
 			sinon.stub( CKEDITOR.plugins.cloudservices, 'cloudServicesLoader', CKEDITOR.fileTools.fileLoader );
 		}
 	} );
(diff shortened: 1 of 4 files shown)


tests/plugins/easyimage/uploadwidget.js
```diff
@@ -438,6 +430,30 @@
 			} );
 
 			wait();
+		},
+
+		'test cloudservices URL can be customized': function( editor ) {
```
Comandeer: Not only URL can be customised, but also loader type and additional request parameters. It'd be nice to have also tests for these.

mlewand:
For reasons unknown to me firefox fires `8` [img load events](https://github.com/ckeditor/ckeditor-dev/blob/77dd5c9609672843acbb07b427a9266e56ec9847/tests/plugins/uploadfile/_helpers/waitForImage.js#L8) in vs `6` in other browsers.

Without getting into details I have changed listener type to `once` instead.

This test suite relies on `img#load` event, with `img[src="data:image/png;base64,..."]` value, so it might hit some quirky edge cases.

Also I have added requested test cases.

Issue #1061, opened 2017-10-17
Failing test /tests/core/ckeditor/basepathglobal on Safari


msamsel:
## Are you reporting a feature request or a bug?

Bug

## Provide detailed reproduction steps (if any)

1. run test `/tests/core/ckeditor/basepathglobal`

### Expected result

Test pass

### Actual result
Below 2 test cases fails:
- test full BASEPATH
- test protocol relative BASEPATH

## Other details

* Browser: Safari Version 11.0 (12604.1.38.1.7), Safari Version 11.0 (13604.1.38.1.6)
* OS: MacOS
* CKEditor version: major (but also stable releases has those problem)
* Installed CKEditor plugins: …

beatadelura:
This test fails if there is no `sub.ckeditor.dev` hostname specified in `/etc/hosts`. So it's a local configuration issue rather than a test.

I'm going to create a new task with regard to this. It would be nice to make this test more flexible and work for custom aliases. See comment [here](https://github.com/ckeditor/ckeditor-dev/pull/1082#pullrequestreview-71501634).

f1ames:
As I took another look on those tests, it looks there is an issue on Safari, more details here https://github.com/ckeditor/ckeditor-dev/pull/1090#discussion_r148548901.

PR #1082, opened 2017-10-24
t/1061 Fixing basepathglobal unit test on Safari

beatadelura:

Fixing failing test `/tests/core/ckeditor/basepathglobal` on Safari.

## What changes did you make?

The reason for failing test was setting `sub.ckeditor.dev` as a current domain. So if, the test was running from other domain was failed. Now, I'm checking current domain.

---

Closes: #1061


beatadelura:
@f1ames You are right. I'll close this issue and create the new one for test improvement.

PR #1085, opened 2017-10-24
Button support in inline toolbar (continuation)

wojtekw92:
This reopen PR https://github.com/ckeditor/ckeditor-dev/pull/992


tests/plugins/inlinetoolbar/buttons.js
```diff
@@ -0,0 +1,34 @@
...
+	bender.test( {
+		'test adding buttion': function() {
+			var panel = new CKEDITOR.ui.inlineToolbar( this.editor );
+			panel.addUIElements( {
+				cut: new CKEDITOR.ui.button( {
+					label: 'test',
+					command: 'cut'
+				} )
+			} );
+			assert.isNotUndefined( panel.getUIElement( 'cut' ), 'The button should exist.' );
```
mlewand: You want to make more precise checks.

```js
assert.isInstanceOf( CKEDITOR.ui.button, panel.getUIElement( 'cut' ), 'Registered button type' );
```

Is a much better choice.

mlewand: Same goes for next TC.

plugins/inlinetoolbar/plugin.js
```diff
@@ -80,25 +101,89 @@
...
 			};
+
+			/**
+			 * Adds an item to the inline toolbar.
+			 *
+			 * @method
+			 * @param {String} name The menu item name.
+			 * @param {Object} element instance of CKEDITOR.ui element
+			 */
+			CKEDITOR.ui.inlineToolbar.prototype.addUIElement = function( name, element ) {
```
mlewand: Could we simplify name to `inlineToolbar.addItem`? I see that you wanted to keep compatibility with [how it's named in dialog](https://docs.ckeditor.com/ckeditor4/docs/#!/api/CKEDITOR.dialog-static-method-addUIElement), however I feel like `addItem` would be a more clear approach.

Same goes for other methods like `addItems`, `getItems`, `deleteItems`.

tests/plugins/inlinetoolbar/manual/kama.html
```diff
@@ -40,11 +40,11 @@
 
 	function instanceReadyListener( evt ) {
 		var image = this.editable().findOne( 'img' ),
-			panel = new CKEDITOR.ui.inlineToolbarView( this );
+			panel = new CKEDITOR.ui.inlineToolbar( this );
 
 		image.on( 'click', function() {
 			panel.create( this );
-			panel.parts.content.setHtml( '<p style="padding: 10px">Clicked element: ' + this.getName() + '</p>' );
+			panel._toolbar.parts.content.setHtml( '<p style="padding: 10px">Clicked element: ' + this.getName() + '</p>' );
```
mlewand: New samples should no longer use old API, it should register real buttons instead.

It would be nice to add for instance 2 buttons, and a rich combo (e.g. bold, font color and font size just for an example).

mlewand: Ofc this change should be propagated to other skin manual tests.

tests/plugins/inlinetoolbar/positioning.js
```diff
@@ -56,7 +56,7 @@
 				assert.ignore();
```
mlewand: Doesn't it actually make sense to work directly on `inlineToolbarView` rather than create a full `inlineToolbar`?

plugins/inlinetoolbar/plugin.js
```diff
@@ -32,13 +32,34 @@
...
+	 * docs for an example definition.
+	 */
+	CKEDITOR.ui.inlineToolbar = function( editor, definition ) {
+		this._toolbar = new CKEDITOR.ui.inlineToolbarView( editor, definition );
+		/**
+		 * Menu items added to inline toolbar
+		 *
+		 * @private
+		 */
+		this._menuItems = [];
```
mlewand: I think we might change this name simply to `_items` as it's not a menu, it's a toolbar.

plugins/inlinetoolbar/plugin.js
```diff
@@ -99,6 +135,62 @@
...
+			CKEDITOR.ui.inlineToolbar.prototype.create = function( element ) {
+				this._view.renderItems( this._items );
+				this._view.create( element );
+			};
+
+			/**
+			 * Adds an item to the inline toolbar.
+			 *
+			 * @param {String} name The menu item name.
+			 * @param {Object} element instance of {CKEDITOR.ui.button/CKEDITOR.ui.richCombo}
```
mlewand: This is not how you mark param type. See other code docs, JSDuck wiki for reference.

plugins/inlinetoolbar/plugin.js
```diff
@@ -99,6 +135,62 @@
 					that.attach( element, false );
 				} ) );
 			};
+
+			/**
+			 * Render items and attach view to DOM element.
```
mlewand: This doc is not enough. `create` method is extremely important (key to create), thus docs need to be good.

`InlineToolbarView.create` method has much better description, but I'd avoid raw copy'n'pasting.

plugins/inlinetoolbar/plugin.js
```diff
@@ -80,6 +103,19 @@
 			};
 
 			/**
+			 * Create html representation of menu items in toolbar.
+			 *
+			 * @param items array of {CKEDITOR.ui.button/CKEDITOR.ui.richCombo} objects.
```
mlewand: This is not a correct `@param` doc tag.

See https://github.com/senchalabs/jsduck/wiki/%40param how to document array of given type correctly.

plugins/inlinetoolbar/plugin.js
```diff
@@ -80,6 +103,19 @@
 			};
 
 			/**
+			 * Create html representation of menu items in toolbar.
```
mlewand: HTML is an abbreviation, as such it should be uppercased, it's HTML, CSS, YAML etc.

plugins/inlinetoolbar/plugin.js
```diff
@@ -32,13 +33,35 @@
...
+	 * @param {Object} definition An object containing the panel definition. See {@link CKEDITOR.ui.balloonPanel}
+	 * docs for an example definition.
+	 */
+	CKEDITOR.ui.inlineToolbar = function( editor, definition ) {
+		this._view = new CKEDITOR.ui.inlineToolbarView( editor, definition );
+
+		/**
+		 * Menu items added to inline toolbar
+		 *
+		 * @private
```
mlewand: Missing type information.

PR #1090, opened 2017-10-25
T/1089 Rename hostname in test

### LINKED ISSUE #1061: Failing test /tests/core/ckeditor/basepathglobal on Safari

### LINKED ISSUE #1089: Test /tests/core/ckeditor/basepathglobal should work for custom aliases

Task

If there is no `sub.ckeditor.dev` specified in the local environment `/etc/hosts` test `/tests/core/ckeditor/basepathglobal` fails on Safari.

It could be more flexible and work for custom aliases instead of using fixed `sub.ckeditor.dev`. See [comment](https://github.com/ckeditor/ckeditor-dev/pull/1082#pullrequestreview-71501634).

beatadelura:

Test `/tests/core/ckeditor/basepathglobal` improvement.

---

Closes #1061.

### DIFF
--- a/tests/core/ckeditor/basepathglobal.js
+++ b/tests/core/ckeditor/basepathglobal.js
@@ -10,7 +10,7 @@
 		folderPath = path + testDir.slice( 0, testDir.lastIndexOf( '/' ) ),
 		expectedEditorPath = '/apps/ckeditor/',
 		query = CKEDITOR.timestamp ? '?t=' + CKEDITOR.timestamp : '',
-		secondDomainName = 'sub.ckeditor.custom',
+		secondDomainName = 'sub.ckeditor.example',
 		port = window.location.port;
 
 
--- a/tests/core/ckeditor/basepathglobal.js
+++ b/tests/core/ckeditor/basepathglobal.js
@@ -10,7 +10,7 @@
 		folderPath = path + testDir.slice( 0, testDir.lastIndexOf( '/' ) ),
 		expectedEditorPath = '/apps/ckeditor/',
 		query = CKEDITOR.timestamp ? '?t=' + CKEDITOR.timestamp : '',
-		secondDomainName = 'sub.ckeditor.dev',
+		secondDomainName = 'sub.' + window.location.hostname,
 		port = window.location.port;
 
 
--- a/tests/core/ckeditor/basepathglobal.js
+++ b/tests/core/ckeditor/basepathglobal.js
@@ -10,7 +10,7 @@
 		folderPath = path + testDir.slice( 0, testDir.lastIndexOf( '/' ) ),
 		expectedEditorPath = '/apps/ckeditor/',
 		query = CKEDITOR.timestamp ? '?t=' + CKEDITOR.timestamp : '',
-		secondDomainName = 'sub.' + window.location.hostname,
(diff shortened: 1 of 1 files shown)


tests/core/ckeditor/basepathglobal.js
```diff
@@ -10,7 +10,7 @@
 		folderPath = path + testDir.slice( 0, testDir.lastIndexOf( '/' ) ),
 		expectedEditorPath = '/apps/ckeditor/',
 		query = CKEDITOR.timestamp ? '?t=' + CKEDITOR.timestamp : '',
-		secondDomainName = 'sub.ckeditor.dev',
+		secondDomainName = 'sub.' + window.location.hostname,
```
Comandeer: This change results in `sub.tests.ckeditor.test` in standard setup and therefore – 404 error in console. What's strange, tests are still passing despite this fact.

f1ames: Because the tests only checks if paths are properly created (on `iframe load event`) and not checking if something was loaded properly (and `/apps/ckeditor.js` which contains `getUrl` method is always loaded from main domain):

image

versus (as `loader.js` failed to load)

image

It passes on Chrome no matter the subdomain used. And fails on Safari... no matter the subdomain used.

I think I need take another look on #1061.

f1ames: Small correction on Safari it fails only with subdomain ending with `.dev` is used.

f1ames: More confusing than I thought on Safari...

Fails for:
* [sub.]ckeditor.dev
* [sub.]ckeditor.test
* [sub.]ckeditor.local
* [sub.]ckeditor.com

Passes for:
* [sub.]ckeditor.custom
* [sub.]ckeditor.pl
* [sub.]ckeditor.bcn

So from what I assume, Safari treats `.dev`, `.test`, `.local` alwasy as non local, but globally registered domains? (because for such domains, like `ckeditor.com`, this test fails also in Chrome).

So tests passes on Chrome/Safari for non-existing domains or locally defined domain which contains proper CKEditor files (unless local domain in Safari ends with `.dev`, `.test`, `.local`).

@Comandeer does it make any sense? (quite confusing for me IMHO)

I pushed fix with the custom domain `sub.ckeditor.custom`.

f1ames:
Reopened #1061 as it was valid, so now this PR `closes #1061`.

tests/core/ckeditor/basepathglobal.js
Comandeer: No, it doesn't make any sense 🤔

Comandeer: Or maybe it does… https://security.stackexchange.com/a/14805

f1ames:
`.example` makes sense and works, I updated the test.

Issue #710, opened 2017-08-01
Widget issues in tables


Comandeer:
## Are you reporting a feature request or a bug?

Bug

## Provide detailed reproduction steps (if any)

All cases are based on https://jsfiddle.net/a8wrevzz/

### Case 1
1. Click image's move handler
2. Try to move image.

#### Expected result

Image is moved.

#### Actual result

Table selection is activated, but image is not moved.

### Case 2

1. Try to resize image using resize handler

#### Expected result

Image is resized.

#### Actual result

* Table selection is activated (especially in case of enlarging image).
* In rare cases image could be not resized at all.

### Case 3

1. Click image.

#### Expected result

Widget is focused.

#### Actual result

Widget becomes focused, but immediately loses focus.

## Other details

* Installed CKEditor plugins: tableselection, image2

mlewand:
Actually this one is for the discussion:

I think that dragging a widget should cancel the `mousedown` event, since it basically takes over the whole process by displaying magic line etc. Also for consistency `mousemove` and `mouseup` should be also prevented (if following the `mousedown` and handled by widget ofc).

Table selection should have a lower priority than a widget selection, and would not be triggered if already handled by widget.

On a side note, table selection should do the same thing as for event cancellation.

Comandeer:
It seems that lowering mouse listeners priorities in `tableselection` plugin and adding `evt.stop` to listeners in `widget` plugin partially fixes the issue. However there is still issue with dragging widget inside the table. After dragging and dropping it the first time, every attempt to write inside table just focuses the widget, making table practically uneditable.

AlfonsoML:
For reference, these problems didn't exist previously (like 4.6.something), it's a regression that hurts the people that wants to upgrade.

msamsel:
Case 2 - Also table selection activate when widget is dragged.

Mgsy:
## Steps to reproduce
1. Go to http://localhost:1030/tests/plugins/clipboard/manual/draganddrop.
2. Select the widget.
3. <kbd>Ctrl</kbd> + <kbd>C</kbd>.
4. Put the caret inside the table cell.
5. <kbd>Ctrl</kbd> + <kbd>V</kbd>.
6. Move to the next cell.
7. <kbd>Ctrl</kbd> + <kbd>V</kbd>.

## Expected result
The image pastes correctly.

## Current result
The image doesn't paste into the cell.

## Other information
OS: Windows 10, MacOS X
Browser: All browsers
[...]

PR #1104, opened 2017-10-30
Widget has now correct selection behaviour inside tables

### LINKED ISSUE #522: Widget selection is not removed when widget is inside of a table cell
Bug

Widgets have a selection mode. When they are clicked on, they get selected. When the insertion point or selected range is changed, they become unselected. This worked in 4.6.x. 
When a widget is inside of a table cell (td), the selection mode fails to unselect when the creates an insertion point selection (range = collapsed). It fails to select when the cell ONLY contains the widget.

To reproduce:
1. Go to http://sdk.ckeditor.com/samples/captionedimage.html
2. Insert a 2 column, 1 row table as shown in the supplied image. Include text in each cell.
3. Select the image widget and use the copy command.
4. Move the insertion point into the first cell.
5. Paste.
The widget is established correctly.
6. Select the widget. It becomes selected.
7. Click the insertion point into the text of the second cell.

widget inside table cell 1

Expected:
Widget becomes unselected.

Actual:
Widget stays selected, while the insertion point is active where you clicked.

8. Type.
Expected:
Typing to appear at the insertion point.
Actual:
Typing is ignored.

## Other details
If you make a selected range in the other cell, unselection happens correctly.
If you omit text in the first cell, after pasting, you can click on the widget but it refuses to establish the selection.

I looked into the source code and can point out the issue.
selection.js/checkSelectionChange. This function now checks for fake selection support on tables through a call to isRealTableSelection. It does not test for fake selection used on widgets.
This is a proposed change:
`if ( !realSel || ( !realSel.isHidden() && ((this.widgets.selected.length > 0) || !isRealTableSelection( realSel, sel )) ) ) {`

However, I don't think its the right implementation because the fake selection engine should support user supplied selections. (I have created one for my installation.) Instead, I believe each consumer of fake selections should register a hook that is executed when realSel.isHidden() returns false. This will benefit users who are not using the tableselection plugin too.

* Browser: Chrome Version 58.0.3029.110 (64-bit)
* OS: Windows 7 Professional SP 1
* CKEditor version: 4.7.0
* Installed CKEditor plugins: While I have the tableselection plugin, the config has it in removePlugins.

### LINKED ISSUE #1027: Bug: cannot add multiple images to table

Bug

## Provide detailed reproduction steps

1. Insert a table
2. Insert an image into a cell of the table
3. Focus on another cell of the same table
4. Press the image button in the toolbar to enter an image into this cell too

### Expected result

Both cells should have an image

### Actual result

Only one image is added to the table

## Screen recording

image in table

### Other details

Browser: Chrome 61, Safari
CKEditor version: 4.7.3

msamsel:

Bug fix

There is no unit test. Selection use quite unfriendly for testing purposes listener mechanism, which use timeouts to execute.
https://github.com/ckeditor/ckeditor-dev/blob/02c35d9ae44656bd75d0efd69028888195dd2a6f/core/selection.js#L272-L299
I wasn't able to properly repeat the problem. I tried to call just `selectionChange` event, also fire `mouseup`, `mousedown` events with target of different table cell. All those things remain selection on widget even tough problem was fixed already.

### This PR contains

- [ ] Unit tests
- [x] Manual tests

## What changes did you make?
- provide method which return information if element is either widget or widget wrapper.
- use this method to omit some rules when table selection is checked. As far as I know only `tableselection` and widget use `fake-selection`, so this solution should be sufficient for this case

close #1027, close #522


plugins/tableselection/plugin.js
```diff
@@ -269,15 +270,19 @@
 		if ( !evt.data.getTarget().getName ) {
 			return;
 		}
+		// Prevent of applying table selection when widget is selected.
+		// Mouseup remain possibility to finish table selection when user release mouse button above widget in table.
+		if ( evt.name !== 'mouseup' && CKEDITOR.plugins.widget && CKEDITOR.plugins.widget.isDomWidget( evt.data.getTarget() ) ) {
```
f1ames: This line introduced one unexpected (I think), behaviour:

Before:
nov-03-2017 14-11-53

After:
nov-03-2017 14-11-35

so when dragging mouse over widgets, cells are not selected with fake selection as they would normally be.

tests/plugins/widget/integration/table/manual/widgetintable.html
```diff
@@ -0,0 +1,108 @@
...
+				requiredContent: 'strong(inline-widget)',
+				allowedContent: 'strong(!inline-widget){*}',
+				upcast: function( element ) {
+					return element.name === 'strong' && element.hasClass( 'inline-widget' );
+				}
+			} );
+		}
+	} );
+
+	CKEDITOR.plugins.add( 'blockWidget', {
```
f1ames: Maybe you could extract those widget definitions to some helper as they are duplicated in both manual tests?

tests/plugins/widget/integration/table/manual/widgetintable.md
```diff
@@ -0,0 +1,17 @@
+@bender-tags: 1027, 4.8.0, widget
+@bender-ui: collapsed
+@bender-ckeditor-plugins: undo,clipboard,basicstyles,toolbar,wysiwygarea,widget,table,resize
+
+----
+
+1. Click into widgets in table
```
f1ames: > Click on the widget in table

tests/plugins/widget/integration/table/manual/widgetintable.md
```diff
@@ -0,0 +1,17 @@
...
+1. Try to select some text **in table**
+1. Perform some operation on selected text ( change, append new text, apply some style, etc. )
+1. Check behaviour inside inner and outer table depend of testing widget
```
f1ames: This step is a little unclear. I assume it means to test for both text in nested and outer table. Maybe it can be skipped and added to the instructions below test scenario?

tests/plugins/widget/integration/table/manual/widgetintable.md
```diff
@@ -0,0 +1,17 @@
...
+1. Try to add new widget in table ( buttons without icon in toolbr )
+
+_Perform above steps for all widgets and both editors._
+
+**Expected:** Widget is deselect when focus is changed to text in table. You are able to perfom operatons on selected text.
```
f1ames: > Widget is deselect when focus is changed to text in table **and text selection is visible**. You are...

tests/plugins/widget/integration/table/manual/widgetintable.md
```diff
@@ -0,0 +1,17 @@
...
+@bender-ui: collapsed
+@bender-ckeditor-plugins: undo,clipboard,basicstyles,toolbar,wysiwygarea,widget,table,resize
+
+----
+
+1. Click into widgets in table
+1. Try to select some text **in table**
+1. Perform some operation on selected text ( change, append new text, apply some style, etc. )
+1. Check behaviour inside inner and outer table depend of testing widget
+1. Try to add new widget in table ( buttons without icon in toolbr )
```
f1ames: This looks a little out of place here. We can treat it as a last step of the scenario, but then it misses **expected** result. You can:

* Extract it as a second test scenario (well, pretty short one).
* Rearrange scenario in a way that the user/tester should add a widget on the beginning and then proceed with the rest of the steps (so added widget will also be a part of further testing).

f1ames:
I see the case with widget drag&drop is reported in #710 (**Case 1**), the question is if the cause is the same it can also be solved?

msamsel:
First case with missing selection. Problem it's not related to this case and might be reproduced without widget. More details here: #1122

Dragging problems (Activation table selection, and problem with dropping to proper cell) are covered by #710 (case 1 and 2)

Problem with not having `tableselection` when going with cursor above the widgets is reported here: #1125

plugins/tableselection/plugin.js
msamsel: Extracted to #1125

tests/plugins/widget/integration/table/manual/widgetintablewithtableselection.md
```diff
@@ -0,0 +1,17 @@
+@bender-tags: 1027, 4.8.0, widget
```
f1ames: This test could be ignored for IE8-10 as `tableselection` is not supported there so it will be exactly the same as `.../manual/widgetintable` test.

tests/tickets/1027/1.md
```diff
@@ -0,0 +1,15 @@
+@bender-tags: 1027, 4.8.0, widget
```
f1ames: This test could be ignored for IE8-10 as tableselection is not supported there.

mlewand:
Gents, just one thing - I'd like to have @Comandeer a second look at it before it gets merged just for extra safety. So @Comandeer please give it a look once it gets a final R+ (I'll mark you already).

tests/plugins/widget/integration/table/manual/widgetintablewithtableselection.md
```diff
@@ -1,17 +1,17 @@
 @bender-tags: 1027, 4.8.0, widget
 @bender-ui: collapsed
-@bender-ckeditor-plugins: undo,clipboard,basicstyles,toolbar,wysiwygarea,widget,table,resize,tableselection
+@bender-include: ../_helpers/testwidgets.js
+@bender-ckeditor-plugins: undo,clipboard,basicstyles,toolbar,wysiwygarea,widget,table,resize
```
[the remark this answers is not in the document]
msamsel: I assume you have `tableselection` on your mind ;)
I've copy paste the description earlier and haven't notice that is missing now  :(

f1ames:
You may now take a look @Comandeer.

core/selection.js
```diff
@@ -96,6 +99,11 @@
 			return table.equals( fakeTable ) || fakeTable.contains( table );
 		}
 
+		// When widget is selected, then definately is not a table (#1027).
```
Comandeer: Typo: "definitely".

plugins/tableselection/plugin.js
```diff
@@ -208,7 +208,8 @@
 			// The selection is inside one cell, so we should allow native selection,
 			// but only in case if no other cell between mousedown and mouseup
 			// was selected.
-			if ( !fakeSelection.dirty && cells.length === 1 ) {
+			// We don't want to clear selection if widget is event target (#1027).
+			if ( !fakeSelection.dirty && cells.length === 1 && !( CKEDITOR.plugins.widget && CKEDITOR.plugins.widget.isDomWidget( evt.data.getTarget() ) ) ) {
```
Comandeer: Maybe we could extract the second part to some function as it's repeated several times in the code?

tests/tickets/1027/1.html
```diff
@@ -0,0 +1,34 @@
+<label>Lena's path: <input style="border:2px solid black" type="text" value="../../_assets/lena.jpg" readonly></label>
```
Comandeer: I'l move this test to `plugins/tableselection/integrations` directory

tests/plugins/widget/integration/table/manual/widgetintable.html
```diff
@@ -0,0 +1,77 @@
...
+<script>
+	CKEDITOR.replace( 'classic', {
+		extraPlugins: 'inlineWidget,blockWidget',
+		removePlugins: 'tableselection',
+	} );
+
+	CKEDITOR.replace( 'divarea', {
+		removePlugins: 'tableselection',
+		extraPlugins: 'divarea,inlineWidget,blockWidget'
+	} )
```
Comandeer: Missing semicolon.

tests/plugins/widget/integration/table/manual/widgetintablewithtableselection.html
```diff
@@ -0,0 +1,82 @@
...
+				if ( !CKEDITOR.plugins.tableselection.isSupportedEnvironment ) {
+					bender.ignore();
+				}
+			}
+		}
+	} );
+
+	CKEDITOR.replace( 'divarea', {
+		extraPlugins: 'divarea,inlineWidget,blockWidget'
+	} )
```
Comandeer: Missing semicolon.

plugins/tableselection/plugin.js
Comandeer: It's repeated both here and in `core/selection.js`.

core/selection.js
```diff
@@ -226,6 +234,7 @@
 			realSel = this.getSelection( 1 );
 			// If real (not locked/stored) selection was moved from hidden container
 			// or is not a table one, then the fake-selection must be invalidated.
+
```
Comandeer: Seems like unnecessary blank line.

core/selection.js
```diff
@@ -11,6 +11,10 @@
 		fillingCharSequenceRegExp = new RegExp( fillingCharSequence + '( )?', 'g' ),
 		isSelectingTable;
 
+	function _isWidget( element ) {
```
Comandeer: Actually there isn't a need to prefix such function with `_` as it's not a method. `isWidget` would be enough.

plugins/tableselection/plugin.js
```diff
@@ -16,6 +16,10 @@
 		insertRow,
 		insertColumn;
 
+	function _isWidget( element ) {
```
Comandeer: Actually there isn't a need to prefix such function with `_` as it's not a method. `isWidget` would be enough.

msamsel:
I've extracted IE8 issue to separate task #1226. It existed before my change.

Yanual:
I just downloaded the tableselection plugin on version 4.8 and got an error: "plugin.js: 20 Uncaught TypeError: CKEDITOR.plugins.widget.isDomWidget is not a function", I look at the functions available in the plugin widget and i try the isDomWidgetElement function and the error does not happen again. I would like to know if my modification is correct and why I have had this problem (am I the only one?) or if there is another sollution.

mlewand:
@Yanual please report this case as a separate issue, according to our issue template (with step-by-step reproduction instruction).

Issue #1047, opened 2017-10-13
Test 'tests/plugins/tableselection/integrations/clipboard/pasteflow' crashes when browser viewport focused


f1ames:
## Are you reporting a feature request or a bug?

Bug

## [Check if the issue is already reported](http://docs.ckeditor.com/#!/guide/dev_issues_readme-section-avoid-duplicates)

Extracted from #962.
Similar to #945.

## Provide detailed reproduction steps (if any)

The test `tests/plugins/tableselection/integrations/clipboard/pasteflow` started to fail (or rather break bender run in the same manner as mentioned in #945), after changes introduced in #962. However it only fails when browser viewport is focused (focusing browser window or browser URL field makes test run smoothly).

Also when changing tests order, from:

```
'test paste flow (tabular content)': function( editor, bot ) {
	testPasteFlow( bot, 'tabular-paste', '2cells1row' );
},

'test paste flow (non-tabular content)': function( editor, bot ) {
	testPasteFlow( bot, 'nontabular-paste', 'paragraph' );
}
```

to

```
'test paste flow (non-tabular content)': function( editor, bot ) {
	testPasteFlow( bot, 'nontabular-paste', 'paragraph' );
},

'test paste flow (tabular content)': function( editor, bot ) {
	testPasteFlow( bot, 'tabular-paste', '2cells1row' );
}	
```

it works fine despite the browser focus.

## Other details

* Browser: **Edge 16+**
* OS: Windows 10
* CKEditor version: 4.7.3 with #962 
* Installed CKEditor plugins: `clipboard`, `tableselection`, `undo`

mlewand:
Interestingly I can no longer repro it with:

```
Microsoft Edge 41.16299.15.0
Microsoft EdgeHTML 16.16299
````

f1ames:
Just wanted to mention that simplest workaround (changing test order) is already present on #1153 (https://github.com/ckeditor/ckeditor-dev/pull/1153/commits/73a07fc0b762b53d1ada469938f662cf7f8f9d9c) and this issue was to analyze the problem in more depth.

I will check if for the former test, the issue is still present on `Edge 16.16299`. If so I suggest to use the already present workaround as @msamsel fix is quite similar. If test will be passing on new Edge version, I will revert those change from #1153.

f1ames:
[...]

Issue #962, opened 2017-09-26
Workaround for Edge not supporting custom MIME types in `dataTransfer.setData`


f1ames:
## Are you reporting a feature request or a bug?

Task

## Provide detailed reproduction steps (if any)

Extracted from #468.
Upstream issue https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/8007622.

The Edge browser still does not support custom MIME types in `dataTransfer.setData`. This are used in `clipboard` plugin for storing `dataTransfer` id (which is used to much dataTransfer objects between events).

The proposal is to apply a workaround which will save custom MIME types in `text/html` MIME type as an HTML comment like:

```
<!--cke-data:DATA-->
```

Where `DATA` is [stringifed](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify) and [encoded](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURIComponent) JSON object.

This will happen dynamically as Edge throws an error/exception when custom MIME types is set. The exception will be caught and fallback used.

f1ames:
There is one specific issue I stumbled upon when working on the fix:

1. Go to http://tests.ckeditor.dev/tests/plugins/clipboard/manual/buttons.
2. Perform copy and paste (as in scenario).

![sep-28-2017 11-55-42](https://user-images.githubusercontent.com/1061942/30971287-c74cf790-a467-11e7-8d79-73f0a26524f4.gif)

`OpenClipboard Failed` error is thrown (in this line - https://github.com/ckeditor/ckeditor-dev/blob/t/468-962/plugins/clipboard/plugin.js#L2095). Looks like some Edge issue with synchronous access to clipboard on paste. Haven't found any reports on https://developer.microsoft.com/en-us/microsoft-edge/platform/issues, but clearly looks like native issue.

---

I wasn't able to reproduce this issue with plain contenteditable, however `OpenClipboard Failed` error seems to be caused by two or more applications accessing the clipboard at the same time (not sure why it happens here still)... Simply changing the code to access `clipboardData` only once solves the issue - see https://github.com/ckeditor/ckeditor-dev/pull/968/commits/2347c18a4136ec9e9761579d5f0a626b9541d254.

f1ames:
[...]

PR #1105, opened 2017-10-31
Fix for failing tests on Edge 16+

msamsel:
Bug fix

fix tests

### This PR contains

- [ ] Unit tests
- [ ] Manual tests

## What changes did you make?
- split test to run in separate editors

Notes:
I've tried to find out where is the problem with this test, unfortunately without any clear result.
From some unknown reason `resume` method wasn't execute properly.
https://github.com/ckeditor/ckeditor-dev/blob/96bf4037cb3ff812fc9b993bce4a304a884f361a/tests/plugins/tableselection/integrations/clipboard/pasteflow.js#L34-L44
I've tried to delay it by some timeout, but without proper reaction.
I also noticed that, when Edge starts to fail, then selection in editor disappears (became collapsed). I test also possibility to change way how selection is made with switching `bot.setHtmlWithSelection` into `bender.tools.selection.setWithHtml`, but second method doesn't support multiple ranges.
https://github.com/ckeditor/ckeditor-dev/blob/96bf4037cb3ff812fc9b993bce4a304a884f361a/tests/plugins/tableselection/integrations/clipboard/pasteflow.js#L27

So to prevent of possible future fails, I've redesign test to run in separate editors. What fix the issue.

Blocked by: #962 
Close: #1047


mlewand:
I believe this PR was closed due to my merge messup in https://github.com/ckeditor/ckeditor-dev/pull/968#issuecomment-343519952

So we'll have to create a new PR and cherry-pick @msamsel commit.

PR #1117, opened 2017-11-03
Balloon panel API change

### LINKED ISSUE #1118: Balloon panel support attach without show

Feature request

Balloon panel should have option to `attach` panel without running `show` function.

We should change `attach` method to accept `options` instead of `focusElement` as a second param.

`options` object should be like this:
```
{
    focusElement: CKEDITOR.dom.element/Boolean
    show: Boolean
}
```

wojtekw92:

New feature

### This PR contains

- [X] Unit tests
- [ ] Manual tests

## What changes did you make?

Balloon panel still support old API in `attach` method but now We are able to give options param:
```
{
    focusElement: CKEDITOR.dom.element/Boolean
    show: Boolean
}
```
Thanks to this we can attach balloon panel without showing it.

Closes #1118

### DIFF
--- a/plugins/balloonpanel/plugin.js
+++ b/plugins/balloonpanel/plugin.js
@@ -392,8 +392,7 @@
 				}
 
 				options = CKEDITOR.tools.extend( options, {
-					show: true,
-					focusElement: undefined
+					show: true
 				} );
 
 				if ( options.show === true ) {
--- a/plugins/balloonpanel/plugin.js
+++ b/plugins/balloonpanel/plugin.js
@@ -320,10 +320,13 @@
 		 *
 		 * @method attach
 		 * @param {CKEDITOR.dom.element} element The element to which the panel is attached.
-		 * @param {CKEDITOR.dom.element/Boolean} [focusElement] The element to be focused after the panel
+		 * @param {Object} options Ballonpanel focus and show options
+		 * @param {CKEDITOR.dom.element/Boolean} [options.focusElement] The element to be focused after the panel
 		 * is attached. By default the `panel` property of {@link #parts} will be focused. You might specify the element
 		 * to be focused by passing any {@link CKEDITOR.dom.element} instance.
 		 * You can also prevent changing focus at all by setting it to `false`.
+		 * Balloonpanel still support old API where focusElement was second param of attach method.
+		 * @param {Boolean} options.show Param defines if we want to show balloonpanel after attach.
 		 */
 		attach: ( function() {
 			var winGlobal, frame, editable, isInline;
@@ -383,8 +386,18 @@
 				left: 'right'
 			};
 
-			return function( element, focusElement ) {
-				this.show();
+			return function( element, options ) {
+				var focusElement;
+				// Support for old API.
+				if ( options instanceof CKEDITOR.dom.element || typeof options === 'boolean' || typeof options === 'undefined' ) {
+					focusElement = options;
+					this.show();
+				} else {
+					if ( options.show ) {
+						this.show();
+					}
+					focusElement = options.focusElement;
+				}
 
 				this.fire( 'attach' );
 
--- a/plugins/balloonpanel/plugin.js
+++ b/plugins/balloonpanel/plugin.js
@@ -326,7 +326,7 @@
 		 * to be focused by passing any {@link CKEDITOR.dom.element} instance.
 		 * You can also prevent changing focus at all by setting it to `false`.
 		 * Balloonpanel still support old API where focusElement was second param of attach method.
-		 * @param {Boolean} options.show Param defines if we want to show balloonpanel after attach.
+		 * @param {Boolean} [options.show] Param defines if we want to show balloonpanel after attach. By default panel will show.
 		 */
 		attach: ( function() {
 			var winGlobal, frame, editable, isInline;
@@ -393,7 +393,7 @@
 					focusElement = options;
 					this.show();
 				} else {
-					if ( options.show ) {
+					if ( options.show === true || typeof options.show === 'undefined' ) {
 						this.show();
 					}
 					focusElement = options.focusElement;
--- a/plugins/balloonpanel/plugin.js
+++ b/plugins/balloonpanel/plugin.js
@@ -320,9 +320,9 @@
 		 *
 		 * @method attach
 		 * @param {CKEDITOR.dom.element} element The element to which the panel is attached.
-		 * @param {Object/CKEDITOR.dom.element/Boolean} [options] **Since 4.8.0** this parameter works as options object.
+		 * @param {Object/CKEDITOR.dom.element/Boolean} [options] **Since 4.8.0** this parameter works as `options` object.
 		 *
-		 * If CKEDITOR.dom.element/Boolean instance is given this parameter acts as a options.focusElement.
+		 * If `{@link CKEDITOR.dom.element}/Boolean` instance is given this parameter acts as a `options.focusElement`.
 		 * @param {CKEDITOR.dom.element/Boolean} [options.focusElement] The element to be focused after the panel
 		 * is attached. By default the `panel` property of {@link #parts} will be focused. You might specify the element
 		 * to be focused by passing any {@link CKEDITOR.dom.element} instance.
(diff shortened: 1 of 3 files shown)


mlewand:
Pull request is based on wrong branch. It's based on `t/933` (Inline Toolbar) while this is a balloonpanel issue, independent from Inline Toolbar.

It should be based on the major branch instead.

plugins/balloonpanel/plugin.js
```diff
@@ -320,10 +320,13 @@
...
 		 * @method attach
 		 * @param {CKEDITOR.dom.element} element The element to which the panel is attached.
-		 * @param {CKEDITOR.dom.element/Boolean} [focusElement] The element to be focused after the panel
+		 * @param {Object} options Ballonpanel focus and show options
+		 * @param {CKEDITOR.dom.element/Boolean} [options.focusElement] The element to be focused after the panel
 		 * is attached. By default the `panel` property of {@link #parts} will be focused. You might specify the element
 		 * to be focused by passing any {@link CKEDITOR.dom.element} instance.
 		 * You can also prevent changing focus at all by setting it to `false`.
+		 * Balloonpanel still support old API where focusElement was second param of attach method.
+		 * @param {Boolean} [options.show] Param defines if we want to show balloonpanel after attach. By default panel will show.
```
mlewand: JSDoc/JSDuck provides a way to indicate what is  the default value of a given property. By using this you will no longer need "By default panel will show." as it will be clear what happens by default.

Here, method works as if `options.show` is set to `true` by default.

plugins/balloonpanel/plugin.js
```diff
@@ -320,10 +320,13 @@
 		 *
 		 * @method attach
 		 * @param {CKEDITOR.dom.element} element The element to which the panel is attached.
-		 * @param {CKEDITOR.dom.element/Boolean} [focusElement] The element to be focused after the panel
+		 * @param {Object} options Ballonpanel focus and show options
+		 * @param {CKEDITOR.dom.element/Boolean} [options.focusElement] The element to be focused after the panel
 		 * is attached. By default the `panel` property of {@link #parts} will be focused. You might specify the element
 		 * to be focused by passing any {@link CKEDITOR.dom.element} instance.
 		 * You can also prevent changing focus at all by setting it to `false`.
+		 * Balloonpanel still support old API where focusElement was second param of attach method.
```
mlewand: This should be actually documented on `options` param itself, which will be tricky to put in a simple way, as it will be something like:

```
		 * @param {Object/CKEDITOR.dom.element/Boolean} options **Since 4.8.0** this parameter works as options object.
		 * If `CKEDITOR.dom.element/Boolean` instance is passed it's used as `options.focusElement`.
```

But it's still very little readable. Think how to simplify it, you can also ask Anna for help.

plugins/balloonpanel/plugin.js
```diff
@@ -320,10 +320,13 @@
 		 *
 		 * @method attach
 		 * @param {CKEDITOR.dom.element} element The element to which the panel is attached.
-		 * @param {CKEDITOR.dom.element/Boolean} [focusElement] The element to be focused after the panel
+		 * @param {Object} options Ballonpanel focus and show options
```
mlewand: This remains to be an optional parameter.

mlewand: Missing dot at the end.

mlewand: Typo "Ballonpanel" -> "Balloonpanel".

mlewand: "Balloonpanel focus and show options" - no need to enumerate option properties. If you do that you're putting yourself in responsibility of constantly maintaining this place (so that it is updated every time).

You don't want that, instead these options are listed automatically right below.

tests/plugins/balloonpanel/balloonpanel.js
```diff
@@ -149,16 +149,47 @@
...
 
 			// Ensure that focus is not changed when giving false.
-			panel.attach( strong, false );
+			panel.attach( strong, { focusElement: false } );
 			assert.isTrue( doc.getActive().equals( doc.getBody() ), 'Focus remains on body element' );
 		},
+		'test panel show option': function() {
+			var panel = new CKEDITOR.ui.balloonPanel( bender.editor, {
+				title: 'Test panel #3'
+			} ),
```
mlewand: Wrong indentation.

tests/plugins/balloonpanel/balloonpanel.js
```diff
@@ -149,16 +149,47 @@
...
 
 			// Blur focused element, so the body will be focused.
 			closeElement.$.blur();
 
 			// Ensure that focus is not changed when giving false.
-			panel.attach( strong, false );
+			panel.attach( strong, { focusElement: false } );
 			assert.isTrue( doc.getActive().equals( doc.getBody() ), 'Focus remains on body element' );
 		},
+		'test panel show option': function() {
```
mlewand: One empty line of spacing above this function would be nice.

tests/plugins/balloonpanel/balloonpanel.js
```diff
@@ -149,16 +149,47 @@
...
+				};
+			} );
+
+
+			panel.attach( strong, { show: false } );
+			assert.isFalse( showEvent, 'Event show should not be fired when show param is false.' );
+
+			panel.attach( strong, { show: true } );
+			assert.isTrue( showEvent, 'Event show should be fired when show param is true.' );
+			panel.hide();
```
mlewand: In this and next line you're actually testing `hide()` method, this is not your concern here.

tests/plugins/balloonpanel/balloonpanel.js
```diff
@@ -149,16 +149,47 @@
...
+			assert.isFalse( showEvent, 'After hide showElement should be false' );
+
+			panel.attach( strong, {} );
+			assert.isTrue( showEvent, 'Event show should be fired when show param is undefined.' );
+			panel.hide();
+
+			panels.push( panel );
```
mlewand: This should be  executed before any asserting happens.

tests/plugins/balloonpanel/balloonpanel.js
```diff
@@ -149,16 +149,47 @@
...
 			assert.isTrue( doc.getActive().equals( doc.getBody() ), 'Focus remains on body element' );
 		},
+		'test panel show option': function() {
+			var panel = new CKEDITOR.ui.balloonPanel( bender.editor, {
+				title: 'Test panel #3'
+			} ),
+				strong = bender.editor.document.findOne( 'strong' ),
+				showEvent = false;
+
+			panel.addShowListener( function() {
```
mlewand: Instead of manually write handling for checking whether or not given method was called, you should use `sinon.stub()`. Familiarize yourself with sinon lib in general, it's a great testing lib that has a lot of good features.

Here's an example how it can look like: https://gist.github.com/mlewand/4e3fef172a69593bb5d5e844b1b4bea0

plugins/balloonpanel/plugin.js
```diff
@@ -383,8 +386,18 @@
 				left: 'right'
 			};
 
-			return function( element, focusElement ) {
-				this.show();
+			return function( element, options ) {
```
mlewand: Code here could be greatly simplified. First of all you want to do as little logic forking as possible, so `this.show();` call should occur only once.

Basically best solution here is to always use `options` in object format, so if second parameter is given as bool/element just make a legit `options` object structure out of it.

plugins/balloonpanel/plugin.js
```diff
@@ -383,8 +386,19 @@
...
+				if ( options instanceof CKEDITOR.dom.element || typeof options === 'boolean' || typeof options === 'undefined' ) {
```
mlewand: Still this construction will allow `null` to slip by, and as `null` gets passed to `CKEDITOR.tools.extend` we get an exception. I have added TC for that in 70c565c54.

The simplest solution here is just change `typeof options === 'boolean' || typeof options === 'undefined'` into a `!options' and that will take care of falsy values.

plugins/balloonpanel/plugin.js
mlewand: This comment has been adressed, but GH did not picked it up.

PR #1129, opened 2017-11-07
Review corrections for inline toolbar grouping

mlewand:

Other

### This PR contains

Proper unit/manual tests are already in place.

## What changes did you make?

Simplified `inlineToolbarView.renderItems` method implementation, created in #1114.


wojtekw92:
It is not fully proper solution but I will fix this in main review.

PR #1130, opened 2017-11-07
API change in inlineToolbarView and inlineToolbar

wojtekw92:

 New feature 

### This PR contains

- [X] Unit tests
- [ ] Manual tests

## What changes did you make?

- Removed `create` method from both classes.
- Added `show` and `hide` methods.
- Added `attach` method in both classes.
- Updated tests and documentation.

### DIFF
--- a/plugins/inlinetoolbar/plugin.js
+++ b/plugins/inlinetoolbar/plugin.js
@@ -105,6 +105,10 @@
 				this.parts.close.remove();
 			};
 
+			/**
+			 * @inheritdoc CKEDITOR.ui.balloonPanel#show
+			 * @member CKEDITOR.ui.inlineToolbarView
+			 */
 			CKEDITOR.ui.inlineToolbarView.prototype.show = function() {
 				if ( this.rect.visible ) {
 					return;
@@ -115,19 +119,21 @@
 
 				this._listeners.push( this.editor.on( 'resize', function() {
 					that.attach( that._element, {
-						focusElement: false,
-						show: true
+						focusElement: false
 					} );
 				} ) );
 				this._listeners.push( editable.attachListener( editable.getDocument(), 'scroll', function() {
 					that.attach( that._element, {
-						focusElement: false,
-						show: true
+						focusElement: false
 					} );
 				} ) );
 				CKEDITOR.ui.balloonPanel.prototype.show.call( this );
 			};
 
+			/**
+			 * @inheritdoc CKEDITOR.ui.balloonPanel#hide
+			 * @member CKEDITOR.ui.inlineToolbarView
+			 */
 			CKEDITOR.ui.inlineToolbarView.prototype.hide = function() {
 				this._detachListeners();
 				CKEDITOR.ui.balloonPanel.prototype.hide.call( this );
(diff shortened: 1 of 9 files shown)


plugins/inlinetoolbar/plugin.js
```diff
@@ -151,36 +184,43 @@
...
+			 * @param {Boolean} [showToolbar=true] Show inline toolbar after attach.
 			 * @member CKEDITOR.ui.inlineToolbar
 			 */
-			CKEDITOR.ui.inlineToolbar.prototype.create = function( element ) {
+			CKEDITOR.ui.inlineToolbar.prototype.attach = function( element, showToolbar ) {
 				this._view.renderItems( this._items );
-				this._view.create( element );
+				this._view.attach( element, {
+					focusElement: false,
+					show: showToolbar === false ? false : true
```
mlewand: I see what you wanted to do here: to set false only if it was explicitly set, however it would be nice to support other _falsy_ values, like `0`.

How about changing `showToolbar` to `hidden`? This value would be marked as `false` by default, and then you could just put `!hidden` as `options.show`. WDYT?

plugins/inlinetoolbar/plugin.js
```diff
@@ -105,6 +104,40 @@
...
+				this._listeners.push( editable.attachListener( editable.getDocument(), 'scroll', function() {
+					that.attach( that._element, {
+						focusElement: false
+					} );
+				} ) );
+				CKEDITOR.ui.balloonPanel.prototype.show.call( this );
+			};
+
+			/**
+			 * @inheritdoc CKEDITOR.ui.balloonPanel#hide
```
mlewand: I might be wrong here but AFAIR if you don't provide this block it will inherit the doc anyway.

plugins/inlinetoolbar/plugin.js
```diff
@@ -105,6 +104,40 @@
...
 
+			/**
+			 * @inheritdoc CKEDITOR.ui.balloonPanel#show
+			 * @member CKEDITOR.ui.inlineToolbarView
+			 */
+			CKEDITOR.ui.inlineToolbarView.prototype.show = function() {
+				if ( this.rect.visible ) {
+					return;
+				}
+				var that = this,
```
mlewand: If you take a look onto our [`.on()` listener adding method](https://docs.ckeditor.com/ckeditor4/docs/#!/api/CKEDITOR.event-method-on) you'll note that it has `scopeObj` as a third argument.

Simply pass `this` as a `scopeObj` and you'll be able to continue to use `this` as a valid context within the function.

plugins/inlinetoolbar/plugin.js
```diff
@@ -151,36 +184,43 @@
...
-				this._detachListeners();
-
-				this._listeners.push( this.editor.on( 'resize', function() {
-					that.attach( element, false );
-				} ) );
-				this._listeners.push( editable.attachListener( editable.getDocument(), 'scroll', function() {
-					that.attach( element, false );
-				} ) );
+			CKEDITOR.ui.inlineToolbarView.prototype.attach = function( element, options ) {
+				this._element = element;
```
mlewand: This member was not documented.

Additionally please, find more specific name for it. How about `_pointedElement`? Var name needs to have a meaning, `element` might be some layout part as well.

plugins/inlinetoolbar/plugin.js
```diff
@@ -151,36 +184,43 @@
 			};
 
 			/**
-			 * Places the inline toolbar next to a specified element so the tip of the toolbar's triangle
-			 * points to that element.
-			 *
+			 * @inheritdoc CKEDITOR.ui.balloonPanel#attach
```
mlewand: Same thing as for `CKEDITOR.ui.inlineToolbarView.prototype.hide`.

plugins/inlinetoolbar/plugin.js
```diff
@@ -151,36 +184,43 @@
...
-				this._view.create( element );
+				this._view.attach( element, {
+					focusElement: false,
+					show: showToolbar === false ? false : true
+				} );
+			};
+
+			/**
+			 * @inheritdoc CKEDITOR.ui.balloonPanel#show
+			 * @member CKEDITOR.ui.inlineToolbar
```
mlewand: I hate to say it but in this case we won't be able to reuse docs for `balloonPanel.show`. As it says:

> Shows the balloon panel.

People are using toolbar, they will have no idea what the heck balloon panel means. 🙂

plugins/inlinetoolbar/plugin.js
```diff
@@ -151,36 +184,43 @@
...
+			 */
+			CKEDITOR.ui.inlineToolbar.prototype.show = function() {
+				this._view.show();
+			};
+
+			/**
+			 * @inheritdoc CKEDITOR.ui.balloonPanel#hide
+			 * @member CKEDITOR.ui.inlineToolbar
```
mlewand: Same thing as `CKEDITOR.ui.inlineToolbar.prototype.show`.

tests/plugins/inlinetoolbar/view.js
```diff
@@ -52,12 +52,37 @@
...
+				view.attach( pointedElement, { focusElement: false, show: false } );
+				assert.isFalse( view.rect.visible, 'Toolbar should not be shown.' );
+				assert.areEqual( 0, view._listeners.length, 'Listensers should not be attached' );
+
+				view.show();
+				assert.isTrue( view.rect.visible, 'Toolbar should be shown after show method' );
+				assert.areEqual( 2, view._listeners.length, 'Listensers should be attached after show method' );
+
+				view.hide();
+				assert.isFalse( view.rect.visible, 'Toolbar should not be shown after hide method.' );
```
mlewand: In assert methods we're not ending string with a dot. Same thing a line below.

plugins/inlinetoolbar/plugin.js
```diff
@@ -25,7 +25,6 @@
 			triangleHeight: 10
 		} );
 		CKEDITOR.ui.balloonPanel.call( this, editor, definition );
-
```
mlewand: Why removing this single line? 😪 It was adding a bit to the readability. 🙂

plugins/inlinetoolbar/plugin.js
```diff
@@ -55,7 +55,7 @@
 	 *				var lastElement = evt.data.path.lastElement;
 	 *
 	 *				if ( lastElement ) {
-	 *					toolbar.create( lastElement );
+	 *					toolbar.attac( lastElement );
```
mlewand: I mean... c'mon 😉

plugins/inlinetoolbar/plugin.js
```diff
@@ -151,36 +176,49 @@
...
 				this._view.renderItems( this._items );
-				this._view.create( element );
+				this._view.attach( element, {
+					focusElement: false,
+					show: !hidden
+				} );
+			};
+
+			/**
+			 * Show inline toolbar.
```
mlewand: We always give one line of spacing in API docs before JSDuck tags.

plugins/inlinetoolbar/plugin.js
wojtekw92: mea culpa. I don't know why, I thought that I missed `create` somewhere... And actually I missed one in example. And forgot about typo.

PR #1136, opened 2017-11-08
Providing fallback CSS for inlinetoolbar

msamsel:

Bug fix

### This PR contains

- [ ] Unit tests
- [X] Manual tests

## What changes did you make?

- Adding css for `inlinetoolbar` and `balloonpanel` (inline toolbar require it to proper work). CSS base on moono-lisa skin.

I thought to force loading `moono-lisa` as fallback, but it might be confusing if user will remove those css folders thinking that are unnecessary for his configuration. That's why I add `-default` suffix and separate files.


tests/plugins/inlinetoolbar/manual/fallbackstyles.html
```diff
@@ -0,0 +1,62 @@
...
+	<p>Hello world</p>
+	<img src="%BASE_PATH%_assets/lena.jpg">
+	<img src="%BASE_PATH%_assets/lena.jpg">
+	<img src="%BASE_PATH%_assets/lena.jpg">
+	<img src="%BASE_PATH%_assets/lena.jpg">
+	<p>&nbsp;</p>
+	<p>&nbsp;</p>
+	<p>&nbsp;</p>
+</div>
+
```
f1ames: You could also add inline editor here.

msamsel:
@mlewand what do you think about that?
> This task is about `inlinetoolbar` and from what I recall the default CSS was supposed to be added only for `inlinetoolbar` plugin. If there are some additional styles from `balloonpanel` needed they probably should be placed in `inlinetoolbar` CSS file too.

Inline toolbar inheritance styles from balloon panel related to:
- drawing triangle below/above balloon
- ballon position (like setting proper display mode, z-index, etc.)

msamsel:
I haven't notice that after rebase there was change in API so `create` method became `attach`.
I've made al changes suggested in your review. I extract from `balloonpanel` only important styles (from perspective of `inlinetoolbar`).

I have to prepare 2 separate test suites with css. Unfortunately skins cannot be changed after loading, so I have to made proper thing at the very begging of each test suite.

tests/plugins/inlinetoolbar/cssskin.js
```diff
@@ -0,0 +1,57 @@
...
+				skin: 'moono'
+			}
+		}
+	};
+
+	// We need to set up skin at the very beginning befor files start to load.
+	CKEDITOR.skinName = 'moono';
+
+	var tests = {
+		'test check default css usage': function( editor ) {
```
f1ames: Should it be `test check specific|moono skin css usage`?

tests/plugins/inlinetoolbar/manual/fallbackstyles.html
```diff
@@ -0,0 +1,104 @@
...
+		} );
+	}
+
+	var appendStyleSheet = sinon.stub( CKEDITOR.dom.document.prototype, 'appendStyleSheet', function( cssFileUrl ) {
+		// Simualate missing css in skin.
+		if ( cssFileUrl.indexOf( 'inlinetoolbar.css' ) > -1 || cssFileUrl.indexOf( 'balloonpanel.css' ) > -1 ) {
+			cssFileUrl = '';
+		}
+
+		if ( this.$.createStyleSheet ) {
```
f1ames: We didn't you use call to original function here same as in unit test?

tests/plugins/inlinetoolbar/manual/fallbackstyles.md
```diff
@@ -0,0 +1,13 @@
+@bender-ui: collapsed
+@bender-tags: 4.8.0, bug, inlinetoolbar
+@bender-ckeditor-plugins: wysiwygarea,toolbar,basicstyles,floatingspace,inlinetoolbar,link,image,resize,language,stylescombo
+
+Perform below steps in both editors:
+1. Click into image.
```
f1ames: 1. _Click **on** image_ or just _Click image_.

tests/plugins/inlinetoolbar/manual/fallbackstyles.html
msamsel: I've made it earlier with stub and forget about it. I think option with calling origin function is simpler, and can be resistant for possible method changes.
I've update this to looks like in Unit test.

msamsel:
As I talk with @mlewand we currently should focus on most popular browsers, and older IE left for further improvements. That's why I ignore tests on IE below 11.
I have add different assertion for the IE envs.

I also a little tune our tools method, which converts rgb to hex to also support rgba. If you consider it as separate issue, give me an info I'll extract it, but this might extend realisation of this issue. As this introduce dependancy here. Another option would be creating specific function just for those test purposes, but I'm not think is a good option, because this will duplicate the code.

plugins/inlinetoolbar/skins/default.css
```diff
@@ -0,0 +1,90 @@
...
+}
+
+/* align: [ hcenter ] */
+.cke_balloon.cke_inlinetoolbar .cke_balloon_triangle_outer.cke_balloon_triangle_align_hcenter
+{
+	left: 50%;
+	margin-left: -20px;
+}
+
+.cke_balloon.cke_inlinetoolbar
```
f1ames: This is the same selector as in line 6 (https://github.com/ckeditor/ckeditor-dev/pull/1136/files#diff-a222e1835c2e1c3b856fdaf98d45e2faR6), it should be merged into one rule.

plugins/inlinetoolbar/skins/default.css
```diff
@@ -0,0 +1,90 @@
...
+	top: -10px;
+}
+
+.cke_balloon.cke_inlinetoolbar .cke_balloon_triangle_inner.cke_balloon_triangle_top
+{
+	border-width: 0 10px 10px;
+}
+
+/* align: [ hcenter ] */
+.cke_balloon.cke_inlinetoolbar .cke_balloon_triangle_outer.cke_balloon_triangle_align_hcenter
```
f1ames: This is the same selector as in line 28 (https://github.com/ckeditor/ckeditor-dev/pull/1136/files#diff-a222e1835c2e1c3b856fdaf98d45e2faR28), it should be merged into one rule.

tests/plugins/inlinetoolbar/_helpers/default.js
```diff
@@ -0,0 +1,19 @@
+/* exported replaceAppendStyleSheet, convertRgbaToRgb */
+
+'use strict';
+
+function replaceAppendStyleSheet() {
```
f1ames: This looks good, however it doesn't provide a possibility to clean up after running tests (restore original function). You could do two things:

* This function might return a function which will restore original one when called:
```
return function() {
    CKEDITOR.dom.document.prototype.appendStyleSheet = oldFn;
}
```
* Use sinon stub which has `stub.restore()` method.

It is good practise to always clean up after test run (especially for unit tests, for manual you may skip it). So after changing it, please add clean up to unit test (you may use `tearDown` function).

msamsel: I decided to not use `sinon.stub`, because I wasn't able to find possibility to fire origin function within stub.
In documentation I found only possibility how to replace the method, with own function.
That's why I wrote some similar wrapper which could execute native method. I'll extend it a little bit to restore native method.

msamsel:
I can use neither `setUp` nor `tearDown`. Those methods are executed too late. Skins are loaded when global `CKEDITOR` object is created and are applied to all editors at given page. That's why test cases are separated between 2 files, and that's why I cannot use it in setUp. Because those methods are executed after main lib is fully loaded.
That's why I made cleaning up in a different approach inside an event, when editor should have all css loaded, so modification shouldn't be required any longer.

tests/plugins/inlinetoolbar/_helpers/default.js
f1ames: It works fine but it's a little over complicated for me. You have to create new instance, it has two methods which you have to call, etc. plus you reinvented a wheel (well, a part of it:D).

> In documentation I found only possibility how to replace the method, with own function.

That's exactly where you can call the original function, the stub could look like this:

```
function stubAppendStyleSheet() {
	var originFn = CKEDITOR.dom.document.prototype.appendStyleSheet;
	return sinon.stub( CKEDITOR.dom.document.prototype, 'appendStyleSheet', function( cssFileUrl ) {
		// Simulate missing css in skin.
		if ( !cssFileUrl.match( /(inlinetoolbar|balloonpanel)\.css/ ) ) {
			originFn.call( CKEDITOR.document, cssFileUrl );
		}
	} );
}
```

and then you can use it like:
```
var stub = stubAppendStyleSheet();
CKEDITOR.once( 'instanceReady', stub.restore );
```

It is shorter and simpler IMHO. Plus for everybody who knows `sinon`, you just take a look at the usage and it should be clear that these two lines stubs `appendStyleSheet` method and then restores it. With custom code you always have to dive in to check what's going on.

PR #1143, opened 2017-11-09
Prototype overwriting workaround in inlinetoolbarView

wojtekw92:
Bug fix 

We are just checking if beforeInit function was already executed.

### DIFF
--- a/plugins/inlinetoolbar/plugin.js
+++ b/plugins/inlinetoolbar/plugin.js
@@ -85,6 +85,89 @@
 		this._items = [];
 	};
 
+	/**
+	 * Displays the inline toolbar, pointing it to the `element`.
+	 *
+	 * @param {CKEDITOR.dom.element} element The element to which the panel is attached.
+	 * @param {Boolean} [hidden=false] Do not show inline toolbar after attach.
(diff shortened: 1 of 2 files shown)


plugins/inlinetoolbar/plugin.js
```diff
@@ -85,12 +85,101 @@
 		this._items = [];
 	};
 
+	/**
```
Comandeer: Why all these methods are moved here as part of this commit?

plugins/inlinetoolbar/plugin.js
```diff
@@ -85,12 +85,101 @@
...
+	};
+	var pluginInit = false;
+
 	CKEDITOR.plugins.add( 'inlinetoolbar', {
 		requires: 'balloonpanel',
 		onLoad: function() {
 			CKEDITOR.document.appendStyleSheet( this.path + 'skins/' + CKEDITOR.skinName + '/inlinetoolbar.css' );
 		},
-		init: function() {
+		beforeInit: function() {
```
Comandeer: What is the reason behind changing `init` to `beforeInit`?

PR #1150, opened 2017-11-10
Styles adjustments

Comandeer:

Bug fix for new feature

### This PR contains

- [x] Unit tests
- [x] Manual tests

## What changes did you make?

I've added styles to ensure that side images won't occupy the same line and full size images won't overlap side images. To do so I was forced to apply classes also on widget's wrapper.


plugins/easyimage/plugin.js
```diff
@@ -148,8 +152,10 @@
 
 					if ( data.type === 'side' ) {
 						this.element.addClass( editor.config.easyimage_sideClass );
+						this.wrapper.addClass( editor.config.easyimage_sideClass );
```
mlewand: By adding exactly same class to the wrapper you're applying the styling twice, once to the wrapper, and then again to the widget element itself.

In this case you used `max-width: 50%`  but the image can never get to 50% (as stated in manual test expectation). See below screen:

Image set to 25% width of the editor with other 25% consumed by blank padding

mlewand: Second thing is that you could avoid this kind of code duplication by calling [`widget.addClass()`](http://docs.ckeditor.test/#!/api/CKEDITOR.plugins.widget-method-addClass) (and `widget.removeClass` respectively) rather than doing this stuff manually.

By doing so your wrapper gets applied class prefixed with `cke_widget_wrapper_`. So you could fix above issue with CSS like that:

```css
.easyimage-side, .cke_widget_wrapper_easyimage-side {
	float: right;
	max-width: 25%;
}

.cke_widget_wrapper_easyimage-side > .easyimage-side {
	float: none;
	max-width: none;
}
```

Although this syntax still looks like a garbage to me (but it's a known fact I'm nor a CSS expert or a CSS lover).

plugins/easyimage/styles/easyimage.css
```diff
@@ -19,6 +20,7 @@ For licensing, see LICENSE.md or http://ckeditor.com/license
 
 .easyimage-side {
 	float: right;
+	max-width: 50%;
```
mlewand: Let's make it 25% per your accidental proposal. 25% looks much better. 🙂

Remember to update manual test description.

tests/plugins/easyimage/manual/styles.md
```diff
@@ -0,0 +1,10 @@
+@bender-tags: 4.8.0, feature, 932
+@bender-ui: collapsed
+@bender-ckeditor-plugins: sourcearea, wysiwygarea, floatingspace, toolbar, easyimage
+
+
+Check if styles are working correctly:
+
+* There is visible difference between full size image and side image.
+* Side image is occupating at most 50% of space.
+* Side images does not occupy the same line.
```
mlewand: How about making it:

> Multiple side images can not be placed in the same horizontal line.

?

plugins/easyimage/plugin.js
Comandeer: Our API still could surprise me 😅

Comandeer: Actually your styles could be simplified to:

```css
.easyimage-side, .cke_widget_wrapper_easyimage-side {
	float: right;
	max-width: 25%;
}
```

mlewand: The reason why I cared for `.easyimage-side` (with no wrapper) alone was that I wanted this stylesheet to be compatible with the content created by CKEditor. So that if someone toggles CKEditor in and out it still is styled nicely using same file.

To see it play with [/tests/plugins/easyimage/manual/styles](http://tests.ckeditor.test:1030/tests/plugins/easyimage/manual/styles) manual test just by executing `CKEDITOR.instances.inline.destroy()`.

Actually I have figured out a better selector that will handle it, without adding too much of an overhead. I'll include it in the review.

PR #1151, opened 2017-11-11
Review corrections to Easy Image linking

mlewand:

Other

### This PR contains

- [x] Unit tests
- [x] Manual tests (manual tests were already in place)

## What changes did you make?

Set of review corrections to #1137.

Most of changes are just:

* Fix for features not to override already registered feature information.
* Renamed `widgetFeatures` to `features`.
* Additional test cases.
* Test simplifications.

### DIFF
--- a/tests/plugins/imagebase/features/link.js
+++ b/tests/plugins/imagebase/features/link.js
@@ -243,7 +243,6 @@
 			testLinkCommand( {
 				bot: bot,
 				html: '<figure><a href="http://foo"><img src="%BASE_PATH%_assets/logo.png"></a></figure>',
-				initialUrl: 'foo',
 				unlinkCmdState: CKEDITOR.TRISTATE_OFF,
 				dialogCallback: function( dialog, widget ) {
 					var data = {};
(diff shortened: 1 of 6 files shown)


tests/plugins/imagebase/features/link.js
```diff
@@ -44,45 +44,32 @@
 		} );
 	}
 
-	function assertLinkWidget( options ) {
+	function assertLinkWidgetStatus( options, linkCount, linkCmdState, unlinkCmdState ) {
```
Comandeer: I don't have today time to do full review, but it's the first thing that caught my eye: why did you expand `options` parameter to 4 parameters?

mlewand: This is just the way diff has shown it 😉 `assertLinkWidget` function is still there, but few lines below. I simply extracted common part of both functions.

Comandeer: OK, but why you expanded `options` then? :P

mlewand: Actually it make sense to reuse options, true.

mlewand: Well, true, it would be a better idea to extend options for that. 🙂

Issue #1145, opened 2017-11-09
Upload widget should allow disabling default notifications


mlewand:
## Are you reporting a feature request or a bug?

Feature request

## Provide detailed reproduction steps (if any)

Currently when extending upload widget [it forces you to use our notifications](https://github.com/ckeditor/ckeditor-dev/blob/e72835fa036f35a4c5657665b4a70e19ca278efc/plugins/uploadwidget/plugin.js?utf8=%E2%9C%93#L199) for progress info, succes / fail msg.

This is good for a quick, default case - but one might want to customize notifications.

For instance in Easy Image we want to put just small custom progress bar in a picture itself.

Thus developer should be able to override default handling.

PR #1152, opened 2017-11-13
Upload progress bar for Easy Image

mlewand:

New feature

### This PR contains

- [x] Unit tests
- [x] Manual tests

## What changes did you make?

This pull requests adds a fancy progress bar to Easy Image widget:

Progress  bar in upload widget

Note that it also registers `uploadWidgetDefinition.skipNotification` which was added in #1145, but it's not in the base of current Easy Image feature branch, thus standard notification will remain in _most cases_.

### DIFF
--- a/plugins/easyimage/plugin.js
+++ b/plugins/easyimage/plugin.js
@@ -11,6 +11,8 @@
 	// jscs:disable maximumLineLength
 	// Black rectangle which is shown before image is loaded.
 	var loadingImage = 'data:image/gif;base64,R0lGODlhDgAOAIAAAAAAAP///yH5BAAAAAAALAAAAAAOAA4AAAIMhI+py+0Po5y02qsKADs=';
+	// Throttling of progress update in ms.
+	var UPLOAD_PROGRESS_THROTTLING = 100;
 	// jscs:enable maximumLineLength
 
 	function addCommands( editor ) {
@@ -196,7 +198,7 @@
 					// Add a progress bar.
 					this.definition._createProgressBar( this );
 
-					progressListeners.push( loader.on( 'update', function() {
+					var updateListener = CKEDITOR.tools.eventsBuffer( UPLOAD_PROGRESS_THROTTLING, function() {
 						var progressBar = this.parts.progressBar.findOne( '.cke_bar' ),
 							percentage;
 
@@ -204,7 +206,9 @@
 							percentage = ( loader.uploaded / loader.uploadTotal ) * 100;
 							progressBar.setStyle( 'width', percentage + '%' );
 						}
-					}, this ) );
+					}, this );
+
+					progressListeners.push( loader.on( 'update', updateListener.input ) );
 
 					progressListeners.push( loader.once( 'abort', removeProgressListeners ) );
 					progressListeners.push( loader.once( 'error', removeProgressListeners ) );
@@ -369,7 +373,7 @@
 			CKEDITOR.dialog.add( 'easyimageAlt', this.path + 'dialogs/easyimagealt.js' );
 
 			CKEDITOR.addCss( '.cke_loader { height: 15px; display: block; background: yellow; position: absolute; left: 0px; right: 0px; }\n' +
-				'.cke_loader .cke_bar { display:block; height: 13px; background: red; width: 0; }' );
+				'.cke_loader .cke_bar { display:block; height: 13px; background: red; width: 0; transition: width ' + UPLOAD_PROGRESS_THROTTLING / 1000 + 's; }' );
 		},
 
 		init: function( editor ) {
--- a/plugins/easyimage/plugin.js
+++ b/plugins/easyimage/plugin.js
@@ -226,6 +226,8 @@
 			// Easy image uses only upload method, as is manually handled in onUploading function.
 			loadMethod: 'upload',
 
+			skipNotifications: true,
+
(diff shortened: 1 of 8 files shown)


Comandeer:
> Note that it also registers uploadWidgetDefinition.skipNotification which was added in #1145, but it's not in the base of current Easy Image feature branch, thus standard notification will remain in most cases.

I've rebased branch to latest `major`, so notifications aren't shown anymore.

plugins/easyimage/plugin.js
```diff
@@ -200,7 +202,11 @@
 				this.replaceWith( '<figure class="' + ( editor.config.easyimage_class || '' ) + '"><img src="' +
 					upload.responseData.response[ 'default' ] + '" srcset="' + srcset + '" sizes="100vw"><figcaption></figcaption></figure>' );
 			}
-		} );
+		};
+
+		addUploadProgressBar( editor, uploadWidgetDefinition );
```
Comandeer: Seems like it could reuse `features` feature. What a pity that it must wait for integration of `features` into widget's core 😞

plugins/easyimage/plugin.js
```diff
@@ -239,15 +245,81 @@
...
+		 * @param {CKEDITOR.plugins.widget} widget
+		 */
+		definition._createProgressBar = function( widget ) {
+			widget.parts.progressBar = CKEDITOR.dom.element.createFromHtml( '<div class="cke_loader">' +
+					'<div class="cke_bar" styles="transition: width ' + UPLOAD_PROGRESS_THROTTLING / 1000 + 's"></div>' +
+				'</div>' );
+			widget.wrapper.append( widget.parts.progressBar, true );
+		};
+
+		editor.on( 'widgetDefinition', function( evt ) {
```
Comandeer: Yeah, it definitely _should_ use `features`.

plugins/easyimage/plugin.js
```diff
@@ -239,15 +245,81 @@
...
 				}
 			}
 
 			data.dataValue = temp.getHtml();
 		} );
 	}
 
+	// Extends given uploadWidget `definition` with an upload progress bar, added within wrapper.
+	function addUploadProgressBar( editor, definition ) {
+		definition.skipNotifications = true;
```
Comandeer: Why not use `CKEDITOR.tools.object.merge`? I don't see so many benefits of mutating definition over returning new, extended one and the code would IMO a lot better.

plugins/easyimage/plugin.js
```diff
@@ -239,15 +245,81 @@
...
+		/*
+		 * Creates a progress bar in a given widget.
+		 *
+		 * Also puts it in it's {@link CKEDITOR.plugins.widget#parts} structure as `progressBar`
+		 *
+		 * @private
+		 * @param {CKEDITOR.plugins.widget} widget
+		 */
+		definition._createProgressBar = function( widget ) {
+			widget.parts.progressBar = CKEDITOR.dom.element.createFromHtml( '<div class="cke_loader">' +
```
Comandeer: I wonder if we shouldn't add some ARIA roles for progressbar (like, uhm, `progressbar`).

plugins/easyimage/styles/easyimage.css
```diff
@@ -42,3 +42,17 @@ For licensing, see LICENSE.md or http://ckeditor.com/license
...
+/* Loaders */
+
+.cke_loader {
+	position: absolute;
+	left: 0px;
+	right: 0px;
+}
+
+.cke_loader .cke_bar {
+	height: 10px;
```
Comandeer: I like CKE5 loader, which is much thinner. Maybe we could also make it thin?

plugins/uploadwidget/plugin.js
```diff
@@ -337,7 +337,15 @@
 				} else {
 					editor.getSelection().selectBookmarks( bookmarks );
 				}
+			},
 
+			/**
+			 * @private
+			 * @returns {CKEDITOR.fileTools.fileLoader/null} Loader associated with this widget instance or `null` if not found.
+			 */
+			_getLoader: function() {
```
Comandeer: I lack unit test for this new private method.

plugins/easyimage/styles/easyimage.css
mlewand: I had a problem with it that it's _barely visible_, I didn't see it at all at first. Let's extract beautifying / styling into a separate user story where Ryszard will help us.

plugins/easyimage/plugin.js
mlewand: A11y will be provided in a followup issue, we should extract it to a separate story (I mean complete a11y, not limited to progress bar).

plugins/easyimage/plugin.js
mlewand: This is what I have been thinking of 🙂 That's the reason why I tried to wrap it in a single function, and make it modular so that it could be easily moved.

plugins/easyimage/plugin.js
mlewand: Don't remember what was my reasoning here. Isn't it that `object.merge` does not modify any object but returns new instead? So that we'll have quite a bit of fun to make it work.

Eventual alternative is `tools.extend`.

Comandeer: Let's skip it now as we're going to rewrite it to `features` anyway.

PR #1158, opened 2017-11-14
Bug fixes for PFW Image with rotation, reflection, alternative text

msamsel:

Bug fix

yes

### This PR contains

- [x] Unit tests
- [x] Manual tests

## What changes did you make?

- improve way how `v:shape` are process by PFW filter.
- remove code which was responsible for returning `null` in case of shape detection. Currently shapes are filtered out earlier, so this part is not required any longer (this cause one of the errors).

related to: #662


tests/plugins/pastefromwordimage/manual/pastewordwithimage.md
```diff
@@ -13,6 +13,9 @@
   2. [Online_and_offline_image.docx](../generated/_fixtures/Online_and_offline_image/Online_and_offline_image.docx)
   3. [Shapes_and_online_and_offline_image.docx](../generated/_fixtures/Shapes_and_online_and_offline_image/Shapes_and_online_and_offline_image.docx)
   4. [Wrapped_image.docx](../generated/_fixtures/Wrapped_image/Wrapped_image.docx)
+  5. [Image_alternative_text.docx](../generated/_fixtures/Image_alternative_text/Image_alternative_text.docx)
+  6. [Image_reflection.docx](../generated/_fixtures/Image_reflection/Image_reflection.docx)
+  7. [Image_rotation.docx](../generated/_fixtures/Image_rotation/Image_rotation.docx)
```
f1ames: When pasting rotated image it is pasted in its original form (not rotated). I'm not sure if it is a proper behaviour or not (as for previous `.docx` images are pasted exactly as they appear in the document). If this is an intended behaviour it should be explain that rotated images are pasted, but without rotation.

tests/plugins/pastefromwordimage/manual/pastewordwithimage.md
```diff
@@ -13,6 +13,9 @@
...
 
 **Expected:** Images are pasted to the editor.<br>
 Note: Keep in mind that shapes (like last object in `Shapes_and_online_and_offline_image.docx`) will not be pasted.
```
f1ames: To be honest I cannot recall what was the status of handling shapes. Now I see that for shapes (e.g. in `Shapes_and_online_and_offline_image.docx`), `img` element with invalid path (pointing to some local/temporary resources) is inserted. Shouldn't it just be skipped? Same happens for canvas elements.

f1ames: I know that this PR is about images, but as a whole implementation of `v:shape` filter changed it influences shapes also.

msamsel: It's a good point. Honestly I don't know. I remain them untouched. I'll talk with @mlewand what to do with it.

msamsel: Ok so img which earlier appear after pasting shapes are now removed. More details how it works is below in main line comment.

msamsel:
Ok so shapes, or more specific img tags belong to shapes, are now removed from PFW output. I change line which tag them with `data-cke-is-shape="true"` to remove them instead.
I also decide to remain shape related tests. Might be useful in the future if we decide to process shapes with a better way. Currently those test shows that img tags are removed from the output. So check if there is not shapes which leaks out as unhandled img tags.
Currently we also don't want to replace those img tags with notification such as "Not supported image" or anything like that.

msamsel:
Extracted issue here: #1183

tests/plugins/pastefromword/manual/shapesaretagged.md
```diff
@@ -6,6 +6,4 @@
 1. Copy its contents and paste to the editor.
 1. Click button `getHtml` below editor.
 
-**Expected:**
-  * For browsers with good `text/html` clipboard processing (Firefox, Chrome, Safari): First image tag had additional attribute `data-cke-is-shape="true"`. Second image is pasted without this attribute.
-  * For other browsers (Edge, IE, Mobile): There won't be `data-cke-is-shape="true"`. Browser will behave in an old way: display blank place or embed image by it's own.
+**Expected:** Single image is pasted to the editor. Shape present in word document is removed during paste process.
```
f1ames: You simplified it too much IMHO. It works as described in Chrome, FF, Safari. On IE11 both shape and image are pasted and on older IEs again only second image.

Additionally AFAIR or Safari and IEs < 11, the image is pasted, but it has invalid src, so basically the broken image is pasted (hard to tell if it is valid behaviour based on description).

TBH, I haven't thought about this earlier, but this test (`shapesaretagged`) was created to check if shapes are tagged with special attribute. Now shapes are removed so there is no purpose in keeping it and it could be simply removed, WDYT @msamsel? The case with removing shapes is already covered in other manual tests.

Sorry for the confusion, but I think it will be the best way, no sense in duplicating similar test.

Issue #1169, opened 2017-11-15
Restrict 'pastebin' usage in widgets plugin only for browsers which really utilizes it


f1ames:
## Are you reporting a feature request or a bug?

Task

## [Check if the issue is already reported](https://docs.ckeditor.com/ckeditor4/docs/#!/guide/dev_issues_readme-section-avoid-duplicates)

Extracted from #1160.

## Provide detailed reproduction steps (if any)

See https://github.com/ckeditor/ckeditor-dev/issues/1160#issuecomment-344580238 for wider context.

The `clipboard` plugin uses `pastebin` only for browsers not supporting Clipboard API (IE browsers), but for widgets it is used always. I think we should consider restricting its usage only from browsers which really needs it, especially that:

1. `copySingleWidget` is called first and it utilizes `CKEDITOR.editable.getHtmlFromRange` to prepare HTML which is copied by the browser to a clipboard/dataTransfer.
1. `clipboard` plugin listeners are executed, which for browsers supporting Clipboard API, creates `CKEDITOR.plugins.clipboard.dataTransfer`.
1. `CKEDITOR.plugins.clipboard.dataTransfer` object upon creation inserts `text/html` data to clipboard (using `CKEDITOR.editor.getSelectedHtml`) overwriting html which was inserted via `pastebin`.

Also `CKEDITOR.editable.getHtmlFromRange` is the base of the `CKEDITOR.editor.getSelectedHtml` method so the output should is the same.

## Other details

* Browser: All
* OS: All
* CKEditor version: 
* Installed CKEditor plugins: `clipboard`, `widgets`

Comandeer:
I wasn't able to activate pastebin in non-IE browsers. It'd be good to check if the issue is still present.

PR #1170, opened 2017-11-15
Cutting the widget scrolls editor to the end (with t/468)

### LINKED ISSUE #1160: Cutting the widget scrolls editor to the end (with t/468)

Bug

1. Go to http://tests.ckeditor.test:1030/tests/plugins/clipboard/manual/draganddrop
2. Select the widget from the first editor.
3. Press CTRL + X.

### Expected result

Editor scrolls to the end.

### Actual result

Editor does not scroll at all or scroll position is restored.

## Other details

* Browser: Edge 16+
* OS: Windows
* CKEditor version: `major` with #468 
* Installed CKEditor plugins: `clipboard`

f1ames:

Bug fix

### This PR contains

- [ ] Unit tests
- [x] Manual tests

## What changes did you make?

Closes #1160.


tests/tickets/gh1160/1.html
```diff
@@ -0,0 +1,80 @@
+<head>
```
Comandeer: Can't these tests be moved to `plugins/widget`? They're widget specific, so that place seems more obvious to me. We got even `plugins/widget/integration` directory.

f1ames: 👍  Moved to `widget` directory.

f1ames:
The only way I was able to workaround this issue is to use the `div` element for blockless editor. Haven't found any negative side effects so it seems to be working fine.

The only problem is that for this 100ms (when pastebin container is present) we have invalid html structure (`div` inside `h1`).

I think the only reliable, not hacky solution would be get rid of `pastebin` as described in #1169.

mlewand:
Just to chime in, there's no chance in the world we'll be able to deliver #1169 in a _reasonable time_. If this PR does the job, let's just use it as a workaround, and then revisit this in future releases.

Just one more question: is this `div` in a `p` thing introduced by this very PR, or was it already in the editor since beginning?

Comandeer:
> Just one more question: is this div in a p thing introduced by this very PR, or was it already in the editor since beginning?

It seems to be there from beginning – that's how our fake selection works: putting hidden `div` at the end of the editable. It _shouldn't_ work, but works flawlessly so far.

mlewand:
So this is nothing critical that we need to fix ASAP in this release. Instead we'll deal with it in the upcoming.

Meanwhile to avoid UX regression for Edge users, let's use this PR as a temporary workaround. Feel free to add a reference in code comments to a #1169 saying that _xyz_ could be safely removed when #1169 is fixed.

BTW:

> It shouldn't work, but works flawlessly so far.

Love that part 😁

PR #1192, opened 2017-11-21
Disable drag function in inline toolbar elements

wojtekw92:

Bug fix

### This PR contains

- [ ] Unit tests
- [X] Manual tests

## What changes did you make?

Added HTML `draggable="false"` flag to all elements inside inline toolbar.

### DIFF
--- a/plugins/inlinetoolbar/plugin.js
+++ b/plugins/inlinetoolbar/plugin.js
@@ -693,6 +693,14 @@
 				}
 
 				this.parts.content.setHtml( output.join( '' ) );
+				this.parts.content.unselectable();
+				var element,
+					elements = this.parts.content.getElementsByTag( '*' );
+
+				for ( var i = 0, count = elements.count() ; i < count ; i++ ) {
+					element = elements.getItem( i );
+					element.setAttribute( 'draggable', 'false' );
+				}
 			};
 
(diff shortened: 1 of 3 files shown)


plugins/inlinetoolbar/plugin.js
```diff
@@ -284,6 +284,14 @@
 				}
 
 				this.parts.content.setHtml( output.join( '' ) );
+				this.parts.content.unselectable();
+				var element,
+					elements = this.parts.content.getElementsByTag( '*' );
```
mlewand: It should be optimized by doing this only for `a` elements.

plugins/inlinetoolbar/plugin.js
```diff
@@ -284,6 +284,14 @@
...
+
+				for ( var i = 0, count = elements.count() ; i < count ; i++ ) {
+					element = elements.getItem( i );
+					element.setAttribute( 'draggable', 'false' );
+				}
```
mlewand: This code could be much simplified, here's an equivalent of it:

```js
CKEDITOR.tools.array.forEach( this.parts.content.find( '*' ).toArray(), function( element ) {
	element.setAttribute( 'draggable', 'false' );
} );
```

PR #1201, opened 2017-11-22
Balloon panel respects body's position attribute

### LINKED ISSUE #1048: Balloon panel body margin handling

Bug

Balloon panel calculate wrong position coordinates when we add margin to body.

1. edit balloon panel manual test test (`balloonpanel/manual/balloonpanel.html`)
2. remove CSS body styles from html
3. run test

### Expected result

balloon panel position is ok.

### Actual result

balloon panel positon is moved for the same length as margin.

## Other details

* Browser: Chrome 62
* OS: MacOS 10
* CKEditor version: 4.7.3
* Installed CKEditor plugins: balloonpanel

msamsel:

Bug fix

yes

### This PR contains

- [x] Unit tests
- [x] Manual tests

## What changes did you make?

- I update way how position of balloonpanel is calculated to respect ancestors tags position attributes. Now if there is non-static ancestor with a margin, its margin will be properly removed from balloon position.
- Balloons now also stick more to editable area if are created with `divarea` plugin

close #1048


tests/plugins/balloonpanel/manual/positioning.md
```diff
@@ -0,0 +1,12 @@
+@bender-ui: collapsed
+@bender-tags: 4.8.0, bug, balloonpanel
+@bender-ckeditor-plugins: wysiwygarea,toolbar,undo,basicstyles,notification,balloonpanel,sourcearea,list,link
+
+----
+
+1. Click into red square.
+2. Check position of displayed balloon.
+
+**Expected:** Balloon point nicely to clicked text.
```
f1ames: I would change it to:
> Balloon points exactly to clicked element.

tests/plugins/balloonpanel/manual/positioning.html
```diff
@@ -0,0 +1,74 @@
+<h2>Divarea editor:</h2>
```
f1ames: This test is fine, however it relies on a fact that bender already has "relative body". I would declare it explicitly here so in case of something changes it still will be valid. Plus you can see what is tested on the first glimpse.

tests/plugins/balloonpanel/positioningnonstatic.js
```diff
@@ -0,0 +1,122 @@
...
+		},
+		'test positioning in nonstatic body': function( editor ) {
+			CKEDITOR.document.getBody().setStyles( {
+				position: 'relative',
+				'margin-left': '300px',
+				'margin-top': '300px',
+				'min-width': '2000px',
+				'min-height': '3000px'
+			} );
+			var BODY_MARGIN_LEFT = parseInt( CKEDITOR.document.getBody().getComputedStyle( 'margin-left' ), 10 ),
```
f1ames: I would stick with proper naming convention here `bodyMarginLeft`.

tests/plugins/balloonpanel/manual/positioning.html
f1ames: You can also add some `margin-top`.

msamsel:
I just removed second test, as it was almost identical. It supposed sth wrong when tests are fired inside this iframe. I check also that using parameters from deleted test in first one, generates passing tests. So it seems to be not related to values, but more for the test environment.

tests/plugins/balloonpanel/positioningnonstatic.js
```diff
@@ -0,0 +1,104 @@
...
+
+		return {
+			elementRect: panel._getAbsoluteRect( element ),
+			triangleTip: balloonTestsTools.getTriangleTipPosition( panel ),
+			panel: panel
+		};
+	}
+
+	var tests = {
+		setUp: function() {
```
f1ames: `setUp` and `tearDown` are run before/after each test. While this is the only place where you add a stub to `stubs` array it looks like there is always only one stub, so it would be easier just to store it globally without `stubs` array. You may rename it to `getViewPaneSizeStub` (or something similar) for clarity.

msamsel:
I'm not sure if the PR branch should be targeted to `major` any longer.

tests/plugins/balloonpanel/_helpers/tools.js
```diff
@@ -57,5 +57,72 @@ var balloonTestsTools = {
 			return document.location.protocol + '//' + document.location.hostname + ( document.location.port ? ':' + document.location.port : '' );
 		}
 		return document.location.origin;
+	},
+
+	getAbsoluteRect: function( editor, element ) {
```
f1ames: Somehow I didn't noticed it earlier... 😞  Since you made `_getAbsoluteRect ` in `balloonpanel` plugin available and it is used in the unit tests, this here is no longer needed and should be removed.

msamsel: done ;)

msamsel:
I also rebased the branch to current major.

f1ames:
It will be merged to `master`/`major` after upcoming release.

PR #1214, opened 2017-11-23
Merge pastefromwordimage plugin to pastefromword

msamsel:

Other, please explain
Code refactoring - Merge pastefromwordimage plugin to pastefromword

There remain all old tests. No new feature or bug fixes was added.

### This PR contains

- [ ] Unit tests
- [ ] Manual tests

## What changes did you make?

- move `paste from word image` plugin into paste from word
- move tests related to this feature
- add configuration option to disable pasting images from word

This is a subtask of #662


plugins/pastefromword/plugin.js
```diff
@@ -110,6 +110,13 @@
 				// this event.
 				isLazyLoad && evt.cancel();
 			}, null, null, 3 );
+
+			// Paste From Word Image:
+			// RTF clipboard is required for embeding images.
+			if ( CKEDITOR.plugins.clipboard.isCustomDataTypesSupported && configInlineImages ) {
+				editor.filter.allow( 'img[src]' );
```
mlewand: Actually in this case we don't want to force adding `img` to the allowed content list.

Instead we can do a bit more clever thing, and detect whether images are allowed. If not there's also no point in doing this logic. For example:

```js
editor.filter.check( 'img[src]' )
```

tests/plugins/pastefromword/generated/PFW_image.js
```diff
@@ -0,0 +1,39 @@
+/* bender-tags: clipboard,pastefromword */
```
mlewand: Please, rename file to lowercased `pfw_image`, actually... `pfw` prefix doesn't have much use here. After all it's a directory for pfw so only its tests are here. 🙂 So how about just `image`?

tests/plugins/pastefromword/generated/PFW_image.js
```diff
@@ -0,0 +1,39 @@
...
+/* bender-ckeditor-plugins: pastefromword,ajax,basicstyles,bidi,font,link,toolbar,colorbutton,image */
+/* bender-ckeditor-plugins: list,liststyle,sourcearea,format,justify,table,tableresize,tabletools,indent,indentblock,div,dialog */
```
mlewand: I think many of this plugins could be safely removed. Could we do that?

msamsel: I reduced it, but few of them still remain.

msamsel:
I wasn't able to easily embed entire logic inside filter. It cause too many test errors. I'll extract it along with other stuff to refactoring ticket.

plugins/pastefromword/filter/default.js
```diff
@@ -38,7 +38,74 @@
...
+		 *
+		 * @private
+		 * @since 4.8.0
+		 * @member CKEDITOR.plugins.pastefromword
+		 * @param {String} rtfContent RTF content to be checked for images.
+		 * @returns {Object[]} An array of images found in the `rtfContent`.
+		 * @returns {String} return.hex Hexadecimal string of an image embedded in `rtfContent`.
+		 * @returns {String} return.type String represent type of image, allowed values: 'image/png', 'image/jpeg'.
+		 */
+		extractImagesFromRtf: function( rtfContent ) {
```
mlewand: I would like to avoid putting this logic at the very beginning of PFW filter file.

This part is not as important to expose it here. Instead I would suggest enclosing it in a dedicated namespace, like `CKEDITOR.plugins.pastefromword.images` and it could contain:

* `extractImagesFromRtf` actually renamed to `extractFromRtf`.
* `extractImgTagsFromHtml` actually renamed to `extractTagsFromHtml`.

Then move it a bit to the bottom of a file.

PR #1216, opened 2017-11-23
Fix for race condition while uploading multiple files with uploadimage plugin

### LINKED ISSUE #1213: Race condition while uploading images to CKFinder using uploadimage plugin

Bug

1. Open a new tab - https://codepen.io/mlewand/pen/RjyoPR.
2. Select all four images rendered in HTML preview.
3. Copy with `ctrl/cmd+c`.
4. Open https://sdk.ckeditor.com/samples/fileupload.html#uploading-dropped-and-pasted-images.
5. Focus and clear contents of editor under "Uploading Dropped and Pasted Images".
6. Paste using `ctrl/cmd+v`.

 ### Expected result

All four images are upload, each image is different.

 ### Actual result

Randomly some images will duplicate/mangle. If you did not reproduced try again, as it's a basically a race condition.

 ## Other details

* Browser: All
* OS: Windows 10, Mac
* CKEditor version: 4.7.3

mlewand:

Bug fix

### This PR contains

- [x] Unit tests
- [ ] Manual tests

## What changes did you make?

This issue is more of a workaround, as the problem is actually on a CKFinder side, but we can't quickly fix it there, so we'll patch client side first.

Now, the problem was that CKFinder had a problem handling multiple files with the same name loaded at once, so we'll generate unique file name for each one. It's a bit dirty, and optimistic, but should do the trick for now.

### Manaul Tests

To be honest I had no good idea how to make a good manual test here. To make it reliable I have set a CKFinder 3 instance on my dev workstation, and added configuration to `config.js`:

By adding:

```
	config.uploadUrl = '/ckfinder/core/connector/php/connector.php?command=QuickUpload&type=Files&responseType=json';
```

And adding `'uploadimage'` to the plugin list.

Closes #1213.

### DIFF
--- a/plugins/uploadimage/plugin.js
+++ b/plugins/uploadimage/plugin.js
@@ -6,6 +6,29 @@
 'use strict';
 
 ( function() {
+	var counter = 0,
+		// Black rectangle which is shown before image is loaded.
+		loadingImage = 'data:image/gif;base64,R0lGODlhDgAOAIAAAAAAAP///yH5BAAAAAAALAAAAAAOAA4AAAIMhI+py+0Po5y02qsKADs=';
+
+	// Returns number as a string. If a number has 1 digit only it returns it prefixed with an extra 0.
+	function padDate( input ) {
+		if ( input <= 9 ) {
+			input = '0' + input;
+		}
+
+		return String( input );
+	}
+
+	// Returns an unique image file name.
+	function getUniqueImageFileName( type ) {
+		var date = new Date(),
+			dateParts = [ date.getFullYear(), date.getMonth() + 1, date.getDate(), date.getHours(), date.getMinutes(), date.getSeconds() ];
+
+		counter += 1;
+
+		return 'image-' + CKEDITOR.tools.array.map( dateParts, padDate ).join( '' ) + '-' + counter + '.' + type;
+	}
+
 	CKEDITOR.plugins.add( 'uploadimage', {
 		requires: 'uploadwidget',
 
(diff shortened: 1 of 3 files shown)


tests/plugins/uploadimage/uploadimage.js
```diff
@@ -550,6 +552,30 @@
 				} );
 			} );
 
+			wait();
+		},
+
+		'test uploads generate unique names (#1213)': function() {
```
Comandeer: Test is always failing:

> TypeError: Attempted to wrap create which is already wrapped

plugins/uploadimage/plugin.js
```diff
@@ -6,6 +6,29 @@
...
+			input = '0' + input;
+		}
+
+		return String( input );
+	}
+
+	// Returns an unique image file name.
+	function getUniqueImageFileName( type ) {
+		var date = new Date(),
+			dateParts = [ date.getFullYear(), date.getMonth() + 1, date.getDate(), date.getHours(), date.getMinutes(), date.getSeconds() ];
```
Comandeer: Couldn't we use simply timestamp here, returned from `Date.now()`/`Date.prototype.getTime()`?

tests/plugins/uploadimage/uploadimage.js
mlewand: Ooops, TBH don't know how I missed that.

plugins/uploadimage/plugin.js
mlewand: We want to give it _some_ human readability, at the same time not making the name too long.

PR #1229, opened 2017-11-25
Fix mobile paste

### LINKED ISSUE #595: Fix pasting (paste/paste plain text/paste from word) experience for mobile browsers
CKEditor's 4.7 plain text paste is broken in Android Firefox and Chrome. It is used not often, but sometimes it is very handy to paste & edit texts from tablet. But in 4.7 paste from clipboard doesn't work at all.

When trying to paste via toolbar icon or editor menu I get "Your browser doesn't allow you to paste this way. Press CTRL+V/CTRL+SHIFT+V to paste" message.

Maybe it is possible to leave old paste dialog as optional plugin? Cause in CKEditor 4.6.3 all types of paste worked fine in mobile and desktop browsers. Seems you fixed a thing that was not broken :)

* Browser: Chrome, Firefox
* OS: Android 7.1.2
* CKEditor version: 4.7

Comandeer:

Bug fix

### This PR contains

- [x] Unit tests
- [x] Manual tests

## What changes did you make?

I've added back our beloved paste dialog. It's shown only if user touches the button (via recognising `touchend` event).

**There are missing language files entries that should be added before merging.** For now only `en` locale is covered.

Closes #595.


plugins/clipboard/plugin.js
```diff
@@ -502,6 +572,29 @@
...
+			if ( editor.ui.addButton ) {
+				// Waiting for editor instance to be ready seems to be the most reliable way to
+				// be sure that paste buttons are already created.
+				editor.once( 'instanceReady', function() {
+					if ( !editor._.pasteButtons ) {
+						return;
+					}
+
+					CKEDITOR.tools.array.forEach( editor._.pasteButtons, function( name ) {
+						var pasteButton = editor.ui.get( name ),
```
mlewand: This function might return `undefined`, for instance you could have clipboard plugin, but no "Paste" (or related) button in the toolbar.

Comandeer: In such case `editor._.pasteButtons` should be empty or undefined.

mlewand: Please check [/tests/plugins/clipboard/manual/missingbutton manual test](http://tests.ckeditor.test:1030/tests/plugins/clipboard/manual/missingbutton) in [t/595-nobutton](https://github.com/ckeditor/ckeditor-dev/tree/t/595-nobutton) branch.

f1ames:
I extracted context menu integration to #1347.

f1ames:
Before merging/closing this ticket #1347 should be covered and merged here.

mlewand:
Rebased onto latest 4.9.0 branch. Changed the base of this PR accordingly.

plugins/clipboard/dialogs/paste.js
```diff
@@ -0,0 +1,252 @@
...
+
+			this.setupContent();
+
+			// Set dialog title to the custom value (set e.g. in editor.openDialog callback) and reset this value.
+			// If custom title not set, use default one.
+			this.parts.title.setHtml( this.customTitle || lang.title );
+			this.customTitle = null;
+
+			// Reset commited indicator.
+			this._.commited = false;
```
mlewand: `commited` -> `committed`

The fix has to be applied across entire PR.

plugins/clipboard/plugin.js
```diff
@@ -427,12 +440,19 @@
 		 * an upcoming major release.
 		 */
 		editor.getClipboardData = function( callbackOrOptions, callback ) {
```
mlewand: Function signature once again changed, now we're using options once again…

That's a messy situation and we should adjust the API docs correctly. I mean that the options argument is again supported, etc.

plugins/clipboard/plugin.js
```diff
@@ -502,6 +569,32 @@
...
+						return;
+					}
+
+					CKEDITOR.tools.array.forEach( editor._.pasteButtons, function( name ) {
+						var pasteButton = editor.ui.get( name );
+						// Check if button was not removed by `removeButtons` config.
+						if ( pasteButton ) {
+							var buttonElement = CKEDITOR.document.getById( pasteButton._.id );
+
+							buttonElement.on( 'touchend', function() {
```
mlewand: I was thinking about optimizing this listener, and putting it once on the entire toolbar and just then check what button was pressed. However after playing a bit with it I start to think that it's too much of effort for the benefit. So I guess it's fine the way it is.

plugins/clipboard/plugin.js
```diff
@@ -502,6 +569,32 @@
...
+								editor._.forcePasteDialog = true;
+							} );
+						}
+					} );
+
+					delete editor._.pasteButtons;
```
mlewand: Let's remove this member in the editor, so it could be reused by other plugins if needed this workarounds for mobile.

plugins/clipboard/plugin.js
```diff
@@ -3002,6 +3123,16 @@
  * @member CKEDITOR.editor
  */
 
+/**
+ * Internal event to open the Paste dialog window.
+ *
+ * @private
+ * @event pasteDialog
```
mlewand: We should note that [`pasteDialog` event](https://github.com/ckeditor/ckeditor-dev/blob/34a59d1/plugins/clipboard/plugin.js#L3130) was not available in 4.7.0 - 4.8.0 versions.

plugins/clipboard/plugin.js
```diff
@@ -1556,6 +1652,31 @@
...
+		 * This method should be called for buttons that should display the Paste Dialog fallback in mobile environments.
+		 * See {@link https://github.com/ckeditor/ckeditor-dev/issues/595#issuecomment-345971174 the rationale} for more
+		 * details.
+		 *
+		 * @since 4.9.0
+		 * @param {CKEDITOR.editor} editor The editor instance.
+		 * @param {String} name Name of the button.
+		 * @param {Object} definition Definition of the button.
+		 */
+		addPasteButton: function( editor, name, definition ) {
```
mlewand: `addPasteButton` should return created `CKEDITOR.ui.button` instance, so that we could use it for instance with Balloon Toolbar.

Comandeer: Due to the super convoluted way of accessing button instance from our UI, we decided to not introduce this change. Info gathered from `editor._.pasteButtons` should be enough to create buttons for Balloon Toolbar.

plugins/clipboard/plugin.js
Comandeer: I've just dropped handling of options as it was used only to set custom title of paste dialog, which wasn't even working.

Comandeer:
> Talking about cleanup, this evt.cancel() call doesn't have much sense as the only place firing the pasteDialogCommit event is paste dialog and it doesn't care whether event was canceled or not. WDYT?

It was a part of whole dialog logic from the beginning and revert of dialog deletion brought it back. Honestly I also don't see much sense in it, but I was just afraid to touch it 😄 However it seems that removing it does not change anything.

Comandeer:
Actually I located the issue: without this `evt.cancel`, `paste` event is called twice. So it's better to leave it untouched.

Comandeer:
I've added unit test for `evt.cancel`, fixed existing tests and add changelog entry.

mlewand:
Thanks for the research. I have rebased onto latest major. There was an issue that on IE unit tests would trigger a security dialog (request for allowing the paste). I have stubbed the `document.execCommand` for that particular TC and it works OK.

I'll wait with a final R+ for #1347.

PR #1242, opened 2017-11-27
Move creation API to the editor directly

### LINKED ISSUE #1237: Move creation API to the editor directly

Task

This is a part of #933 issue.

We would like to simplify the API and docs by moving Balloon Toolbar creation API to the `editor` class directly.

f1ames:

Task: API refactoring

### This PR contains

- [ ] Unit tests
- [ ] Manual tests

## What changes did you make?

Closes #1237.

### DIFF
--- a/plugins/balloontoolbar/plugin.js
+++ b/plugins/balloontoolbar/plugin.js
@@ -376,9 +376,16 @@
 	}
 
 	ContextManager.prototype = {
-
 		/**
-		 * Creates {@link CKEDITOR.plugins.balloontoolbar.context} from the given options.
+		 * Creates a toolbar context based on provided `options`, registers it in the manager and returns.
+		 *
+		 *		// Shows bold and underline button for any selection.
+		 *		var context = contextManager.create( {
+		 *			buttons: 'Bold,Underline',
+		 *			refresh: function() {
+		 *			 	return true;
+		 *			}
+		 *		} );
 		 *
 		 * @param {CKEDITOR.plugins.balloontoolbar.contextDefinition} options Config object that determines the conditions used to display the toolbar.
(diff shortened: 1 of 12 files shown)


plugins/balloontoolbar/plugin.js
```diff
@@ -536,45 +551,23 @@
 		init: function( editor ) {
 
 			/**
-			 * Set of instance-specific public APIs exposed by the [Balloon Toolbar](https://ckeditor.com/cke4/addon/balloontoolbar) plugin.
+			 * Balloon Toolbar manager for a given editor instance. It ensures that there's only one toolbar visible at a time.
+			 *
+			 * It is the simplest way to create a Balloon Toolbar. It uses {@link CKEDITOR.plugins.balloontoolbar.contextManager#create} underneath.
```
mlewand: I think this line could be simplified something like:

> * Use {@link CKEDITOR.plugins.balloontoolbar.contextManager#create} method to register a new toolbar context.

plugins/balloontoolbar/plugin.js
```diff
@@ -376,6 +376,21 @@
 	}
 
 	ContextManager.prototype = {
+
+		/**
+		 * Creates {@link CKEDITOR.plugins.balloontoolbar.context} from the given options.
```
mlewand: I don't think we should repeat returned type explicitly here. It's enough it's in the `@returns` tag.

Let's just make it something like "Creates a toolbar context", so that we reduce change that future maintainer will change return value in one place, and not the other.

It would be also nice to add the information that this method (as opposed to add method) creates, adds and returns the context.

I would also love to see a short code listing here (yea I know that one was in the `editor.balloonToolbars` property, but I think it won't do harm).

mlewand:
Just a couple of minor doc details. Proposed my changes in #1245.

f1ames:
Merged changes and rebased on`t/933`, ready for R.

PR #1327, opened 2017-12-05
Balloon toolbar now register its items in focusmanager

### LINKED ISSUE #1232: Balloon Toolbar buttons should be registered as focusable elements

Bug

1. Open [tests/plugins/balloontoolbar/context/manual/context test](http://tests.ckeditor.dev:1030/tests/plugins/balloontoolbar/context/manual/context).
2. Click inside any strong element to bring up the toolbar.
3. Long press left mouse button on "underline" button.

### Expected result

* Underline command should be activated.
* Toolbar should remain visible.
* Editor should be focused.

### Actual result

* Underline command is not activated.
* Toolbar hides.

## Other details

* Browser: Any
* OS: Any
* CKEditor version: 4.8.0

msamsel:

Bug fix

yes

### This PR contains

- [x] Unit tests
- [x] Manual tests

## What changes did you make?

- Currently when balloon toolbar will be created then all added items will be registered in focus manager
- There is added additional array which store reference to elements which was registered is focus manager.

close #1232

### DIFF
--- a/plugins/balloontoolbar/plugin.js
+++ b/plugins/balloontoolbar/plugin.js
@@ -663,7 +663,7 @@
 			};
 
 			CKEDITOR.ui.balloonToolbarView.prototype.destroy = function() {
-				this.deregisterFocusableItems();
+				this._deregisterItemFocusables();
 				CKEDITOR.ui.balloonPanel.prototype.destroy.call( this );
 				this._detachListeners();
 			};
@@ -680,7 +680,7 @@
 					groupStarted = false;
 
 				// When we rerender toolbar we want to clear focusable in case of removing some items.
-				this.deregisterFocusableItems();
+				this._deregisterItemFocusables();
 
 				CKEDITOR.tools.array.forEach( keys, function( itemKey ) {
 
@@ -728,19 +728,17 @@
 			};
 
 			/**
-			 * Method deregister all focusable items from balloontoolbar.
+			 * Method deregisters {@link #focusables} registered by UI items, like buttons.
 			 *
 			 * @private
 			 * @since 4.8.1
 			 * @member CKEDITOR.ui.balloonToolbarView
 			 */
-			CKEDITOR.ui.balloonToolbarView.prototype.deregisterFocusableItems = function() {
+			CKEDITOR.ui.balloonToolbarView.prototype._deregisterItemFocusables = function() {
 				var focusables = this.focusables;
 				for ( var id in focusables ) {
-					if ( focusables.hasOwnProperty( id ) ) {
-						if ( focusables[ id ].getId() && focusables[ id ].getName() === 'a' ) {
-							this.deregisterFocusable( focusables[ id ] );
-						}
+					if ( this.parts.content.contains( focusables[ id ] ) ) {
+						this.deregisterFocusable( focusables[ id ] );
 					}
 				}
 			};
--- a/plugins/balloontoolbar/plugin.js
+++ b/plugins/balloontoolbar/plugin.js
@@ -34,12 +34,6 @@
 		 * @private
 		 */
 		this._listeners = [];
-
-		/**
-		 * Reference to {@link CKEDITOR.dom.element} belong to rendered items inside {@link CKEDITOR.ui.balloonToolbar}.
-		 * Array is necessary to clear out those items from {@link CKEDITOR.focusManager}.
-		 */
-		this._focusablesItems = [];
 	};
 
 	/**
@@ -667,7 +661,7 @@
 			};
 
 			CKEDITOR.ui.balloonToolbarView.prototype.destroy = function() {
-				deregisterAllFocusableItems.call( this, this._focusablesItems );
+				this.deregisterFocusableItems();
 				CKEDITOR.ui.balloonPanel.prototype.destroy.call( this );
 				this._detachListeners();
 			};
@@ -684,7 +678,7 @@
 					groupStarted = false;
 
 				// When we rerender toolbar we want to clear focusable in case of removing some items.
-				deregisterAllFocusableItems.call( this, this._focusablesItems );
+				this.deregisterFocusableItems();
 
 				CKEDITOR.tools.array.forEach( keys, function( itemKey ) {
 
@@ -711,7 +705,6 @@
 				CKEDITOR.tools.array.forEach( this.parts.content.find( 'a' ).toArray(), function( element ) {
 					element.setAttribute( 'draggable', 'false' );
 					this.registerFocusable( element );
-					this._focusablesItems.push( element );
 				}, this );
 			};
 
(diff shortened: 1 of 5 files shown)


tests/plugins/balloontoolbar/context/manual/focusablebuttons.md
```diff
@@ -0,0 +1,12 @@
+@bender-ui: collapsed
+@bender-tags: 4.8.1, bug, balloontoolbar
+@bender-ckeditor-plugins: wysiwygarea,toolbar,basicstyles,floatingspace,balloontoolbar,link,image
+
+----
+
+1. Click into `emphasize` or `strong` element.
+2. When ballon toolbar show click into unselected button and keep mosedown for a few seconds.
```
mlewand: Typo ballon -> balloon
and mosedown -> mousedown

tests/plugins/balloontoolbar/context/focusablebuttons.js
```diff
@@ -0,0 +1,87 @@
...
+
+			before = CKEDITOR.tools.objectKeys( toolbar._view.focusables );
+
+			toolbar.attach( editor.editable().findOne( 'strong' ) );
+			afterAdding = CKEDITOR.tools.objectKeys( toolbar._view.focusables );
+
+			toolbar.deleteItem( 'Bold' );
+			afterDeleting = CKEDITOR.tools.objectKeys( toolbar._view.focusables );
+
+			assert.isFalse( CKEDITOR.tools.arrayCompare( before, afterAdding ), 'After adding new button, there should appear new focusables.' );
```
mlewand: Assertion messages should not end with a dot character. Please fix also other occurences.

plugins/balloontoolbar/plugin.js
```diff
@@ -723,6 +736,16 @@
 
 				CKEDITOR.ui.balloonPanel.prototype.attach.call( this, element, options );
 			};
+
+			// We don't want to expose this method as it should be called only in very specific moments:
+			// 1. Rendering view to clear out toolbar from possible deleted items.
+			// 2. Destroying view.
+			function deregisterAllFocusableItems( elements ) {
```
mlewand: `deregisterFocusableItems` is a simpler name that still reflects the purpose.

plugins/balloontoolbar/plugin.js
```diff
@@ -723,6 +736,16 @@
 
 				CKEDITOR.ui.balloonPanel.prototype.attach.call( this, element, options );
 			};
+
+			// We don't want to expose this method as it should be called only in very specific moments:
```
mlewand: This is why you should make it a private method (please don't forget to include `@since` information).

As a result you won't need to do things like `CKEDITOR.ui.balloonPanel.prototype.deregisterFocusable.call( this, element );` - the code will be simplified.

plugins/balloontoolbar/plugin.js
```diff
@@ -704,7 +715,9 @@
 				this.parts.content.unselectable();
 				CKEDITOR.tools.array.forEach( this.parts.content.find( 'a' ).toArray(), function( element ) {
 					element.setAttribute( 'draggable', 'false' );
-				} );
+					this.registerFocusable( element );
+					this._focusablesItems.push( element );
```
mlewand: Actually we could solve this issue without introducing `this._focusablesItems` - which duplicates [`focusables`](https://docs.ckeditor.com/ckeditor4/docs/#!/api/CKEDITOR.ui.balloonPanel-property-focusables) property. And this is the biggest problem introduced by this approach, we have 2 duplicated places where we keep this information.

Instead we should reuse and rely on `focusables`. The only place where we really need to do something custom is here:

https://github.com/ckeditor/ckeditor-dev/blob/d567e2e45a6ff206c36753a1c3978689ba26cb39/plugins/balloontoolbar/plugin.js#L691-L692

We have to remove focusables that are within `content` part.

To do that we could simply iterate over it and test each focusable if it is a child of content part and remove it if it actually is (do this check for each focusable). And we won't have to worry about logic duplication.

tests/plugins/balloontoolbar/context/focusablebuttons.js
```diff
@@ -0,0 +1,87 @@
...
+
+	bender.editor = {
+		startupData: '<p>Hello <strong>world</strong>.</p>',
+		config: {
+			extraAllowedContent: 'strong em'
+		}
+	};
+
+	bender.test( {
+		'test check if balloon toolbar elements are registered as focusable': function() {
```
mlewand: This TC name could be simplified to "test if balloon toolbar elements are registered as focusable".

tests/plugins/balloontoolbar/context/focusablebuttons.js
```diff
@@ -0,0 +1,87 @@
...
+			var editor = this.editor,
+				toolbar = new CKEDITOR.ui.balloonToolbar( editor ),
+				before,
+				afterAdding,
```
mlewand: Why adding extra variables like that instead asserting right after given action happened? E.g. instead of putting it into `afterAdding` variable it should be simply checked right after attaching the toolbar.

tests/plugins/balloontoolbar/context/manual/focusablebuttons.md
```diff
@@ -0,0 +1,12 @@
+@bender-ui: collapsed
+@bender-tags: 4.8.1, bug, balloontoolbar
```
mlewand: Missing issue number reference.

mlewand:
Pushed a commit with typos fixes in manual test.

plugins/balloontoolbar/plugin.js
```diff
@@ -723,6 +729,24 @@
...
+			/**
+			 * Method deregister all focusable items from balloontoolbar.
+			 *
+			 * @private
+			 * @since 4.8.1
+			 * @member CKEDITOR.ui.balloonToolbarView
+			 */
+			CKEDITOR.ui.balloonToolbarView.prototype.deregisterFocusableItems = function() {
+				var focusables = this.focusables;
+				for ( var id in focusables ) {
```
msamsel: I have to iterate here over focusables to get actually registered elements in focus manager. It appears that reference to elements is kept in `this.panels.content` even when those elements are removed from DOM. This cause an error in test where additional clear was called as `tearDown`, because another attempt to deregister focusable with those same elements have happen.

plugins/balloontoolbar/plugin.js
```diff
@@ -723,6 +729,24 @@
...
+					if ( focusables.hasOwnProperty( id ) ) {
+						if ( focusables[ id ].getId() && focusables[ id ].getName() === 'a' ) {
```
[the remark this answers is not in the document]
mlewand: In https://github.com/ckeditor/ckeditor-dev/pull/1327#discussion_r157462771 I have asked for slightly different thing:

> We have to remove focusables that are within content part.
>
> To do that we could simply iterate over it and test each focusable if it is a child of content part and remove it if it actually is (do this check for each focusable). And we won't have to worry about logic duplication.

This could solved simply with code like that:

```js
var focusables = this.focusables;
for ( var id in focusables ) {
	if ( this.parts.content.contains( focusables[ id ] ) ) {
		this.deregisterFocusable( focusables[ id ] );
	}
}
```

PR #1352, opened 2017-12-15
Use XHR as possible uploading mechanism in filebrowser plugin

### LINKED ISSUE #643: File browser plugin should allow for setting custom HTTP headers

Feature

## Check if the issue is already reported

* https://dev.ckeditor.com/ticket/13791
* https://github.com/cksource/ckfinder-connector-php/issues/245#issuecomment-296366359

[Uploading dropped files](https://docs.ckeditor.com/ckeditor4/docs/#!/guide/dev_file_upload-section-editor-side-configuration) allows for creating cross-origin AJAX requests by adding custom HTTP requests.

Unfortunately this is not possible in the [File Browser](https://ckeditor.com/cke4/addon/filebrowser) plugin.

This is caused by the File Browser plugin currently using a standard `<form>` element that is submitted upon clicking the "Send it to the Server" button.

Mouse cursor over the "Send it to the Server" button

## Solution

We'll implement an **additional upload method** for the File Browser plugin that uses `XMLHttpRequest` through the [File Tools](https://ckeditor.com/cke4/addon/filetools) plugin.

The new upload method could be enabled using the `config.filebrowserUploadMethod` property.

The old method using the plain `<form>` element will still be supported as it is required by old browsers.

For 4.8.x, the `<form>`element will still be used by default since we cannot make backward incompatible changes in a minor release. Early adopters will already be able to enable it by setting:

```js
config.filebrowserUploadMethod = 'xhr';
```

The default behavior will be changed in version 4.9.0 (#1365).

## Browser Compatibility

All [officially supported browsers](https://docs.ckeditor.com/ckeditor4/docs/#!/guide/dev_browsers) except Internet Explorer 8 and 9.

The new feature will not be compatible with IE8 and IE9 as there is no Form Data type available there. These browsers will, however, gracefully fall back to the old `<form>`-based solution. As a result no custom headers will be sent, but the editor will still attempt to upload the file.

## Other details

As stated in ticket https://dev.ckeditor.com/ticket/13791, it will probably be easier  to create a new plugin then redesign the existing one.

* Browser: Any
* OS: Any
* CKEditor version: 4.0+
* Installed CKEditor plugins: `filebrowser`

msamsel:

New feature

### This PR contains

- [x] Unit tests
- [x] Manual tests

## What changes did you make?

- change way how filebrowser upload files - prefer using XHR
- allow on providing custom headers to XHR used when files are uploaded
- add config flag which will force to use `form submit` behaviour when files are uploaded

Closes #643.

### DIFF
--- a/plugins/filetools/plugin.js
+++ b/plugins/filetools/plugin.js
@@ -49,6 +49,7 @@
 				var fileLoader = evt.data.fileLoader,
 					$formData = new FormData(),
 					requestData = evt.data.requestData,
+					configXHRHeaders = editor.config.xmlHttpRequestHeaders,
 					header;
 
 				for ( var name in requestData ) {
@@ -65,9 +66,9 @@
 				// Append token preventing CSRF attacks.
 				$formData.append( 'ckCsrfToken', CKEDITOR.tools.getCsrfToken() );
 
-				if ( editor.config.xmlHttpRequestHeaders ) {
-					for ( header in editor.config.xmlHttpRequestHeaders ) {
-						fileLoader.xhr.setRequestHeader( header, editor.config.xmlHttpRequestHeaders[ header ] );
+				if ( configXHRHeaders ) {
+					for ( header in configXHRHeaders ) {
+						fileLoader.xhr.setRequestHeader( header, configXHRHeaders[ header ] );
 					}
 				}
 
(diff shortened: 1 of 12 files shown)


plugins/filetools/plugin.js
```diff
@@ -885,3 +911,16 @@
...
+/**
+ * Additional headers of XMLHttpRequest used during file upload with plugins: {@link CKEDITOR.fileTools} and filebrowser.
+ *
+ * 	config.xmlHttpRequestHeaders = {
+ * 		'Cache-Control': 'no-cache',
+ * 		'X-CUSTOM': 'HEADER'
+ * 	};
+ *
+ * @since 4.8.1
+ * @cfg {Object} [xmlHttpRequestHeaders=null]
```
mlewand: This config property should be prefixed with `fileTools_` just like the other filetools config property.

mlewand:
I'll take over this one from here.

PR #1369, opened 2017-12-21
Corrections to XHR support for File Browser plugin uploads

mlewand:

New feature

### This PR contains

- [x] Unit tests
- [x] Manual tests

## What changes did you make?

Corrections to #1352 PR.

There are multiple minor changes, like:

* renaming config options,
* adding missing API doc tags,
* adding some unit tests,
* multiple simplifications.

### DIFF
--- a/plugins/filetools/plugin.js
+++ b/plugins/filetools/plugin.js
@@ -49,7 +49,7 @@
 				var fileLoader = evt.data.fileLoader,
 					$formData = new FormData(),
 					requestData = evt.data.requestData,
-					configXHRHeaders = editor.config.fileTools_requestHeaders,
+					configXhrHeaders = editor.config.fileTools_requestHeaders,
 					header;
 
 				for ( var name in requestData ) {
@@ -66,9 +66,9 @@
 				// Append token preventing CSRF attacks.
 				$formData.append( 'ckCsrfToken', CKEDITOR.tools.getCsrfToken() );
 
-				if ( configXHRHeaders ) {
-					for ( header in configXHRHeaders ) {
-						fileLoader.xhr.setRequestHeader( header, configXHRHeaders[ header ] );
+				if ( configXhrHeaders ) {
+					for ( header in configXhrHeaders ) {
+						fileLoader.xhr.setRequestHeader( header, configXhrHeaders[ header ] );
 					}
 				}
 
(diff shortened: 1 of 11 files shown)


tests/plugins/filebrowser/manual/uploadxhrwitherror.html
```diff
@@ -1,17 +1,14 @@
...
-
 </textarea>
 
 <div id="output" style="background-color:lightgreen;"></div>
 
 <script>
-	var xhr, requests;
-
-	if ( CKEDITOR.env.ie && CKEDITOR.env.version < 10 ) {
+	if ( !CKEDITOR.fileTools.isFileUploadSupported ) {
```
Comandeer: Manual test is not working due to this change.

tests/plugins/filebrowser/manual/uploadxhr.html
```diff
@@ -1,13 +1,12 @@
...
-<div id="output" style="background-color:lightgreen;"></div>
+<pre id="output" style="background-color:lightgreen;"></pre>
 
 <script>
-	var xhr, requests;
+	var xhr,
+		requests;
 
-	if ( CKEDITOR.env.ie && CKEDITOR.env.version < 10 ) {
+	if ( !CKEDITOR.fileTools.isFileUploadSupported ) {
```
Comandeer: Manual test is not working due to this change.

plugins/filetools/plugin.js
```diff
@@ -867,21 +867,18 @@
...
 		/**
-		 * Property inform if current browser poses native methods used by {@link CKEDITOR.fileTools} to upload files.
+		 * Feature detection indicating whether current browser supports methods essential to send files over XHR request.
 		 *
+		 * @since 4.8.1
 		 * @property {Boolean} isFileUploadSupported
 		 */
 		isFileUploadSupported: ( function() {
-			if ( typeof FileReader === 'function' &&
+			return typeof FileReader === 'function' &&
```
Comandeer: Actually `FileReader` is not needed alongside `FormData`. Making XHR request using `FormData` with passed file input should be easy to do. I wonder if we could modify our file loaders accordingly.

Comandeer: Because ATM our file loader transforms `Blob` (so binary format ready to be send via xhr) to Data URI, which is probably longer and weights more than binary format. This operation is also completely redundant in case we already have _better_ format.

mlewand: That's true, however it's a separate issue that is beyond scope of this one.

mlewand:
Failing tests fixed.

PR #1387, opened 2017-12-29
Prevent image stretching in Easy Image

Comandeer:

Bug fix

### This PR contains

- [x] Unit tests
- [x] Manual tests

## What changes did you make?

I've added `[width]` attribute with natural dimensions of the image to the widget, which prevents stretching image if it's too small to fit editor's width.

### DIFF
--- a/plugins/easyimage/plugin.js
+++ b/plugins/easyimage/plugin.js
@@ -170,6 +170,21 @@
 	}
 
 	function registerUploadWidget( editor ) {
+		// Natural width of the image can be fetched only after image is loaded.
+		// However cached images won't fire `load` event, but just mark themselves
+		// as complete.
+		function getNaturalWidth( image, callback ) {
+			var $image = image.$;
+
+			if ( $image.complete && $image.naturalWidth ) {
+				return callback( $image.naturalWidth );
+			}
+
+			image.once( 'load', function() {
+				callback( $image.naturalWidth );
+			} );
+		}
+
 		var uploadWidgetDefinition = {
 			supportedTypes: /image\/(jpeg|png|gif|bmp)/,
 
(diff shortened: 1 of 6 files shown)


mlewand:
As for the tests I think there are at least two reasons for tests passing on IE11/Edge:

* [We're using the same `blob:` URL for every test](https://github.com/ckeditor/ckeditor-dev/blob/6a9b60d/tests/plugins/easyimage/uploadwidget.js#L13), where browser can cache the image (as if it would do with regular HTTP(S) images).
* We're making assertions after the main image was loaded.

These are just hints, as I'd have to do some more checks before making a definitive claim, but I hope this is a valuable hint.

plugins/easyimage/plugin.js
```diff
@@ -197,10 +212,13 @@
...
 			onUploaded: function( upload ) {
-				var srcset = CKEDITOR.plugins.easyimage._parseSrcSet( upload.responseData.response );
+				var srcset = CKEDITOR.plugins.easyimage._parseSrcSet( upload.responseData.response ),
+					widget = this;
 
-				this.replaceWith( '<figure class="' + ( editor.config.easyimage_class || '' ) + '"><img src="' +
-					upload.responseData.response[ 'default' ] + '" srcset="' + srcset + '" sizes="100vw"><figcaption></figcaption></figure>' );
+				getNaturalWidth( widget.parts.img, function( width ) {
+					widget.replaceWith( '<figure class="' + ( editor.config.easyimage_class || '' ) + '"><img src="' +
+					upload.responseData.response[ 'default' ] + '" srcset="' + srcset + '" sizes="100vw" width="' + width + '"><figcaption></figcaption></figure>' );
```
mlewand: It should be indented. Generally this HTML string should be a bit more readable.

PR #1402, opened 2018-01-04
Balloon Toolbar API is now available in plugin.init()

### LINKED ISSUE #1346: Balloon Toolbar API should be available in plugin.init()

Feature request

While making a [PoC plugin, reusing Balloon Toolbar](https://github.com/ckeditor/ckeditor4-plugin-linkballoon/blob/7088d7137551ba5e9fcf9aa9b8a0cc63cb6e0886/plugin.js#L19-L22), I was surprised that the API is not exposed in `plugin.init()`, thus I had to use `afterInit()` which is not an elegant workaround.

It's the result of #1142 in a long run, but maybe we could fix it without waiting for #1142 fix? For instance init Balloon Toolbar APIs in `beforeInit()`.

## Other details

* Browser: Any
* OS: Any
* CKEditor version: 4.8.0

jacekbogdanski:

Feature

### This PR contains

- [x] Unit tests
- [x] Manual tests

## What changes did you make?

Added balloon toolbar api context initialization to plugin.beforeInit().

Closes #1346

#1346: Fixed: [Balloon Toolbar API](https://docs.ckeditor.com/ckeditor4/docs/#!/api/CKEDITOR.ui.balloonToolbar) should be exposed in plugin.init() of a [required](https://docs.ckeditor.com/ckeditor4/docs/#!/api/CKEDITOR.pluginDefinition-property-requires) plugin.


tests/plugins/balloontoolbar/manual/context-api.md
```diff
@@ -0,0 +1,11 @@
+@bender-ui: collapsed
+@bender-tags: 4.8.0, feature, 1346, balloontoolbar
```
f1ames: As `4.8.0` was already release, the target version of this fix is `4.8.1`.

tests/plugins/balloontoolbar/manual/context-api.md
```diff
@@ -0,0 +1,11 @@
...
+@bender-ckeditor-plugins: wysiwygarea,toolbar,undo,basicstyles,notification,floatingspace,sourcearea,list,link
```
f1ames: When creating tests, remember to include plugins which are only necessary for the tests to pass/fail. In this case, you could just skip all plugins as there is not user interaction with the editor.
However, it is better to make editor properly load (so it doesn't looks like test is broken) so you could use `wysiwygarea,toolbar,basicstyles`.

tests/plugins/balloontoolbar/manual/context-api.md
```diff
@@ -0,0 +1,11 @@
...
+
+Check if `balloontoolbar` api is available in plugin.init().
+
+This test should not throw error after page load.
+
+Open console and check if `editor.balloonToolbars should be initialized` error has been thrown.
+
+If not, test passed.
```
f1ames: When creating tests description you may try to follow our common pattern for tests scenarios like:

image

tests/plugins/balloontoolbar/plugin.js
```diff
@@ -0,0 +1,26 @@
+/* bender-tags: balloontoolbar, 1346, context */
```
f1ames: I think `context` it to detailed, I would go with `editor, balloontoolbar, 1346`.

tests/plugins/balloontoolbar/plugin.js
```diff
@@ -0,0 +1,26 @@
...
+/* bender-include: _helpers/default.js */
+/* global ignoreUnsupportedEnvironment */
+
+( function() {
+	'use strict';
+
+	bender.editor = {};
```
f1ames: `bender.editor = {};` creates default (based on bender tags) CKEditor instance, before any tests are executed. As you are already creating an editor instance in test (`bender.editorBot.create`) you may simply remove this line.

tests/plugins/balloontoolbar/plugin.js
```diff
@@ -0,0 +1,26 @@
...
+		'test context api is available in plugin.init()': function() {
+			CKEDITOR.plugins.add( 'testplugin', {
+				requires: 'balloontoolbar',
+				init: function( editor ) {
+					assert.isNotUndefined( editor.balloonToolbars );
+				}
+			} );
+
+			bender.editorBot.create(
+			{ name: 'editor2', config: { extraPlugins: 'testplugin' } }, function noop(  ) {} );
```
f1ames: This line should be indented (or you may simply put everything in the one line). Also I don't see any benefit in naming this `noop` function. You may simply use `bender.editorBot.create( { ... }, function() {} );`.

tests/plugins/balloontoolbar/plugin.js
f1ames: `plugin.js` is not very descriptive name for test suite, maybe something like `contextinit.js` (well, it will be nice if the naming will be consistent with manual test name, so you could go with `contextapi`)?

f1ames:
Could you also add suggested changelog entry?

f1ames:
Rebased on latest `master`, adjusted and moved changelog entry to end of the list (according to docs).

PR #728, opened 2017-08-04
Styles definition always stores its classes in lexicographical order

Comandeer:
## What is the purpose of this pull request?

Bug fix

## Does your PR contain necessary tests?

All patches which change the editor code must include tests. You can always read more
on [PR testing](http://docs.ckeditor.com/#!/guide/dev_contributing_code-section-tests),
[how to set the testing environment](http://docs.ckeditor.com/#!/guide/dev_tests) and
[how to create tests](http://docs.ckeditor.com/#!/guide/dev_tests-section-creating-your-own-test)
in the official CKEditor documentation.

### This PR contains

- [x] Unit tests
- [ ] Manual tests

## What changes did you make?

Closes #727.

Issue #989, opened 2017-09-29
Cannot avoid classes being sorted alphabetically, breaking Semantic UI


vphantom:
## Are you reporting a feature request or a bug?

Bug

This may be related to, but is distinct from, #727.

## Detailed reproduction steps

1. In config.extraAllowedContent, allow classes '*'.
2. In Source edition mode, create `<div class="ui three column grid"></div>`
3. Switch to WYSIWYG.
4. Switch back to Source edition to see filtered HTML.

### Expected result

I expect the HTML to remain `<div class="ui three column grid"></div>`.

### Actual result

HTML was transformed to `<div class="column grid three ui"></div>`, breaking Semantic UI's order-sensitive CSS rules.

The problem here is that I need to let my CMS users make use of some of our framework's building blocks (for flex columns, for example).  I was hoping there was a flag to disable this behavior, but couldn't find anything relevant.

I understand per #727 and others that for _internal_ comparison with toggle-able styles, sorting is necessary, but that doesn't need to be reflected in the actual HTML output.

## Other details

* Browser: Chrome 61.0.3163.91 (Official Build) (64-bit)
* OS: Debian 4.9.18-1~bpo8+1 (2017-04-10) x86_64 GNU/Linux
* CKEditor version: from the CDN, 4.7.3 (revision dee99e2)
* Installed CKEditor plugins: your stock "full" build

gabiudrescu:
any ideas / suggestions on how to avoid sorting the classes on save?

PR #1403, opened 2018-01-04
Checking elements in filter regardless of class order

### LINKED ISSUE #727: Styles combo disabled when using style with multiple, not sorted classes

Bug

https://dev.ckeditor.com/ticket/13206

1. Open: https://jsfiddle.net/z90dfv8n/
2. Click on the styles combo.

### Alternative steps

1. Create a minimal CKEditor instance that includes the `stylescombo` plugin and the `styleSet` configuration option set to:

```
[ { name: 'Marker', element: 'span', attributes: { 'class': 'b-test a-test' } } ]
```
2. Click on the styles combo.

### Expected result

Styles combo shows one style that's named "Marker".

### Actual result

Styles combo is empty and disables itself upon clicking on it.

## Other details

* With the class order switched (sorted lexically) the bug does not occur.
* The (bug-inducing) style is recognized but in this form the ACF filters it out from the styles list.
* Pasting this snippet: 

```
<p><span class="b-test a-test">Test</span></p>
```

into the source and switching modes back and forth produces a sorted output like this: 

```
<p><span class="a-test b-test">Test</span></p>
```

* Browser: All
* OS: All
* CKEditor version: 4.3
* Installed CKEditor plugins: `stylescombo`

msamsel:

Bug fix

### This PR contains

- [x] Unit tests
- [x] Manual tests

## What changes did you make?

- ~pretty much I've made change as it was in #728 and add manual tests~
- ~I couldn't re-use #728 :( there is unknown source repository, so I wasn't able to start from that PR.~
- I propose another solution which sorts class during check and do not change users markup

Close #727


tests/core/filter/manual/classcheckinlexicalorder.html
```diff
@@ -0,0 +1,19 @@
+<textarea id="classic"><p>Hello world</p></textarea>
+<div id="inline" contenteditable="true"><p>Hello world</p></div>
+
+<script>
+	CKEDITOR.replace( 'classic', {
+		stylesSet: [
```
engineering-this: When each pair of key and value is in new line it is much more easier to read. Also I'd consider extracting styleSet to variable, as it is used more than once.

tests/core/filter/manual/classcheckinlexicalorder.md
```diff
@@ -0,0 +1,22 @@
...
+3. Check what is displayed in styles combo
+
+**Expected:** Styles combo should display 3 different styles
+
+----
+
+1. Select entire paragraph
+2. Apply style `Marker 2` to the paragraph
+3. Do not change selection or reselect the same text
+3. Re-open stylescombo
```
engineering-this: Typo: step '3.' is twice. You could just use '1.' in every step and let markdown parser to create correct lists.

tests/core/filter/manual/classcheckinlexicalorder.md
```diff
@@ -0,0 +1,22 @@
...
+1. Select paragraph
+2. Open stylecombo
+3. Check what is displayed in styles combo
+
+**Expected:** Styles combo should display 3 different styles
+
+----
+
+1. Select entire paragraph
+2. Apply style `Marker 2` to the paragraph
```
engineering-this: I'd add explicit information to use the styles combo to apply certain style.
Eg. 'Apply style `Marker 2` from styles combo'

msamsel:
> I think the problem is more general, than just disabled styles combo. This is just case when the issue is visible, but our approach should be wider, to cover all cases with multiple classes.

In my opinion is not the scope of this issue. It's more likely feature request where we start support classes without lexical order in entire editor. Currently there are some situation when editor's content is destroyed and recreated from the scratch. During this process class are re-created alphabetically. That's why to support described cases we should store somehow order of the classes and properly restore it. Comparing this to issue that style cannot be obtain from styles combo is much more wider task.
This conclusions are also somehow related to issues: #1671, #989.

@Comandeer WDYT about going into general and wider approach with this issue?

msamsel:
I've rebased branch and change target to master.

Comandeer:
> WDYT about going into general and wider approach with this issue?

It'd be nice, but seems like to big change without really visible advantages. IMO this PR should fix only the rich combo issue.

Issue #1360, opened 2017-12-19
Sample wrongly assumes that wysiwygarea is available


Comandeer:
## Are you reporting a feature request or a bug?

Bug

## Provide detailed reproduction steps (if any)

1. Comment out `wysiwygarea` plugin in `config.js`.
2. Open `samples/index.html`.

### Expected result

Editor is in inline mode due to non-availability of `wysiwygarea` plugin.

### Actual result

Editor is in classic mode, but it doesn't work as there is no `wysiwygarea`.

Because of this bug, editor in source mode downloaded from [online builder](https://ckeditor.com/cke4/builder) and without `wysiwygarea` has not-working example.

## Other details

* CKEditor version: dev version

Issue is caused by `isWysiwygareaAvailable` function claiming that development mode ensures availability of `wysiwygarea` plugin.

PR #1437, opened 2018-01-11
Fixed plugin detection in samples

jacekbogdanski:

Bug fix.

### This PR contains

- [ ] Unit tests
- [ ] Manual tests

## What changes did you make?

Changed samples initialisation code to resolve plugins in development mode based on `config.js` file.
Closing #1360

### DIFF
--- a/samples/js/sample.js
+++ b/samples/js/sample.js
@@ -14,14 +14,11 @@ CKEDITOR.config.height = 150;
 CKEDITOR.config.width = 'auto';
 
 var initSample = ( function() {
-	var wysiwygareaAvailable = isWysiwygareaAvailable(),
-		isBBCodeBuiltIn = !!CKEDITOR.plugins.get( 'bbcode' );
-
 	return function() {
 		var editorElement = CKEDITOR.document.getById( 'editor' );
 
-		// :(((
-		if ( isBBCodeBuiltIn ) {
+		// Sample content for bbcode.
+		if ( hasPlugin( 'bbcode' ) ) {
 			editorElement.setHtml(
 				'Hello world!\n\n' +
 				'I\'m an instance of [url=https://ckeditor.com]CKEditor[/url].'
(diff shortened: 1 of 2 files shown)


samples/js/sample.js
```diff
@@ -14,40 +14,48 @@ CKEDITOR.config.height = 150;
...
 		if ( CKEDITOR.revision == ( '%RE' + 'V%' ) ) {
-			return true;
+			var config = {};
+			CKEDITOR.editorConfig( config );
+
+			return config.plugins.split( ',' ).indexOf( name ) > -1;
 		}
 
-		return !!CKEDITOR.plugins.get( 'wysiwygarea' );
+		return Boolean( CKEDITOR.plugins.get( name ) );
```
engineering-this: We do use double negation in our code style. Don't change it to Boolean constructor.

jacekbogdanski: TBO I cannot disagree more. In my opinion, it's better to be explicit, not implicit. It's more verbose and more performant (although performance is not an important statement here). I think we should improve our code if we can, not follow bad code styles. For me, double negation is as bad as a bitwise operator (ya, I was trying to smuggle this one into our code base 😭). However, I think it would be nice to hear our team opinion thus it may be not objective.

jacekbogdanski:
I rebased it into the latest upstream.

samples/js/sample.js
engineering-this: I understand your point of view, but I think its more important to abide [code style guide](https://github.com/ckeditor/ckeditor-idiomatic.js)👮. According to it:

>All code in any code-base should look like a single person typed it, no matter how many people contributed.
>"Arguments over style are pointless. There should be a style guide, and you should follow it"

If you disagree with it, you should start with raising discussion among team. And until we decide to change it please follow current rules. Note that we have only 5 occurrences of phrase 'Boolean( ' on our `master` branch and 2 of them are coming for your commits.

jacekbogdanski: In the big part of our code, we also have a single line if statements without brackets. What about it?

jacekbogdanski: I believe that this issue is not so much important that we could firstly resolve the issue with code styles.

jacekbogdanski: It seems like our team agrees that we should follow already defined rules. I will refactor this code soon.

PR #1456, opened 2018-01-15
Easy Image - Improve object URL management

f1ames:

Task

### This PR contains

- [x] Unit tests
- [x] Manual tests

## What changes did you make?

Improved object URL management.


tests/plugins/easyimage/manual/objecturlerror.html
```diff
@@ -0,0 +1,28 @@
+<div id="classic">
```
Comandeer: You included `floatingspace` plugin, but not added inline editor. Please add it.

plugins/easyimage/plugin.js
```diff
@@ -202,6 +202,9 @@
 	}
 
 	function registerUploadWidget( editor ) {
+		var objectUrlCache = {},
```
Comandeer: I'm wondering if it won't be better to create nice class for managing object URLs.

f1ames: Yes, it will, but as I discussed with @mlewand we wanted the "minimalistic" approach here and creating proper manager (probably integrated somehow with `uploadwidget`) will be a follow up task.

tests/plugins/easyimage/manual/objecturlerror.html
f1ames: Some leftover 😞

plugins/easyimage/plugin.js
mlewand: Proper manager will be implemented together with Mateusz hackathon task. Which will be as @f1ames mentioned a followup task, as it will require much more time than the simple one.

mlewand:
Guys please let me have a look at it before it's merged, so after R+✅ please ping me in the comments and leave PR open.

Comandeer:
@mlewand you could take a look now.

mlewand:
That being said I started to doubt whether we're in a (desperate) need for this feature.

In the end browser will clean up the resource for us as the document is destroyed.

This feature will be basically blocked by this clipboard ticket (#1563).

f1ames:
I have revisited this PR and as suggested in the https://github.com/ckeditor/ckeditor-dev/pull/1456#issuecomment-362229083:

> That being said I started to doubt whether we're in a (desperate) need for this feature.
>
> In the end browser will clean up the resource for us as the document is destroyed.

it seems this is not a crucial feature. This improvement is in fact invisible to end user (unless one checks for his RAM usage 🙈) and the value it brings may not be clear. It is quite big refactor and might be difficult to get back to it smoothly (especially, that I don't see any issue which describes well what was the reason for this improvement) which means it will require significant amount of time. Let's close it for now.

PR #1463, opened 2018-01-16
Fixed 'uploadWidgetDefinition.onAbort' method is not called when loader is aborted

### LINKED ISSUE #1454: 'uploadWidgetDefinition.onAbort' method is not called when loader is aborted

Bug

According to SDK there are few [`on*`](https://docs.ckeditor.com/ckeditor4/docs/#!/api/CKEDITOR.fileTools.uploadWidgetDefinition-property-onAbort) methods in [uploadWidgetDefinition](https://docs.ckeditor.com/ckeditor4/docs/#!/api/CKEDITOR.fileTools.uploadWidgetDefinition) which are fired when loader status changes.

However, even if `onAbort` is defined in `uploadWidgetDefinition` it is not fired when upload is aborted (widget removed or editor switched to `Source mode` during upload).

### Quick reproduction steps

1. Go to https://sdk.ckeditor.com/samples/fileupload.html.
2. Paste in the browser dev console:
```
CKEDITOR.instances.editor2.widgets.registered.uploadimage.onAbort = function() { console.log( 'abort' ) }
```
3. Drop image into second editor on page.
4. Delete image widget or switch to source mode during upload.

### Proper reproduction steps

`uploadWidgetDefinition` containing `onAbort` method is needed. Easy way may be to [adjust `uploadimage`](https://github.com/ckeditor/ckeditor-dev/blob/393ab9cb89a64a15b34188eb3c3b4e848fe77399/plugins/uploadimage/plugin.js#L57) plugin (by adding `onAbort` method) and then test it on e.g.  `/tests/plugins/uploadwidget/manual/image2` manual test.

### Expected result

`onAbort` method from `uploadWidgetDefinition` is fired when uploading aborted.

### Actual result

`onAbort` method from `uploadWidgetDefinition` is not fired when uploading aborted.

## Other details

* Browser: All (tested on Chrome and FF)
* OS: All
* CKEditor version: -
* Installed CKEditor plugins: `uploadwidget`

jacekbogdanski:

Bug fix.

### This PR contains

- [x] Unit tests
- [x] Manual tests

## What changes did you make?

Changed initialisation of the [upload Widget Definition](https://docs.ckeditor.com/ckeditor4/docs/#!/api/CKEDITOR.fileTools.uploadWidgetDefinition) to include [`onAbort`](https://docs.ckeditor.com/ckeditor4/docs/#!/api/CKEDITOR.fileTools.uploadWidgetDefinition-property-onAbort) method when upload status changes to `abort`.

Closes #1454

### DIFF
--- a/tests/plugins/uploadwidget/uploadwidget.js
+++ b/tests/plugins/uploadwidget/uploadwidget.js
@@ -156,6 +156,48 @@
 			} );
 		},
 
+		'test abort': function() {
+			var bot = this.editorBot,
+				editor = bot.editor;
+
+			var stub = sinon.stub().returns( true );
+
+			addTestUploadWidget( editor, 'testuploadwidget', {
+				onAbort: stub
+			} );
+
+			bot.setData( '', function() {
+				pasteFiles( editor, [ bender.tools.getTestPngFile() ] );
+
+				var loader = editor.uploadRepository.loaders[ 0 ];
+				loader.changeStatus( 'abort' );
+
+				assert.isTrue( stub.calledOnce );
+			} );
+		},
+
+		'test abort can be called only once': function() {
+			var bot = this.editorBot,
+				editor = bot.editor;
+
+			var stub = sinon.stub().returns( true );
+
+			addTestUploadWidget( editor, 'testuploadwidget', {
+				onAbort: stub
+			} );
+
+			bot.setData( '', function() {
+				pasteFiles( editor, [ bender.tools.getTestPngFile() ] );
+
+				var loader = editor.uploadRepository.loaders[ 0 ];
+
+				loader.changeStatus( 'abort' );
+				loader.changeStatus( 'abort' );
+
+				assert.isTrue( stub.calledOnce );
+			} );
+		},
+
 		'test markElement': function() {
 			var element = new CKEDITOR.dom.element( 'p' );
 			CKEDITOR.fileTools.markElement( element, 'widgetName', 1 );
--- a/tests/plugins/uploadwidget/uploadwidget.js
+++ b/tests/plugins/uploadwidget/uploadwidget.js
@@ -171,7 +171,7 @@
 
 				var loader = editor.uploadRepository.loaders[ 0 ];
 
-				editor.widgets.destroyAll();
+				bender.tools.objToArray( editor.widgets.instances )[ 0 ].wrapper = null;
 				editor.editable().findOne( '[data-cke-upload-id="' + loader.id + '"]' ).remove();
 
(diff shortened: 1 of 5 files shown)


jacekbogdanski:
Because we are in state when soon there will be major release I rebased it to the major. If it's an issue I will rebase it to the master.

Comandeer:
Rebased onto latest `master` and changed base.

tests/plugins/uploadwidget/uploadwidget.js
```diff
@@ -156,6 +156,59 @@
...
+				editor.editable().findOne( '[data-cke-upload-id="' + loader.id + '"]' ).remove();
+
+				loader.changeStatus( 'abort' );
+
+				assert.isTrue( stub.calledOnce );
+			} );
+		},
+
+		// (#1454)
+		'test abort can be called only once': function() {
```
Comandeer: Is this test realistic? Is such situation possible to happen? According to our docs, `abort` event is fired only once for every loader.

In fact this test uses private method, `changeStatus`, which does not prevent double setting the same status; yet public `abort` method does it.

Comandeer: Of course it could be fixed with slight change to the listener by using `once` instead of `on`.

tests/plugins/uploadwidget/uploadwidget.js
```diff
@@ -156,6 +156,49 @@
 			} );
 		},
 
+		// (#1454)
+		'test onAbort': function() {
```
Comandeer: This test passes also without the fix. The source of the issue is simple: `onAbort` does not fire only when widget is removed from the editor (due to https://github.com/ckeditor/ckeditor-dev/blob/6b32b1b109d85ff998df79861f59e78c0add17aa/plugins/uploadwidget/plugin.js#L256 ). This test does not remove widget from the editor, therefore it doesn't test anything. The same is true for the other test as well.

This discovery also made me wondering, if the fix could be even more simplified, e.g. moving handling of abort to the beginning of the `update` listener? Not sure about it however.

jacekbogdanski: You are right. When refactoring test I inadvertently removed code responsible to test the issue.

jacekbogdanski:
It looks like moving abort handling at the beginning of the `update` event fixes the issue.

jacekbogdanski:
I spend some time trying to simplify this function. However, it's pretty messed up and it may be better to skip refactoring to avoid regression.

Comandeer:
Rebased onto latest `major` and changed base.

PR #1473, opened 2018-01-18
Bug's workaround for flickering scrollbar in chrome

msamsel:

Bug fix

### This PR contains

- [ ] Unit tests
- [x] Manual tests

## What changes did you make?

- correct styles which might generate [flickering error](https://bugs.chromium.org/p/chromium/issues/detail?id=803045) in Chrome
- bug fix only for classic editor to not break divarea and inline cases when `vw` will be calculated for entire page not for editor

close: #1550

### DIFF
--- a/plugins/easyimage/styles/easyimage.css
+++ b/plugins/easyimage/styles/easyimage.css
@@ -49,7 +49,8 @@ The outline is not a part of the element's dimensions, we have to use a border a
 }
 
 /*
-Overwrite 25% in classic editor to prevent existing this bug https://bugs.chromium.org/p/chromium/issues/detail?id=803045
+Overwrite max-width: 25% with max-width: 25vw to prevent native Chrome bug
+https://bugs.chromium.org/p/chromium/issues/detail?id=803045 (#1550).
 */
 body.cke_editable > .cke_widget_wrapper_easyimage-side {
 	max-width: 25vw;
--- a/plugins/easyimage/styles/easyimage.css
+++ b/plugins/easyimage/styles/easyimage.css
@@ -51,7 +51,7 @@ The outline is not a part of the element's dimensions, we have to use a border a
 /*
 Overwrite 25% in classic editor to prevent existing this bug https://bugs.chromium.org/p/chromium/issues/detail?id=803045
 */
-body.cke_editable .cke_widget_wrapper_easyimage-side {
+body.cke_editable > .cke_widget_wrapper_easyimage-side {
 	max-width: 25vw;
 }
 
--- a/plugins/easyimage/styles/easyimage.css
+++ b/plugins/easyimage/styles/easyimage.css
@@ -48,6 +48,13 @@ The outline is not a part of the element's dimensions, we have to use a border a
 	min-width: 10em;
 }
 
+/*
+Overwrite 25% in classic editor to prevent existing this bug https://bugs.chromium.org/p/chromium/issues/detail?id=803045
+*/
+body.cke_editable .cke_widget_wrapper_easyimage-side {
+	max-width: 25vw;
+}
+
 .easyimage .cke_widget_editable {
 	background-color: #f7f7f7;
 	/* Add border so when caption is focused, blue border does not cause flickering. */
--- a/plugins/easyimage/styles/easyimage.css
+++ b/plugins/easyimage/styles/easyimage.css
@@ -49,7 +49,7 @@ The outline is not a part of the element's dimensions, we have to use a border a
 }
 
 /*
-Overwrite max-width: 25% with max-width: 25vw to prevent native Chrome bug
+Overwrite 'max-width: 25%' with 'max-width: 25vw' to prevent native Chrome bug
 https://bugs.chromium.org/p/chromium/issues/detail?id=803045 (#1550).
 */
 body.cke_editable > .cke_widget_wrapper_easyimage-side {
--- a/tests/plugins/easyimage/manual/scrollbarchromebug.html
+++ b/tests/plugins/easyimage/manual/scrollbarchromebug.html
@@ -1,0 +1,30 @@
+<textarea id="classic">
+	<h1>Apollo 11</h1>
+	<figure class="easyimage">
+		<img alt="CKEditor Sample" src="%BASE_PATH%_assets/logo.png" />
+	</figure>
+	<p><strong>Apollo 11</strong> was the spaceflight that landed the first humans, Americans <a href="http://en.wikipedia.org/wiki/Neil_Armstrong">Neil Armstrong</a> and <a href="http://en.wikipedia.org/wiki/Buzz_Aldrin">Buzz Aldrin</a>, on the Moon on July 20, 1969, at 20:18 UTC. Armstrong became the first to step onto the lunar surface 6 hours later on July 21 at 02:56 UTC.</p>
+	<figure class="easyimage easyimage-side">
+		<img alt="Saturn V carrying Apollo 11" src="%BASE_PATH%_assets/lena.jpg">
+		<figcaption>Saturn V carrying Apollo 11</figcaption>
+	</figure>
+	<p>Armstrong spent about <s>three and a half</s> two and a half hours outside the spacecraft, Aldrin slightly less; and together they collected 47.5 pounds (21.5&nbsp;kg) of lunar material for return to Earth. A third member of the mission, <a href="http://en.wikipedia.org/wiki/Michael_Collins_(astronaut)">Michael Collins</a>, piloted the <a href="http://en.wikipedia.org/wiki/Apollo_Command/Service_Module">command</a> spacecraft alone in lunar orbit until Armstrong and Aldrin returned to it for the trip back to Earth.</p>
+</textarea>
+
+<script>
+	if ( CKEDITOR.env.ie && CKEDITOR.env.version < 11 ) {
+		bender.ignore();
+	}
+
+	getToken( function( token ) {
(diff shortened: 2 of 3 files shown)


tests/plugins/easyimage/manual/scrollbarchromebug.md
```diff
@@ -0,0 +1,13 @@
+@bender-tags: 4.9.0, bug, 932
+@bender-ui: collapsed
+@bender-ckeditor-plugins: sourcearea, wysiwygarea, toolbar, easyimage, link
+@bender-include: ./_helpers/tools.js
```
f1ames: Not sure why this helpers are needed (if for CS think you don't need CS connection here)?

tests/plugins/easyimage/manual/scrollbarchromebug.html
```diff
@@ -0,0 +1,30 @@
...
+	</figure>
+	<p>Armstrong spent about <s>three and a half</s> two and a half hours outside the spacecraft, Aldrin slightly less; and together they collected 47.5 pounds (21.5&nbsp;kg) of lunar material for return to Earth. A third member of the mission, <a href="http://en.wikipedia.org/wiki/Michael_Collins_(astronaut)">Michael Collins</a>, piloted the <a href="http://en.wikipedia.org/wiki/Apollo_Command/Service_Module">command</a> spacecraft alone in lunar orbit until Armstrong and Aldrin returned to it for the trip back to Earth.</p>
+</textarea>
+
+<script>
+	if ( CKEDITOR.env.ie && CKEDITOR.env.version < 11 ) {
+		bender.ignore();
+	}
+
+	getToken( function( token ) {
```
f1ames: Same here, connection to CS is probably not needed in this test.

tests/plugins/easyimage/manual/scrollbarchromebug.md
```diff
@@ -0,0 +1,13 @@
...
+
+----
+1. Move cursor over image.
+
+**Expected:** Nothing unusal happen.
+
+**Unexpected:**
+  * Empty scrollbar shows up.
+  * Scrollbar flicker when cursor is over image.
```
f1ames: > when cursor is **moved** over image.

?

plugins/easyimage/styles/easyimage.css
```diff
@@ -34,7 +34,14 @@ The outline is not a part of the element's dimensions, we have to use a border a
...
 	float: right;
-	max-width: 25%;
+	max-width: 25%
+}
+
+/*
+Overwrite 25% in classic editor to prevent existing this bug https://bugs.chromium.org/p/chromium/issues/detail?id=803045
+*/
+body.cke_editable .cke_widget_wrapper_easyimage-side {
+	max-width: 25vw;
```
f1ames: There are some problems with `vw` unit. It was discussed during #1492 fix (see TP ticket for details). In #1492 we decided to go with `max-width: %` and `min-width: em`, but I'm not sure if it helps in this case (you may test it on #1492).

The problem with `vw` is that the image will not scale if placed inside some other container (like table), so maybe the way to go is to apply `vw` only if image is a direct body child and use `%` in other cases? WDYT?

tests/plugins/easyimage/manual/scrollbarchromebug.html
```diff
@@ -0,0 +1,30 @@
...
+		<img alt="CKEditor Sample" src="%BASE_PATH%_assets/logo.png" />
+	</figure>
+	<p><strong>Apollo 11</strong> was the spaceflight that landed the first humans, Americans <a href="http://en.wikipedia.org/wiki/Neil_Armstrong">Neil Armstrong</a> and <a href="http://en.wikipedia.org/wiki/Buzz_Aldrin">Buzz Aldrin</a>, on the Moon on July 20, 1969, at 20:18 UTC. Armstrong became the first to step onto the lunar surface 6 hours later on July 21 at 02:56 UTC.</p>
+	<figure class="easyimage easyimage-side">
+		<img alt="Saturn V carrying Apollo 11" src="%BASE_PATH%_assets/lena.jpg">
+		<figcaption>Saturn V carrying Apollo 11</figcaption>
+	</figure>
+	<p>Armstrong spent about <s>three and a half</s> two and a half hours outside the spacecraft, Aldrin slightly less; and together they collected 47.5 pounds (21.5&nbsp;kg) of lunar material for return to Earth. A third member of the mission, <a href="http://en.wikipedia.org/wiki/Michael_Collins_(astronaut)">Michael Collins</a>, piloted the <a href="http://en.wikipedia.org/wiki/Apollo_Command/Service_Module">command</a> spacecraft alone in lunar orbit until Armstrong and Aldrin returned to it for the trip back to Earth.</p>
+</textarea>
```
f1ames: I'm for adding `divarea` editor too, so it is visible if something is styled wrong due to applied CSS change.

msamsel:
I tried also to find some good solution which would also cover situation in case of divarea editor. But all ideas which I test was unsuccessful. I test increasing body size, providing `::after` element and styling it, some possibilities to replace float: right/left to some other things.
So it remain only fix for classic editor with widget placed directly in body.

tests/plugins/easyimage/manual/scrollbarchromebug.md
msamsel: Probably I copied entire description from other test and missed it :(

plugins/easyimage/styles/easyimage.css
```diff
@@ -43,6 +43,13 @@ The outline is not a part of the element's dimensions, we have to use a border a
 	max-width: 25%;
 }
 
+/*
+Overwrite 25% in classic editor to prevent existing this bug https://bugs.chromium.org/p/chromium/issues/detail?id=803045
```
f1ames: I would go with something like:

> Overwrite `max-width: 25%` with `max-width: 25vw` to prevent native Chrome bug - https://bugs.chromium.org/p/chromium/issues/detail?id=803045 (#1550).

tests/plugins/easyimage/manual/scrollbarchromebug.html
```diff
@@ -0,0 +1,47 @@
...
+	<p><strong>Apollo 11</strong> was the spaceflight that landed the first humans, Americans <a href="http://en.wikipedia.org/wiki/Neil_Armstrong">Neil Armstrong</a> and <a href="http://en.wikipedia.org/wiki/Buzz_Aldrin">Buzz Aldrin</a>, on the Moon on July 20, 1969, at 20:18 UTC. Armstrong became the first to step onto the lunar surface 6 hours later on July 21 at 02:56 UTC.</p>
+	<figure class="easyimage easyimage-side">
+		<img alt="Saturn V carrying Apollo 11" src="%BASE_PATH%_assets/lena.jpg">
+		<figcaption>Saturn V carrying Apollo 11</figcaption>
+	</figure>
+	<p>Armstrong spent about <s>three and a half</s> two and a half hours outside the spacecraft, Aldrin slightly less; and together they collected 47.5 pounds (21.5&nbsp;kg) of lunar material for return to Earth. A third member of the mission, <a href="http://en.wikipedia.org/wiki/Michael_Collins_(astronaut)">Michael Collins</a>, piloted the <a href="http://en.wikipedia.org/wiki/Apollo_Command/Service_Module">command</a> spacecraft alone in lunar orbit until Armstrong and Aldrin returned to it for the trip back to Earth.</p>
+</div>
+
+<script>
+	if ( CKEDITOR.env.ie && CKEDITOR.env.version < 11 ) {
```
f1ames: 🤔  Shouldn't this test be ignored for all browser but Chrome (as Chrome is the only browser when this bug occurs)?

tests/plugins/easyimage/manual/scrollbarchromebug.md
```diff
@@ -0,0 +1,14 @@
...
+----
+Note: Problem exists only in **Chrome** browsers. All other browsers should work fine and flickering scrollbar shouldn't be present.
+
+1. Move cursor over image in both editors.
+
+**Expected:** Nothing unusal happen in 1st editor. Unfortunatelly bug is not fixed in divarea editor, so flickering should be observable in bottom editor, what is expected behaviour here.
+
+**Unexpected:**
+  * Empty scrollbar shows up.
+  * Scrollbar flicker when cursor is moved over the images.
```
f1ames: I find this test description a little messy. First there is a note about Chrome (making it Chrome only would solve this one - see previous comment). Then **expected** saying that it shouldn't flicker in first editor but should in second. And then **unexpected** saying that it shouldn't flicker. It's a little bit confusing. 

Maybe something like:

> 1.  Move cursor over image in both editors.
>
> **Expected**: Nothing unusual happens, scrollbar is not flickering.
> 
> **Unexpected**: 
> * Empty scrollbar shows up.
> * Scrollbar flicker when cursor is moved over the images.
>
> **Note**: In **Chrome** browsers in second editor, scrollbars flickers as the issue is not fixed for `divarea` > editor (you may add link to upstream issue here).

If the test will be Chrome only you may adjust (shorten) description event more.

tests/plugins/easyimage/manual/scrollbarchromebug.html
msamsel: From the other hand, fix will be available in every browser not only Chrome which use classic editor and easyimage. So theoretically this fix somehow might affect on other browsers styles. In my opinion my be better to test it on each of them to have sure that nothing strange is happen there. That's why there is condition to ignore only browsers not supported by easyimage.

PR #1482, opened 2018-01-22
Balloon toolbar on destroy doesn't destroy its content

### LINKED ISSUE #1477: Balloon toolbar should destroy its content properly

Bug

## Description

When calling [`destroy`](https://github.com/ckeditor/ckeditor-dev/blob/e68fca3fc67a0a0af55ccb467e4b1b617663e10c/plugins/balloontoolbar/plugin.js#L174) on `balloonToolbar` it doesn't invoke destroy on its children elements (`_items`). 

Current situation might cause some issues with eg. `richCombo` which itself doesn't have such method. 

We should add new method `destroy` to `richCombo`, and call it within `balloonToolbar` destroy.

engineering-this:

Bugfix

### This PR contains

- [x] Unit tests
- [x] Manual tests

## What changes did you make?

I've made balloonToolbar `destroy` to call also `destroy` on each of its `_items`. And added `destroy` method to richCombo, which removes all listeners from it.

Closes #1477

### DIFF
--- a/plugins/balloontoolbar/plugin.js
+++ b/plugins/balloontoolbar/plugin.js
@@ -170,24 +170,16 @@
 	};
 
 	/**
-	 * Cleares Balloon Toolbar from each menu items.
-	 *
-	 * @since 4.9.0
+	 * Hides the toolbar, removes it from the DOM, and clears all its items.
 	 */
-	CKEDITOR.ui.balloonToolbar.prototype.clearItems = function() {
+	CKEDITOR.ui.balloonToolbar.prototype.destroy = function() {
 		for ( var key in this._items ) {
 			if ( this._items[ key ].destroy ) {
 				this._items[ key ].destroy();
 			}
 			this.deleteItem( key );
 		}
-	};
 
-	/**
-	 * Hides the toolbar, removes it from the DOM, and clears all its items.
-	 */
-	CKEDITOR.ui.balloonToolbar.prototype.destroy = function() {
-		this.clearItems();
 		this._pointedElement = null;
 		this._view.destroy();
 	};
--- a/plugins/balloontoolbar/plugin.js
+++ b/plugins/balloontoolbar/plugin.js
@@ -131,6 +131,7 @@
 	 * @param {CKEDITOR.ui.button/CKEDITOR.ui.richCombo} element An instance of the UI element.
 	 */
 	CKEDITOR.ui.balloonToolbar.prototype.addItem = function( name, element ) {
+		// console.log( this._items[ name ] );
 		this._items[ name ] = element;
 	};
 
@@ -170,9 +171,24 @@
 	};
 
 	/**
-	 * Hides the toolbar and removes it from the DOM.
+	 * Cleares Balloon Toolbar from each menu items.
+	 *
+	 * @since 4.9.0
+	 */
+	CKEDITOR.ui.balloonToolbar.prototype.clearItems = function() {
+		for ( var key in this._items ) {
+			if ( this._items[ key ].destroy ) {
+				this._items[ key ].destroy();
+			}
+			this.deleteItem( key );
+		}
+	};
+
+	/**
+	 * Hides the toolbar, removes it from the DOM, and clears all its items.
 	 */
 	CKEDITOR.ui.balloonToolbar.prototype.destroy = function() {
+		this.clearItems();
 		this._pointedElement = null;
 		this._view.destroy();
 	};
--- a/plugins/balloontoolbar/plugin.js
+++ b/plugins/balloontoolbar/plugin.js
@@ -131,7 +131,6 @@
 	 * @param {CKEDITOR.ui.button/CKEDITOR.ui.richCombo} element An instance of the UI element.
 	 */
 	CKEDITOR.ui.balloonToolbar.prototype.addItem = function( name, element ) {
-		// console.log( this._items[ name ] );
 		this._items[ name ] = element;
 	};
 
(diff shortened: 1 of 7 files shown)


plugins/balloontoolbar/plugin.js
```diff
@@ -169,9 +170,24 @@
 	};
 
 	/**
-	 * Hides the toolbar and removes it from the DOM.
+	 * Cleares Balloon Toolbar from each menu items.
+	 *
+	 * @since 4.9.0
+	 */
+	CKEDITOR.ui.balloonToolbar.prototype.clearItems = function() {
```
Comandeer: I'm wondering if this method should be public. I don't see any real use-cases for it except of destroying the toolbar.

plugins/richcombo/plugin.js
```diff
@@ -381,6 +383,19 @@ CKEDITOR.plugins.add( 'richcombo', {
...
 					this.setState( CKEDITOR.TRISTATE_DISABLED );
 				}
+			},
+			/**
+			 * Removes all listeners from richCombo element.
+			 *
+			 * @since 4.9.0
+			 */
+			destroy: function() {
+				if ( this._listeners.length ) {
```
Comandeer: IMO this check is totally unnecessary. `forEach` loop would work totally fine with empty array.

tests/plugins/balloontoolbar/manual/destroy.md
```diff
@@ -0,0 +1,15 @@
...
+1. Click on list item.
+1. Click on another list item.
+
+## Expected:
+
+Balloon toolbar should appear next to last clicked list item.
+
+## Unexpected:
+
+1. Console throws an error.
```
Comandeer: If we use console here, it's nice to include opening console inside the procedure. Also it makes this test mobile unfriendly, so it should be ignored in mobile environments.

plugins/balloontoolbar/plugin.js
```diff
@@ -170,9 +170,16 @@
 	};
 
 	/**
-	 * Hides the toolbar and removes it from the DOM.
+	 * Hides the toolbar, removes it from the DOM, and clears all its items.
 	 */
 	CKEDITOR.ui.balloonToolbar.prototype.destroy = function() {
+		for ( var key in this._items ) {
```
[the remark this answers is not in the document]
engineering-this: I think that `for in` is fine for iterating over obj properties. However when pushing them into an array I think `map` would be best choice.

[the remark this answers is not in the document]
engineering-this: Actually I adapted your suggestion in other place. However I think `for in` loop is fine here.

plugins/richcombo/plugin.js
```diff
@@ -96,6 +96,13 @@ CKEDITOR.plugins.add( 'richcombo', {
 				panelDefinition: panelDefinition,
 				items: {}
 			};
+
+			/**
+			 * Array containing richCombos registered listeners.
+			 *
+			 * @private
+			 */
```
Comandeer: There is missing type of this property. It'd be ideal to use dedicated type (see #2383), but some more generic approach would be also enough.

tests/plugins/balloontoolbar/destroy.js
```diff
@@ -0,0 +1,50 @@
...
+/* bender-ckeditor-plugins: balloontoolbar,button,basicstyles,list,richcombo,stylescombo,wysiwygarea,toolbar */
+
+( function() {
+	'use strict';
+
+	bender.editor = {
+	};
+
+	var tests = {
+		'test destroy toolbar with richcombo': function() {
```
Comandeer: This test does not test anything. Array passed to `CKEDITOR.tools.array.every` is empty, therefore test always passes.

tests/plugins/balloontoolbar/destroy.js
```diff
@@ -0,0 +1,50 @@
...
+
+			// Storing items from panel in another variable, because calling destroy() on panel will remove all item references.
+			items = CKEDITOR.tools.array.map( CKEDITOR.tools.objectKeys( this._items ), function( key ) {
+				return panel._items[ key ];
+			} );
+
+			panel.destroy();
+
+			var listenersDeleted = CKEDITOR.tools.array.every( items, function( item ) {
+				return !( item instanceof CKEDITOR.ui.richCombo ) || item._listeners.length === 0;
```
Comandeer: I discovered it by this typo: ~~`item._listeners`~~ instead of `item._.listeners`.

tests/plugins/richcombo/richcombo.js
```diff
@@ -38,5 +38,21 @@ bender.test( {
 			anchorEl = CKEDITOR.document.getById( 'cke_' + combo.id ).findOne( 'a' );
 
 		assert.areEqual( anchorEl.getAttribute( 'aria-haspopup' ), 'listbox' );
+	},
+	'test destroy removes combo listeners': function() {
```
Comandeer: Please add issue reference.

plugins/richcombo/plugin.js
```diff
@@ -96,6 +96,14 @@ CKEDITOR.plugins.add( 'richcombo', {
...
 				items: {}
 			};
+
+			/**
+			 * Array containing richCombos registered listeners.
+			 *
+			 * @type {Array}
+			 * @private
+			 */
+			this._.listeners = [];
```
Comandeer: If we add it to `_` namespace, API docs are not needed. It would be even weird to document only one property of it. Let's just add `listeners` array to the object literal above.

engineering-this: I guess every piece of extra information might be helpful in future especially if it's already written, but I guess this property is so obvious we can drop docs, and move it to the object literal.

PR #1492, opened 2018-01-24
Added 'min-width' for easy image widget

f1ames:

Bug fix

### This PR contains

- [ ] Unit tests
- [x] Manual tests

## What changes did you make?

Added 'min-width' for easy image widget so it does not scale to very tiny sizes when placed in smaller containers like table cells.

Closes: #1553


tests/plugins/easyimage/manual/sideimagebig.html
```diff
@@ -0,0 +1,97 @@
+<h2>Classic editor</h2>
```
mlewand: This test is pretty much broken on IE11. Table resize does not work here,

tests/plugins/easyimage/manual/sideimagebig.md
```diff
@@ -0,0 +1,13 @@
+@bender-tags: 4.9.0, feature, 932
```
mlewand: Both tests are bug tests more than a feature.

tests/plugins/easyimage/manual/sideimagebig.html
f1ames: Looks like it is not the issue with added styles, but basically an issue with IE11 and easyimage widget inside table. Without `min-width` and even `max-width` table is still not resizable.

mlewand: That's an interesting issue, could you extract this? I think that we simply should change manual test concept here to have it solved.

f1ames: Change the test to use smaller image so resizing is possible (but image still scales) and add proper info.

mlewand:
I have rebased the branch onto latest `t/932-2961`.

f1ames:
@mlewand I adjusted tests description, now I think it should fit such cases.

PR #1499, opened 2018-01-25
Added method 'getClientRects()' to CKEDITOR.dom.range

engineering-this:
## What is the purpose of this pull request?

New feature

## Does your PR contain necessary tests?

All patches which change the editor code must include tests. You can always read more
on [PR testing](https://docs.ckeditor.com/ckeditor4/docs/#!/guide/dev_contributing_code-section-tests),
[how to set the testing environment](https://docs.ckeditor.com/ckeditor4/docs/#!/guide/dev_tests) and
[how to create tests](https://docs.ckeditor.com/ckeditor4/docs/#!/guide/dev_tests-section-creating-your-own-test)
in the official CKEditor documentation.

### This PR contains

- [x] Unit tests
- [ ] Manual tests 
This feature will be used in my upcoming PR on balloonPanel and this usage will be covered in manual tests.

## What changes did you make?

I've added `getClientRects` method which uses native method with same name, also implemented fallback for browsers that doesn't support it.

Closes #1498

PR #1526, opened 2018-01-30
Added possibility to attach balloon panel to selection instead of el…

### LINKED ISSUE #1176: Allow attaching Balloon Panel to a range/selection

Feature request

It would be great if the Balloon Panel supported attaching to a range object, so that it could be placed literally anywhere in the text and so on.

In result [balloonPanel.attach](https://docs.ckeditor.com/ckeditor4/docs/#!/api/CKEDITOR.ui.balloonPanel-method-attach) would support first parameter as a `CKEDITOR.dom.range` too.

Currently the limitation is that the balloon can only be attached to the whole element, so selection like that (in case of balloon toolbar) allows us only to position it relatively to the whole element:

image

Instead we'd like to attach it to the center of a selection.

engineering-this:
Note:
~This branch is based on branch from PR #1499 ( t/1498 ).~

New feature

### This PR contains

- [x] Unit tests
- [x] Manual tests

## What changes did you make?

I've added option to pass a selection as first argument of `CKEDITOR.ui.balloonPanel.attach()`. In that case balloonPanel will attach itself to selection instead of element.

Closes #1176


Comandeer:
Please rebase to latest `major` as `CKEDITOR.dom.range.getClientRects` is already merged.

Comandeer:
Rebased onto latest `major`.

plugins/balloonpanel/plugin.js
```diff
@@ -363,6 +408,34 @@
 			};
 
 			return function( element, options ) {
+				if ( element instanceof CKEDITOR.dom.selection ) {
+					var ranges = element.getRanges(),
+						rectList = ranges[ ranges.length - 1 ].getClientRects( true );
+
+					// If selection is fake we have selected widget or table.
+					if ( element.isFake ) {
```
Comandeer: After the quick test in Chrome, it seems that this step could be omitted for widget ranges at all, as the new `rectList` after the loop is the same as the `rectList` before it. We have https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_dom_selection.html#method-isInTable method to check if selection is in table.

IMO it would be even better to refactor this whole code to something like this:

```javascript
if ( element.isFake && element.isInTable() ) {
	// loop and prepare rects
} else {
	// just getClientRects of appropriate range
}
```

After this refactor rectlist will be calculated only once in every situation.

plugins/balloonpanel/plugin.js
```diff
@@ -363,6 +408,34 @@
 			};
 
 			return function( element, options ) {
+				if ( element instanceof CKEDITOR.dom.selection ) {
```
Comandeer: It's no longer `element`, it's `elementOrSelection`. Subtle change, but it will increase the readability of the rest of the code (`element.isFake` vs `elementOrSelection.isFake`).

plugins/balloonpanel/plugin.js
```diff
@@ -424,12 +497,29 @@
...
+					this._adjustElementRect( elementRect, isInline ? allowedRect : editorRect );
+				}
 
 				// The area of the panel.
 				var panelArea = panelWidth * panelHeight,
-					alignments = this._getAlignments( elementRect, panelWidth, panelHeight ),
-					minDifferenceAlignment, alignmentRect, areaDifference;
+					alignments, minDifferenceAlignment, alignmentRect, areaDifference;
+
+				if ( selectionRects ) {
```
Comandeer: Can't this conditional be merged with the previous one? There are divided only by defining some variables.

plugins/balloonpanel/plugin.js
```diff
@@ -734,6 +825,16 @@
 					left: elementRect.right - elementRect.width / 2 - panelWidth + this.triangleMinDistance
 				}
 			};
+
+			if ( requestedAlignments ) {
```
Comandeer: I'd think about conditionally calculating alignments than deleting the unnecessary ones.

OTOH is it really so crucial to return only the subset of alignments? Can't the unnecessary ones be just ignored?

PR #1531, opened 2018-01-30
Added possibility to use custom icons in buttons

### LINKED ISSUE #1530: Possibility to use custom icons in buttons

Feature request

It should be possible to set custom icon image when creating button like:
```
editorInstance.ui.addButton( 'MyBold', {
      label: 'My Bold',
      command: 'bold',
      toolbar: 'basicstyles,1',
      icon: 'icon_path.png',
      iconHidpi: 'hidpi_icon_path.png'
} );
```

f1ames:

New Feature

### This PR contains

- [x] Unit tests
- [x] Manual tests

## What changes did you make?

Added possibility to use custom icons in buttons as described in #1530.

The only problem here was that there already had benn an undocumented `definition.icon` property which was [used to display icon defined by another plugin](https://github.com/ckeditor/ckeditor-dev/blob/7ecc15bc26aef53fadb7f3ec342510ca2d736236/plugins/button/plugin.js#L260).
This caused `definition.icon` property to be more complex and docs get a little more complex too (well, there wasn't any before).
Still, this behaviour has been preserved and plays nicely with the new one. I also added tests covering both behaviours.

Closes #1530.


tests/plugins/button/buttonicon.js
```diff
@@ -0,0 +1,225 @@
+/* bender-tags: editor */
```
Comandeer: There is no test for adding custom icon for standard buttons (e.g. `Bold`). It should be possible by overwriting `button.icon` property before rendering.

tests/plugins/button/buttonicon.js
```diff
@@ -0,0 +1,225 @@
...
+/* bender-ckeditor-plugins: button,toolbar */
+
+var editorCounter = 0,
```
Comandeer: Won't it be better to encapsulate such variables inside IIFE?

tests/plugins/button/buttonicon.js
```diff
@@ -0,0 +1,225 @@
...
+	'test default button icon': function() {
+		resolutionStandard();
+		assertIcon(
+			{ extraPlugins: 'link' },
+			'Link',
+			isDev ? /plugins\/link\/icons\/link\.png/gi : /plugins\/icons\.png/gi
+		);
+	},
+
+	'test default button icon (hidpi)': function() {
```
Comandeer: Hm, personally I'd simplify tests and encapsulate switching between standard/hidpi inside `createIconTest` or something like that.

f1ames: This is problematic due to a fact how icons are registered. Icon is registered on plugin init so when plugin is loaded on standard resolution standard icons are initialized. Then switching to HiDPI and creating new editor (with same plugin) would not reregister the icons so the standard one is still in use.

This is the reason standard an HiDPI tests use different plugins. This is also a reason why generating two tests (standard one and HiDPI) from one editor config doesn't make sense thus auto generation doesn't make much sense (since you have to pass two different set of settings).

I refactored the tests creation to be more automated, however still standard and HiDPI tests are not generated form one set of settings.

f1ames:
😭  Strangely, it should have standard (not HiDPI) set of icons... Probably for built version, setting manually `CKEDITOR.env.hidpi` works different (or not at all) than in dev one.

f1ames:
The cause of failing tests is the way how tests are run with build version, check this comment for more details - https://github.com/ckeditor/ckeditor-dev/pull/1531/commits/b2d9f31f6a35ea41357d1ef7818994cb8b9cb592#diff-6dfbc137c0c9a9951601ec7b03d6033aR17.

f1ames:
Change sample icons color to red.

---

Also I found one issue. After the https://github.com/ckeditor/ckeditor-dev/pull/1531/commits/bc3293fb763acde5f3669228958d919138f3eaae change, where `iconPath` was used as `iconName` for custom icons, it generated invalid CSS class names:

screen shot 2018-02-05 at 17 12 51

as class is based on `iconName` which had `iconPath` value. I updated the code so the `iconPath` and `iconName` are now properly used, see https://github.com/ckeditor/ckeditor-dev/pull/1531/commits/482016379d7d0ce5622c26a0266ff665c353fced.

Comandeer:
Actually this is valid CSS class, that could be [represented by escape sequence](https://www.w3.org/TR/CSS21/syndata.html#value-def-identifier). What's more, it's also [valid HTML class](https://html.spec.whatwg.org/multipage/common-microsyntaxes.html#set-of-space-separated-tokens).

E.g. `trust_me/I'm.engineer` is valid HTML class, which could be represented by `.trust_me\/I\'m\.engineer` in CSS. It's interpreted correctly even in IE8.

So this change wasn't necessary, but at least we have nicer classes now ;)

PR #1518, opened 2018-01-29
Styles system for Easy Image

Comandeer:
## What is the purpose of this pull request?

New feature

## Does your PR contain necessary tests?

All patches which change the editor code must include tests. You can always read more
on [PR testing](https://docs.ckeditor.com/ckeditor4/docs/#!/guide/dev_contributing_code-section-tests),
[how to set the testing environment](https://docs.ckeditor.com/ckeditor4/docs/#!/guide/dev_tests) and
[how to create tests](https://docs.ckeditor.com/ckeditor4/docs/#!/guide/dev_tests-section-creating-your-own-test)
in the official CKEditor documentation.

### This PR contains

- [x] Unit tests
- [x] Manual tests

## What changes did you make?

Rewrite of whole commands system for EI, enabling to add custom styles. Additionaly existing commands (`easyimageFull` and `easyimageSide`) were moved into the new style system.

PR #1555, opened 2018-01-31
Integrate button icons with easyimage styles

f1ames:

Task

### This PR contains

- [x] Unit tests
- [x] Manual tests

## What changes did you make?

This PR integrates #1531 with #1518 adding a proper icon paths passing from easyimage styles config object to button constructor, so that icons are visible in easyimage balloon toolbar.

To work properly it needs both #1531 and #1518 merged (unit test passes as it checks if icons path are properly propagated, however manual test is integrational one and it needs #1531 to work properly).

Now it targets `t/932-3393` branch (#1518 PR), but can be also merged to `t/932-2961` (main easyimage branch ATM) after #1518 is merged.

### DIFF
--- a/tests/plugins/easyimage/customstyleicons.js
+++ b/tests/plugins/easyimage/customstyleicons.js
@@ -17,6 +17,10 @@
 
 	bender.test( {
 		setUp: function() {
+			if ( CKEDITOR.env.ie && CKEDITOR.env.version < 11 ) {
+				assert.ignore();
+			}
+
 			this.addButtonStub = sinon.stub( CKEDITOR.ui.prototype, 'addButton', function( name, definition ) {
 				if ( definition.command === 'easyimageTest' ) {
 					buttonCreated = true;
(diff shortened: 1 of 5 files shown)


Comandeer:
@f1ames could you rebase to the newest `t/932-2961`?

f1ames:
@Comandeer here you go!

tests/plugins/easyimage/customstyleicons.js
```diff
@@ -0,0 +1,73 @@
...
+			easyimage_toolbar: [ 'EasyimageTest' ]
+		};
+
+	bender.test( {
+		setUp: function() {
+			if ( CKEDITOR.env.ie && CKEDITOR.env.version < 11 ) {
+				assert.ignore();
+			}
+
+			this.addButtonStub = sinon.stub( CKEDITOR.ui.prototype, 'addButton', function( name, definition ) {
```
Comandeer: Button stub causes tests to fail on built version of the editor due to the fact that `contextmenu` plugin is loaded and `addContextMenuItems` uses `editor.ui.items` to get info from toolbar's buttons.

Probably spying instead of stubbing will fix the issue.

f1ames: `Spy` doesn't give an easy solution to do this assert same way as with stub (or I am just bad at searching). I just added call to original function in the stub so it does not affect a normal flow in any way.

PR #1559, opened 2018-01-31
Integrate caption with basicstyles plugin

Comandeer:
## What is the purpose of this pull request?

Bug fix

## Does your PR contain necessary tests?

All patches which change the editor code must include tests. You can always read more
on [PR testing](https://docs.ckeditor.com/ckeditor4/docs/#!/guide/dev_contributing_code-section-tests),
[how to set the testing environment](https://docs.ckeditor.com/ckeditor4/docs/#!/guide/dev_tests) and
[how to create tests](https://docs.ckeditor.com/ckeditor4/docs/#!/guide/dev_tests-section-creating-your-own-test)
in the official CKEditor documentation.

### This PR contains

- [x] Unit tests
- [x] Manual tests

## What changes did you make?

I've broaden check for caption (now it's check if an element is a caption or is inside a caption). I've also introduced mechanism to force treating widget as a focused one (via the second parameter to the `_refreshCaption` private method).

Closes #1646.

PR #1560, opened 2018-01-31
Fixed: Empty caption doesn't hide after moving focus to another caption

### LINKED ISSUE #1543: Empty caption doesn't hide after moving focus to another caption with Shift + Tab

Bug, follow up of https://github.com/ckeditor/ckeditor-dev/pull/1175#issuecomment-357923601

1. Open the sample with 2 images with captions.
2. Focus first image's caption.
3. Press <kbd>Tab</kbd>.
4. Select caption of second image with <kbd>Ctrl</kbd> + <kbd>A</kbd>.
5. Press <kbd>Delete</kbd>.
6. Press <kbd>Shift</kbd> + <kbd>Tab</kbd>.

### Expected result

Second image's caption hides.

### Actual result

Second image's empty caption is visible. 

## Other details

**GIF**

bug_cke4

* Browser: All browsers
* OS: All platforms

engineering-this:

Bug fix

### This PR contains

- [x] Unit tests
- [x] Manual tests

## What changes did you make?

The root of this bug was that imagebase didn't update state of all widgets with active capsions. Just the first one.
After my fix editor will find all active widgets and update them when needed.

## Note

I cherry-picked 2 commits from other PR that has not been reviewed yet, but it solves issue with Edge.

Closes #1543


plugins/imagebase/plugin.js
```diff
@@ -554,9 +560,11 @@
...
 					}
 
-					if ( previous && hasWidgetFeature( previous, 'caption' ) ) {
-						previous._refreshCaption( sender );
-					}
+					activeWidgets.forEach( function( item ) {
+						if ( item && hasWidgetFeature( item, 'caption' ) ) {
+							item._refreshCaption( sender );
+						}
+					} );
```
f1ames: It seems that the intention of previous code is to have currently focused (`focused`) and previously focused (`previous`) widget reference to update its captions which makes sense (as you are not able to focus more than one widget having last two is sufficient). However, getting `previous` instance based on `'figcaption[data-cke-caption-active]'` does not work.

I see 3 solutions here:

* One proposed above (the only problem is that it will refresh captions of all widgets with active captions not only those recently focused/selected).
* Stay with `previous` approach, but take the variable out of the scope of listener (so it will be defined inside `setUp` function together with `listeners`). Last thing happening in the listener should be updating it like `previous = focused;` so next time listener is called, `previous` will really point to previously focused widget.
* Third option is getting `previous` like `previous = editor.widgets.getByElement( editor.editable().findOne( 'figcaption.cke_widget_editable_focused[data-cke-caption-active]' ) );`. It is based on the fact that above listener has `9` priority and [`selectionChange` one in widgets which updates `cke_widget_editable_focused` has `10` priority](https://github.com/ckeditor/ckeditor-dev/blob/7a1b01f3cbd7a0ad7c41c3f20a09b5e6de15810b/plugins/widget/plugin.js#L2746) (lower). So when above listener is called, previously focused widget will still have `cke_widget_editable_focused` class and can be obtained this way.

Using `previous` approach seems a little smarter (and more efficient), but I am not sure about it. Asking for opinion, WDYT @mlewand @engineering-this ?

plugins/imagebase/plugin.js
```diff
@@ -541,8 +541,14 @@
 					var path = evt.name === 'blur' ? editor.elementPath() : evt.data.path,
 						sender = path ? path.lastElement : null,
 						focused = getFocusedWidget( editor ),
-						previous = editor.widgets.getByElement(
-							editor.editable().findOne( 'figcaption[data-cke-caption-active]' ) );
+						activeWidgets = [];
+
+					// When there are many active widgets we need to find them find them all and update status of each one.
```
f1ames: Remember about issue referencing in such cases, like:

> // When there are many active widgets we need to find them find them all and update status of each one (#1543).

plugins/imagebase/plugin.js
mlewand: Quite a few propositions @f1ames 😉

> One proposed above (the only problem is that it will refresh captions of all widgets with active captions not only those recently focused/selected).

I'd rather avoid iterating over all the widgets, it's unnecessary and puts a performance footprint on CKE where we can avoid it.

> Stay with previous approach (…) previously focused widget.

It smells pretty much like a global variable :D However since it would be declared in `setup` it would be one variable per editor, which should do a good job keeping track of last, actual caption. So I think there's a potential in this idea.

Maybe with that we would be able to get rid of `[data-cke-caption-active]` attributes all along?

> Third option is getting previous like (…)

I see - it would, yes - but we can't do that kind of changes that relies on such internals of other code (if that can be avoided). So I'm against it.

f1ames: Thanks @mlewand for sharing your views. I fully agree, the first one is not as efficient as it could be and the third one is smart (or rather hacky), tricky and subject to some internal changes. Thus **we should go with the second solution**.

plugins/imagebase/plugin.js
```diff
@@ -557,9 +560,10 @@
 					if ( previous && hasWidgetFeature( previous, 'caption' ) ) {
 						previous._refreshCaption( sender );
 					}
+					previous = focused;
 				}
 
-				listeners.push( editor.on( 'selectionChange', listener , null, null, 9 ) );
```
engineering-this: Not part of this issue, but...

plugins/imagebase/plugin.js
```diff
@@ -535,14 +538,14 @@
 		 */
 		return {
 			setUp: function( editor ) {
-				var listeners = [];
+				var listeners = [],
+					// Changing the scope of previous so it actually stores previously focused element (#1543).
```
f1ames: This comment is understandable only in relation to previous code. As this code is not super complex I would skip this comment here.

tests/plugins/imagebase/integration/easyimage/manual/caption.md
```diff
@@ -0,0 +1,19 @@
+@bender-tags: 4.9.1, bug, 1543
```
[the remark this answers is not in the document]
engineering-this: I cherry-picked 2 commits from other PR that has not been reviewed yet, but it solves issue with Edge

engineering-this:
Added self assignment to remember to revive this PR once #1559 is merged.

PR #1573, opened 2018-02-02
Added style definition docs

### LINKED ISSUE #1567: CKEDITOR.style definition is undocumented

Task

[`CKEDITOR.style`'s constructor](https://docs.ckeditor.com/ckeditor4/latest/api/CKEDITOR_style.html#constructor) takes style definition as an argument, but this type is left undocumented.

For instance look at [`widget.repository.add` method docs](https://docs.ckeditor.com/ckeditor4/latest/api/CKEDITOR_plugins_widget_repository.html#add), note how `widgetDef` param is beautifully documented - we need the same for styles.

Remember that this type should be marked as abstract, as it is not a real type.

jacekbogdanski:

Task

## What changes did you make?

Added missing [style](https://docs.ckeditor.com/ckeditor4/latest/api/CKEDITOR_style.html) definition docs.
Closes #1567

### DIFF
--- a/core/style.js
+++ b/core/style.js
@@ -2120,9 +2120,12 @@ CKEDITOR.tools.extend( CKEDITOR.editor.prototype, {
 
 /**
  * Abstract class describing definition of a style.
- * It is used as one of {@link CKEDITOR.style#constructor}'s arguments.
+ *
+ * This virtual class illustrates the properties that developers can use to define and create
+ * style definitions.
  *
  * Style definition object represents a style as a set of properties describing CSS style rules and HTML attributes.
+ * It also store element type which can be used to provide valid HTML element as a wrapper or replacement for a selection's HTML.
  * The {@link CKEDITOR.style} based on such definition can be applied to and removed from selection
  * through various {@link CKEDITOR.style} methods.
  *
@@ -2143,10 +2146,10 @@ CKEDITOR.tools.extend( CKEDITOR.editor.prototype, {
  * Defines style type.
  *
  * There are three standard style types - {@link CKEDITOR#STYLE_INLINE}, {@link CKEDITOR#STYLE_BLOCK}
- * or {@link CKEDITOR#STYLE_OBJECT}, each one related to the element used in the style rule andq types of
+ * or {@link CKEDITOR#STYLE_OBJECT}, each one related to the element used in the style rule and types of
  * elements to which specific style can be applied.
  *
- * Plugins may define {@link CKEDITOR.style.customHandlers special style handlers} which can be applied in certain situations.
+ * Plugins may define {@link CKEDITOR.style.customHandlers special style handlers} to customize style operations.
  * To use special style handler, `type` property should be set as a name of the style handler, e.g. `widget`.
  *
  * More on style types can be found in the [Style Types section of the Styles guide](#!/guide/dev_styles-section-style-types).
@@ -2155,7 +2158,8 @@ CKEDITOR.tools.extend( CKEDITOR.editor.prototype, {
  */
 
 /**
- * An unique name assigned to a style definition.
+ * An unique name assigned to a style definition. An name is used to differentiate style definitions e.g.
+ * it's used as an label in style combo dropdown when using [Styles Combo](https://ckeditor.com/cke4/addon/stylescombo) plugin.
  *
  *		{ name: 'Special title' }
  *
@@ -2173,7 +2177,7 @@ CKEDITOR.tools.extend( CKEDITOR.editor.prototype, {
  *			}
  *		}
  *
- * @property {Object} attributes
+ * @property {Object.<String, String>} attributes
  */
 
 /**
@@ -2187,11 +2191,13 @@ CKEDITOR.tools.extend( CKEDITOR.editor.prototype, {
 /**
  * A set of properties specifying CSS style rules.
  *
- *		{
- *			color: 'red',
- *			'font-size': '12px',
- *			'font-weight': 'bold'
- *		}
+ * 		{
+ * 			styles: {
+ * 				color: 'red',
+ * 				'font-size': '12px'
+ * 				'font-weight': 'bold'
+ * 			}
+ * 		}
  *
- * @property {Object} styles
+ * @property {Object.<String, String>} styles
  */
--- a/core/style.js
+++ b/core/style.js
@@ -1,4 +1,4 @@
-﻿/**
+/**
  * @license Copyright (c) 2003-2018, CKSource - Frederico Knabben. All rights reserved.
  * For licensing, see LICENSE.md or https://ckeditor.com/legal/ckeditor-oss-license
  */
@@ -136,7 +136,7 @@ CKEDITOR.STYLE_OBJECT = 3;
 	 *
 	 * @class
 	 * @constructor Creates a style class instance.
-	 * @param styleDefinition
+	 * @param {CKEDITOR.style.definition} styleDefinition
 	 * @param variablesValues
 	 */
 	CKEDITOR.style = function( styleDefinition, variablesValues ) {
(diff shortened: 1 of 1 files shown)


f1ames:
@AnnaTomanek could you also take a look before we merge it?

core/style.js
```diff
@@ -2117,3 +2117,81 @@ CKEDITOR.tools.extend( CKEDITOR.editor.prototype, {
  * @cfg {String/Array/Boolean} [stylesSet='default']
  * @member CKEDITOR.config
  */
+
+/**
+ * Abstract class describing definition of a style.
```
mlewand: There should be an information how we treat "abstract" classes like that. Generally just put them for information purpose, while they are not existing in the API.

It's important because otherwise one might be trying to instantiate it.

See how other members do that:

* Balloon Panel: https://github.com/ckeditor/ckeditor-dev/blob/5a7a0f441778413a520ab7deead8ce9aebcbb838/plugins/balloonpanel/plugin.js#L834-L835
* Dialog: https://github.com/ckeditor/ckeditor-dev/blob/5a7a0f441778413a520ab7deead8ce9aebcbb838/plugins/dialog/dialogDefinition.js#L15-L16

core/style.js
```diff
@@ -2117,3 +2117,81 @@ CKEDITOR.tools.extend( CKEDITOR.editor.prototype, {
...
+ * It is used as one of {@link CKEDITOR.style#constructor}'s arguments.
+ *
+ * Style definition object represents a style as a set of properties describing CSS style rules and HTML attributes.
```
mlewand: > (…) properties describing CSS style rules and HTML attributes.

We lost one information here - it could also store element type, which is neither HTML attr nor is it a CSS style.

Though I don't have a clear idea how to write it here 😜

core/style.js
```diff
@@ -2117,3 +2117,81 @@ CKEDITOR.tools.extend( CKEDITOR.editor.prototype, {
...
+ */
+
+/**
+ * Defines style type.
+ *
+ * There are three standard style types - {@link CKEDITOR#STYLE_INLINE}, {@link CKEDITOR#STYLE_BLOCK}
+ * or {@link CKEDITOR#STYLE_OBJECT}, each one related to the element used in the style rule andq types of
+ * elements to which specific style can be applied.
+ *
+ * Plugins may define {@link CKEDITOR.style.customHandlers special style handlers} which can be applied in certain situations.
```
mlewand: > which can be applied in certain situations

How about:

> to customize style operations.

?

My point is that it affects all the operations doable with styles:

* checking
* application
* removal

core/style.js
```diff
@@ -2117,3 +2117,81 @@ CKEDITOR.tools.extend( CKEDITOR.editor.prototype, {
...
+ * To use special style handler, `type` property should be set as a name of the style handler, e.g. `widget`.
+ *
+ * More on style types can be found in the [Style Types section of the Styles guide](#!/guide/dev_styles-section-style-types).
+ *
+ * @property {String/Number} type=CKEDITOR.STYLE_INLINE
+ */
+
+/**
+ * An unique name assigned to a style definition.
```
mlewand: AFAIR this property is used in style combo dropdown (stylescombo plugin) - but double check it 🙂 So you can give it as an example that this label is used there for end users.

core/style.js
```diff
@@ -2117,3 +2117,81 @@ CKEDITOR.tools.extend( CKEDITOR.editor.prototype, {
...
+ * If `style` attribute is present it will be merged with existing {@link CKEDITOR.style.definition#styles} property.
+ *
+ *		{
+ *			attributes: {
+ *				style: 'color: red',
+ *				class: 'link'
+ *			}
+ *		}
+ *
+ * @property {Object} attributes
```
mlewand: We could use new dictionary notation that we agreed on TM some time ago. `@property {Object.<String, String>}` to indicate expected type of keys/values.

core/style.js
```diff
@@ -2117,3 +2117,81 @@ CKEDITOR.tools.extend( CKEDITOR.editor.prototype, {
...
+/**
+ * A set of properties specifying CSS style rules.
+ *
+ *		{
+ *			color: 'red',
+ *			'font-size': '12px',
+ *			'font-weight': 'bold'
+ *		}
+ *
+ * @property {Object} styles
```
mlewand: Same thing regarding `@property` type as for `attributes` property.

core/style.js
```diff
@@ -2117,3 +2117,81 @@ CKEDITOR.tools.extend( CKEDITOR.editor.prototype, {
...
+ *
+ *		{ element: 'h1' }
+ *
+ * @property {String} element
+ */
+
+/**
+ * A set of properties specifying CSS style rules.
+ *
+ *		{
```
mlewand: It's inconsistent with code listings in `attributes` and `element`. You should wrap it with another object and add `styles` property, and then there should be an object that we see here.

core/style.js
f1ames: Maybe something like:

> Style definition object represents a style as a set of properties describing CSS style rules, HTML attributes, style element and name.

or maybe more general one:

> Style definition object represents a style as a set of properties defining element structure, its attributes and styles/CSS properties.

f1ames: And then we could skip next sentence

> It also store element type which can be used to provide valid HTML element as a wrapper or replacement for a selection's HTML.

as it seems a little redundant.

core/style.js
```diff
@@ -2117,3 +2117,87 @@ CKEDITOR.tools.extend( CKEDITOR.editor.prototype, {
...
+ * Plugins may define {@link CKEDITOR.style.customHandlers special style handlers} to customize style operations.
+ * To use special style handler, `type` property should be set as a name of the style handler, e.g. `widget`.
+ *
+ * More on style types can be found in the [Style Types section of the Styles guide](#!/guide/dev_styles-section-style-types).
+ *
+ * @property {String/Number} type=CKEDITOR.STYLE_INLINE
+ */
+
+/**
+ * An unique name assigned to a style definition. An name is used to differentiate style definitions e.g.
```
f1ames: Maybe something a little shorter like:

>  An unique style definition name. It could be used to differentiate definitions, like in Styles Combo plugin dropdown where it represents item labels.

?

core/style.js
```diff
@@ -2117,3 +2117,87 @@ CKEDITOR.tools.extend( CKEDITOR.editor.prototype, {
...
+ *				style: 'color: red',
+ *				class: 'link'
+ *			}
+ *		}
+ *
+ * @property {Object.<String, String>} attributes
+ */
+
+/**
+ * Element type which will be applied to a selection when applying a style. It should be a valid HTML element, for example `span`.
```
f1ames: Not sure, but i thinks one `a` is unnecessary:

> be applied to ~a~ selection when applying a style...

core/style.js
```diff
@@ -2117,3 +2117,87 @@ CKEDITOR.tools.extend( CKEDITOR.editor.prototype, {
...
+ *
+ *		{ element: 'h1' }
+ *
+ * @property {String} element
+ */
+
+/**
+ * A set of properties specifying CSS style rules.
```
f1ames: I would add

> A set of properties specifying CSS style rules of the HTML style element.

so it is consistent with what we have for `attributes`.

f1ames:
@AnnaTomanek could you take a look and if everything is ok give R+ so we can proceed to close this?

mlewand:
@f1ames I wanted to avoid changes on `major` since @AnnaTomanek already reviewed all the docs and we have a freeze there.

I'm fine with either adding it to 4.9.1 or pushing it still into 4.9.0 if it looks good now.

@AnnaTomanek please let us know what will work better for you.

AnnaTomanek:
Sorry for keeping you waiting, I agreed with @mlewand a few days ago that I'd prefer if you just pushed whatever you feel is correct to the `major`, and I'll fix language issues/reword stuff there. Just forgot to mention it here :)

Go ahead and I'll improve whatever lands in `major` when it's there, it's easier for me than "commenting" every minor typo/rewording here.

Issue #1532, opened 2018-01-30
Easyimage integration with undo manager


msamsel:
## Are you reporting a feature request or a bug?

feature

## Description

For the time being EI does not have a good undo integration.
 
For instance, progressbar can cause extra snapshots to be created (as it adds extra elements to the wrapper, and modify width style).

f1ames:
The problem with progress bar was addressed in #1602.

Other problems with undo integrations are:

* When image is dropped/pasted it is focused by default. When during upload focus changes, the Easy Image caption element is hidden which triggers undo snapshot creation.
* There is also an issue that uploaded image needs to undo steps to be removed (so there are two undo steps instead of one).

PR #1602, opened 2018-02-07
Remove progress bar during snapshot creation

f1ames:

Task

### This PR contains

- [x] Unit tests
- [x] Manual tests

## What changes did you make?

Progress bar is now removed from HTML during snapshot creation so it does not pollute the images causing undo manager to think something changes.

### DIFF
--- a/plugins/imagebase/plugin.js
+++ b/plugins/imagebase/plugin.js
@@ -811,6 +811,7 @@
 		 * however it could be used via child classes to apply some more advanced behaviour, like listening to various
 		 * editor events.
 		 *
+		 * @since 4.10.0
 		 * @param {CKEDITOR.editor} editor
 		 */
 		bindEditor: function() {}
@@ -854,7 +855,7 @@
 			parent = this.wrapper.getParent();
 
 		// Hide progressbar during snapshot creation so it does not trigger saving new snapshot
-		// (as content will be changed due to progressbar progress).
+		// (as content will be changed due to progressbar progress) (#1532).
 		undoListeners.push(
 			editor.on( 'beforeUndoImage', function() {
 				if ( !this.wrapper.getParent() ) {
--- a/plugins/imagebase/plugin.js
+++ b/plugins/imagebase/plugin.js
@@ -857,13 +857,12 @@
 		// (as content will be changed due to progressbar progress).
 		undoListeners.push(
 			editor.on( 'beforeUndoImage', function() {
-				if ( this.wrapper.getParent() ) {
-					if ( !parent ) {
-						parent = this.wrapper.getParent();
-					}
-					// Detach wrapper only if it is attached.
-					this.wrapper.remove();
+				if ( !this.wrapper.getParent() ) {
+					// If wrapper is not attached do not try to detach it.
+					return;
 				}
+				parent = parent || this.wrapper.getParent();
+				this.wrapper.remove();
 			}, this )
 		);
 
@@ -879,10 +878,10 @@
 		var remove = this.remove;
 		this.remove = function() {
 			// Remove undo listeners when progress bar is removed.
-			CKEDITOR.tools.array.filter( undoListeners, function( listener ) {
+			CKEDITOR.tools.array.forEach( undoListeners, function( listener ) {
 				listener.removeListener();
-				return false;
 			} );
+			undoListeners.length = 0;
 			remove.call( this );
 		};
 	};
--- a/plugins/imagebase/plugin.js
+++ b/plugins/imagebase/plugin.js
@@ -409,6 +409,7 @@
 						// Progress reporter has only sense if widget is in progress.
 						var progress = new widget.progressReporterType();
 						widget.wrapper.append( progress.wrapper );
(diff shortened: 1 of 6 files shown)


f1ames:
As Easy Image was merged to `major` already, I targeted this PR there too.

f1ames:
This PR does not solve all udno integration problems, but only one in which progress bar HTML (which changes due to tracking upload progress), causes undoManager to save new snapshots, because of changes in HTML (caused by progress bar progress).

See more in #1532.

tests/plugins/easyimage/manual/progressbarundo.md
```diff
@@ -0,0 +1,14 @@
...
+## Easy Image Progressbar Undo
+
+1. Upload some images (using drag and drop, copy/paste, etc).
+2. During image upload manipulate selection (click somewhere, select some editor content, etc).
+
+### Expected
+
+**Saved undo steps** counter should not increment on selection manipulation.
+
+**Important**: Selecting or deselecting widget will create an undo step.
```
Comandeer: Honestly I don't understand these instructions.

First of all, upload is too quick to allow sensible "selection manipulation". Maybe mocking it with slowed progress would be a good idea?

Comandeer: Honestly I don't understand these instructions.

First of all, upload is too quick to allow sensible "selection manipulation". Maybe mocking it with slowed progress would be a good idea?

What's more, how to check if counter does not change, when every selection change will change it?

f1ames: Slowing down an upload makes sense, as network speed throttling is only available on Chrome (and probably you will need 10mb image to test it properly without it).

As for selection change adding snapshot, I'm not sure, tested it on Chrome, FF, Safari, IE11 and it seems pure selection changes does not cause new undo step:

feb-08-2018 11-40-48

Maybe there are some particular cases in which id does and some in which it doesn't?

f1ames: There is a case with selecting/deselecting widget which triggers undo snapshot, but it is already covered in tests description.

f1ames: There are also other cases, like select all. Relaying on undo snapshots count may be misleading. I redesigned the test to check if latest created snapshot contains progressbar.

mlewand:
We won't be able to contain this in the upcomin release. Let's move it to 4.9.1 instead.

plugins/imagebase/plugin.js
```diff
@@ -832,6 +842,44 @@
...
+					// Attach wrapper only if it is detached.
+					parent.append( this.wrapper );
+				}
+			}, this )
+		);
+
+		var remove = this.remove;
+		this.remove = function() {
+			// Remove undo listeners when progress bar is removed.
+			CKEDITOR.tools.array.filter( undoListeners, function( listener ) {
```
Comandeer: Why `filter` is used here? It does not modify original array, so it's left untouched, with all listeners inside. `CKEDITOR.tools.array.forEach` + `undoListener.length = 0` seems like a better alternative.

plugins/imagebase/plugin.js
```diff
@@ -832,6 +842,44 @@
...
 
+	ProgressBar.prototype.bindEditor = function( editor ) {
+		var undoListeners = [],
+			parent = this.wrapper.getParent();
+
+		// Hide progressbar during snapshot creation so it does not trigger saving new snapshot
+		// (as content will be changed due to progressbar progress).
+		undoListeners.push(
+			editor.on( 'beforeUndoImage', function() {
+				if ( this.wrapper.getParent() ) {
```
Comandeer: This function could be simplified by early return if `this.wrapper` does not have a parent.

tests/plugins/imagebase/features/upload.js
```diff
@@ -608,6 +646,56 @@
...
+					fullLoad: false,
+					callback: function( widgets ) {
+						var initialSnapshots = editor.undoManager.snapshots.length,
+							loader = editor.uploadRepository.loaders[ 0 ];
+
+						assert.areSame( 1, widgets.length, 'Widgets count' );
+
+						loader.uploadTotal = 10;
+
+						updateLoader( loader, 2, function() {
```
Comandeer: This whole part of tests is very repetitve. I'd extract it to dedicated helper function.

plugins/imagebase/plugin.js
f1ames: Looks like some leftover after poor copy/paste 😭

Comandeer:
I've rebased to latest major and update tests.

mlewand:
Taking over the review.

f1ames:
I have revisited this PR and it looks like some minor thing. Since Easy Image plugin is publicly available for some significant amount of time already and we didn't have any similar issue reports it is a low priority.

As for the solution itself we can continue with the above (it's 1,5 year old so may need some updates) or try to address this with `undoManager.filter` (#3245) mechanism which was recently introduced.

PR #1604, opened 2018-02-07
Widgets are now correctly destroyed on selection delete

### LINKED ISSUE #1452: Widget 'destroy' event is not fired directly after removing widget

Bug

1. Go to https://codepen.io/f1ames/pen/EopRoP?editors=1010.
1. Open browser dev console (widget destroy event is logged there).
1. Delete one of the widgets.
1. Do some text editing.
1. Switch to `Source mode`.

### Expected result

Widget `destroy` event should be fired right after deleting/removing widget (**step 3**).

### Actual result

Widget destroy event is fired after switching to `Source mode` (**step 5**) - both for widgets present and already deleted from editable area (makes sense for present widgets as they are destroyed upon switching to `Source mode`, not sure about those which has been already removed).

Alternatively, in **step 4** you may change list type (for existing one or just create new one and change its type). After this operation, widget `destroy` event is fired (for the widget removed earlier).

This scenario can also be reproduced on https://sdk.ckeditor.com/samples/captionedimage.html.

## Other details

* Browser: All (checked on Chrome, FF, IE11)
* OS: All
* CKEditor version: `4.4.0` (probably was introduced in `4.3.0` together with widgets)
* Installed CKEditor plugins: `widget`

jacekbogdanski:

Bug fix.

### This PR contains

- [x] Unit tests
- [x] Manual tests

## What changes did you make?

Handled destroying widgets on selection delete for native and fake selections.

Closes #1452

### DIFF
--- a/core/selection.js
+++ b/core/selection.js
@@ -586,6 +586,10 @@
 			var editor = evt.editor,
 				selectedWidgets = editor.widgets.selected;
 
+			if ( editor.readOnly ) {
+				return;
+			}
+
 			for ( var i = 0; i < selectedWidgets.length; i++ ) {
 				selectedWidgets[ i ].destroy( true );
 			}
--- a/core/selection.js
+++ b/core/selection.js
@@ -531,6 +531,9 @@
 					return;
 				}
 
+				// Cache selected widgets.
+				var selectedWidgets = editor.widgets.selected.slice();
+
 				// If haven't found place for caret on the default side,
 				// try to find it on the other side.
 				if ( !( found = range.moveToClosestEditablePosition( evt.selected, right ) ) )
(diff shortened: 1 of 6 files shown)


core/selection.js
```diff
@@ -531,6 +531,9 @@
 					return;
 				}
 
+				// Cache selected widgets.
+				var selectedWidgets = editor.widgets.selected.slice();
```
Comandeer: `editor.widgets` is not always present! Due to not checking at least _several hundreds_ of tests are failing.

jacekbogdanski:
There is a spinoff for cut events (#1607) but probably it could be fixed with this PR. After some thoughts I think we should also fire `destroy` event for any input with selection because it will also delete widgets.

jacekbogdanski:
The issue has been postponed because of its implementation problems and small importance. See #1452 for more information.

Issue #1566, opened 2018-02-01
Integration between widgets and CKEDITOR.style


Comandeer:
## Are you reporting a feature request or a bug?

Feature request

We have special class for widget styles, [`CKEDITOR.style.customHandlers.widget`](https://docs.ckeditor.com/ckeditor4/latest/api/CKEDITOR_style_customHandlers_widget.html), however it doesn't contain whole functionality of basic `CKEDITOR.style` and it's limited only to operating on classes.

It's caused by the fact that internally it uses widget methods to apply and remove styles:

* [`CKEDITOR.plugins.widget#applyStyle`](https://docs.ckeditor.com/ckeditor4/latest/api/CKEDITOR_plugins_widget.html#applyStyle)
* [`CKEDITOR.plugins.widget#removeStyle`](https://docs.ckeditor.com/ckeditor4/latest/api/CKEDITOR_plugins_widget.html#removeStyle)

Such subclass of `CKEDITOR.style` should support all ways to add styles, especially inline styles.

mlewand:
This would be great addition for upcoming [Easy Image plugin](https://github.com/ckeditor/ckeditor-dev/issues/932).

engineering-this:
Self assigning to remember to clean this once we are done with #1674

PR #1681, opened 2018-02-15
Adding inline styles and attributes to widget with applyStyle

### LINKED ISSUE #1674: Adding inline styles to widget
Feature request

This issue is follow up to #1566 

When using 
`CKEDITOR.applyStyle( style )`
 or 
`CKEDITOR.removeStyle( style )`

 inline styles from `CKEDITOR.style` instance should be toggled on `widget.element`.

engineering-this:

Feature

### This PR contains

- [x] Unit tests
- [x] Manual tests

## What changes did you make?

I have extended `widget.applyRemoveStyle` method so when passed `CKEDITOR.style` instance has property 'style' or 'attribute' it will add or remove by those properties to widget.

Closes #1674


engineering-this:
Initially I was supposed to work on adding/removing styles and attributes as separate tickets, but since doing them both looks almost the same I squashed it into one PR.

plugins/widget/plugin.js
```diff
@@ -3078,28 +3078,46 @@
...
+			if ( changed )
+				widget.setData( 'classes', updatedClasses );
+		}
+
+		function toggleStyleAttribute ( element, property, StyleOrAttr, apply ) {
+			apply = apply ? 'set' : 'remove';
+			StyleOrAttr = StyleOrAttr ? 'Style' : 'Attribute';
+			StyleOrAttr = apply + StyleOrAttr;
+
+			for ( var key in property ) {
```
f1ames: Instead of iterating you may use `apply/removeStyles` and `apply/removeAttributes`.

engineering-this: 1. There is no method `removeStyles` only `removeStyle`.
2. Method `setAttributes` takes `object` containing pairs as attribute, while `removeAttributes` takes `array` as argument.

Applying all that will make our logic more complicated than it is right now so I will stay with current approach.

plugins/widget/plugin.js
```diff
@@ -1066,17 +1064,11 @@
...
 		 */
 		checkStyleActive: function( style ) {
-			var classes = getStyleClasses( style ),
-				cl;
-
-			if ( !classes )
-				return false;
+			style = CKEDITOR.tools.copy( style );
+			style.element = this.element.getName();
+			style._.definition.ignoreReadonly = true;
```
f1ames: Here is already an example the modifying copy (`style._.definition.ignoreReadonly`), will also modify source object. See below:

image

plugins/widget/plugin.js
```diff
@@ -1066,17 +1064,11 @@
 		 * @returns {Boolean} Whether the style is applied to this widget.
 		 */
 		checkStyleActive: function( style ) {
-			var classes = getStyleClasses( style ),
-				cl;
-
-			if ( !classes )
-				return false;
+			style = CKEDITOR.tools.copy( style );
```
f1ames: `CKEDITOR.tools.copy` does a shallow copy so it is not completely safe to use it (as changing copy may change source object). There are two solutions here:

* Use `CKEDITOR.tools.clone` which does deep copy.
* Create new style using definition from source one: `new CKEDITOR.style( CKEDITOR.tools.clone( style._.definition ) )` (still we need to clone the definition).

I'm for the second one as cloning whole `CKEDITOR.style` object may be expensive. And if you look on [`CKEDITOR.style` constructor](https://github.com/ckeditor/ckeditor-dev/blob/9409e28dadf1564ec35987bcd888dfe6df23ef0d/core/style.js#L142) it doesn't perform anything fancy or computational heavy so it should be faster than cloning whole object.

plugins/widget/plugin.js
```diff
@@ -3078,28 +3071,46 @@
...
+			if ( changed ) {
+				widget.setData( 'classes', updatedClasses );
+			}
+		}
+
+		toggleStyleAttribute( widget.element, styles, 1, apply );
+		toggleStyleAttribute( widget.element, attributes, 0, apply );
+	}
+
+	function toggleStyleAttribute ( element, property, styleOrAttr, apply ) {
```
f1ames: Having a `Boolean` property with name like `styleOrAttr` is misleading. Without looking at function code is not possible to assume what `styleOrAttr=false` or `styleOrAttr=true` really means. But, it is also quite difficult to come up with proper name in such cases 🤔 

You may go same way as `apply` param - if `true` it means apply/set if not - remove. So here it would be `attributes` - `true` means apply attributes, false - styles (far from perfect IMHO).

The more readable way (but a little bit "inflated" too) would be something like:

```
function toogleAttributes( element, properties, apply ) { 
	toogleProperties( element, properties, apply, 'Attribute' );
}

function toogleStyles( element, properties, apply ) {
	toogleProperties( element, properties, apply, 'Style' );
}

function toogleProperties( element, properties, apply, propertyType ) { ... }
``` 
or actually, you may just stay with `toogleProperties` (skipping `toogleAttributes` and `toggleStyles`) and call it with proper params in the first place.

f1ames: One more thing, `property` param, should be `properties` as it is an object containing multiple properties.

tests/plugins/widget/customstyleinlineandattribute.js
```diff
@@ -0,0 +1,131 @@
...
+			creator: 'inline',
+			config: {
+				extraPlugins: 'testWidget',
+				allowedContent: true
+			}
+		}
+	};
+
+	bender.test( {
+		'test apply remove inline style': function() {
```
f1ames: I would add few tests here checking how it behaves when widget already has some attributes, styles defined. You may check if different attributes are preserved and same ones are replaced, if styles are merged with existing once and same values replaced, etc.

f1ames:
After F2F talk, we came up with two conclusions:
* Style object (it may contain styles and attributes) should be applied to widget wrapper (the outermost wrapper element). Main reason for this is that styles applied to widget element may not work as expected (take `float` for example, which will reposition element inside widget instead of the whole widget).
* When downcasting the widget all styles should be applied to widget element (as wrapper is removed). In order to do so, one idea is to store the merged definition  of all applied styles so it can be easily applied. On upcast, definition should be created based on existing styles/attributes. When merging styles definitions, keep in mind there might be conflicts so styles applied later have priority as they overwrites what was before in case of conflicts.

f1ames:
As there were some doubts if we should store all applied styles as a one, merged object or just as an array of all applied styles, after looking how `CKEDITOR.style` works, going with one object with all applied/removed styles properties merged seems to be a proper approach.

First of all it works the same way as just applying styles to an element. If a style contains property which is already present in the element, the new value overwrites old one of existing property (same will happen when merging newly applied style object to cached style definition).

Two short scenarios just to illustrate how it should work (again based on `CKEDITOR.style`):

### Example 1

```
var style1 = new CKEDITOR.style( { element: 'p', styles: { color: 'red' } } ),
    style2 = new CKEDITOR.style( { element: 'p', styles: { color: 'blue' } } )
    style3 = new CKEDITOR.style( { element: 'p', styles: { color: 'blue', 'font-size': '10px' } } );

// Here we assume that the styles were applied (in order of declaration)
// to current selection/element so we can operate on it.
// It has the following styles: 'style="color: blue; font-size: 10px;"'
editor.removeStyle( s1 ); // Does nothing
editor.removeStyle( s2 ); // Would remove "color: blue" leaving 'style="font-size: 10px"'
editor.removeStyle( s3 ); // Would remove whole styles
```

Of course if we run `editor.removeStyle( s2 );` and then `editor.removeStyle( s3 );`, the second one will have no effect as the style (as a whole) is not present on the selection/element.

### Example 2

Having styles with some common property, like:

```
var style1 = new CKEDITOR.style( { element: 'p', styles: { color: 'blue', 'font-size': '10px' } } ),
    style2 = new CKEDITOR.style( { element: 'p', styles: { color: 'blue', float: 'left' } } );
```
It is possible to only remove one style:

```
editor.removeStyle( s1 ); // Leaves 'style="float: left"'.
editor.removeStyle( s2 ); // Does nothing.

// Or in the reverse order.
editor.removeStyle( s2 ); // Leaves 'style="font-size: 10px"'.
editor.removeStyle( s1 ); // Does nothing.
```

For quick experiments with styles (especially removing) you may use [this codpen](https://codepen.io/f1ames/pen/ddKqpR?editors=1011).

---

One more thing, after F2F talk with @engineering-this, we decide to not reuse [`CKEDITOR.style.checkElementMatch`](https://docs.ckeditor.com/ckeditor4/latest/api/CKEDITOR_style.html#method-checkElementMatch) when checking if style is applied to widget, but go with simpler approch whcih will just compare given style to widget current style cache (which holds current "styling state").

engineering-this:
Currently I'm working on getting styles from `widget.element` and passing them to `widget.wrapper` on upcast and reverse on downcast. There is a case when the widget is generated from `img` and then its linked. When downcasting widget.element is now `a`, so I can't pass styles back to widget that way, because our `source` output will be different than input. Also when such widget would be unlinked it would lose all its styles because of that.

f1ames:
The problem @engineering-this mentioned in https://github.com/ckeditor/ckeditor-dev/pull/1681#issuecomment-368042055 is quite tricky. We were thinking about storing somehow (in e.g. `styleDefiniton.element`) element name (so when it gets linked we still now what was the element), but it won't work in some cases as you can have linked widget on start or paste and then you don't have any element name stored from earlier.

I see two solutions:
1. Find a reliable way to determine which element is a proper `widget.element` which should have styles applied (skipping `a` is not an option ofc as widget can have it in its structure by default).
1. Use styles in a way that utilizes `styleDefinition.element` so even if widget gets linked (or its HTML changes somehow) styles will be applied to the element pointed by styles (btw. it looks like it should work like this as `styleDefiniton.element` plays important role in other places where styles are used).

plugins/widget/plugin.js
```diff
@@ -1066,17 +1064,11 @@
...
+			style = CKEDITOR.tools.copy( style );
+			style.element = this.element.getName();
+			style._.definition.ignoreReadonly = true;
 
-			while ( ( cl = classes.pop() ) ) {
-				if ( !this.hasClass( cl ) )
-					return false;
-			}
-			return true;
+			return CKEDITOR.style.prototype.checkElementMatch.call( style, this.element );
```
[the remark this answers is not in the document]
engineering-this: I'm using `style.prototype` because using it like that would refer to `style.customHandler.widget.checkElementMatch` instead. But after reworking my PR I won't call it.

plugins/widget/plugin.js
```diff
@@ -1055,9 +1054,10 @@
 		 * {@link CKEDITOR.style#checkActive} method instead of using this method directly,
 		 * because unlike style's method, this one does not perform any checks.
```
engineering-this: Not sure if this part should stay or not. Now this method checks much more than just classes.

plugins/widget/plugin.js
```diff
@@ -1067,14 +1067,42 @@
...
+			if ( classes ) {
+				while ( ( item = classes.pop() ) ) {
+					if ( !this.hasClass( item ) )
+						return false;
+				}
+			}
+			if ( attributes ) {
+				for ( item in attributes ) {
+					if ( item !== 'class' && item !== 'style' ) {
+						if ( !( item in styleDefinition.attributes ) ) {
```
engineering-this: Could make two `if`'s as one, but it is easier to read this way.

plugins/widget/plugin.js
```diff
@@ -2901,6 +2944,82 @@
...
+			delete element.attributes.style;
+		}
+		if ( style ) {
+			wrapper.attributes.style = style;
+			styleDefinition.styles = CKEDITOR.tools.parseCssText( style );
+		} else {
+			// If there are no styles remove it to make sure we don't have `styles=''` in our output.
+			delete styleDefinition.styles;
+		}
+		wrapper.attributes[ 'data-style-definition' ] = JSON.stringify( styleDefinition );
```
engineering-this: This might be considered as 'ugly' solution, but I couldn't find better way to store `widget.styleDefinition` before widget is initialized.

engineering-this:
This could possibly close #1566 too. After review I'd like other devs to discuss if we need any further integration between `widget` and `CKEDITOR.style`.

plugins/widget/plugin.js
```diff
@@ -948,6 +952,15 @@
 			 */
 			repository: widgetsRepo,
 
+			/**
+			 * The widgets style definition.
+			 *
+			 * @readonly
+			 * @since: 4.10.0
+			 * @property {CKEDITOR.plugins.widget.styleDefinition}
```
Comandeer: There is no such type.

plugins/widget/plugin.js
Comandeer: At least it should be internal CKE attribute, `[data**-cke**-style-definition]`.

What's more, information about partial styles (classes) are inside widget's data. Maybe we should reuse this mechanism and include all styles there.

plugins/image2/plugin.js
```diff
@@ -932,6 +938,10 @@
 
 				// Image can be wrapped in link <a><img/></a>.
 				image = el.getFirst( 'img' ) || el.getFirst( 'a' ).getFirst( 'img' );
+
+				el.lockedStyle = el.lockedStyle || {};
```
Comandeer: Shouldn't it be bound to the widget, not the element itself?

engineering-this: At this point widget hasn't been initialized and there is no instance to which I could bind this attribute.

Comandeer: But upcast method takes `data` as a second parameter. These locked styles could be treated as a part of widget's data.

plugins/widget/plugin.js
```diff
@@ -2917,9 +3071,10 @@
 			// Clean up and wrap all queued elements.
 			while ( ( toBeWrapped = upcastIterator.toBeWrapped.pop() ) ) {
 				cleanUpWidgetElement( toBeWrapped[ 0 ] );
-				widgetsRepo.wrapElement( toBeWrapped[ 0 ], toBeWrapped[ 1 ] );
-			}
 
+				var wrapper = widgetsRepo.wrapElement( toBeWrapped[ 0 ], toBeWrapped[ 1 ] );
+				handleStylesOnUpcast( toBeWrapped[ 0 ], wrapper );
```
Comandeer: I wonder if it shouldn't be moved into upcast iterator, constructed in `createUpcastIterator`. Thanks to that we'd be able to replace some magic properties (e.g. `lockedStyles`) with reusing widget's data.

engineering-this: I've moved `lockedStyle` to `data` and accessed it from here.

engineering-this:
@Comandeer @mlewand 

>I wonder if styleDefinition also couldn't be moved to widget data, similar to lockedStyle, and whole style handling logic would then be placed inside initial data handler. WDYT

Doing so results in many tests failing
http://localhost:1030/#tests/is:unit,path:/tests/plugins/widget,path:/tests/plugins/image2,path:/tests/plugins/imagebase,path:/tests/plugins/easyimage
I'd rather avoid changing that. Currently I'm removing this attribute right when I don't use it. Putting it in data would affect tests that have hardcoded expected html output.

plugins/widget/plugin.js
```diff
@@ -3071,35 +3224,85 @@
...
+
+		toggleStyleAttribute( widget, styles, 0, apply );
+		toggleStyleAttribute( widget, attributes, 1, apply );
+	}
+
+	// Togles styles or attributes on widget.
+	// @param {CKEDITOR.plugins.widget} Widget instance.
+	// @param {Object} Pairs of inline styles or attributes.
+	// @param {Boolean} whenever to apply or remove style.
+	function toggleStyleAttribute( widget, properties, attribute, apply ) {
```
Comandeer: Parameters names and use of this function is rather unclear. `toggleStyleAttribute` should… well, toggle `[style]` attribute. But it seems that this function is used to apply/remove attributes/styles.

plugins/widget/plugin.js
```diff
@@ -3071,35 +3224,85 @@
...
+				if ( apply ) {
+					if ( !updatedClasses[ cl ] )
+						changed = updatedClasses[ cl ] = 1;
+				} else if ( updatedClasses[ cl ] ) {
 					delete updatedClasses[ cl ];
 					changed = 1;
 				}
 			}
+			if ( changed ) {
+				widget.setData( 'classes', updatedClasses );
```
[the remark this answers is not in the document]
engineering-this: Getting rid of data.classes wouldn't break users custom plugins that extend widget?  I'm not sure how to approach this.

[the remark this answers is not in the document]
Comandeer: I'm quite sure that this data is used only internally by our plugin to know, which styles are applied to the widget. I don't even see any info about `data.classes` in our docs. @mlewand, am I right or I'm missing something?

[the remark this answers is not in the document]
engineering-this: https://docs.ckeditor.com/ckeditor4/latest/api/CKEDITOR_plugins_widget.html#method-setData

I've found only this references to `data.classes`.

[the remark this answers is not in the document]
mlewand: We can't get rid of `data.classes` it has been already exposed. Among all it's used by the `image2` plugin.

Community had already a good time on picking it up, and relying on that. Because of that we should keep it for compatibility reasons.

plugins/widget/plugin.js
```diff
@@ -2180,8 +2225,39 @@
...
+					if ( key !== 'style' && key !== 'class' && key.substring( 0, 4 ).toLowerCase() !== 'data' ) {
+						styleDefinition.attributes[ key ] = attributes[ key ];
+					} else if ( key === 'class' ) {
+						classes = attributes[ key ].split( ' ' );
+						classes = CKEDITOR.tools.array.filter( classes, function( item ) {
+							return item.substring( 0, 3 ) !== 'cke';
+						} );
+						classes = classes.join( ' ' );
+						if ( classes ) {
+							styleDefinition.attributes[ 'class' ] = classes;
```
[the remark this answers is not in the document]
engineering-this: It is, but as `styleDefinition` is meant to hold all widgets styles it should contain also classes. While we can't get rid of `data.classes` it will be duplicated...

plugins/widget/plugin.js
```diff
@@ -1067,15 +1083,32 @@
...
+			}
+
+			if ( attributes && !checkAttributes( attributes, styleDefinition ) ) {
+				return false;
 			}
+
+			if ( styles && !checkStyles( styles, styleDefinition ) ) {
+				return false;
+			}
+
```
engineering-this: This should have much better readability now.

plugins/widget/plugin.js
```diff
@@ -2901,6 +3004,80 @@
 		} );
 	}
 
+	function handleStylesOnDowncast( widget, widgetElement ) {
+		var styledElement;
+
+		if ( widgetElement ) {
```
Comandeer: This could be simplified by reversing conditional and implementing early `return`.

plugins/widget/plugin.js
```diff
@@ -2901,6 +3004,80 @@
 		} );
 	}
 
+	function handleStylesOnDowncast( widget, widgetElement ) {
```
Comandeer: Where are classes handled? I'm wondering if it won't be better to put them also in the same place as other methods of styling.

plugins/widget/plugin.js
```diff
@@ -2901,6 +3004,80 @@
...
+
+		if ( widgetElement ) {
+			if ( widget.styleDefinition && widgetElement.name !== widget.styleDefinition.element ) {
+				styledElement = widgetElement.getFirst( widget.styleDefinition.element );
+			}
+			if ( !styledElement ) {
+				styledElement = widgetElement;
+			}
+
+			styledElement.attributes = styledElement.attributes || {};
```
Comandeer: What is the purpose of this line?

plugins/widget/plugin.js
```diff
@@ -2901,6 +3004,80 @@
...
+
+	function handleStylesOnUpcast( element, wrapper ) {
+		var style = element.attributes.style,
+			lockedStyle = {},
+			outputStyle = {},
+			styleDefinition = {},
+			key;
+
+		if ( element.attributes[ 'data-cke-widget-data' ] ) {
+			lockedStyle = JSON.parse( decodeURIComponent( element.attributes[ 'data-cke-widget-data' ] ) ).lockedStyle || {};
```
Comandeer: Probably it would be safer to delete `lockedStyle` property from widget's data after handling styles as it could pollute it. Maybe it would be even better to move locked styles to dedicated `[data-cke-lockedstyle]` property? 🤔

plugins/widget/plugin.js
```diff
@@ -2901,6 +3004,80 @@
...
+		}
+
+		styleDefinition = styleDefinition || { attributes: {} };
+		styleDefinition.attributes = styleDefinition.attributes || {};
+		styleDefinition.element = element.name;
+
+		for ( key in element.attributes ) {
+			if ( key !== 'style' && key.substring( 0, 4 ).toLowerCase() !== 'data' ) {
```
Comandeer: In this way, some `[data-*]` attributes could be lost.

plugins/widget/plugin.js
```diff
@@ -3071,35 +3249,86 @@
...
+		}
+
+		applyRemoveStylesOrAttributes( widget, styles, 0, apply );
+		applyRemoveStylesOrAttributes( widget, attributes, 1, apply );
+	}
+
+	// Applies or removes widgets styles/attributes.
+	// @param {CKEDITOR.plugins.widget} Widget instance.
+	// @param {Object} Pairs of inline styles or attributes.
+	// @param {Boolean} whenever to apply or remove style.
```
Comandeer: There are four parameters, yet comment describes only three and additionally not in order.

plugins/widget/plugin.js
```diff
@@ -3256,11 +3485,15 @@
 
 			for ( cl in previousClasses ) {
 				// Avoid removing and adding classes again.
-				if ( !( newClasses && newClasses[ cl ] ) )
+				if ( !( newClasses && newClasses[ cl ] ) ) {
 					this.removeClass( cl );
+					addRemoveClassToStyleDef( this, cl, 1 );
```
Comandeer: Shouldn't it be moved into `removeClass`?

plugins/widget/plugin.js
```diff
@@ -3256,11 +3485,15 @@
...
+				}
 			}
-			for ( cl in newClasses )
+			for ( cl in newClasses ) {
 				this.addClass( cl );
+				addRemoveClassToStyleDef( this, cl );
```
Comandeer: Shouldn't it be moved into `addClass`?

plugins/widget/plugin.js
engineering-this: Classes don't need special treatment here. They are already on element just like other attributes.

plugins/widget/plugin.js
engineering-this: Some tests were failing without this, but after other changes to code it's redundant leftover. I'm removing that.

plugins/widget/plugin.js
engineering-this: As we discussed I'm moving `lockedStyle` to `styleDefinition.lockedStyle`.

tests/plugins/image2/internal.js
```diff
@@ -4,6 +4,7 @@
 
 ( function() {
 	'use strict';
+	var fixHtml = image2TestsTools.fixHtml;
```
engineering-this: Linters...

engineering-this:
> In case of captioned images in image2 plugin there is a mismatch between data.classes and styleDefinition.attributes.class. The first one does not contain image value, yet the second one has it always.

Even on major `widget.element` has class 'image' and `widget.data` doesn't. This sounds like a bug since data.classes should represent actual classes?

Comandeer:
It seems like one, but please mind that `.image` class is more a way to identify image widget than to style it (styles are bound directly to `figure`). Therefore it isn't exactly a style. Still: it's a class and should be reflected.

engineering-this:
@Comandeer 
According to this [comment](https://github.com/ckeditor/ckeditor-dev/blob/major/plugins/image2/plugin.js#L481) it is intended. However I don't like this approach especially when this class is in source output.

engineering-this:
>I even wonder if we could make styleDefiniton a dictionary of style definitions, with styles dedicated to elements (parts?) of the widget and styles dedicated for widget's wrapper. In case of aligned image2 it could look like:

I had that in my mind, but then what in case of two elements with same html tag in widget. For that case I am thinking about something like that:
```
wrapper: {
	name: 'div',
	styleDefinition: {},
	0: {
		name: 'p',
		styleDefinition: {}
	},
	1: {
		name: 'p',
		styleDefinition: {},
		0: { some other nested child },
		1: { some other nested child },
	}
}
```

But that looks way too over complicated, and we are basically creating some kind copy of DOM tree.
@mlewand WDYT?

plugins/widget/plugin.js
```diff
@@ -2901,6 +3016,95 @@
...
+		// From now style definition will be only on wrapper.
+		delete element.attributes[ 'data-cke-style-definition' ];
+		delete styleDefinition.lockedStyle;
+
+		if ( style ) {
+			wrapper.attributes.style = style;
+			styleDefinition.styles = CKEDITOR.tools.parseCssText( style );
+		} else {
+			// If there are no styles remove it to make sure we don't have `styles=''` in our output.
+			// delete styleDefinition.styles;
```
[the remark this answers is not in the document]
engineering-this: Oh, I cheated myself here. I was looking for this line in my code and couldn't find it... It took me a minute to remember. I fixed this and amended, but didn't push 😁

engineering-this:
I've updated manual test and added new test. So @mlewand you can take a look on this. In the meantime I will fix this one:

>Easy Image has the same issue as image2: data.classes in newly created widget does not contain .easyimage class, yet styleDefinition – does. Maybe such behavior is hardcoded into widget plugin itself?

plugins/widget/plugin.js
engineering-this: I'm reverting this change because of 

>Easy Image has the same issue as image2: data.classes in newly created widget does not contain .easyimage class, yet styleDefinition – does. Maybe such behavior is hardcoded into widget plugin itself?

`styleDefinition.attributes.class` should be updated on `data`, not on every call of `addClass` or `removeClass`

plugins/widget/plugin.js
engineering-this: Same as https://github.com/ckeditor/ckeditor-dev/pull/1681#discussion_r183987783

awasson:
I see that this issue has been frozen. 

Does that mean that the issue of image tags not being able to take classes from the StylesCombo because IMG becomes IMAGE in the editor has been resolved? It has been an issue with the CKEditor that is bundled with Drupal 8 from day one and I'm keen about resolving it. 

Cheers,
Andrew

Dumluregn:
Hello!

Unfortunately, the issue itself is still valid (#1674). We've just frozen this PR as the work done here wasn't finished two years ago and our codebase changed a lot since then, so it would take a lot of time to just rebase this PR and resolve all the conflicts.

awasson:
Hello @Dumluregn,  

Thanks for the information. I've been following the issue from the point of view coming from Drupal 8 where CKEditor is bundled. There is an issue in the Drupal Issue Queue ([2642808](https://www.drupal.org/project/drupal/issues/2642808)) that goes back to 2015 and it has stalled in anticipation of a resolution from the CKEditor side. 

On Drupal we can preprocess the stylesSet array before rendering the editor so we are able to loop through the img classes and create a duplicate set for the image widget, formatted in such a way that CKEditor can work with them. The following example of this takes a class for an img tag and duplicates it for an image widget. 

Original stylesSet

`"stylesSet":[
   {
      "name":"This Class",
      "element":"img",
      "attributes":{
         "class":"this-class"
      }
   }
]`

StylesSet rendered in CKEditor.
`"stylesSet":[
   {
      "name":"This Class",
      "element":"img",
      "attributes":{
         "class":"this-class"
      }
   },
   {
      "name":"This Class Widget",
      "type":"widget",
      "widget":"image",
      "attributes":{
         "class":"this-class"
      }
   }
]`

I'm not sure if this will help with the native CKEditor but this works within Drupal.

PR #1687, opened 2018-02-16
Removed BOM encoding and Lossless image compression

drsassafras:

Removed BOM Encoding = Bugfix
All files are identical aside from the UTF-BOM encoding flag that has been removed. All files are now UTF, with is a much more standard encoding. BOM encoding can cause a bitwise shift causing strange errors if it is not handled correctly.

Lossless Image Compression

All images are identical, with the exception of size. Google requires this for best search engine rankings, and who wants to transfer more kb if they don't have to?

I have reviewed each file to ensure that they are all identical. This patch does not change the functionality of the editor.

### This PR contains

- [ ] Unit tests
- [X ] Manual tests


Comandeer:
Created follow-up PR in: https://github.com/ckeditor/ckeditor-dev/pull/2320

Issue #1690, opened 2018-02-18
Move objToArray function to bender tools


jacekbogdanski:
## Are you reporting a feature request or a bug?

Feature request

### Expected result

New function `bender.tools.objToArray`.

### Actual result

`objToArray` private function exists inside `imagebase/features/_helpers/tools.js`.

## Other details

The main reason for this change is the `objToArray` function could be used in many unit tests because when we are using the same editor for single test file `editor.widgets.instances` are removed and added for each test case. Editor instance is not destroyed between tests so widget instances have different ids. With this function it's easy to write unit tests using `editor.widgets.instances` regardless of other tests.

Comandeer:
We already have `obj2Array` implementation inside our widget testing tools → https://github.com/ckeditor/ckeditor-dev/blob/9409e28dadf1564ec35987bcd888dfe6df23ef0d/tests/plugins/widget/_helpers/tools.js#L146-L152

Instead of moving it to `bender.tools`, I'd rather update old implementation with the new one as we don't use it anywhere outside widgets.

jacekbogdanski:
After our talk F2F we decided to move `objToArray` to `bender.tools` and refactor widget tools to use new function because we cannot import modules between `_helpers/tools.js`.

PR #1692, opened 2018-02-18
Fixed a bug that was causing certain selection in read only mode to report an error

### LINKED ISSUE #1184: Copy/paste data in read-only mode in IE8-11 results in error

Bug

These are not duplicates but very similar issues:
* https://github.com/ckeditor/ckeditor-dev/issues/449
* https://dev.ckeditor.com/ticket/11825
* https://github.com/ckeditor/ckeditor-dev/issues/888

1. Open the read-only sample
2. Make editor read-only 
3. Select some text in editor (after making it read-only)
4. Press Ctrl+V

### Expected result

Nothing happens

### Actual result

After pressing `Ctrl` error is thrown:

**Error:**
Message: Unspecified error
Line: 2185
URI: ckeditor/code/selection.js
Code:
```
// Select the range.
sel.addRange( nativeRange );
```

## Other details

* Browser: IE8-11
* OS: Windows
* CKEditor version: IE11 - 4.7.1; IE8 - 4.7.0
* Installed CKEditor plugins: full preset was used

### LINKED ISSUE #1632: Unit tests plugins/widget/widgetsintegration are failing on Edge

Failing test

Run suite:

http://tests.ckeditor.test:1030/tests/plugins/widget/widgetsintegration

**Note** test case that was causing all the issues was ignored in 4.9.0. Shuold be fixed with this ticket.

Several tests are failing:

```
test cutting single focused widget with readonly mode
selection was moved to the copybin
Expected: true (boolean)
Actual:   false (boolean)

test pasting a widget with lots of extra markup and mixed HTML case
widget is selected
Expected: [object Object] (object)
Actual:   null (object)

test widget blocking/passing keystrokes
letter X was prevented
Expected: 2 (number)
Actual:   1 (number)

forwarding keys to focused widget
Keydown on widget 1 was caught
Expected: 1 (number)
Actual:   0 (number)

blocking forwarded keys by widget listener
Editor#key listener was not fired
Expected: false (boolean)
Actual:   true (boolean)

all not handled keys are blocked by default on focused widget
Values should be the same.
Expected: 1 (number)
Actual:   0 (number)
```

### LINKED ISSUE #1916: IE9-11 - setting editor to readOnly mode and pressing Delete throws unspecified error

Bug

1. Put attached HTML file on server and run it in a IE9-11 browser. 
**EDIT:** You can also use this [codePen](https://codepen.io/j_swiderski/pen/XqVRxG) if you need online example.
2. Click inside e.g. `Heading 1` and press Delete

[selection_error.zip](https://github.com/ckeditor/ckeditor-dev/files/1938059/selection_error.zip)
### Expected result

Nothing happens

### Actual result

Many [selection-fake-reset](https://docs.ckeditor.com/ckeditor4/latest/guide/dev_errors.html#selection-fake-reset), [selection-fake-reset](https://docs.ckeditor.com/ckeditor4/latest/guide/dev_errors.html#selection-fake-reset) and [range-startcontainer](https://docs.ckeditor.com/ckeditor4/latest/guide/dev_errors.html#range-startcontainer) errors are reported.

Also for IEs: Unspecified error gets thrown in `selection.js` in line `2202`. Code throwing error is `sel.addRange( nativeRange );`

## Other details

According to the reporter:  
> In IE, native selection would throw this error if nativeRange wraps around an invisible node (display: none). 

It might a good tip to investigate.

Please note that problem can be reproduced from CKEditor 4.7.0. so this is rather new issue.

* Browser: IE9-11, Edge
* OS: Windows
* CKEditor version: 4.7.0+

jacekbogdanski:

Bug fix

### This PR contains

- [x] Unit tests
- [x] Manual tests

## What changes did you make?

The first idea was to skip changing selection on keydown events for read-only mode which was little naive. I found out there is a change proposed by @mlewand which works pretty well. In this version, we are selecting widget only if it's not an entire editable for keydown events. I added an additional unit test. 
I'm not sure how to test this case differently than in already existing manual test for [original issue](https://dev.ckeditor.com/ticket/14407) at `tests/core/selection/manual/iecontenteditablefalse`. Let me know if you think there should be an additional manual test.

This PR depends on #1690

Closes #1632, closes #1184, closes #1916

### DIFF
--- a/core/selection.js
+++ b/core/selection.js
@@ -929,6 +929,7 @@
 					ascendant = nonEditableAscendant( sel );
 
 				if ( ascendant ) {
+					// Prevent changing selection when an ascendant is an entire editable (#1632).
 					if ( !ascendant.equals( editable ) ) {
 						sel.selectElement( ascendant );
 					}
--- a/core/selection.js
+++ b/core/selection.js
@@ -923,20 +923,23 @@
 			// 2. After the accomplish of keyboard and mouse events.
 			editable.attachListener( editable, 'selectionchange', checkSelectionChange, editor );
 			editable.attachListener( editable, 'keyup', checkSelectionChangeTimeout, editor );
-			// https://dev.ckeditor.com/ticket/14407 - Don't even let anything happen if the selection is in a non-editable element.
-			editable.attachListener( editable, 'keydown', function( evt ) {
-				var sel = this.getSelection( 1 ),
-					ascendant = nonEditableAscendant( sel );
 
-				if ( ascendant ) {
-					// Prevent changing selection when an ascendant is an entire editable (#1632).
-					if ( !ascendant.equals( editable ) ) {
-						sel.selectElement( ascendant );
+			if ( CKEDITOR.env.ie ) {
+				// https://dev.ckeditor.com/ticket/14407 - Don't even let anything happen if the selection is in a non-editable element.
+				editable.attachListener( editable, 'keydown', function( evt ) {
+					var sel = this.getSelection( 1 ),
+						ascendant = nonEditableAscendant( sel );
+
+					if ( ascendant ) {
+						// Prevent changing selection when an ascendant is an entire editable (#1632).
+						if ( !ascendant.equals( editable ) ) {
+							sel.selectElement( ascendant );
+							evt.data.preventDefault();
+						}
 					}
+				}, editor );
+			}
 
-					evt.data.preventDefault();
-				}
-			}, editor );
 			// Always fire the selection change on focus gain.
 			// On Webkit do this on DOMFocusIn, because the selection is unlocked on it too and
 			// we need synchronization between those listeners to not lost cached editor._.previousActive property
@@ -1024,15 +1027,13 @@
 			}
 
 			function nonEditableAscendant( sel ) {
-				if ( CKEDITOR.env.ie ) {
-					var range = sel.getRanges()[ 0 ],
-						ascendant = range ? range.startContainer.getAscendant( function( parent ) {
-								return parent.type == CKEDITOR.NODE_ELEMENT &&
-									( parent.getAttribute( 'contenteditable' ) == 'false' || parent.getAttribute( 'contenteditable' ) == 'true' );
-							}, true ) : null ;
+				var range = sel.getRanges()[ 0 ],
+					ascendant = range ? range.startContainer.getAscendant( function( parent ) {
+						return parent.type == CKEDITOR.NODE_ELEMENT &&
+							( parent.getAttribute( 'contenteditable' ) == 'false' || parent.getAttribute( 'contenteditable' ) == 'true' );
+					}, true ) : null ;
 
-					return range && ascendant.getAttribute( 'contenteditable' ) == 'false' && ascendant;
-				}
+				return range && ascendant.getAttribute( 'contenteditable' ) == 'false' && ascendant;
 			}
 		} );
 
--- a/core/selection.js
+++ b/core/selection.js
@@ -925,6 +925,10 @@
 			editable.attachListener( editable, 'keyup', checkSelectionChangeTimeout, editor );
 			// https://dev.ckeditor.com/ticket/14407 - Don't even let anything happen if the selection is in a non-editable element.
 			editable.attachListener( editable, 'keydown', function( evt ) {
+				// Prevent changing selection for readonly mode where entire editable is non editable (#1632).
+				if ( this.readOnly ) {
+					return;
+				}
 				var sel = this.getSelection( 1 );
 				if ( nonEditableAscendant( sel ) ) {
 					sel.selectElement( nonEditableAscendant( sel ) );
--- a/core/selection.js
+++ b/core/selection.js
@@ -925,13 +925,14 @@
 			editable.attachListener( editable, 'keyup', checkSelectionChangeTimeout, editor );
 			// https://dev.ckeditor.com/ticket/14407 - Don't even let anything happen if the selection is in a non-editable element.
 			editable.attachListener( editable, 'keydown', function( evt ) {
-				// Prevent changing selection for readonly mode where entire editable is non editable (#1632).
-				if ( this.readOnly ) {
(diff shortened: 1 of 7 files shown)


jacekbogdanski:
Rebased to the latest major.

Comandeer:
Rebased to latest `major`.

core/selection.js
```diff
@@ -925,9 +925,15 @@
 			editable.attachListener( editable, 'keyup', checkSelectionChangeTimeout, editor );
 			// https://dev.ckeditor.com/ticket/14407 - Don't even let anything happen if the selection is in a non-editable element.
 			editable.attachListener( editable, 'keydown', function( evt ) {
```
Comandeer: If it's IE only issue, I'd move it inside appropriate conditional.

core/selection.js
```diff
@@ -1018,15 +1027,13 @@
...
-
-					return range && ascendant.getAttribute( 'contenteditable' ) == 'false' && ascendant;
-				}
+				var range = sel.getRanges()[ 0 ],
+					ascendant = range ? range.startContainer.getAscendant( function( parent ) {
+						return parent.type == CKEDITOR.NODE_ELEMENT &&
+							( parent.getAttribute( 'contenteditable' ) == 'false' || parent.getAttribute( 'contenteditable' ) == 'true' );
+					}, true ) : null ;
+
+				return range && ascendant.getAttribute( 'contenteditable' ) == 'false' && ascendant;
```
mokona88: I'd recommend revise this line. if 'ascendant' is null, ascendant.getAttribute() should throw exception. truthiness check for 'ascendant' should be done before calling getAttribute.

Comandeer:
Rebased onto latest `master`.

core/selection.js
Comandeer: @mokona88 is right – current condition is wrong. Existence of `ascendant` should be checked before checking its attribute.

tests/core/selection/manual/iereadonly.md
```diff
@@ -0,0 +1,15 @@
+@bender-tags: bug, 4.10.0, 1632, selection, widget
```
Comandeer: Please update version tag.

jacekbogdanski:
There is another issue which will be fixed with this PR: #1916

jacekbogdanski:
It doesn't matter which keydown event has been pressed for this issue. Although I added `ctrl` and `delete` key unit tests to explicit cover reported issues. Also, assertions are now more precise.

tests/plugins/widget/widgetsintegration.js
```diff
@@ -673,6 +664,8 @@
 					assert.isTrue( !!getWidgetById( editor, 'w1' ), 'widget has not been deleted' );
 					assert.isFalse( !!editor.getSelection().isFake, 'selection is not faked' );
 					assert.isFalse( !!editor.document.getById( 'cke_copybin' ), 'copybin was removed' );
+
+					this.editor.setReadOnly( false );
```
Comandeer: If some of the assertions above failed, then editor will remain read-only after this test. It's safer to keep it in `tearDown` method.

core/selection.js
```diff
@@ -923,14 +923,23 @@
...
+
+			if ( CKEDITOR.env.ie ) {
+				// https://dev.ckeditor.com/ticket/14407 - Don't even let anything happen if the selection is in a non-editable element.
+				editable.attachListener( editable, 'keydown', function( evt ) {
+					var sel = this.getSelection( 1 ),
+						ascendant = nonEditableAscendant( sel );
+
+					if ( ascendant ) {
+						// Prevent changing selection when an ascendant is an entire editable (#1632).
+						if ( !ascendant.equals( editable ) ) {
```
Comandeer: Why this condition is not joined with the one above?

core/selection.js
```diff
@@ -1018,15 +1027,13 @@
...
-					var range = sel.getRanges()[ 0 ],
-						ascendant = range ? range.startContainer.getAscendant( function( parent ) {
-								return parent.type == CKEDITOR.NODE_ELEMENT &&
-									( parent.getAttribute( 'contenteditable' ) == 'false' || parent.getAttribute( 'contenteditable' ) == 'true' );
-							}, true ) : null ;
-
-					return range && ascendant.getAttribute( 'contenteditable' ) == 'false' && ascendant;
-				}
+				var range = sel.getRanges()[ 0 ],
+					ascendant = range ? range.startContainer.getAscendant( function( parent ) {
```
Comandeer: Actually I find this whole function really weird. We want to get `[contenteditable=false]` ascendant for the range, yet this function does it in super convoluted way… Why can't it be simplified to something like:

```javascript
function nonEditableAscendant( sel ) {
	var range = sel.getRanges()[ 0 ];

	if( !range ) {
		return null;
	}

	return range.startContainer.getAscendant( function( element ) {
		return element.type === CKEDITOR.NODE_ELEMENT && element.getAttribute( 'contenteditable' ) === 'false';
	}, true );
}
```

Should be working perfectly fine.

mokona88: I have similar question when skimming through this code couple months ago. This check
`parent.getAttribute( 'contenteditable' ) == 'false' || parent.getAttribute( 'contenteditable' ) == 'true'` gave me 2 thoughts:
1. this function actually retrieved an ancestor element with attribute 'contentEditable' that is literally defined. getAttribute() would return 'null' if contentEditable is not defined in html attribute, correct? Otherwise, a better approach to attribute is using native element property .isContentEditable. This property is a boolean, and it is inherited from parent element, or can be set if we literally defined attribute 'contentEditable'.
2. the function's name does not reflect its functionality. I'd recommend a small documentation about what the function does. That would provide a context when skimming through codes.

Thank you.

Comandeer: ad. 1) That's not really an issue as every non-editable element in CKEditor's instance has `[contenteditable=false]` directly set.
ad. 2) I think that the name is obvious enough to not include the documentation. However I must agree that it could be better, e.g. `getNonEditableAscendant`. As it's used only in this one place, such change should be safe.

PR #1697, opened 2018-02-20
Added support to run tests in isolation

### LINKED ISSUE #1696: Add option to create new editor instance for each test case using bender.tools.createTestsForEditors

Test feature request

### Expected result

`bender.tools.createTestsForEditors` with additional `isolateTests` flag creates new editor for each test case

### Actual result

`bender.tools.createTestsForEditors` is using the same editor instance for each test case

## Other details

This function already exists inside `plugins/easyimage/_helpers/tools.js` but because I need to use it to test different plugin I thing it's good idea to move it into `bender.tools`.

jacekbogdanski:

New test tools feature

## What changes did you make?

Updated `bender.tools.createTestsForEditors` to allow run tests in isolation by creating new editor instance for each test case. Refactored easyimage tests to use new function.

Closes #1696

### DIFF
--- a/tests/_benderjs/ckeditor/static/tools.js
+++ b/tests/_benderjs/ckeditor/static/tools.js
@@ -1116,7 +1116,7 @@
 									inputTests[ testName ]( bot.editor, bot );
 								} );
 							};
-						} )( testName, editorName, editorCounter++ );
+						} )( testName, editorName, editorCounter );
 					} else {
 						outputTests[ specificTestName ] = ( function( testName, editorName ) {
 							return function() {
--- a/tests/_benderjs/ckeditor/static/tools.js
+++ b/tests/_benderjs/ckeditor/static/tools.js
@@ -1067,9 +1067,9 @@
 		/**
 		 * Multiplies inputTests for every editor.
 		 *
-		 * @param {Object} editorsDefinitions editors definitions.
+		 * @param {Object} editorsNames editors definitions.
 		 * @param {Object} inputTests Tests to apply on every editor.
-		 * @param {Boolean} isolateTests creates new editor instance for each test case.
+		 * @param {Boolean} isolateTests If set to `true` each test is run on new editor instance.
 		 * @returns {Object} Created tests for every editor.
 		 */
 		createTestsForEditors: function( editorsNames, inputTests, isolateTests ) {
--- a/tests/_benderjs/ckeditor/static/tools.js
+++ b/tests/_benderjs/ckeditor/static/tools.js
@@ -1069,9 +1069,10 @@
 		 *
 		 * @param {Object} editorsDefinitions editors definitions.
 		 * @param {Object} inputTests Tests to apply on every editor.
+		 * @param {Boolean} isolateTests creates new editor instance for each test case.
 		 * @returns {Object} Created tests for every editor.
 		 */
-		createTestsForEditors: function( editorsNames, inputTests ) {
+		createTestsForEditors: function( editorsNames, inputTests, isolateTests ) {
 			var outputTests = {},
 				specificTestName,
 				specialMethods = {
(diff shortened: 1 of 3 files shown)


tests/plugins/easyimage/_helpers/tools.js
```diff
@@ -6,35 +6,7 @@ var easyImageTools = ( function() {
 
 	// Force Edge to run every test in new CKEditor's instance.
 	function createTestsForEditors( editors, tests ) {
```
f1ames: This function is now used only in one test in easyimage - `commands.js`, so you may just remove this helper (as it acts as very simpler proxy now) and add `bender.tools.createTestsForEditors( editors, tests, CKEDITOR.env.edge );` call directly in that test.

tests/_benderjs/ckeditor/static/tools.js
```diff
@@ -1103,11 +1104,26 @@
 						throw new Error( 'Test named "' + specificTestName + '" already exists' );
 					}
 
-					outputTests[ specificTestName ] = ( function( testName, editorName ) {
-						return function() {
-							inputTests[ testName ]( bender.editors[ editorName ], bender.editorBots [ editorName ] );
-						};
-					} )( testName, editorName );
+					// We are creating new editor instance to isolate each test case (#1696).
```
f1ames: It works fine, however there could be one optimalization applied. The fact that this method is based on `bender.editors` means that when `isolateTests=true` we create new editor instance for each test base on `bender.editors`.

However, bender itself already created editor instances based on `bender.editors` (it happens before test file is executed). So we end up here with few (ofc depends on `bender.editors` content) not used editor instance. I would suggest alternating this method in a way that for the first iteration it uses already created editor instances and then it creates new ones.

tests/_benderjs/ckeditor/static/tools.js
```diff
@@ -1080,7 +1081,7 @@
 					'setUp': 1,
 					'tearDown': 1
 				},
-				i, editorName;
+				i, editorName, postfix = 1;
```
f1ames: The `postfix` could be something like `editorNumber`/`editorCounter` which is more meaningful.

f1ames:
Strangely, my review comment disappeared somewhere 😬 I will add it as a regular comment.

tests/_benderjs/ckeditor/static/tools.js
```diff
@@ -1103,11 +1104,26 @@
 						throw new Error( 'Test named "' + specificTestName + '" already exists' );
 					}
 
-					outputTests[ specificTestName ] = ( function( testName, editorName ) {
-						return function() {
-							inputTests[ testName ]( bender.editors[ editorName ], bender.editorBots [ editorName ] );
-						};
-					} )( testName, editorName );
+					// We are creating new editor instance to isolate each test case (#1696).
+					if ( isolateTests && editorCounter++ ) {
```
f1ames: `editorCounter++` is called here and then the same `editorCounter++` is called while passing to iife function, which produces editors like `editor2`, `editor4`, `editor6`, etc instead of `editor2`, `editor3`, `editor4`, etc.

tests/_benderjs/ckeditor/static/tools.js
```diff
@@ -1080,15 +1081,15 @@
...
+				i, editorName, editorCounter;
 
 			for ( var method in specialMethods ) {
 				if ( inputTests[ method ] ) {
 					outputTests[ method ] = inputTests[ method ];
 				}
 			}
 
-			for ( i = 0; i < editorsNames.length; i++ ) {
+			for ( i = 0, editorCounter = 0; i < editorsNames.length; i++, editorCounter = 0 ) {
```
f1ames: Looks a little unusual at first sight, but works fine and is quite straightforward and minimalistic approach which is good 👍

f1ames:
Rebased on latest `major`.

mlewand:
I have rebased the branch and ran the test on couple of browsers to see that it works correctly. Since it already has a R+ and was waiting for branch to be unlocked, I'm merging.

PR #1716, opened 2018-02-23
Removed duplicated lang entries from easyimage and justify

### LINKED ISSUE #1715: Alignment translations in Easy Image are duplicates from Justify plugin

Bug

Our policy is to reuse language entries wherever we can.
With Easy Image we added new language strings for alignment options while, in fact, we already have these terms translated in the Justify plugin.

Justify:
```
	center: 'Center',
	left: 'Align Left',
	right: 'Align Right'
```

Easy Image:
```
		alignLeft: 'Align Left',
		alignCenter: 'Align Center',
		alignRight: 'Align Right',
```

Note that we can easily use "Center" instead of "Align Center", too.

### Expected result

If an identical language string appears twice, it should land in `core` and be reused. Make sure we preserve existing translations!

### Actual result

All localizations unnecessarily have an untranslated entry where, in fact, there is often the correct translation already available. We also bloat the code with no good reason at all.

jacekbogdanski:

Task

## What changes did you make?

We had duplicated language entries in easyimage and justify plugins. In the issue description I was asked to reuse justify lang entries in easyimage plugin but it could't properly work because easyimage plugin doesn't depend on justify plugin. I find out we have already defined duplicated entries in `editor.lang.common` property which I used to remove duplications in easyimage and justify plugins.

Closes #1715


AnnaTomanek:
The entries already defined in `common` are different, though?

Easy Image and Justify use "Align Right", "Align Left", "Center".
Common has "Right", "Left", "Center".

Would this be enough?

f1ames:
@AnnaTomanek True, and Easy Image has "Align Center" and Justify has just "Center". But if we don't go with commons there will be an issue which @jacekbogdanski mentioned:

> In the issue description I was asked to reuse justify lang entries in easyimage plugin but it could't properly work because easyimage plugin doesn't depend on justify plugin.

So we may leave it as is, extract to commons (with `Align` prefix) or go with what we have (so entries without `Align` prefix). WDYT @AnnaTomanek ? Which would be to most reasonable?

f1ames:
@AnnaTomanek I noticed what you put in **Expected** of issue description:

> If an identical language string appears twice, it should land in core and be reused. Make sure we preserve existing translations!

So it seems that it will be a good idea to put something like:
```
alignLeft: 'Align Left',
alignRight: 'Align Right'
// We already have ones for center and justify.
```
in the common lang file (just small problem with naming as we already have `alignLeft`/`alignRight` there 🤔).

AnnaTomanek:
It would make most sense if the "Left", "Right", "Center" were just called `left`, `right`, `center` really, because truth is, the entries themselves do not mention align at all...

And then the new ones would be `alignRight` etc.

Would we be able to clean it up this way or is it too much?

jacekbogdanski:
We are not using alignment entries in many places so it shouldn't be a big deal to refactor existing ones to use new `left`, `right`, `center` properties.

jacekbogdanski:
Because of the above reasons my suggestion is to just give up changing names and keep `Align left` `Align Right` as proposed before.

f1ames:
So the proposition is to rename `alignLeft`, `alignRight`, `alignCenter`, `alignJustify` (other like `Top`, `Middle`, `Bottom`, `None` may be done as a follow up to keep things consistent) to `left`, `rigth`, `center`, `justify` as @AnnaTomanek mention and it makes perfect sense:

> It would make most sense if the "Left", "Right", "Center" were just called left, right, center really, because truth is, the entries themselves do not mention align at all...

And then add `alignLeft`, `alignRight`, `alignCenter` entries with "Align Left", "Align Right", "Align Center". And to keep translations consistent with what was before, `easyimage` would use new `alignLeft`, `alignRight`, `alignCenter` entries. And `justify` plugin would use new `alignLeft`, `alignRight` entries and previous `center`, `justify`.

@jacekbogdanski As for meta description for `align*` you may use something like _Left alignment option text used mainly for button tooltips. It contains whole "Align Left" phrase._

> And once more, remember not to lose any existing translations when working on this :)

@jacekbogdanski One important thing, when you rename the lang entry which already exists and has translation in language files, you should also rename entry in language files so langtool will not overwrite it thinking it is missing. Also when creating common `alignLeft` and `alignRight` please reuse lang entries from `justify` plugin for the same reason. So the only lang entry which will not have translations should be `alignCenter`

Issue #1234, opened 2017-11-27
Behavior of changing lists from unordered to ordered and back


msevcenko:
## Are you reporting a feature request or a bug?

Feature request? (consistent but unexpected behavior)

## Provide detailed reproduction steps (if any)

1. Create unordered list with at least two items
2. Put caret to any of the items
3. Press the ordered list button

### Expected result

Most word processing software (incl. MS Word, Google document, Tiny MCE) would turn the whole list from unordered to ordered, so I guess this is what average user would expect.

### Actual result

Only the current list item is turned into one-item ordered list, splitting the list before and/or after the item.

## Other details

* Installed CKEditor plugins: Relates to Link plugin I guess

PR #1725, opened 2018-02-28
Updated changeListType behavior to be more simple & consistent

pajasevi:

This pull request solves issue #1234 - makes changing of a list type more consistent.

This PR does not have any tests.

### This PR contains

- [x] Unit tests
- [ ] Manual tests

## What changes did you make?

`changeListType` method's behavior was unconsistent for the users. New behavior is more aligned with classic editors like MS Word, Google document, Tiny MCE. New behavior can be seen on my screen-cap video [here](https://anonfile.com/a027G1ddbb/new-chaneListType-behavior.mov).

Anyway, I never wrote new behavior for CKEditor, so if I missed anything, I apologize and I will fix it promptly.

As they say: best code is deleted code...

### DIFF
--- a/plugins/list/plugin.js
+++ b/plugins/list/plugin.js
@@ -292,53 +292,15 @@
 		}
 	};
 
-	function changeListType( editor, groupObj, database, listsCreated ) {
-		// This case is easy...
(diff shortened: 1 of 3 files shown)


jacekbogdanski:

PR #1727, opened 2018-03-01
Exposed getAbsoluteClientRect function

### LINKED ISSUE #1724: Expose getAbsoluteRect private function

Feature request

### Expected result

`CKEDITOR.tools.dom.getElementRect` function returns the position of the element on the screen.

### Actual result

`_getAbsoluteRect` function is hidden inside `balloonpanel` plugin.

## Other details

I need to use this function in other place so I think it's good idea to expose it as a public API. This function takes editor and element instances as arguments to calculate element position. It doesn't fit well in `core/element` nor `core/editor` so after some debate with @f1ames we think tools are best place to expose it.

jacekbogdanski:

New feature

### This PR contains

- [x] Unit tests
- [x] Manual tests

## What changes did you make?

Moved `_getAbsoluteRect` balloon toolbar private function to `core/element` as `getAbsoluteClientPosition`

I wasn't sure how to unit test this function because it behaves differently depending of used browser, monitor etc (because it depends on rendered pixels).  A tests would fail either if you open console etc. Because of that I only added manual test, but if you have some idea how to unit test this function, I would be grateful for some tips.

Closes #1724

### DIFF
--- a/core/dom/element.js
+++ b/core/dom/element.js
@@ -493,6 +493,7 @@ CKEDITOR.dom.element.clearMarkers = function( database, element, removeFromDatab
 		 * frame.getClientRect( true );
 		 * ```
 		 *
+		 * @since 4.10.0
 		 * @param {Boolean} [isAbsolute=false] The function will retrieve absolute rectangle of the currect element i.e. including scroll position.
 		 * @returns {Object.<String, Number>} The dimensions of the DOM element.
 		 * @returns {Number} return.top Top offset.
--- a/core/dom/element.js
+++ b/core/dom/element.js
@@ -502,6 +502,8 @@ CKEDITOR.dom.element.clearMarkers = function( database, element, removeFromDatab
 		 * @returns {Number} return.right Right offset.
 		 * @returns {Number} return.height Element height.
 		 * @returns {Number} return.width Element width.
+		 * @returns {Number} return.x X coordinate value. This property is not available in Internet Explorer or Edge browsers.
+		 * @returns {Number} return.y Y coordinate value. This property is not available in Internet Explorer or Edge browsers.
 		 */
 		getClientRect: function( isAbsolute ) {
 			// http://help.dottoro.com/ljvmcrrn.php
@@ -521,6 +523,9 @@ CKEDITOR.dom.element.clearMarkers = function( database, element, removeFromDatab
 			elementRect.top += winGlobalScroll.y;
 			elementRect.left += winGlobalScroll.x;
 
+			elementRect.y += winGlobalScroll.y;
+			elementRect.x += winGlobalScroll.x;
+
 			elementRect.right = elementRect.left + elementRect.width;
 			elementRect.bottom = elementRect.top + elementRect.height;
 
@@ -535,6 +540,9 @@ CKEDITOR.dom.element.clearMarkers = function( database, element, removeFromDatab
 
 				elementRect.top += frameRect.top;
 				elementRect.left += frameRect.left;
+
+				elementRect.x += frameRect.x;
+				elementRect.y += frameRect.y;
 
 				appendParentFramePosition( frame.getWindow().getFrame() );
 			}
--- a/core/dom/element.js
+++ b/core/dom/element.js
@@ -487,7 +487,7 @@ CKEDITOR.dom.element.clearMarkers = function( database, element, removeFromDatab
 		 * As an example you could use this function with {@link CKEDITOR.dom.window#getFrame editor's window frame} to
 		 * calculate the bounding rectangle of the visible area of the editor and the viewport.
 		 * The retrieved bounding rectangle could be used to position elements like toolbars or notifications (elements outside editor)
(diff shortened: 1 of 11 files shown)


jacekbogdanski:
Do you have some advice how to write unit test for this function? As I mentioned in PR description it's not that easy because we don't have good reference point - the position of the rectangle depends on used browser, screen resolution etc.

Comandeer:
After discussion, we decided that it could be hidden behind new, optional `isAbsolute` parameter to `getClientRect`.

What's more, we also decided to drop `editor` parameter as it is used only to get editor's frame. It could be simply replaced by recursively fetching nested frames and documents.

core/dom/element.js
```diff
@@ -480,17 +480,63 @@ CKEDITOR.dom.element.clearMarkers = function( database, element, removeFromDatab
...
+		 * calculate the absolute rectangle of the visible area of the editor's viewport.
+		 * The retrieved absolute rectangle could be used to position elements like toolbars or notifications (elements outside an editor)
+		 * to always keep them inside an editor viewport independently from the scroll position.
+		 *
+		 * ```javascript
+		 * var frame = editor.window.getFrame();
+		 * frame.getClientRect( true );
+		 * ```
+		 *
+		 * @param {Boolean} [isAbsolute=false] The function will retrieve absolute rectangle of the currect element i.e. including scroll position.
```
Comandeer: Please add information when this parameter was added.

core/dom/element.js
```diff
@@ -480,17 +480,63 @@ CKEDITOR.dom.element.clearMarkers = function( database, element, removeFromDatab
...
+		 * @returns {Object.<String, Number>} The dimensions of the DOM element.
+		 * @returns {Number} return.top Top offset.
+		 * @returns {Number} return.bottom Bottom offset.
+		 * @returns {Number} return.left Left offset.
+		 * @returns {Number} return.right Right offset.
+		 * @returns {Number} return.height Element height.
+		 * @returns {Number} return.width Element width.
```
Comandeer: Please add information about returned `x` and `y` properties (as they're even used in test).

tests/core/dom/element/absoluterect.js
```diff
@@ -0,0 +1,45 @@
...
+
+			assert.isNumberInRange( absoluteRect.top, 10, 12, 'top' );
+			assert.isNumberInRange( absoluteRect.bottom, 30, 32, 'bottom' );
+			assert.isNumberInRange( absoluteRect.left, 0, 2, 'left' );
+			assert.isNumberInRange( absoluteRect.right, 15, 17, 'right' );
+			assert.isNumberInRange( absoluteRect.height, 19, 21, 'height' );
+		},
+
+		// (#1724)
+		'test nested iframe element has correct absolute rect': function() {
```
Comandeer: There is no test for deeper nesting (e.g. `iframe` in `iframe`).

core/dom/element.js
```diff
@@ -480,17 +480,72 @@ CKEDITOR.dom.element.clearMarkers = function( database, element, removeFromDatab
...
+		 * The retrieved absolute rectangle could be used to position elements like toolbars or notifications (elements outside an editor)
+		 * to always keep them inside an editor viewport independently from the scroll position.
+		 *
+		 * ```javascript
+		 * var frame = editor.window.getFrame();
+		 * frame.getClientRect( true );
+		 * ```
+		 *
+		 * @since 4.10.0
+		 * @param {Boolean} [isAbsolute=false] The function will retrieve absolute rectangle of the currect element i.e. including scroll position.
```
mlewand: Typo currect -> current. Actually I think we can skip this word all along, it won't degradate it's meaning, as this is a method in element context.

mlewand: > including scroll position

Does it really include the scroll? It is my udnerstanding that it actually skips the scroll, it does not care about scroll.

Say, you have an iframe with 20 000 px of height, and you have scrolltop offset, which is positioned on top of the container. If you scroll the inner frame by 18 000 px, and request for absolute position of a point in top left are of this window, it will return "a little" of pixels :D it **will not** include these 18 000.

image

We need to be very, very careful about the wording it needs to clear and understandable.

Instead I'd suggest complete reword, and say that it will return an offset of element relative to top left corner of a **viewport** (do I recall correctly that it returns to a topmost viewport and not the document?).

mlewand:
Also please rebase the branch, and guys let's focus, this is very small task let's have it done.

core/dom/element.js
Comandeer: > Does it really include the scroll? It is my udnerstanding that it actually skips the scroll, it does not care about scroll.

Probably the better wording would be "including the scroll position of the topmost frame"?

But yeah, the most readable version would be "position relative to the topmost viewport" or just simply "absolute position".

jacekbogdanski:
Sorry, it looks like I forgot to include commit to this PR with valid `x` and `y` unit tests. I would like to wait for the review until tomorrow so I will be able to run tests on Windows.

PR #1733, opened 2018-03-02
 First autocomplete option is now focused when typing

jacekbogdanski:

New feature

### This PR contains

- [x] Unit tests
- [x] Manual tests

## What changes did you make?

Added auto-selection for first autocomplete option on view open.

### DIFF
--- a/tests/plugins/autocomplete/autocomplete.js
+++ b/tests/plugins/autocomplete/autocomplete.js
@@ -31,6 +31,20 @@
 			ac.destroy();
 		},
 
+		'test autocomplete starts with the first item selected': function() {
+			var editor = this.editor, bot = this.editorBot,
+				ac = new CKEDITOR.plugins.autocomplete( this.editor, matchTestCallback, dataCallback );
+
+			bot.setHtmlWithSelection( '' );
+
+			editor.editable().fire( 'keyup', new CKEDITOR.dom.event( {} ) );
+
+			assertViewOpened( ac, true );
+			assert.isTrue( ac.view.getItemById( 1 ).hasClass( 'cke_autocomplete_selected' ) );
+
+			ac.destroy();
+		},
+
 		'test arrow down selects next item': function() {
 			var editor = this.editor, bot = this.editorBot, editable = editor.editable(),
 				ac = new CKEDITOR.plugins.autocomplete( this.editor, matchTestCallback, dataCallback );
@@ -47,6 +61,16 @@
 			assertViewOpened( ac, true );
 			assert.isTrue( ac.view.getItemById( 2 ).hasClass( 'cke_autocomplete_selected' ) );
 
+			editable.fire( 'keydown', new CKEDITOR.dom.event( { keyCode: 40 } ) );
+
+			assertViewOpened( ac, true );
+			assert.isTrue( ac.view.getItemById( 3 ).hasClass( 'cke_autocomplete_selected' ) );
+
+			editable.fire( 'keydown', new CKEDITOR.dom.event( { keyCode: 40 } ) );
+
+			assertViewOpened( ac, true );
+			assert.isTrue( ac.view.getItemById( 1 ).hasClass( 'cke_autocomplete_selected' ) );
+
 			ac.destroy();
 		},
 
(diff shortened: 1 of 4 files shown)


tests/plugins/autocomplete/manual/optionselection.md
```diff
@@ -0,0 +1,15 @@
+@bender-tags: 4.10.0, feature, tp3558
```
mlewand: It's a bug, not a feature.

tests/plugins/autocomplete/manual/optionselection.md
```diff
@@ -0,0 +1,15 @@
...
+2. Type `@`
+
+## Expected
+
+First autocomplete option is selected.
+
+## Unexpected
+
+No option is selected.
+
```
mlewand: There are two empty line at the end (even though GH shows only one), let's reduce it just to one.

tests/plugins/autocomplete/autocomplete.js
```diff
@@ -89,7 +77,6 @@
 
 			editor.editable().fire( 'keyup', new CKEDITOR.dom.event( {} ) );
 
-			editable.fire( 'keydown', new CKEDITOR.dom.event( { keyCode: 40 } ) );
```
mlewand: This change is not related to this issue 🤔

tests/plugins/autocomplete/autocomplete.js
```diff
@@ -65,18 +59,12 @@
...
 			assertViewOpened( ac, true );
-			assert.isTrue( ac.view.getItemById( 2 ).hasClass( 'cke_autocomplete_selected' ) );
-
-			editable.fire( 'keydown', new CKEDITOR.dom.event( { keyCode: 38 } ) );
-
-			assertViewOpened( ac, true );
-
-			assert.isFalse( ac.view.getItemById( 2 ).hasClass( 'cke_autocomplete_selected' ) );
-			assert.isTrue( ac.view.getItemById( 1 ).hasClass( 'cke_autocomplete_selected' ) );
+			assert.isTrue( ac.view.getItemById( 3 ).hasClass( 'cke_autocomplete_selected' ) );
```
mlewand: Could we check another subsequent key up press? Here the selection cycled to the end of list (which is important case to test), we're missing the "regular" go up case.

tests/plugins/autocomplete/autocomplete.js
```diff
@@ -39,18 +39,12 @@
 
 			editor.editable().fire( 'keyup', new CKEDITOR.dom.event( {} ) );
 
-			assertViewOpened( ac, true );
-
-			editable.fire( 'keydown', new CKEDITOR.dom.event( { keyCode: 40 } ) );
-
 			assertViewOpened( ac, true );
 			assert.isTrue( ac.view.getItemById( 1 ).hasClass( 'cke_autocomplete_selected' ) );
```
mlewand: You're asserting that the first item is focused in "test arrow down selects next item". Instead the bug that you have fixed should get a dedicated TC, like "test autocomplete starts with the first item selected".

Same issue happens in the "test arrow up selects previous item" TC, you're testing whether the fist item was focused there too.

tests/plugins/autocomplete/autocomplete.js
jacekbogdanski: After this bug fix I just simplified above test a little. There is no need to pass arrow down keyCode because an option is selected already.

mlewand: Ah, OK, in that case it makes perfect sense - thanks for clarification!

tests/plugins/autocomplete/manual/__template__.html
```diff
@@ -30,6 +26,10 @@ <h1>Lorem ipsum dolor sit amet, consectetur.</h1>
 		}
 	} );
 
+	// The body of the root html element has additional margin-left to display bender manual-sidebar which has fixed position.
+	// Our autcomplete view lives inside root html so with its margin it has invalid position relative to the editor.
+	// We have to change position of our editor relative to the root html by setting position:static ( see style tag ) and move it to the center of the page
+	// so it won't be hidden below sidebar.
```
mlewand: > autcomplete view lives inside root html

Technically it sits in `body` which is not a HTML root element.

mlewand: Still this workaround is way over-complicated. Am I correct to say that simply setting:
```html
<style>
	body {
		margin-left: 0px;
		padding-left: 350px;
	}
</style>
```

Solves the problem? Make sure to still leave a short comment for the reason of the workaround.

tests/plugins/autocomplete/autocomplete.js
```diff
@@ -40,18 +54,22 @@
 			editor.editable().fire( 'keyup', new CKEDITOR.dom.event( {} ) );
 
 			assertViewOpened( ac, true );
+			assert.isTrue( ac.view.getItemById( 1 ).hasClass( 'cke_autocomplete_selected' ) );
```
mlewand: Again, you're asserting here same thing as in "test autocomplete starts with the first item selected" TC. This is not needed.

tests/plugins/autocomplete/autocomplete.js
```diff
@@ -65,19 +83,28 @@
...
```
mlewand: Again, you're asserting here same thing as in "test autocomplete starts with the first item selected" TC. This is not needed.

PR #1739, opened 2018-03-05
 Added option to set completing key codes

jacekbogdanski:

Feature request

### This PR contains

- [x] Unit tests
- [x] Manual tests

## What changes did you make?

Added global configuration and autocomplete options to choose completing key codes. Default key codes are now `enter` and `tab`.

### DIFF
--- a/plugins/autocomplete/plugin.js
+++ b/plugins/autocomplete/plugin.js
@@ -188,6 +188,15 @@
 		this.textWatcher = this.getTextWatcher( textTestCallback );
 
 		/**
+		 * The autocomplete key codes used to finish autocompletion with selected view item.
+		 * The property is using {@link CKEDITOR.config#autocomplete_completingKeyCodes} configuration option as default key codes.
+		 * You can change this property to set individual key codes for plugin instance.
+		 *
+		 * @property {Number[]}
+		 */
+		this.completingKeyCodes = CKEDITOR.config.autocomplete_completingKeyCodes.slice(); // Remove reference to configuration property.
+
+		/**
 		 * Listeners registered by this autocomplete instance.
 		 *
 		 * @private
@@ -427,8 +436,8 @@
 			} else if ( keyCode == 38 ) {
 				this.model.selectPrevious();
 				handled = true;
-			// Enter key.
-			} else if ( keyCode == 13 ) {
+			// Completition keys.
+			} else if ( ~this.completingKeyCodes.indexOf( keyCode ) ) {
 				this.commit();
 				this.textWatcher.unmatch();
 				handled = true;
@@ -1076,4 +1085,14 @@
 	CKEDITOR.plugins.autocomplete = Autocomplete;
 	Autocomplete.view = View;
 	Autocomplete.model = Model;
+
+	/**
+	 * The autocomplete key codes used to finish autocompletion with selected view item.
+	 * This setting will set completing key codes for each autocomplete plugin respectively.
+	 * To change completing key codes individually use {@link CKEDITOR.plugins.autocomplete#completingKeyCodes} plugin property.
+	 *
+	 * @cfg {Number[]} [autocomplete_completingKeyCodes=[9, 13] (9 = tab, 13 = enter, empty array = disabled)]
+	 * @member CKEDITOR.config
+	 */
+	CKEDITOR.config.autocomplete_completingKeyCodes = [ 9, 13 ];
 } )();
--- a/plugins/autocomplete/plugin.js
+++ b/plugins/autocomplete/plugin.js
@@ -155,6 +155,9 @@
 	 * {@link CKEDITOR.plugins.autocomplete.model.item} interface.
 	 */
 	function Autocomplete( editor, textTestCallback, dataCallback ) {
+		var commitKeystroke = CKEDITOR.config.autocomplete_commitKeystroke;
+		commitKeystroke = CKEDITOR.tools.array.isArray( commitKeystroke ) ? commitKeystroke.slice() : commitKeystroke;
+
 		/**
 		 * The editor instance to which this autocomplete is attached (meaning &mdash; on which it listens).
 		 *
@@ -188,13 +191,13 @@
 		this.textWatcher = this.getTextWatcher( textTestCallback );
 
 		/**
-		 * The autocomplete key codes used to finish autocompletion with selected view item.
-		 * The property is using {@link CKEDITOR.config#autocomplete_completingKeyCodes} configuration option as default key codes.
-		 * You can change this property to set individual key codes for plugin instance.
+		 * The autocomplete keystrokes used to finish autocompletion with selected view item.
+		 * The property is using {@link CKEDITOR.config#autocomplete_commitKeystroke} configuration option as default keystrokes.
+		 * You can change this property to set individual keystrokes for plugin instance.
 		 *
-		 * @property {Number[]}
+		 * @property {Number/Number[]}
 		 */
-		this.completingKeyCodes = CKEDITOR.config.autocomplete_completingKeyCodes.slice(); // Remove reference to configuration property.
+		this.commitKeystroke = commitKeystroke;
 
 		/**
 		 * Listeners registered by this autocomplete instance.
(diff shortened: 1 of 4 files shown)


plugins/autocomplete/plugin.js
```diff
@@ -1070,4 +1079,14 @@
 	CKEDITOR.plugins.autocomplete = Autocomplete;
 	Autocomplete.view = View;
 	Autocomplete.model = Model;
+
+	/**
+	 * The autocomplete key codes used to finish autocompletion with selected view item.
+	 * This setting will set completing key codes for each autocomplete plugin respectively.
+	 * To change completing key codes individually use {@link CKEDITOR.plugins.autocomplete#completingKeyCodes} plugin property.
+	 *
+	 * @cfg {Number[]} [autocomplete_completingKeyCodes=[9, 13] (9 = tab, 13 = enter, empty array = disabled)]
```
mlewand: Missing `@since` tag.

mlewand: How about simplifying it to `autocomplete_commitKeystroke`? Somehow better to me, you could do a quick voting on Slack or sth 🙂. And in that case it could be either a `Number` (convienient) or `Number[]`.

jacekbogdanski: Looks like `autocomplete_commitKeystroke` won 😄

tests/plugins/autocomplete/manual/completingkeycodes.md
```diff
@@ -0,0 +1,18 @@
+@bender-tags: 4.10.0, feature, tp3560
+@bender-ui: collapsed
+@bender-ckeditor-plugins: wysiwygarea, toolbar, basicstyles, autocomplete, textmatch
```
mlewand: After rebasing the branch, the `textmatch` plugin will be loaded out of the box by `__template__.html`.

tests/plugins/autocomplete/manual/completingkeycodes.md
```diff
@@ -0,0 +1,18 @@
...
+
+1. Focus the editor.
+2. Type `@`.
+3. Press `arrow down`.
+3. Press `enter`.
```
mlewand: There are two `3.`s - either use subsequent numbers, or just use markdown magic and go with 

```markdown
1. Focus the editor.
1. Type `@`.
1. Press `arrow down`.
1. Press `enter`.
1. Repeat 1-3 and press `tab`.
```

## Result

1. Focus the editor.
1. Type `@`.
1. Press `arrow down`.
1. Press `enter`.
1. Repeat 1-3 and press `tab`.

tests/plugins/autocomplete/manual/completingkeycodes.md
```diff
@@ -0,0 +1,18 @@
...
+4. Repeat 1-3 and press `tab`.
+
+## Expected
+
+Match has been completed for `tab` and `enter` keys.
+
+## Unexpected
+
+Match has not been completed.
+
```
mlewand: Two subsequent end of line chars. We're sticking with just one.

tests/plugins/autocomplete/autocomplete.js
```diff
@@ -97,6 +97,99 @@
 			ac.destroy();
 		},
 
+		'test tab inserts match': function() {
+			var editor = this.editor, bot = this.editorBot, editable = editor.editable(),
```
mlewand: Don't define/declare multiple vars in a single line. Do only one var, per one line - it's a more diff friendly approach.

mlewand: There's no need to store `bot` variable, you're using it only once, and it's very short. The same should be applied to other test cases.

tests/plugins/autocomplete/autocomplete.js
```diff
@@ -97,6 +97,99 @@
...
+			editable.fire( 'keydown', new CKEDITOR.dom.event( { keyCode: 16 } ) ); // SHIFT
+
+			assert.areEqual( '<p>item1</p>', editor.getData() );
+
+			ac.destroy();
+		},
+
+		'test custom keys inserts match (global configuration)': function() {
+			var editor = this.editor, bot = this.editorBot, editable = editor.editable(),
+				configKeyCodes = CKEDITOR.config.autocomplete_commitKeystroke.slice();
```
mlewand: Instead of trickery like that, just use a separate editor for that.

Make one "clean" editor that will be used in the tests that are already created, and add new editor(s) for your custom config. See [Easy Image tests](https://github.com/ckeditor/ckeditor-dev/blob/82801e3/tests/plugins/easyimage/config.js) as an example.

That solution will be much cleaner.

tests/plugins/autocomplete/autocomplete.js
```diff
@@ -97,6 +97,99 @@
...
+			editor.editable().fire( 'keyup', new CKEDITOR.dom.event( {} ) );
+
+			editable.fire( 'keydown', new CKEDITOR.dom.event( { keyCode: 40 } ) ); // ARROW DOWN
+			editable.fire( 'keydown', new CKEDITOR.dom.event( { keyCode: 16 } ) ); // SHIFT
+
+			assert.areEqual( '<p>item1</p>', editor.getData() );
+
+			ac.destroy();
+		},
+
```
mlewand: Lastly I don't see a case that checks that if `autocomplete_commitKeystroke` was changed to something like `config.autocomplete_commitKeystroke = [ x, y ]` (here both are other than `9` which is tab key) then tab key no longer causes the insertion.

plugins/autocomplete/plugin.js
```diff
@@ -150,6 +150,9 @@
 	 * {@link CKEDITOR.plugins.autocomplete.model.item} interface.
 	 */
 	function Autocomplete( editor, textTestCallback, dataCallback ) {
+		var commitKeystroke = CKEDITOR.config.autocomplete_commitKeystroke;
+		commitKeystroke = CKEDITOR.tools.array.isArray( commitKeystroke ) ? commitKeystroke.slice() : commitKeystroke;
```
mlewand: This is a stray line. I see why you used `commitKeystroke` - simply because `CKEDITOR.config.autocomplete_commitKeystroke` is long and you want to make an alias - good decision.

However the above line is stray and it could be inlined to `this.commitKeystroke` initialization, like so:

```js
this.commitKeystroke = CKEDITOR.tools.array.isArray( configKeystroke ) ? configKeystroke.slice() : configKeystroke;
```

And with this change I find `configKeystroke` name to be more appropriate.

plugins/autocomplete/plugin.js
```diff
@@ -182,6 +185,15 @@
 		 */
 		this.textWatcher = this.getTextWatcher( textTestCallback );
 
+		/**
+		 * The autocomplete keystrokes used to finish autocompletion with selected view item.
+		 * The property is using {@link CKEDITOR.config#autocomplete_commitKeystroke} configuration option as default keystrokes.
+		 * You can change this property to set individual keystrokes for plugin instance.
```
mlewand: Praise: nice API docs description, I find this very clear 👍

plugins/autocomplete/plugin.js
```diff
@@ -421,8 +433,8 @@
 			} else if ( keyCode == 38 ) {
 				this.model.selectPrevious();
 				handled = true;
-			// Enter key.
-			} else if ( keyCode == 13 ) {
+			// Completition keys.
+			} else if ( this.commitKeystroke == keyCode || ~CKEDITOR.tools.indexOf( this.commitKeystroke, keyCode ) ) {
```
mlewand: Just use `!== -1` instead of bitwise not operator, we're rarely using these in our codebase.

mlewand: Maybe it would be a better idea to make this.commitKeystroke always an array? So that you could use just `CKEDITOR.tools.indexOf()` here? That would simplify things a bit - just a suggestion, not an important thing.

jacekbogdanski: I was thinking about it earlier but with this change API could be a little inconsistent. Currently we are accepting `Number` and `Number[]` as an keystroke (by configuration) so if we change this property to an array we will be forced to use always `Number[]` with `autocomplete` instance. Not sure if it's big deal. WDYT @mlewand ?

mlewand: Not really, you could still accept `Number/Number[]` in `config.autocomplete_commitKeystroke` - but only `this.commitKeystroke` would always store array (`Number[]`). Meaning that you'd have to normalize it simply during `this.commitKeystroke` initialization.

PR #1747, opened 2018-03-05
Fix refresh problem of balloontolbar in Edge

### LINKED ISSUE #1580: Easy Image button state after applying the style is not refreshed on IE/Edge in inline editors

Bug

1. Open <http://tests.ckeditor.test:1030/tests/plugins/easyimage/manual/balloontoolbar/balloontoolbar> manual test.
2. Use the "Divarea editor".
3. Focus the image.
1. Change type to a "Full Size Image".

### Expected result

"Full Size Image" button in the Balloon Toolbar gets activated, formerly selected side image button is no longer active.

### Actual result

Side image button remains active, full size image is not marked as active.

Widget itself is transformed as expected though.

## Other details

Same thing happens for the "Inline editor". It's not reproducible with "Classic editor" though.

It does look as if our [`forceSelectionCheck` trick](https://github.com/ckeditor/ckeditor-dev/blob/a458f76801f2644497b8aa8e3890281a6deaefa1/plugins/easyimage/plugin.js#L46-L47) was not working.

* Browser: Edge, IE11
* OS: Windows 10
* CKEditor version: 4.8.0

msamsel:

Bug fix

### This PR contains

- [x] Unit tests
- [x] Manual tests

## What changes did you make?

- Edge clears out selection of hidden element in when balloontoolbar is clicked. Now i such situation happen we will try to restore selection in hidden element instead of immediately clearing it out.

close #1580

### DIFF
--- a/plugins/easyimage/plugin.js
+++ b/plugins/easyimage/plugin.js
@@ -131,8 +131,21 @@
 			editor.on( 'afterCommandExec', function( evt ) {
 				if ( getStyleNameFromCommand( evt.data.name, styles ) ) {
 					editor.forceNextSelectionCheck();
-					editor.selectionChange( true );
+					runSelectionChange();
 				}
+
+				function runSelectionChange() {
+					// In div-type editors Edge removes selection from hidden element when balloon toolbar is clicked.
+					// That's why we preserve this selection after command's execution (#1580).
+					if ( CKEDITOR.env.edge && editor.editable().isInline() ) {
+						var range = editor.getSelection().getRanges()[ 0 ];
+						editor.selectionChange( true );
+						range.select();
+					} else {
+						editor.selectionChange( true );
+					}
+				}
+
 			} );
 
 			editor.on( 'beforeCommandExec', function( evt ) {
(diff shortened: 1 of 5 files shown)


msamsel:
I've change a lot. I wasn't able to figure out any nice solution to fix the real issue, so I only cover it with some ugly hack.

More generic approach extracted to #1897.

plugins/easyimage/plugin.js
```diff
@@ -131,8 +131,21 @@
...
+					runSelectionChange();
 				}
+
+				function runSelectionChange() {
+					// In div-type editors Edge removes selection from hidden element when balloon toolbar is clicked.
+					// That's why we preserve this selection after command's execution (#1580).
+					if ( CKEDITOR.env.edge && editor.editable().isInline() ) {
+						var range = editor.getSelection().getRanges()[ 0 ];
+						editor.selectionChange( true );
+						range.select();
```
Comandeer: Couldn't it be done by locking and unlocking selection? 🤔

msamsel: Unfortunately not. It's required to have proper native selection. Locking selection doesn't stop removing selection in Edge browser.

msamsel:
As my fix is only apply to inline editors:
https://github.com/ckeditor/ckeditor-dev/blob/4006ed3828e2bcdc6467139bcaec85a41aeab2a2/plugins/easyimage/plugin.js#L140
I've ignore the test for non inline editors.

PR #1748, opened 2018-03-06
Add onHide doc

sunnyone:

Add lacking documentation. The "onHide" property of dialogDefinition is used (for example, plugins/link/dialogs/anchor.js) but it's not described in dialogDefinition.js.

No. This PR add a comment only.

## What changes did you make?

Add a doc comment to dialogDefinition.js.

### DIFF
--- a/plugins/dialog/dialogDefinition.js
+++ b/plugins/dialog/dialogDefinition.js
@@ -129,6 +129,12 @@
  */
 
 /**
+ * The function to execute when the dialog is hidden (executed every time the dialog is closed).
+ *
+ * @property {Function} onHide
+ */
+
+/**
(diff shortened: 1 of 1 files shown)


plugins/dialog/dialogDefinition.js
```diff
@@ -128,6 +128,12 @@
  * @property {Function} onShow
  */
 
+/**
+ * The function to execute when the dialog is hidden (executed every time the dialog is closed).
```
mlewand: This could be simplified down to:

> The function executed every time the dialog is closed.

I'll correct this while merging.

mlewand:
Merged with 011af640c - it will be published in 4.10.0. Thank you!

PR #1760, opened 2018-03-07
Added token polling in CKEditor Cloud Services connector plugin

### LINKED ISSUE #1755: Cloud Services plugin should download token on its own
The names of configuration options for Cloud Services are wrong.

## CKEditor 5

* (cloudServices.)tokenUrl 
* (cloudServices.)uploadUrl

both names are perfect as they precisely describe what kind of data is expected there and what is its purpose: https://docs.ckeditor.com/ckeditor5/latest/api/module_cloudservices_cloudservices-CloudServicesConfig.html 

## CKEditor 4

* cloudServices_token : String
~The token used for CKEditor Cloud Services authentication.~

The explanation is wrong. What the user is supposed to be set there is the token end point URL (URL that will issue tokens for users). Just copy the description from CKEditor 5. 

* cloudServices_url : String
The endpoint URL for CKEditor Cloud Services uploads. Note that this is **the URL only for file uploads**. CS exposes more URLs. That's why it should be named "uploadUrl" not just "url".

So:

* cloudServices_tokenUrl
* cloudServices_uploadUrl 

are the valid names for configuration options.

mlewand:

New feature

### This PR contains

- [x] Unit tests
- [x] Manual tests (existing tests for EI and CS has to be used)

## What changes did you make?

This PR:

* Changes the way CS connector gets the token. Formerly it was expecting to get a token directly. Now it expects to be given a URL where token can be downloade and fetches it.
* Changes CS upload / token config property names.
* Adds a dependency on ajax plugin in Cloud Service connector plugin (to simplify XHR calls).
* Changes unit tests for CS connector, due to this change.

Actually unit tests for `cloudservices` plugin required substantial amount of work to make it work, as internals changed noticably and handling XHR fired as soon as plugin is inited is not the nicest thing to do in our framework.

Config name changes needs to be rflected in ckeditor-docs repo, I'll create a PR for that shortly.

Closes #1755.


mlewand:
Docs PR in ckeditor/ckeditor-docs#141.

tests/plugins/cloudservices/cloudservices.js
```diff
@@ -90,9 +103,9 @@
 			}
 		},
 		'test no URL error': function() {
```
Comandeer: There is no test for `no-token-url` error.

tests/plugins/cloudservices/cloudservices.js
```diff
@@ -101,18 +114,6 @@
 			listener.removeListener();
 		},
 
-		'test no TOKEN error': function() {
```
Comandeer: I don't understand why this test was removed as `no-token` error is still present in editor.

tests/plugins/cloudservices/manual/errors.md
```diff
@@ -4,10 +4,19 @@
 @bender-include: ../../easyimage/_helpers/tools.js
 
 1. Open browser console.
+
+	## Expected
+
+	In console we should have information about error with code `cloudservices-no-url`.
```
Comandeer: This instruction is incorrect as shown error is for `no-token-url`.

plugins/cloudservices/plugin.js
```diff
@@ -75,22 +75,55 @@
...
+
+					refreshToken: function() {
+						CKEDITOR.ajax.load( tokenUrl, function( token ) {
+							if ( token ) {
+								tokenFetcher.token = token;
+							}
+						} );
+					},
+
+					init: function() {
```
Comandeer: This method could be simplified:

* first call to `refreshToken` could use `this`
* `setInterval` could just take `this.refreshToken` and `this.REFRESH_INTERVAL`. There is no need to introduce anonymous function or `that` variable.

plugins/cloudservices/plugin.js
```diff
@@ -75,22 +75,55 @@
...
+			var tokenUrl = editor.config.cloudServices_tokenUrl,
+				tokenFetcher = {
+					token: null,
+
+					// Allow external code (tests) to affect token refresh interval if needed.
+					REFRESH_INTERVAL: editor.CLOUD_SERVICES_TOKEN_INTERVAL || 3600000,
+
+					refreshToken: function() {
+						CKEDITOR.ajax.load( tokenUrl, function( token ) {
+							if ( token ) {
```
Comandeer: Maybe we should introduce some basic error handling? E.g. logging info to the console to inform the dev.

plugins/cloudservices/plugin.js
```diff
@@ -48,12 +48,12 @@
...
+			 * @param {String} [url] The upload URL. If not provided, {@link CKEDITOR.config#cloudServices_uploadUrl} will be used.
 			 * @param {Object} [additionalRequestParameters] Additional data that would be passed to the
 			 * {@link CKEDITOR.editor#fileUploadRequest} event.
 			 */
 			CloudServicesLoader.prototype.upload = function( url, additionalRequestParameters ) {
-				url = url || this.editor.config.cloudServices_url;
+				url = url || this.editor.config.cloudServices_uploadUrl;
 
 				if ( !url ) {
 					CKEDITOR.error( 'cloudservices-no-url' );
```
Comandeer: Shouldn't it be `no-upload-url` now?

tests/plugins/cloudservices/tokenregeneration.js
```diff
@@ -0,0 +1,71 @@
...
+			cloudServices_tokenUrl: TOKEN_URL
+		}
+	};
+
+	CKEDITOR.once( 'instanceCreated', function( evt ) {
+		// Lower the polling rate.
+		evt.editor.CLOUD_SERVICES_TOKEN_INTERVAL = 100;
+	} );
+
+	xhrServer.respondWith( function( req ) {
```
Comandeer: Code for creating fake token endpoint is present in several tests and it tends to be different for every test, yet – it always does the same. I propose creating generic `createFakeTokenEndpoint` function and extracting it to tests' helpers.

tests/plugins/cloudservices/cloudservices.js
mlewand: It was no longer viable in the former implementation and would have to be rewritten.

plugins/cloudservices/plugin.js
mlewand: Good point.

plugins/cloudservices/plugin.js
mlewand: This is already in place. If no token is returned dev will get `cloudservices-no-token` log. This behavior is covered by unit tests.

tests/plugins/cloudservices/tokenregeneration.js
mlewand: I have intentionally not do that, as it would create another layer of abstraction on top of sinon that would be used in that way only in a single place, making it harder to understand for dev reading the code.

However given the amount of work I had to put into tests now, I decided to change the approach  in test suite, thus merging three test suites back into one.

mlewand:
While doing these fixes I found one important bit - token polling was not cancelled when the related editor was destroyed. I added a proper code for that.

In addition to that I have found out that one test case was redundant so it was removed.

tests/plugins/imagebase/features/manual/upload.html
```diff
@@ -10,7 +10,7 @@
 
 	easyImageTools.getToken( function( token ) {
```
Comandeer: This function does not exist any longer.

plugins/cloudservices/plugin.js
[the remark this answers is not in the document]
mlewand: This is a different scenario, and it is a request for handling invalid tokens. Same thing happens if your link produced an invalid token.

PR #1762, opened 2018-03-08
 extraPlugins and removePlugins should allow for whitespaces

### LINKED ISSUE #1712: extraPlugins configuration option should allow whitespaces between values
## Feature Request

After pulling my hair out trying to debug why I could only add one external plugin at a time, I realized that the config option `extraPlugins` doesn't accept whitespace between values.

Not a huge deal, but *no* other config options seem to have that requirement, and there is no documentation around this required format.

I feel like this is confusing, as `removePlugins` is fine with whitespace, but `extraPlugins` isnt:

image

Accepting whitespace would be the better solution, but a note in the docs would be the quicker one, I'm assuming.

Happy to open a PR to trim the whitespace.

jacekbogdanski:

New feature

### This PR contains

- [x] Unit tests
- [x] Manual tests

## What changes did you make?

Removed whitespaces from a user input for `extraPlugins` and `removePlugins`. `removePlugins` was working previously with whitespaces but it looks like it wasn't on purpose so with strings like `plugin1, plugin2[space]` it wasn't removing plugins.

Closes #1712


core/editor.js
```diff
@@ -497,6 +497,8 @@
 			removePlugins = config.removePlugins;
 
 		if ( extraPlugins ) {
+			// Remove whitespaces (#1712).
+			extraPlugins = extraPlugins.replace( / /g, '' );
```
Comandeer: And what about non-breakable spaces? It would be safer to use `s` group.

tests/core/editor/editor.js
```diff
@@ -377,5 +377,31 @@ bender.test( {
...
+				plugins = CKEDITOR.tools.objectKeys( editor.plugins );
+
+			assert.isFalse( contains( plugins, 'basicstyles' ) );
+			assert.isFalse( contains( plugins, 'image2' ) );
+			assert.isFalse( contains( plugins, 'toolbar' ) );
+		} );
 	}
 } );
+
+function contains( array, value ) {
```
Comandeer: We have our `CKEDITOR.tools.indexOf` function. Please mind that IE8- does not support native `Array#indexOf`.

tests/core/editor/editor.js
```diff
@@ -377,5 +377,31 @@ bender.test( {
...
+	return Boolean( ~array.indexOf( value ) );
```
Comandeer: Let's not use such tricks – they're decreasing the readability of the code.

core/editor.js
```diff
@@ -492,9 +492,10 @@
 
 	function loadPlugins( editor ) {
 		var config = editor.config,
-			plugins = config.plugins,
-			extraPlugins = config.extraPlugins,
-			removePlugins = config.removePlugins;
+			// We have to trim plugins so they will be correctly processed by our regex.
+			plugins = CKEDITOR.tools.trim( config.plugins ),
+			extraPlugins = CKEDITOR.tools.trim( config.extraPlugins ),
+			removePlugins = CKEDITOR.tools.trim( config.removePlugins );
```
Comandeer: Wouldn't it be better to replace all whitespace here and then simplify `split` and `replace` statements as there won't be any whitespace to deal with? Because now we have separate treating for spaces at the beginning and end of the strings and separate for spaces inside strings.

tests/core/editor/editor.js
```diff
@@ -377,5 +377,42 @@ bender.test( {
 		assert.areSame( 'foo', insertHtml.firstCall.args[ 0 ], 'insertHtml dataValue' );
 		assert.areSame( 'html', insertHtml.firstCall.args[ 1 ], 'insertHtml mode' );
 		assert.areSame( range, insertHtml.firstCall.args[ 2 ], 'insertHtml range' );
+	},
+
+	'test extraPlugins allows whitespaces': function() {
```
Comandeer: There is `tests/core/config` directory – maybe it would be better/more readable to move these unit tests to dedicated file (`plugins.js`) in this directory?

Comandeer: I'll also add ticket references to tests.

core/editor.js
```diff
@@ -492,9 +492,10 @@
 
 	function loadPlugins( editor ) {
 		var config = editor.config,
-			plugins = config.plugins,
-			extraPlugins = config.extraPlugins,
-			removePlugins = config.removePlugins;
+			// We have to remove whitespaces (#1712).
```
Comandeer: If we remove whitespaces here, then we could simplify later regexs.

core/editor.js
```diff
@@ -492,20 +492,21 @@
...
+			plugins = config.plugins.replace( /\s/g, '' ),
+			extraPlugins = config.extraPlugins.replace( /\s/g, '' ),
```
Comandeer: Now checking if such config option is set is skipped and we assume from the start that it's string. It causes several tests to fail (e.g. `tests/plugins/image2/image2`) when `extraPlugins`/`removePlugins` is set to `undefined` explicitly. Probably `replace` for every option should be moved into appropriate `if`.

PR #1766, opened 2018-03-08
New plugin: splitbutton

### LINKED ISSUE #1679: Split button UI component

Feature request

We need a UI component that would allow to group multiple related options in a single button (in that sense it's similar to menu button). Once some option is activated it should change the icon to the active option.

For example think about alignment buttons. You have four buttons, but only one is active at a time. Having 4 buttons consumes a lot of space, instead we could group it into a single button. Important thing is that upon refresh, the active button is displayed.

We did a similar thing on our CKSource XMass hackathon.

An would look like that:

Split button with justify option active

When thinking about that I'm wondering if we really need a new component, or whether we could enhance menubutton in a way that it would conveniently handle this case.

engineering-this:

New feature

### This PR contains

- [x] Unit tests
- [x] Manual tests

## What changes did you make?

I've added the `splitbutton` plugin which is extending the `menubutton` class. Also I've slightly updated the `button`

Closes #1679.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -5,7 +5,8 @@
 
 New Features:
 
-* [#1679](https://github.com/ckeditor/ckeditor-dev/issues/1679): Introduced the [Split Button](https://ckeditor.com/cke4/addon/splitbutton plugin. Also added option to set initial style for button when creating it, and create button without icon.
+
+* [#1679](https://github.com/ckeditor/ckeditor-dev/issues/1679): Introduced the [Split Button](https://ckeditor.com/cke4/addon/splitbutton) plugin. Also added option to set initial style for button when creating it, and create button without icon.
 
 ## CKEditor 4.11.2
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -5,7 +5,7 @@
 
 New Features:
 
-* [#1679](https://github.com/ckeditor/ckeditor-dev/issues/1679): Introduced the Split Button plugin. Also added option to set initial style for button when creating it, and create button without icon.
+* [#1679](https://github.com/ckeditor/ckeditor-dev/issues/1679): Introduced the [Split Button](https://ckeditor.com/cke4/addon/splitbutton plugin. Also added option to set initial style for button when creating it, and create button without icon.
 
 ## CKEditor 4.11.2
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -3,6 +3,10 @@
 
 ## CKEditor 4.12
 
+New Features:
+
+* [#1679](https://github.com/ckeditor/ckeditor-dev/issues/1679): Introduced the Split Button plugin. Also added option to set initial style for button when creating it, and create button without icon.
+
 ## CKEditor 4.11.2
 
 ## CKEditor 4.11.1
--- a/plugins/menu/plugin.js
+++ b/plugins/menu/plugin.js
@@ -188,7 +188,7 @@ CKEDITOR.plugins.add( 'menu', {
 							var item = this.editor.getMenuItem( itemName );
 
 							if ( item && ( !item.command || this.editor.getCommand( item.command ).state ) ) {
-								item.state = ( item.stateFn && item.stateFn() ) || listenerItems[ itemName ];
+								item.state = listenerItems[ itemName ];
 								this.add( item );
 							}
 						}
--- a/plugins/menu/plugin.js
+++ b/plugins/menu/plugin.js
@@ -188,7 +188,7 @@ CKEDITOR.plugins.add( 'menu', {
 							var item = this.editor.getMenuItem( itemName );
 
 							if ( item && ( !item.command || this.editor.getCommand( item.command ).state ) ) {
-								item.state = listenerItems[ itemName ];
+								item.state = ( item.stateFn && item.stateFn() ) || listenerItems[ itemName ];
 								this.add( item );
 							}
 						}
--- a/plugins/splitbutton/plugin.js
+++ b/plugins/splitbutton/plugin.js
@@ -2,159 +2,172 @@
  * @license Copyright (c) 2003-2018, CKSource - Frederico Knabben. All rights reserved.
  * For licensing, see LICENSE.md or https://ckeditor.com/legal/ckeditor-oss-license
  */
+( function() {
+	'use strict';
 
-CKEDITOR.plugins.add( 'splitbutton', {
-	requires: 'menubutton',
+	CKEDITOR.plugins.add( 'splitbutton', {
+		requires: 'menubutton',
 
-	beforeInit: function( editor ) {
-		/**
-		 * @class
-		 * @extends CKEDITOR.ui.menuButton
-		 */
-		CKEDITOR.ui.splitButton = CKEDITOR.tools.createClass( {
-			base: CKEDITOR.ui.menuButton,
-
+		beforeInit: function( editor ) {
 			/**
-			 * Creates a menuButton class instance.
-			 *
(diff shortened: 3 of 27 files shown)


tests/plugins/splitbutton/splitbutton.js
```diff
@@ -0,0 +1,60 @@
...
+						icon: 'superscript'
+					} ]
+				} );
+			}
+		}
+	}
+};
+
+bender.test( {
+	'test splitbutton in toolbar': function() {
```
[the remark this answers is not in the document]
Comandeer: There are also missing tests for changing the state of menu items.

tests/plugins/splitbutton/manual/splitbutton.html
```diff
@@ -0,0 +1,54 @@
...
+			} );
+		}
+	} );
+
+	CKEDITOR.replace( 'editor', {
+		extraPlugins: 'teststylesplit'
+	} );
+
+	// Just hiding buttons that are irrelevant to test.
+	CKEDITOR.instances.editor.on( 'contentDom', function() {
```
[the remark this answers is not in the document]
engineering-this: Hej, @mlewand what is your view on this?

engineering-this:
@Comandeer 

>There is issue with plugins with options that are exclusive (e.g. justify) – menu does not correctly update states.

It looks fine with justify on sample
https://ckeditor4-preview.ckeditor.com/t/1679/plugins/splitbutton/samples/splitbutton.html

>The other concern I have is the fact how menu works: it focuses the first element upon opening. However our Styles or Format combos focuses the first selected option. I'm wondering what would be more user-friendly. Speaking in terms of ARIA: menu or listbox?

I think we need a bit of discussion here if first item should be initially focused or not?

@mlewand What is your tough on that?

>It also uncovers very important issue with the whole implementation: as #678 is not resolved, splitbutton needs other buttons to be added to be able to use their commands. That makes the whole feature de facto not usable at all.

Actually it is caused by ACF. I had to set `config.allowedContent = true;` and everything works. My question here is how should we deal with that?

engineering-this:
After some talking first item will be focused on menu open, just like other menu buttons.

tests/plugins/splitbutton/manual/splitbutton.html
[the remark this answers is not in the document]
mlewand: For sure we should not workaround this like that. We need a proper fix mentioned by @Comandeer.

I mean, workaround can be placed for the time being, but the issue can not be merged to stable release with such workaround in place.

@engineering-this please, put a reference to #678 anywhere where such workaround is used.

plugins/splitbutton/plugin.js
```diff
@@ -0,0 +1,186 @@
...
+
+				if ( activeButton ) {
+					CKEDITOR.document.getById( properties.buttons[ defaultButton.id ]._.id ).setStyle( 'display', 'none' );
+					properties.buttons[ defaultButton.id ].hidden = true;
+
+					CKEDITOR.document.getById( properties.buttons[ activeButton.id ]._.id ).removeStyle( 'display' );
+					properties.buttons[ activeButton.id ].hidden = false;
+					previousButton = activeButton;
+				} else {
+					CKEDITOR.document.getById( properties.buttons[ defaultButton.id ]._.id ).removeStyle( 'display' );
```
Comandeer: This code is the same as code above and similar to code, in which `display` is set to `none`. It'd move it to separate, dedicated function.

plugins/splitbutton/plugin.js
```diff
@@ -0,0 +1,186 @@
...
+			// First button as default. It might be overwritten by actual default button.
+			if ( !defaultButton ) {
+				previousButton = properties.buttons[ item.id ];
+				defaultButton = properties.buttons[ item.id ];
+			} else {
+				if ( item[ 'default' ] ) {
+					defaultButton = properties.buttons[ item.id ];
+				} else {
+					properties.buttons[ item.id ].style = 'display: none;';
+					properties.buttons[ item.id ].hidden = true;
```
Comandeer: I understand that it's a workaround for #678 to add features by adding buttons and hiding them?

engineering-this: Correct.

engineering-this:
>Actually we haven't defined any behavior for that, but changing icons in such way seems to be confusing. Microsoft Word has always the same icon in split button face and it changes only its state, e.g. "Numbering" has always the same icon despite which numbering option is selected. The face is used for switching feature on/off, not for representing currently selected option.

This was the our idea of split button, to create 'splitbutton' with static face we could just add one button for default behaviour and next to it menu button for options.

Comandeer:
But now the face is not static, it's dynamically changing and I think it's quite confusing to the end user.

engineering-this:
Well, I think we could add option to lock face on default button.

CHANGES.md
```diff
@@ -6,6 +6,7 @@
 New Features:
 
 * [#1761](https://github.com/ckeditor/ckeditor-dev/issues/1761): [Autolink](https://ckeditor.com/cke4/addon/autolink) plugin supports email links.
+* [#1679](https://github.com/ckeditor/ckeditor-dev/issues/1679): Introduced the Split Button plugin. Also added option to set initial style for button when creating it, and create button without icon.
```
Comandeer: Please add link to the plugin.

plugins/splitbutton/plugin.js
Comandeer: Then I'd think about making `hide` method on button or any other toolbar element.

plugins/splitbutton/plugin.js
```diff
@@ -0,0 +1,213 @@
...
+			item.role = 'menuitemcheckbox';
+
+			// We need both 'click' and 'onClick', because this definition can be used for setting both buttons and menu items.
+			// Button uses 'click' property, menu item uses 'onClick' property.
+			item.onClick = item.onClick || item.click ||  function() {
+				editor.execCommand( this.command, this.commandData );
+			};
+			item.click = item.onClick;
+
+			editor.getCommand( item.command ).on( 'state', function() {
```
Comandeer: It's still not possible to bind only click to button, without any command.

Comandeer: There aren't also tests for such case.

plugins/menu/plugin.js
```diff
@@ -188,7 +188,7 @@ CKEDITOR.plugins.add( 'menu', {
 							var item = this.editor.getMenuItem( itemName );
 
 							if ( item && ( !item.command || this.editor.getCommand( item.command ).state ) ) {
-								item.state = listenerItems[ itemName ];
+								item.state = ( item.stateFn && item.stateFn() ) || listenerItems[ itemName ];
```
Comandeer: I'm not sure if a better solution won't be to just substitute listener, not add another property directly to `menu`.

engineering-this:
>I'm wondering also about one thing: why do we need to render and hide all buttons? Can't we have one button that would act as face and its icon and click actions would be just replaced ad hoc? Is the current solution connected with our convoluted way of registering commands alongside the buttons? If yes, then I really think if #678 isn't blocking.

@Comandeer: We need to render buttons, otherwise we won't register commands. And if thats necessary then I'm reusing all rendered buttons. It might be possible to remove them after rendering and just switch face onClick, icon and label, but that would just make all the logic even more complex. Imho this would be too haxy.

engineering-this:
Rebased onto the latest major and updated version references to 4.11.

engineering-this:
Rebased onto latest `major`.

PR #1770, opened 2018-03-09
Removed memory leak caused by CKEDITOR.filter.instances

### LINKED ISSUE #1722: Fix memory leaks caused by CKEDITOR.filter.instances storing nested editables' filters forever

Bug

## References
 * https://dev.ckeditor.com/ticket/11010

Every nested editable which has allowed content rules defined has its filter instance. These instances are stored in nested editable instances, which most likely are correctly garbage collected. But filter instances are also stored in CKEDITOR.filter.instances and they are kept there forever.

We could have `filter.destroy()` method removing instance from the instances hash and doing other cleanup if necessary. Then we could destroy filters when destroying widgets.

### Expected result

All filter instances should be removed.

### Actual result

Filter instances are also stored in CKEDITOR.filter.instances and they are kept there forever.

## Other details

* Browser: …
* OS: …
* CKEditor version: 4.3.0+

### LINKED ISSUE #2466: CKEDITOR.filter should expose a way to bind standalone instance with editor to prevent memory leaks
## Type of report

Feature request

Currently, standalone CKEDITOR.filter instance cannot be removed during destroying editor because it doesn't contain a reference to the editor itself. We should add an optional way to bind filter and editor. Because of the polymorphic constructor, this request is pretty tricky, although we could move `rules` to the second constructor position as an optional parameter:

`new CKEDITOR.filter( editorOrRules, [rules] )`

jacekbogdanski:

Bug fix

### This PR contains

- [x] Unit tests
- [x] Manual tests

## What changes did you make?

Added additional option to `CKEDITOR.filter` constructor to pass an editor owning filter instance. Changed `editor.destroy` function to remove correct editors and refactored plugins using `CKEDITOR.filter`. With this change we are now able to remove orphaned editor filters and prevent memory leak with `CKEDITOR.filter.instances`.

Additional parameter doesn't look most beautiful because we should probably go with something like `new CKEDITOR.filter( editor, rules)` but it would require to provide breaking change for our API.

Based on #1690

Closes #1722, closes #2466

### DIFF
--- a/core/filter.js
+++ b/core/filter.js
@@ -79,7 +79,7 @@
 	 * @class
 	 * @constructor Creates a filter class instance.
 	 * @param {CKEDITOR.editor/CKEDITOR.filter.allowedContentRules} editorOrRules
-	 * @param {CKEDITOR.editor} owner Editor owning the filter instance.
+	 * @param {CKEDITOR.editor} owner Editor owning the filter instance. This parameter is available since 4.10.0.
 	 */
 	CKEDITOR.filter = function( editorOrRules, owner ) {
 		/**
--- a/core/filter.js
+++ b/core/filter.js
@@ -59,19 +59,29 @@
 	 * {@link CKEDITOR.filter.allowedContentRules} instead of {@link CKEDITOR.editor}
 	 * to the constructor:
 	 *
-	 *		var filter = new CKEDITOR.filter( 'b' );
+	 * ```javascript
+	 * var filter = new CKEDITOR.filter( 'b' );
 	 *
-	 *		filter.check( 'b' ); // -> true
-	 *		filter.check( 'i' ); // -> false
-	 *		filter.allow( 'i' );
-	 *		filter.check( 'i' ); // -> true
+	 * filter.check( 'b' ); // -> true
+	 * filter.check( 'i' ); // -> false
+	 * filter.allow( 'i' );
+	 * filter.check( 'i' ); // -> true
+	 * ```
+	 *
+	 * If filter is only used by a single editor instance, you should pass additional owner editor instance
+	 * so the filter could be removed with {@link CKEDITOR.editor#destroy} method.
+	 *
+	 * ```javascript
+	 * var filter = new CKEDITOR.filter( 'b', editor );
+	 * ```
 	 *
 	 * @since 4.1
 	 * @class
 	 * @constructor Creates a filter class instance.
 	 * @param {CKEDITOR.editor/CKEDITOR.filter.allowedContentRules} editorOrRules
+	 * @param {CKEDITOR.editor} owner Editor owning the filter instance.
 	 */
-	CKEDITOR.filter = function( editorOrRules ) {
+	CKEDITOR.filter = function( editorOrRules, owner ) {
 		/**
 		 * Whether custom {@link CKEDITOR.config#allowedContent} was set.
 		 *
@@ -191,6 +201,7 @@
 		// Rules object passed in editorOrRules argument - initialize standalone filter.
 		else {
 			this.customConfig = false;
+			this.editor = owner;
 			this.allow( editorOrRules, 'default', 1 );
 		}
 	};
--- a/core/filter.js
+++ b/core/filter.js
@@ -68,20 +68,20 @@
 	 * filter.check( 'i' ); // -> true
 	 * ```
 	 *
-	 * If filter is only used by a single editor instance, you should pass additional owner editor instance
+	 * If filter is only used by a single editor instance, you should pass editor instance alongside with the rules
 	 * so the filter could be removed with {@link CKEDITOR.editor#destroy} method.
 	 *
 	 * ```javascript
-	 * var filter = new CKEDITOR.filter( 'b', editor );
+	 * var filter = new CKEDITOR.filter( editor, 'b' );
 	 * ```
 	 *
 	 * @since 4.1
 	 * @class
 	 * @constructor Creates a filter class instance.
 	 * @param {CKEDITOR.editor/CKEDITOR.filter.allowedContentRules} editorOrRules
-	 * @param {CKEDITOR.editor} owner Editor owning the filter instance. This parameter is available since 4.10.0.
+	 * @param {CKEDITOR.editor} rules This parameter is available since 4.10.0.
 	 */
-	CKEDITOR.filter = function( editorOrRules, owner ) {
+	CKEDITOR.filter = function( editorOrRules, rules ) {
 		/**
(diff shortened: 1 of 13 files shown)


tests/_benderjs/ckeditor/static/tools.js
```diff
@@ -59,6 +59,19 @@
 		} );
 
 	bender.tools = {
+		/**
+		 * Creates an array from an object.
+		 *
+		 * @param  {Object} obj
+		 * @return {Array} object values.
+		 */
+		objToArray: function( obj ) {
```
Comandeer: This change does not seem to be directly connected with the fix. Why is it present here?

core/filter.js
```diff
@@ -59,19 +59,29 @@
...
+	 *
+	 * ```javascript
+	 * var filter = new CKEDITOR.filter( 'b', editor );
+	 * ```
 	 *
 	 * @since 4.1
 	 * @class
 	 * @constructor Creates a filter class instance.
 	 * @param {CKEDITOR.editor/CKEDITOR.filter.allowedContentRules} editorOrRules
+	 * @param {CKEDITOR.editor} owner Editor owning the filter instance.
```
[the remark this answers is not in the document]
jacekbogdanski: I agree that it's not beautiful. But let's say you would like to create empty filter and setup it by `CKEDITOR.filter#allow` function e.g
```javascript
var filter = new CKEDITOR.filter();
filter.allow( {...} );
```
In that case, you won't be able to create a filter with owning `editor` - if you try to pass `editor` as an argument you will create a filter based on `editor` (https://github.com/ckeditor/ckeditor-dev/blob/major/core/filter.js#L168).

We have that case in our code (https://github.com/ckeditor/ckeditor-dev/blob/major/plugins/clipboard/plugin.js#L1347) and it's possible that people could use it in that way also.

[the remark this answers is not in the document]
jacekbogdanski: I agree it can be done but with that change, we will provide breaking change for our API. If we are ok with that I also prefer `new CKEDITOR.filter( editorOrRules, rules )` signature. WDYT @Comandeer?

jacekbogdanski:
Rebased into the latest `major`.

jacekbogdanski:
I left `CKEDITOR.filter` signature unchanged - please refer to https://github.com/ckeditor/ckeditor-dev/pull/1770#discussion_r183010701 and if you think we can provide breaking change for API I will refactor the code. 
About destroying nested editables inside `widget` plugin - a problem with this approach is that we are sharing the same filter for each widget instance:
https://github.com/ckeditor/ckeditor-dev/blob/045ae8e15cbfc85872748d737e8e54fb414acbcd/plugins/widget/plugin.js#L2202-L2224
So if we destroy nested editable filter on widget destroy it will become unavailable for other widgets of the same type. It's not a problem when destroying editor because in that case widgets are already  / will be destroyed so filters are not needed anymore. I could implement some workaround like destroying filter only if it's not used by widgets but not sure if it's correct way.

Comandeer:
Hmm, the main ticket is about destroying filters in nested editables, so it would be nice to actually implement it. I think that workaround to detect any nested editable still using this filter should be enough.

core/filter.js
```diff
@@ -59,19 +59,29 @@
...
+	 *
+	 * ```javascript
+	 * var filter = new CKEDITOR.filter( editor, 'b' );
+	 * ```
 	 *
 	 * @since 4.1
 	 * @class
 	 * @constructor Creates a filter class instance.
 	 * @param {CKEDITOR.editor/CKEDITOR.filter.allowedContentRules} editorOrRules
+	 * @param {CKEDITOR.editor} rules This parameter is available since 4.10.0.
```
Comandeer: Please update version tag.

core/filter.js
```diff
@@ -165,11 +175,12 @@
 		// Register filter instance.
 		CKEDITOR.filter.instances[ this.id ] = this;
 
-		if ( editorOrRules instanceof CKEDITOR.editor ) {
-			var editor = this.editor = editorOrRules;
+		this.editor = editorOrRules instanceof CKEDITOR.editor ? editorOrRules : null;
```
Comandeer: Why not:

```javascript
var editor = this.editor = editorOrRules instanceof CKEDITOR.editor ? editorOrRules : null;
```
? With such version there won't be need to change all references to the editor.

tests/core/editor/destroy.js
```diff
@@ -1,84 +1,112 @@
...
+
+				assert.areEqual( 0, countFilters( editor ) );
+				assert.areEqual( filters + 1, countFilters() );
+			} );
+		}
+	} );
+
+	function countFilters( editor ) {
+		var filters = bender.tools.objToArray( CKEDITOR.filter.instances );
+		return editor ? filters.filter( function( filter ) {
```
Comandeer: This function won't work in IE8, as there is no `Array.prototype.filter`. Our own `CKEDITOR.tools.array.filter` should be used here.

tests/core/editor/manual/filterdestroy.md
```diff
@@ -0,0 +1,14 @@
+@bender-tags: bug, 4.9.1, 1722
```
Comandeer: Please update version tag.

tests/plugins/widget/manual/nestededitablefilters.md
```diff
@@ -0,0 +1,14 @@
+@bender-tags: widget, bug, 4.10.0, 1722
```
Comandeer: Please update version tag.

core/filter.js
Comandeer: Also this parameter is not of `CKEDITOR.editor` type.

Comandeer:
Rebased onto latest `major` and changed base.

core/filter.js
```diff
@@ -59,19 +59,29 @@
...
-	 *		filter.allow( 'i' );
-	 *		filter.check( 'i' ); // -> true
+	 * filter.check( 'b' ); // -> true
+	 * filter.check( 'i' ); // -> false
+	 * filter.allow( 'i' );
+	 * filter.check( 'i' ); // -> true
+	 * ```
+	 *
+	 * If filter is only used by a single editor instance, you should pass editor instance alongside with the rules
+	 * so the filter could be removed with {@link CKEDITOR.editor#destroy} method.
```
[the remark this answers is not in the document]
jacekbogdanski: Isn't enough described by the whole https://github.com/ckeditor/ckeditor-dev/blob/013af363d892680f9c0a2f11765aee45ec09ca2b/core/filter.js#L33-L56 docs? :thinking:  However, I added additional info that passing only editor ( without rules ) will also prevent memory leaks.

PR #1773, opened 2018-03-09
Fixed widgetselection plugin receiving keystrokes from outside of the editable

theNailz:
https://github.com/ckeditor/ckeditor-dev/issues/1719

Bugfix for https://github.com/ckeditor/ckeditor-dev/issues/1719

No

### This PR contains

- [ ] Unit tests
- [ ] Manual tests

## What changes did you make?

Applied the fix by @casjonker as requested by @jacekbogdanski.

### DIFF
--- a/plugins/widgetselection/plugin.js
+++ b/plugins/widgetselection/plugin.js
@@ -121,6 +121,10 @@
 		 */
 		addFillers: function( editable ) {
 			var editor = editable.editor;
+			
+			if (!editor.focusManager.hasFocus) {
+    			return false;
+			}
 
 			// Whole content should be selected, if not fix the selection manually.
 			if ( !this.isWholeContentSelected( editable ) && editable.getChildCount() > 0 ) {


plugins/widgetselection/plugin.js
```diff
@@ -121,6 +121,10 @@
 		 */
 		addFillers: function( editable ) {
 			var editor = editable.editor;
+			
+			if (!editor.focusManager.hasFocus) {
```
mlewand: This change fixes the symptom of the bug, but does not fix the source of this issue.

If you inspect the trace of the problem, it is caused that the listener is added to the document of the editor here: https://github.com/ckeditor/ckeditor-dev/blob/88afdc651576e4ef54b7545dd20d456cf8c52940/plugins/widgetselection/plugin.js#L28

This works fine for classic editor, as it creates an `iframe` with a dedicated document. However inline editors (`divarea` or instance created with `CKEDITOR.inline`) share same document with the rest of the page. The listener should be clearly limited to the editable itself, that will guarantee that no external key presses will be picked up.

PR #1793, opened 2018-03-14
Fixed blurring empty image base caption

### LINKED ISSUE #1776: Empty placeholder not hidden when blurred

Bug

1. Open http://tests.ckeditor.test:1030/tests/plugins/imagebase/features/manual/caption
2. Focus widget with empty caption.
3. Put some content into the empty caption.
4. Focus widget with non-empty caption (the last one in the editor).
5. Select all content inside the caption and delete it.
6. Blur the widget.

### Expected result

Caption for the last widget is hidden.

### Actual result

Caption is still visible, despite the fact it's empty.

## Other details

* CKEditor version: 4.9.0
* Installed CKEditor plugins: `easyimage`

jacekbogdanski:

Bug fix.

### This PR contains

- [x] Unit tests
- [x] Manual tests

## What changes did you make?

Changed `imagebase` plugin to refresh every caption featured widget on `selectionChange`.

Closes #1776

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -9,7 +9,7 @@ Fixed Issues:
 * [#1034](https://github.com/ckeditor/ckeditor-dev/issues/1034): Fixed: JAWS leaves forms mode after pressing <kbd>Enter</kbd> key in an inline CKEditor instance.
 * [#1748](https://github.com/ckeditor/ckeditor-dev/pull/1748): Fixed: Added missing [`CKEDITOR.dialog.definition.onHide`](https://docs.ckeditor.com/ckeditor4/latest/api/CKEDITOR_dialog_definition.html#property-onHide) API documentation. Thanks to [sunnyone](https://github.com/sunnyone)!
 * [#1321](https://github.com/ckeditor/ckeditor-dev/issues/1321): Ideographic space character `\u3000` is lost when pasting text.
-* [#1776](https://github.com/ckeditor/ckeditor-dev/issues/1776): Image base empty caption placeholder is not hidden when blurred.
+* [#1776](https://github.com/ckeditor/ckeditor-dev/issues/1776): [Image Base](https://ckeditor.com/cke4/addon/imagebase) empty caption placeholder is not hidden when blurred.
 
 ## CKEditor 4.9
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,7 +1,7 @@
 ﻿CKEditor 4 Changelog
 ====================
 
-## CKEditor 4.10.0
+## CKEditor 4.10
 
 Fixed Issues:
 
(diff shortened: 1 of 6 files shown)


CHANGES.md
```diff
@@ -1,6 +1,12 @@
 ﻿CKEditor 4 Changelog
 ====================
 
+## CKEditor 4.9.1
```
Comandeer: Please fix version – as 4.9.1 is dropped, it should be 4.10.0

plugins/imagebase/plugin.js
```diff
@@ -614,6 +608,16 @@
 		link: getLinkFeature()
 	};
 
+	function getWidgetsWithFeature( widgets, feature ) {
```
Comandeer: Functions should be declared before they are used in the code.

plugins/imagebase/plugin.js
```diff
@@ -614,6 +608,16 @@
...
+		return CKEDITOR.tools.array.reduce( CKEDITOR.tools.objectKeys( widgets ), function( featuredWidgets, id ) {
```
Comandeer: Shouldn't it be `filter`, not `reduce`? Then the whole function could be simplified to:

```javascript
return hasWidgetFeature( widgets[ id ], feature );
```

tests/plugins/imagebase/features/manual/captionemptyplaceholder.md
```diff
@@ -0,0 +1,15 @@
+@bender-tags: 4.9.1, bug, 1776
```
Comandeer: Please update version tag.

tests/plugins/imagebase/features/manual/captionemptyplaceholder.md
```diff
@@ -0,0 +1,15 @@
...
+@bender-ui: collapsed
+@bender-ckeditor-plugins: wysiwygarea, toolbar, basicstyles, imagebase
+@bender-include: %BASE_PATH%/plugins/easyimage/_helpers/tools.js
+
+1. Focus first image.
```
Comandeer: I'm not sure if this procedure is enough to reproduce the bug. It seems that typing and deleting – as stated in #1776 - is necessary.

plugins/imagebase/plugin.js
jacekbogdanski: I'm not sure how to write it as a filter function? `editor.widgets.instances` is an object so we will have to convert it to an array to use filter function which requires additional code. If I simply replace `reduce` with `filter` here I will get widget ids, not instances. WDYT @Comandeer ?

tests/plugins/imagebase/features/manual/captionemptyplaceholder.md
jacekbogdanski: The HTML of this manual test already have empty `figcaption` element to reproduce removing caption content. I reproduced manual test steps without my changes and the test was broken so I think it works fine. I created it to shorten test steps but if you think it's too terse I will refactor the test.

CHANGES.md
jacekbogdanski: I updated it to `4.10` to keep consistent with the rest of the file.

plugins/imagebase/plugin.js
Comandeer: True, I didn't think about it. So let's stay with `reduce` as it's the best option here.

tests/plugins/imagebase/features/manual/captionemptyplaceholder.md
Comandeer: Strangely, I could reproduce it in your test, but no success in `/tests/plugins/imagebase/features/manual/caption` 🤔

plugins/imagebase/plugin.js
jacekbogdanski: I see now that functions in this file are declared at the top and I moved my function also. But are we going with this code style in whole CKEditor codebase? Linter didn't complain about it and I think it's cleaner when your private functions are below the main login. Also I didn't catch this code style in different files. Should I follow this rule in other files?

PR #1829, opened 2018-03-21
Provide integration between tableselection and enterkey plugins

### LINKED ISSUE #1816: Enter key breaks table structure when pressed in a table selection

Bug

1. Enter https://ckeditor.com/ckeditor-4/
2. Add a table with the default dialog configuration
3. Select some cells and press enter. Like that:

Sample table selection

### Expected result

* All cells are emptied.
* New line is inserted into the first selected cell.
* Fake selection is removed.

### Actual result

After press enter: image

## Other details

~There was a very similar issue fixed for Firefox in #415. I wonder why Chrome does not benefit from this fix?~

~I wonder if we had this issue from the launch, or is it a result of Chrome team doing some updates under the hood?~

Edit: I actually see that all browser do not support this case well it has to be checked out.

* Browser: Any
* OS: Any
* CKEditor version: since 4.7.0

msamsel:

Bug fix

### This PR contains

- [x] Unit tests
- [x] Manual tests

## What changes did you make?

- Add new keyCode which will be processed by tableSelection keystroke handler. As this event prevent of firing further listeners, then enterkey's listeners won't be execute. This allows on keeping coherent behaviour of pressing <kbd>Enter</kbd> over table selection with <kbd>Delete</kbd> and <kbd>Backspace</kbd>
- Update generic test for tableselection to contain `enterkey` plugin.

Close #1816


plugins/tableselection/plugin.js
```diff
@@ -891,7 +891,8 @@
 						39: 1, // Right Arrow,
 						40: 1, // Down Arrow
 						8: 1, // Backspace
-						46: 1 // Delete
+						46: 1, // Delete
+						13: 1 // Enter
```
Comandeer: Comment before the function is no longer true (now also Enter is handled).

plugins/tableselection/plugin.js
```diff
@@ -1008,12 +1009,35 @@
 							ranges[ 0 ].moveToElementEditablePosition( firstCell );
 						}
 
+						if ( keystroke === 13 ) {
+							insertNewLineInsideFirstRange( editor, evt.data, ranges );
```
Comandeer: It feels dirty to reimplement this logic… Isn't there any sensible way to not destroy table, but at the same time – delegate <kbd>Enter</kbd> handling to our existing mechanics?

plugins/enterkey/plugin.js
```diff
@@ -405,6 +414,42 @@
...
+				}
+
+				if ( CKEDITOR.tools.indexOf( [ 'td', 'th' ], path.lastElement.getName() ) === -1 ) {
+					return false;
+				}
+				if ( path.lastElement.getChildCount() !== 1 ) {
+					return false;
+				}
+
+				innerElement = path.lastElement.getChild( 0 ).clone( true );
```
Comandeer: Is this cloning logic really needed? I've tested in Firefox and IE 11 with following code:

```javascript
var element = CKEDITOR.dom.element.createFromHtml( '<div><br/></div>' );
element.getBogus(); // <br/>
element.getText(); // ''
element.getText().length; // 0
```

Seems like browsers (at least Firefox & IE 11) treat bogus element as non-existent.

plugins/tableselection/plugin.js
```diff
@@ -986,10 +987,13 @@
 					firstCell = ranges[ 0 ]._getTableElement();
 					lastCell = ranges[ ranges.length - 1 ]._getTableElement();
 
-					evt.data.preventDefault();
-					evt.cancel();
+					// Pass enter event to be handled by enterkey.
```
Comandeer: Please add reference to the issue.

plugins/enterkey/plugin.js
msamsel: At least in chrome it is problematic, because paragraph might be removed in such situation, and cell became unelectable. That's why using cloned element seems to be more proper here.
apr-30-2018 10-52-28

plugins/tableselection/plugin.js
```diff
@@ -1009,6 +1011,14 @@
 						}
 
 						selection.selectRanges( ranges );
+
+						// We need to lock undoManager to consider clearing table and inserting new paragraph as single operation, and have only one undo step (#1816).
+						if ( key === 13 && editor.plugins.enterkey ) {
+							editor.undoManager.lock();
```
Comandeer: It should be done by firing appropriate events (`lockSnapshot` and `unlockSnapshot`) as it's possible to use `tableselection` with `enterkey` plugin, but without `undo` one. In such case `editor.undoManager` won't be defined.

msamsel: I missed them during searching :(

msamsel:
> There is no test for handling <kbd>Enter</kbd> without `enterkey` plugin. 

I don't see any benefits of such tests.
Please remember, that originally reported bug exists **only** when `tableselection` was present together with `enterkey`. And this PR contain fix for integrations between those two plugins.
That's why I don't understand how test case without `enterkey` might be helpful here. It sounds like testing totally different case (integration of `tableselection` with native enter behaviour).

Comandeer:
> I don't see any benefits of such tests.

The first iteration of your fix AFAIR _broke_ the default behaviour of <kbd>Enter</kbd>. Therefore it's possible that next iterations also could break it and we don't have any unit test to check if that's the case.

plugins/tableselection/plugin.js
```diff
@@ -1009,6 +1020,13 @@
 						}
 
 						selection.selectRanges( ranges );
+
+						// We need to lock undoManager to consider clearing table and inserting new paragraph as single operation, and have only one undo step (#1816).
+						if ( key === 13 && editor.plugins.enterkey ) {
```
Comandeer: Probably the whole handling of undo should be refactored as now native <kbd>Enter</kbd> also produces two steps (works correctly on `master`).

Comandeer: It also means that we need one more unit test for native <kbd>Enter</kbd>, this time connected with undo ;)

PR #1834, opened 2018-03-23
Fixed: Column resize cursor shows up inside merged cells

### LINKED ISSUE #1467: Column resize cursor shows up inside merged cells

Bug

1. Open sample https://sdk.ckeditor.com/samples/table.html
2. Hover over "Full Year (unaudited)" header cell or merge one of content cells inside the table and hover between two cells.

### Expected result

Cursor should be standard text cursor. There should be no resize cursor. 

### Actual result

Resize cursor shows up and it is possible to resize columns. 
This can cause problems when user wants to select text which ends or starts where resized column starts/ends.

## Other details

* Browser: Any
* OS: Any
* CKEditor version: 4.0+
* Installed CKEditor plugins: tableresize

table resizer

engineering-this:

Bugfix

### This PR contains

- [ ] Unit tests
- [x] Manual tests

In theory I could cover this with unit test that would fail before this patch. But such test could just check if resizer appears on one spot. This should be tested manually to ensure that resizer fully acts as expected and doesn't show up in other unexpected places.

## What changes did you make?

Previously resize plugin created one `pillar` after each column which is basically reference to coordinates range in which resizer should showup once mouse is there. Each pillar had height of whole table and resizer would appear only when mouse coordinates were same as pillar coordinates + pillar width. After my change there is pillar after each cell from each row. Each of new pillars has now height same as rows height, so pillars won't leak to merged cells. Resizer although will appear covering all pillars that aligned with hovered pillar.

Closes #1467


mlewand:
I'd be for displaying a continuous pillar, despite the fact that some cells are missing inbetween.

Any thoughts on https://github.com/ckeditor/ckeditor-dev/pull/1834#pullrequestreview-107661605 @dkonopka @oleq @m-kr?

engineering-this:
After some thinking I decided to go this way. Splitted splitted pillar might suggest that you are resizing whole column not just one cell. Especially if there were few merged rows in between.

PR #1861, opened 2018-04-03
Fixed: Typing text on iOS scrolls editable

### LINKED ISSUE #498: iOS Safari typing text scrolls the ckeditor
Hi there was some old issues in the previous tracker reporting bad scroll behavior on iOS. The issue is still here.

To reproduce this issue:
On the main page of ckeditor.com add a few lines to the end of the first ckEditor instantiated (Article Editor) then place your cursor to the top of the article and start typing.

The CKEditor will scroll down and you will not see what you are typing.

Tested on Safari iOS 10 iPad and iPhone.

Regards.

engineering-this:

Bugfix

### This PR contains

- [ ] Unit tests
- [x] Manual tests

## What changes did you make?
I'm calling `scrollIntoView` on current selection ranges on iOS. Doing that on `keydown` and on `textInput`. Reason to use two listeners is because iOS screen keyboard doesn't fire `keydown` event when using autocomplete. And using backspace or new line doesn't call `textInput`.

One problem with this solution is that editor is scrolled right after scroll invoked by browsers native methods. Hence there is visible flickering.

Closes #498

### DIFF
--- a/core/editable.js
+++ b/core/editable.js
@@ -1216,12 +1216,12 @@
 				if ( !this.isInline() && CKEDITOR.env.iOS && CKEDITOR.env.safari ) {
 					var touchListenerStatus, scrollListenerStatus;
 
-					this.on( 'textInput', scrollToRangesOnScroll, this, null, 10000 );
+					this.on( 'input', scrollToRangesOnScroll, this, null, 10000 );
 					this.on( 'keydown', scrollToRangesOnScroll, this, null, 10000 );
 				}
 
 				function scrollToRangesOnScroll() {
-					var contentsWrapper = CKEDITOR.document.findOne( '#' + CKEDITOR.instances.editor.id + '_contents' );
+					var contentsWrapper = CKEDITOR.document.findOne( '#' + editor.id + '_contents' );
 
 					function scrollListener() {
 						this.editor.getSelection().getRanges()[ 0 ].scrollIntoView();
--- a/core/editable.js
+++ b/core/editable.js
@@ -1211,6 +1211,38 @@
 						return false;
 					}, this, null, 100 ); // Later is better – do not override existing listeners.
 				}
+
+				// Scroll to current ranges due to bug on iOS #498
+				if ( !this.isInline() && CKEDITOR.env.iOS && CKEDITOR.env.safari ) {
+					var touchListenerStatus, scrollListenerStatus;
+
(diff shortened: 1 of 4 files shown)


core/editable.js
```diff
@@ -1211,6 +1211,38 @@
 						return false;
 					}, this, null, 100 ); // Later is better – do not override existing listeners.
 				}
+
+				// Scroll to current ranges due to bug on iOS #498
+				if ( !this.isInline() && CKEDITOR.env.iOS && CKEDITOR.env.safari ) {
+					var touchListenerStatus, scrollListenerStatus;
+
+					this.on( 'textInput', scrollToRangesOnScroll, this, null, 10000 );
```
Comandeer: Why isn't standard `input` event used here?

core/editable.js
```diff
@@ -1211,6 +1211,38 @@
...
+					this.on( 'keydown', scrollToRangesOnScroll, this, null, 10000 );
+				}
+
+				function scrollToRangesOnScroll() {
+					var contentsWrapper = CKEDITOR.document.findOne( '#' + CKEDITOR.instances.editor.id + '_contents' );
```
Comandeer: Can't we use `editor.container` here?

What's more, this code works **only** in instance name `editor`! The editor's instance is hardcoded.

engineering-this:
I've been digging this issue a lot. Only way to avoid this flickering is to prevent `keydown` event, but then we would have to create fake event and manually push input to editor. Imo we should treat this fix as temporary until webkit fixes this bug.

mlewand:
@Comandeer I'll ask you to help on this issue. Please do a research what are the options for getting around this issue as it does not look like it's going to get any attention form the browser vendor.

Comandeer:
Well… I managed to get rid of flickering, but I created a beast instead.

I noticed that adding `[scrolling=no]` attribute to the `iframe` and setting it height to the exact value of the wrapping `div` eliminates scrolling. But that's just ½ of the solution, working only for the first line. To apply this workaround to the lines in the middle of the editor, I had to add negative `margin-top` to editable itself to put the line being edited into view.

After `touchstart` OTOH I'm resetting everything and scrolling editor to appropriate line (as removing `[scrolling=no]` and `height` scrolls to the beginning of the editor).

The solution is just insane, but _it works_ without flickering. However I feel dirty now…

Pushed my solution to `t/498-scrolling-no`.

PR #1867, opened 2018-04-06
Add CKEDITOR.dom.rect abstract type

### LINKED ISSUE #1866: Create an abstract API doc type for DOM rectangle

Task

Some methods in our API are returning objects, similar to DOMRect type.

For instance currently we have [`element.getClientRect()`](https://docs.ckeditor.com/ckeditor4/latest/api/CKEDITOR_dom_element.html#method-getClientRect). Currently we're implementing `range.getClientRects()` (#1499) which will also use it.

The best name for it seems to be `CKEDITOR.dom.rect`. It should be an abstract type.

Based on https://github.com/ckeditor/ckeditor-dev/pull/1627#discussion_r178267720 comment.

Comandeer:

Task

**Nope**.

## What changes did you make?

I've added abstract `CKEDITOR.dom.rect` class and modify return type for `CKEDITOR.dom.element.getClientRect`.

Closes #1866.

### DIFF
--- a/core/dom/rect.js
+++ b/core/dom/rect.js
@@ -1,0 +1,69 @@
+/**
+ * @license Copyright (c) 2003-2018, CKSource - Frederico Knabben. All rights reserved.
+ * For licensing, see LICENSE.md or https://ckeditor.com/legal/ckeditor-oss-license
+ */
+
+/**
+ * @fileOverview Defines the "virtual" {@link CKEDITOR.dom.rect} class
+ * which contains the defintion of an element's DOM rectangle. This file is for
+ * documentation purposes only.
+ */
+
+/**
+ * Virtual class that illustrates the {@link CKEDITOR.dom.element} DOM rectangle (dimensions and coordinates
+ * of the area, which element occupies on the page).
+ *
+ * @class CKEDITOR.dom.rect
+ * @abstract
+ * @since 4.10.0
+ */
+
+/**
+ * Element's offset from the top.
+ *
+ * @property {Number} top
+ */
+
+/**
+ * Element's offset from the bottom.
+ *
+ * @property {Number} bottom
+ */
+
+/**
+ * Element's offset from the left.
+ *
+ * @property {Number} left
+ */
+
+/**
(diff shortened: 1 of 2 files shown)


core/dom/rect.js
```diff
@@ -0,0 +1,69 @@
...
+ */
+
+/**
+ * Element's offset from the top.
+ *
+ * @property {Number} top
+ */
+
+/**
+ * Element's offset from the bottom.
```
mlewand: By offset from bottom I would understand distance to the bottom edge of the viewport. This is not the case.

Bottom contains offset from the top edge of the viewport.

core/dom/rect.js
```diff
@@ -0,0 +1,69 @@
...
+ */
+
+/**
+ * Element's offset from the left.
+ *
+ * @property {Number} left
+ */
+
+/**
+ * Element's offset from the right.
```
mlewand: It's a relative value to the left edge of the viewport.

core/dom/rect.js
```diff
@@ -0,0 +1,69 @@
...
+ */
+
+/**
+ * Element's width.
+ *
+ * @property {Number} width
+ */
+
+/**
+ * Element's offset from the viewport's left edge.
```
mlewand: Here you included information that it's relative to the viewport 👍

core/dom/rect.js
```diff
@@ -0,0 +1,69 @@
...
+ * documentation purposes only.
+ */
+
+/**
+ * Virtual class that illustrates the {@link CKEDITOR.dom.element} DOM rectangle (dimensions and coordinates
+ * of the area, which element occupies on the page).
+ *
+ * @class CKEDITOR.dom.rect
+ * @abstract
+ * @since 4.10.0
```
mlewand: This abstract type was not introduced in 4.10.0 - it has been around for ages. We can safely skip the `@since` tag.

core/dom/rect.js
mlewand: Also I'd love to se an extra line between these two.

Issue #1881, opened 2018-04-10
Provide possibility to not merge identical ancestor elements after text removing


msamsel:
## Are you reporting a feature request or a bug?

Feature

## Provide detailed reproduction steps (if any)

Issue #541 editor start to use `extractSelectedHtml` to handle webkit quirks. This method has implemented cleaning up, which tries to merge identical elements. This will cause situation when blocks will be merged after deletion. For example if you use markup as below, select content between brackets and remove it, then outer divs might combine into one:
```
<div>
    <p>fo[o</p>
</div>
<div>
    <p>b]ar</p>
</div>
```

It could be nice to introduce way to prevent such situations. It might be useful in some edge cases, when content is styles with some specific way, for example `div`s could have 100px margin.

## Other details

* Browser: Chrome
* OS: MacOS
* CKEditor version: 4.10.0+
* Installed CKEditor plugins: …

Issue #1880, opened 2018-04-10
extractSelectedHtml wrongly remove nested tables


msamsel:
## Are you reporting a feature request or a bug?

BUG

## Provide detailed reproduction steps (if any)

1. Open CKEditor
2. Use this source as editor input:
```
<p>Fo[o</p>

<table border="1">
	<tbody>
		<tr>
			<td>AA</td>
			<td>
			<table border="1">
				<tbody>
					<tr>
						<td>1]1</td>
						<td>33</td>
					</tr>
					<tr>
						<td>22</td>
						<td>44</td>
					</tr>
				</tbody>
			</table>
			</td>
		</tr>
		<tr>
			<td>BB</td>
			<td>&nbsp;</td>
		</tr>
	</tbody>
</table>

<p>Bar</p>
```
3. Make selection according to square brackets
4. Execute command `editor.extractSelectedHtml()`.

### Expected result

Table's cells will be properly cleared out, but won't be removed.

### Actual result

Table cells are removed

## Other details
Extracted from #541

* Browser: …
* OS: …
* CKEditor version: …
* Installed CKEditor plugins: …

PR #1882, opened 2018-04-10
Fix for wrong deletion of mixed text and table content

### LINKED ISSUE #541: Partial selection outside and inside the table cause cells to be deleted when pressing Del or Backspace

Bug

https://dev.ckeditor.com/ticket/16863

1. Open attached HTML and paste it into editor.
2. Make a selection staring in `^except DSO` and ending in `dk^`
3. Press delete or backspace 

### Expected result

Table contents get deleted but table stays 3X3 

### Actual result

Cells 1.1 get removed 

## Other details

**EDIT**: **Problem can also be reproduced when selecting content between cells. Select whole cell 1.1 and part of cell 1.2**. 

* Browser: Webkit 
* OS: Windows, Mac
* CKEditor version: 4.4.2

msamsel:

Bug fix

### This PR contains

- [x] Unit tests
- [x] Manual tests

## What changes did you make?
* use `editor.extractSelectedHtml` instead of current logic to remove content from the editor after pressing <kbd>Backspace</kbd> or <kbd>Delete</kbd>
* correct bunch of test
* Unfortunately this solution has 2 minor things:
    * nested tables are still wrongly processed what is caused by this issue: #1880
    * identical elements might be merged #1881 (because of that test `#15` and `#16` are changed)

Close #541

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -16,6 +16,7 @@ Fixed Issues:
 * [#1776](https://github.com/ckeditor/ckeditor-dev/issues/1776): [Image Base](https://ckeditor.com/cke4/addon/imagebase) empty caption placeholder is not hidden when blurred.
 * [#1592](https://github.com/ckeditor/ckeditor-dev/issues/1592): [Image Base](https://ckeditor.com/cke4/addon/imagebase) caption is not visible after paste.
 * [#620](https://github.com/ckeditor/ckeditor-dev/issues/620): Fixed: [`forcePasteAsPlainText`](https://docs.ckeditor.com/ckeditor4/latest/api/CKEDITOR_config.html#cfg-forcePasteAsPlainText) will be respected when internal and cross-editor pasting happens.
+* [#541](https://github.com/ckeditor/ckeditor-dev/issues/1592): Selection containing text and tables, will now properly support <kbd>Backspace</kbd> and <kbd>Delete</kbd> keys.
 
 API Changes:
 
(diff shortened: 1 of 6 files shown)


mlewand:
## Firefox

I had a look with @msamsel during a f2f talk and the Firefox thing looks clearly like their feature (if you have selected same element on both ends, here paragraph, and delete the content then merge it). We should not worry about that at this point.

## Edge / IE

As of Edge it is clearly an upstream issue in Edge/IE. Normally I'd say that all we need to do here is report upstream issue and not worry about it, but it will look kind of silly if we say in changelog that we fixed this case, but it does not work in Edge / IE.

With above in mind, please:

* Create a feature request to add handling for IE / Edge.
* Report an upstream issue, and link it properly.

So this PR should not worry about IE/Edge support, that will be done in a follow-up PR.

msamsel:
> There is also one thing that makes me wonder: why didn't we use editor.extractSelectedHtml at first place?

I think because `editor.extractSelectedHtml` was developed 3 years ago, and this corrected part was developed 4 years ago ;P

msamsel:
@mlewand could you provide feedback what behaviour should be implemented in CKE4?

mlewand:
Note that you could also give a try with `document.execCommand` approach, e.g. `document.execComand( 'delete' )`. Although, for me, it should be exact sam logic as pressing the <kbd>delete</kbd> key, after brief checking it looks like Chrome implements actually a different logic there.

mlewand:
After f2f talk with @msamsel seems that `delete` command looks promising, the problem is that `range.scrollIntoView()` breaks the selection if anchored in a text (thus #1956). @msamsel as for fighting the dissapearing selection temporarily, could you try to make bookmark? That should help mitigate the issue.

PR #1898, opened 2018-04-18
Add Latin small letter e with macron to specialchar plugin

danielkorte:

Add Latin small letter e with macron to specialchar plugin.


mlewand:
Hi, sorry for the delay, it's been a busy period for us during 4.10.0 release 🙂

Now regarding this pull request, thanks for the code, however if you look at our [config for Special Chars plugin](https://docs.ckeditor.com/ckeditor4/latest/api/CKEDITOR_config.html#cfg-specialChars) we made a dedicated config option just for this case. Values in this dialog are a subject of opinion and particular use case. We found that current values are good for start, but we know that it won't fit all use cases, thus the config option.

Please use just the config option for your customization (you could also remove options that are not needed for your use case) 😉

Issue #1046, opened 2017-10-13
Link dialog: id not set after the first time


TomBosmans:
## Are you reporting a feature request or a bug?

type: Bug

Referece link: http://nightly.ckeditor.com/17-10-13-06-04/full/samples/

## reproduction steps

1. Create a link with id (in the advance tap)
2. Edit the link (the id is still filled in)
3. Create a second link with id
4. Edit the second link

### Expected result
The id is filled in

### Actual result
The id is empty

## Other details
* Browser: Firefox
* OS: MacOs High Sierra
* CKEditor version: version of the sample, maybe older ones as well?
* Installed CKEditor plugins: Nothing special, same as the sample

beatadelura:
I'm able to reproduce this issue from CKEditor version 4.3.0.

PR #1914, opened 2018-04-20
Make sure the ID is always inserted if entered in the advanced dialog

nathansamson:

Fixes issue #1046

Not yet - will add later

### This PR contains

- [ ] Unit tests
- [ ] Manual tests

## What changes did you make?

When creating a new element based on style it would remote the ID attribute to prevent duplicates.
This bit of code was added in 2011 to prevent an issue with copying links (according to the git blame and git log 235dc2a7edb21a6b8514ce189121e721b5ab2298) - I did not retest this use case as I did not know where to find this issue
I modified it so it only removes ID's if the chosen ID already exists in the targetDocument

### DIFF
--- a/core/style.js
+++ b/core/style.js
@@ -1662,7 +1662,8 @@ CKEDITOR.STYLE_OBJECT = 3;
 
 		// Avoid ID duplication.
 		if ( targetDocument.getCustomData( 'doc_processing_style' ) && el.hasAttribute( 'id' ) )
-			el.removeAttribute( 'id' );
+			if ( targetDocument.getById(el.getAttribute('id')) != null )
+				el.removeAttribute( 'id' );
 		else
 			targetDocument.setCustomData( 'doc_processing_style', 1 );
 


nathansamson:
I have trouble understanding the test suite and how to write an appropriate test for this...

core/style.js
```diff
@@ -1662,7 +1662,8 @@ CKEDITOR.STYLE_OBJECT = 3;
 
 		// Avoid ID duplication.
 		if ( targetDocument.getCustomData( 'doc_processing_style' ) && el.hasAttribute( 'id' ) )
-			el.removeAttribute( 'id' );
+			if ( targetDocument.getById(el.getAttribute('id')) != null )
```
mlewand: This is not a proper fix for this. Actually the problem originates from bad `doc_processing_style` custom data management.

If you trace the code changes, you'll notice that it was moved from `setupElement` function to `applyStyleOnSelection` - and this is exactly when the bug was introduced.

This attribute was used to, sort of, mark a single transaction - but in case of editing an existing link this function wasn't even called, thus it didn't clear this "switch".

All that is needed here is to restore it and it should work correctly.

mlewand: Also the code style is inconsistent, see spacing next to parenthesis.

mlewand:
Thanks for the contribution, since it has been a while since you pushed the PR we'll take it from here and add the missing parts in a follow-up pull request.

PR #1917, opened 2018-04-23
Get absolute rects via range.getClientRects

### LINKED ISSUE #1902: Add tests for range.getClientRects() with fake selection

Task

One part that we have missed during #1498 is an integration test with fake selection.

Fake selection happens in several places:

* When selecting/focusing a widget.
* When selecting a table with `tableselection` plugin.

Let's add a manual test that works similarly to existing [getclientrects](https://github.com/ckeditor/ckeditor-dev/blob/major/tests/core/dom/range/manual/getclientrects.md) test but includes `codesnippet` and `tableselection` plugins.

### LINKED ISSUE #1915: Add possibility to retrieve absolute rects via `range.getClientRects`

Feature

`CKEDITOR.dom.element.getClientRect` returns absolute positioned rect when passing first argument as true. We should add same option to `CKEDITOR.range.getClientRects` too.

engineering-this:

Feature/task

### This PR contains

- [ ] Unit tests
- [x] Manual tests

## What changes did you make?

I've added option to get absolute rects via `range.getClientRects` and added manual test with table selection and code snippet plugins.

Closes #1902 
Closes #1915

### DIFF
--- a/core/dom/range.js
+++ b/core/dom/range.js
@@ -2849,11 +2849,12 @@ CKEDITOR.dom.range = function( root ) {
 		 * <img src="https://33333.cdn.cke-cs.com/rc1DFuFpHqcR3Mah6y0e/images/90893fcc6c323c10023e73ebfc1fbaa622b48b29c066f7af_ie-rects.png">
 		 *
 		 * @since 4.10.0
+		 * @param {Boolean} [isAbsolute=false] The function will retrieve an absolute rectangle of the element i.e. position relative to the upper-left corner of the topmost viewport.
 		 * @returns {CKEDITOR.dom.rect[]}
 		 */
 		getClientRects: ( function() {
 			if ( this.document.getSelection !== undefined ) {
-				return function() {
+				return function( isAbsolute ) {
 					// We need to create native range so we can call native getClientRects.
 					var range = document.createRange(),
 						rectArray;
@@ -2862,8 +2863,8 @@ CKEDITOR.dom.range = function( root ) {
 					range.setEnd( this.endContainer.$, this.endOffset );
 
 					rectArray = CKEDITOR.tools.array.map( range.getClientRects(), function( item ) {
-						return convertRect( item );
-					} );
+						return convertRect( item, isAbsolute, this );
+					}, this );
 
 					// In some cases ( eg. ranges contain only image ) IE will return empty rectList.
 					if ( !range.collapsed && !rectArray.length ) {
@@ -2875,18 +2876,50 @@ CKEDITOR.dom.range = function( root ) {
 					return rectArray;
 				};
 			} else {
-				return function() {
-					return [ getRect( this.createBookmark() ) ];
+				return function( isAbsolute ) {
+					return [ convertRect( getRect( this.createBookmark() ), isAbsolute, this ) ];
 				};
 			}
 
 			// Extending empty object with rect, to prevent inheriting from DOMRect, same approach as in CKEDITOR.dom.element.getClientRect().
-			function convertRect( rect ) {
+			function convertRect( rect, isAbsolute, context ) {
 				var newRect = CKEDITOR.tools.extend( {}, rect );
+
+				if ( isAbsolute ) {
+					appendParentFramePosition( context.document.getWindow().getFrame() );
+
+					var winGlobalScroll = CKEDITOR.document.getWindow().getScrollPosition();
+
+					newRect.top += winGlobalScroll.y;
+					newRect.left += winGlobalScroll.x;
+
+					newRect.y += winGlobalScroll.y;
+					newRect.x += winGlobalScroll.x;
+
+					newRect.right = newRect.left + newRect.width;
+					newRect.bottom = newRect.top + newRect.height;
+				}
+
 				// Some browsers might not return width and height.
 				!newRect.width && ( newRect.width = newRect.right - newRect.left );
 				!newRect.height && ( newRect.height = newRect.bottom - newRect.top );
 				return newRect;
+
+				function appendParentFramePosition( frame ) {
+					if ( !frame ) {
+						return;
+					}
+
+					var frameRect = frame.getClientRect();
+
+					newRect.top += frameRect.top;
+					newRect.left += frameRect.left;
+
+					newRect.x += frameRect.x;
+					newRect.y += frameRect.y;
+
+					appendParentFramePosition( frame.getWindow().getFrame() );
+				}
 			}
 
 			// Fallback helper for browsers that don't support native getClientRects().
--- a/tests/core/dom/range/manual/getclientrectsfakesel.html
+++ b/tests/core/dom/range/manual/getclientrectsfakesel.html
@@ -1,0 +1,139 @@
+<style>
+	body {
+		padding-top: 150px;
+		margin-left: 0px;
+		padding-left: 330px;
+	}
+
+	#editable {
+		position: absolute;
+		width: 300px;
+		left: 600px;
+	}
+
+	#editable img {
+		width: 100px;
+		height: 30px;
+		vertical-align: text-bottom;
+	}
+
+	.marker {
+		position: absolute;
+		outline: 1px solid red;
+		pointer-events: none;
+	}
+
+	#output {
+		width: 250px;
+	}
+
+</style>
+
+<textarea name="output" id="output" cols="30" rows="10" readonly="readonly"></textarea>
+
+<div id="editor" contenteditable="true">
+	<p>Some text.</p>
+	<table border="1" cellspacing="1" cellpadding="1" style="width:250px">
+		<tbody>
+		<tr>
+			<td>Cell</td>
+			<td>Cell</td>
+			<td>Cell</td>
+		</tr>
+		<tr>
+			<td>Cell</td>
+			<td>Cell</td>
+			<td>Cell</td>
+		</tr>
+		<tr>
+			<td>Cell</td>
+			<td>Cell</td>
+			<td>Cell</td>
+		</tr>
+		</tbody>
+	</table>
+	<pre>
+		<code class="language-javascript">console.log( 'Hello World!' );</code>
+	</pre>
+</div>
+
+
+<script>
+	( function() {
+		// IE 8 polyfill doesn't work when selecting block elements like widget. Also it's unsupported by table selection.
+		if ( CKEDITOR.env.ie && CKEDITOR.env.version === 8 ) {
+			bender.ignore()
+		}
+
+		CKEDITOR.replace( 'editor', {
+			height: 500,
+			on: {
+				instanceReady: function() {
+					var rectMarkers = [];
+
+					var doc = CKEDITOR.document,
+						editable = this.editable(),
+						par = editable.findOne( 'p' ),
+						rng = new CKEDITOR.dom.range( editable ),
+						output = doc.getById( 'output' ),
+						editor = this;
+
+					rng.setStart( par.getParent(), par.getIndex() );
+					rng.setEnd( par.getParent(), par.getIndex() + 1 );
+
+					this.getSelection().selectRanges( [ rng ] );
+
+					window.setInterval( function() {
+						var ranges = editor.getSelection().getRanges(),
+							rects = ranges && CKEDITOR.tools.array.reduce( ranges, function( total, current ) {
+								return total.concat( current.getClientRects( true ) );
+							}, [] );
+
+						output.setValue( 'Selection rect\n---\nrectangle count: ' + ( rects ? rects.length : rects ) );
+
+						removeRectangles();
+
+						if ( !rng || !rects || !rects.length ) {
+							return;
+						}
+
+						output.setValue( output.getValue() + '\nwidth:' + rects[ 0 ].width + '\nheight: ' + rects[ 0 ].height );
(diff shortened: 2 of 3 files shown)


core/dom/range.js
```diff
@@ -2849,11 +2849,12 @@ CKEDITOR.dom.range = function( root ) {
 		 * <img src="https://33333.cdn.cke-cs.com/rc1DFuFpHqcR3Mah6y0e/images/90893fcc6c323c10023e73ebfc1fbaa622b48b29c066f7af_ie-rects.png">
 		 *
 		 * @since 4.10.0
+		 * @param {Boolean} [isAbsolute=false] The function will retrieve an absolute rectangle of the element i.e. position relative to the upper-left corner of the topmost viewport.
```
mlewand: Please split the line into two, try not exceed 160 chars in a line.

engineering-this: Ok, but shouldn't linters check for that value?

mlewand:
@Comandeer feel free to take over the R of this PR

core/dom/range.js
```diff
@@ -2843,50 +2843,125 @@ CKEDITOR.dom.range = function( root ) {
...
 
 					range.detach();
 
-					return rectArray;
+					return CKEDITOR.tools.array.map( rectList, function( item ) {
+						return convertRect( item, isAbsolute, this );
+					}, this );
 				};
 			} else {
-				return function() {
```
Comandeer: Can't this version of `getClientRects` be just deleted? It looks exactly like line 2909. In such case I'd think that the whole `getClientRects` could be further simplified.

core/dom/range.js
```diff
@@ -2843,50 +2843,125 @@ CKEDITOR.dom.range = function( root ) {
...
+					return [ convertRect( getRect( this.createBookmark() ), isAbsolute, this ) ];
 				};
 			}
 
 			// Extending empty object with rect, to prevent inheriting from DOMRect, same approach as in CKEDITOR.dom.element.getClientRect().
-			function convertRect( rect ) {
+			function convertRect( rect, isAbsolute, context ) {
 				var newRect = CKEDITOR.tools.extend( {}, rect );
+
+				if ( isAbsolute ) {
```
Comandeer: This is exactly the same logic as in `CKEDITOR.dom.element#getClientRects( true )`. I'm thinking about moving it into one place. Maybe to `CKEDITOR.dom.rect`? 🤔 @mlewand @engineering-this WDYT?

mlewand: Yup this looks like a copy'n'pasta so it should be extracted to a common place. I don't think that `CKEDITOR.dom.rect` is a good place though.

I'd start with a helper function for that, for instance in tools. As we'd get more members like that we can put them into a dedicated type, and deprecate old methods.

engineering-this:
>I'm not sure, but selecting the widget should create only one rect, for the whole widget. This is the behavior in Firefox.

@mlewand @Comandeer 
I can't think of anything easier than:

1. Find all widgets in ranges.
2. For each widget create new native ranges which contains only widget and call `getClientRets`.
3a. For initial rect list filter out every rect that are same as the ones from step 2 except for the first one which represents the wrapper.

Alternative that might be a bit better on performance
3b. If step 2 returns list of 7 rects, then find index of first rect inside our rect list, and remove all rects with indexes from i + 1 to i + 6. But here we are assuming that rect list on our ranges has same order as rect list from ranges with only widget inside.

core/dom/range.js
engineering-this: It can be deleted only if some changes are added to code. Every call of native `range.getClientRects()` would require a check if that method exist, since IE8 doesn't support it. Possibly other checks would be needed too. Why perform all the logic on that browser when it won't do anything useful?

CHANGES.md
```diff
@@ -16,13 +16,16 @@ Fixed Issues:
 * [#1776](https://github.com/ckeditor/ckeditor-dev/issues/1776): [Image Base](https://ckeditor.com/cke4/addon/imagebase) empty caption placeholder is not hidden when blurred.
 * [#1592](https://github.com/ckeditor/ckeditor-dev/issues/1592): [Image Base](https://ckeditor.com/cke4/addon/imagebase) caption is not visible after paste.
 * [#620](https://github.com/ckeditor/ckeditor-dev/issues/620): Fixed: [`forcePasteAsPlainText`](https://docs.ckeditor.com/ckeditor4/latest/api/CKEDITOR_config.html#cfg-forcePasteAsPlainText) will be respected when internal and cross-editor pasting happens.
+* [#1925](https://github.com/ckeditor/ckeditor-dev/issues/1925): Fixed: `range.getClientRects` doesn't work correctly on IE9-IE10 with collapsed selection at the end of line.
```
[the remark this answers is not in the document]
mlewand: Yes, we should skip the changelog in case it goes in 4.10.0.

mlewand:
We can start by supporting case when a single widget is selected (when it gets a blue outline). As for selection that contains multiple widgets, we could skip it for now.

Comandeer:
In case of one widget I'm not sure if more naive approach wouldn't be even more suitable, e.g. `widgetElement.getClientRects`.

engineering-this:
https://github.com/ckeditor/ckeditor-dev/pull/1917#issuecomment-388008336

@mlewand 
By the time of this comment I had it done. Multiple widgets it works well. However after some testing I found that MS browsers including Edge which adds one more rect for selection part of parent element.

@Comandeer 
For issues with IE8 we should create separate ticket for it. Fixing IE8 polyfill would require to rewrite it and add multiple cases of ranges.

core/tools.js
```diff
@@ -2152,6 +2152,47 @@
 
 				return copy1;
 			}
+		},
+
+		/**
+		 * Returns rect with absolute position.
```
Comandeer: Please be more precise and describe what absolute position means.

core/dom/range.js
```diff
@@ -2843,46 +2843,161 @@ CKEDITOR.dom.range = function( root ) {
...
+					widgetRange,
+					nodeList;
+
+				// Lets iterate over each element inside ranges to find widgets.
+				while ( element = iterator.getNextParagraph() ) {
+					nodeList = element.find( '[data-widget]' );
+					// Store each widget element in array.
+					if ( element.hasAttribute( 'data-widget' ) ) {
+						widgetElements.push( element );
+
```
Comandeer: Redundant empty line.

core/dom/range.js
```diff
@@ -2843,46 +2843,161 @@ CKEDITOR.dom.range = function( root ) {
...
+					} else if ( nodeList.$.length ) {
+						widgetElements = widgetElements.concat( CKEDITOR.tools.array.map( nodeList.$, function( item ) {
```
Comandeer: Why don't you use our node list, but native one?

core/dom/range.js
```diff
@@ -2843,46 +2843,161 @@ CKEDITOR.dom.range = function( root ) {
...
+					widgetElements = [],
+					widgetRects,
+					widgetRange,
+					nodeList;
+
+				// Lets iterate over each element inside ranges to find widgets.
+				while ( element = iterator.getNextParagraph() ) {
+					nodeList = element.find( '[data-widget]' );
+					// Store each widget element in array.
+					if ( element.hasAttribute( 'data-widget' ) ) {
```
Comandeer: `CKEDITOR.dom.node#getAscendant` with second parameter set to `true` should be able to simplify this logic.

core/tools.js
```diff
@@ -2152,6 +2152,47 @@
...
+		 *
+		 * @since 4.10.0
+		 * @param { CKEDITOR.dom.window } Window containing element for which rect is passed.
+		 * @param { CKEDITOR.dom.rect } Rect with relative position.
```
Comandeer: These parameters lack names in API docs.

core/tools.js
```diff
@@ -2152,6 +2152,47 @@
...
+		 * @returns { CKEDITOR.dom.rect } Rect with absolute position.
+		 */
+		getAbsoluteRectPosition: function( window, rect ) {
```
Comandeer: I'll add unit test for this method.

core/dom/range.js
engineering-this: `getAscendant` returns ancestor, and I'm iterating over children, not sure how I can use this method here.

Comandeer: Setting the second parameter to `true` also fetches the element itself. But now I see that this logic is more complicated than I previously thought as it has to find all widgets inside the widget.

engineering-this:
>In Chrome trying to select these two widget from the end results in tab's crash.

This happens on major, so its unrelated.

tests/core/dom/range/getclientrectswidget.js
```diff
@@ -0,0 +1,67 @@
...
+				allowedContent: true
+			}
+		},
+		inline: {
+			name: 'inline',
+			creator: 'inline'
+		}
+	};
+
+	var tests = {
```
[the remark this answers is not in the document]
Comandeer: There are also no tests for nested editables and inline widgets.

tests/core/dom/range/manual/getclientrectswitheditor.html
```diff
@@ -0,0 +1,135 @@
...
+	</table>
+	<pre>
+		<code class="language-javascript">console.log( 'Hello World!' );</code>
+	</pre>
+</div>
+
+
+<script>
+	( function() {
+		// IE 8 polyfill doesn't work when selecting block elements like widget. Also it's unsupported by table selection.
```
Comandeer: Is this reported as separate issue?

config.js
```diff
@@ -65,6 +65,7 @@ CKEDITOR.editorConfig = function( config ) {
 		'tableselection,' +
 		'tabletools,' +
 		'templates,' +
+		'placeholder,' +
```
Comandeer: That change shouldn't be committed.

tests/core/dom/range/manual/getclientrectswitheditor.md
```diff
@@ -0,0 +1,27 @@
+@bender-tags: 4.10.0, feature, range, selection, 1902, 1915, 1925, 1930, tableselection
+@bender-ckeditor-plugins: wysiwygarea, toolbar, codesnippet, image2, table, tableselection, sourcearea, placeholder
```
Comandeer: It would be nice to have also second editor in this test, without `tableselection` plugin as native table selection is also broken.

tests/core/tools/absoluterectposition.js
```diff
@@ -0,0 +1,48 @@
+/* bender-tags: editor, tools */
+
+( function() {
+	'use strict';
+	bender.editor = {
+		name: 'classic'
+	};
+
+	bender.test( {
+		'test getting absolute rect': function() {
```
Comandeer: It would be nice to have test for getting absolute rect also for widget.

core/dom/range.js
```diff
@@ -2843,46 +2843,167 @@ CKEDITOR.dom.range = function( root ) {
...
 		 * Internet Explorer 8 doesn't have implemented native `range.getClientRects()` which is base for this method.
-		 * As a workaround it will return an array containing only one rectangle which would start in the top left hand corner of the selection and end in the bottom right hand corner.
+		 * As a workaround it will return an array containing only one rectangle which would
+		 * start in the top left hand corner of the selection and end in the bottom right hand corner.
 		 * Possible cases when returned rectangle doesn't fully cover ranges are presented below:
 		 *
 		 * <img src="https://33333.cdn.cke-cs.com/rc1DFuFpHqcR3Mah6y0e/images/90893fcc6c323c10023e73ebfc1fbaa622b48b29c066f7af_ie-rects.png">
 		 *
 		 * @since 4.10.0
+		 * @param {Boolean} [isAbsolute=false] The function will retrieve an absolute
```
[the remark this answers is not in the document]
Comandeer: Still not addressed.

core/dom/range.js
```diff
@@ -2843,46 +2843,231 @@ CKEDITOR.dom.range = function( root ) {
...
+					newRange.setStart( context.startContainer.getParent(), context.startContainer.getIndex() );
+				}
+				if ( moveEnd ) {
+					newRange.setEnd( context.endContainer.getParent(), context.endContainer.getIndex() + 1 );
+				}
+				if ( moveStart || moveEnd ) {
+					context = newRange;
+				}
+
+				documentFragment = context.cloneContents();
```
Comandeer: Can't this logic be greatly simplified by omitting walker altogether? Something like:

```javascript
var selectedNodes = context.cloneContents(),
	widgets = CKEDITOR.dom.document.prototype.find.call( selectedNodes, '[data-cke-widget-id]' ).toArray();

CKEDITOR.tools.forEach( widgets, function() {
	// Logic for every widget.
} );
```

engineering-this:
>In IE 10 when inline widgets are selected, there is still rect for paragraph that contains them:
>However I'm not sure if it's fixable in any sensible way.

It's just how it works on MS browsers, same for Edge. Lets just stay with how it is. With our uses of `getClientRects` I don't think it would be problem.

tests/core/tools/absoluterectposition.js
```diff
@@ -0,0 +1,55 @@
+/* bender-tags: editor, tools */
+
+( function() {
+	'use strict';
+	bender.editor = {
+		name: 'classic'
```
Comandeer: It'll be nice to add inline editor also in this unit test.

PR #1757, opened 2018-03-07
Introduced mentions plugin

jacekbogdanski:
## What is the purpose of this pull request?

New feature
## Does your PR contain necessary tests?

All patches which change the editor code must include tests. You can always read more
on [PR testing](https://docs.ckeditor.com/ckeditor4/docs/#!/guide/dev_contributing_code-section-tests),
[how to set the testing environment](https://docs.ckeditor.com/ckeditor4/docs/#!/guide/dev_tests) and
[how to create tests](https://docs.ckeditor.com/ckeditor4/docs/#!/guide/dev_tests-section-creating-your-own-test)
in the official CKEditor documentation.

### This PR contains

- [x] Unit tests
- [x] Manual tests

## What changes did you make?

Introduced mentions plugin.

~Based on #1727~

Closes #1703

PR #1922, opened 2018-04-25
Introduced autocomplete plugin

### LINKED ISSUE #1751: Autocomplete plugin

Feature request

We need a component that provides an ability to display a dropdown with options matched to the currently typed text. This feature is needed for Mentions #1703 and Emoji #1746 features.

This feature should provide a flexible API, it should work both synchronously and asynchronously.

It should allow for customizing the way entries are displayed.

## Preview

Users interested in seeing how the feature looks today can preview it on [our Autocomplete feature branch preview](https://ckeditor4-preview.ckeditor.com/t/1751/plugins/autocomplete/samples/smiley.html).

## Browser Support

Currently we're looking to support the feature in all [supported browsers](https://docs.ckeditor.com/ckeditor4/latest/guide/dev_browsers.html#officially-supported-browsers) except IE8, QM and CM.

jacekbogdanski:

New feature

### This PR contains

- [x] Unit tests
- [x] Manual tests

## What changes did you make?

This PR continues work on #1751 plugin. See #1757 for more details.

Closes #1751


plugins/autocomplete/plugin.js
```diff
@@ -0,0 +1,1204 @@
...
+	 *			CKEDITOR.plugins.customAutocomplete = CustomAutocomplete;
+	 *		}
+	 *	} );
+	 * ```
+	 *
+	 * @class CKEDITOR.plugins.autocomplete
+	 * @since 4.10.0
+	 * @constructor Creates the new instance of autocomplete and attaches it to the editor.
+	 * @param {CKEDITOR.editor} editor The editor to watch.
+	 * @param {Function} textTestCallback Callback executed to check if a text next to the selection should open
```
msamsel: I think there is missing information that `textTestCallback` is called with `range` argument, what is showed in example in general description of the plugin.

plugins/autocomplete/plugin.js
```diff
@@ -0,0 +1,1204 @@
...
+
+		/**
+		 * Returns HTML that should be inserted into the editor when the item is committed.
+		 *
+		 * See also the {@link #commit} method.
+		 *
+		 * @param {CKEDITOR.plugins.autocomplete.model.item} item
+		 * @returns {String} The HTML to insert.
+		 */
+		getHtmlToInsert: function( item ) {
```
msamsel: This method force to use `item.name` to be inserted in editor. Where `item.name` is not required parameter. As result of this situation, you cannot use items without name, because those won't be inserted in editor.

mlewand: @msamsel I don't understand what you mean exactly.

1. `item` param is *required* for this mentod.
1. It does not force you to use `item.name` as you're free to override it.

msamsel: @mlewand 
I'm aware that I need to overwrite entire method.
First I thought that there might exists template for inserted content similar to templet for displaying autocomplete item. I hoped that those two parts will work in a similar way.

plugins/autocomplete/plugin.js
```diff
@@ -0,0 +1,1204 @@
...
+
+	/**
+	 * The unique ID of the item. The ID should not change with time, so two
+	 * {@link CKEDITOR.plugins.autocomplete.model#dataCallback}
+	 * calls should always result in the same ID for the same logical item.
+	 * This can for example allow to keep the same item selected when
+	 * the data changes.
+	 *
+	 * **Note:** When using a string as an item make sure that the string does not
+	 * contain any special characters (above all `"[]` characters). This limitation is
```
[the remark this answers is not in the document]
jacekbogdanski: View item IDs are stored in HTML so they shouldn't contain any special characters or they will break HTML. I think that maybe we could escape IDs before inserting into DOM. WDYT @mlewand?

plugins/textwatcher/plugin.js
```diff
@@ -0,0 +1,254 @@
...
+		 *
+		 * @readonly
+		 * @property {Function}
+		 */
+		this.callback = callback;
+
+		/**
+		 * Keys that should be ignored by the {@link #check} method.
+		 *
+		 * @readonly
```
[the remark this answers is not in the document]
jacekbogdanski: I think it doesn't make sense to use object here. I will refactor this property into an array.

plugins/autocomplete/plugin.js
jacekbogdanski: There is also information: `See the {@link CKEDITOR.plugins.textWatcher}'s `callback` argument` because this callback will be passed directly into `textwatcher`. `textwatcher` already contains information about `range` parameter. I'm not sure if I should duplicate it. WDYT?

plugins/autocomplete/plugin.js
```diff
@@ -0,0 +1,1208 @@
...
+			};
+		},
+
+		/**
+		 * Gets the item element by the item ID.
+		 *
+		 * @param {Number/String} itemId
+		 * @returns {CKEDITOR.dom.element} The item element.
+		 */
+		getItemById: function( itemId ) {
```
msamsel: Method doesn't handle situation when `itemId` contains specific characters. 
e.g.
Try to use as id this string: `:Antigua_&amp;_Barbuda:`

plugins/autocomplete/plugin.js
```diff
@@ -0,0 +1,1208 @@
...
+		init: function() {
+			if ( !stylesLoaded ) {
+				CKEDITOR.document.appendStyleSheet( this.path + 'skins/' + CKEDITOR.skin.name + '/autocomplete.css' );
+				stylesLoaded = true;
+			}
+		}
+	} );
+
+	/**
+	 * The main class implementing a generic {@link https://ckeditor.com/cke4/addon/autocomplete autocomplete} feature in the editor.
```
msamsel: This invalid way of creating links in documentation (Umberto throws an error for this link).
`{@link ...}` is used to link inside docs. Outside docs should be used regular markdown syntax.

jacekbogdanski:
I rebased into the latest `major`. I also replaced `caretposition` plugin with `range.getClientRects` function. No change noticed, except minus almost 500 lines of code! Great job @engineering-this :tada:

plugins/autocomplete/plugin.js
```diff
@@ -0,0 +1,1422 @@
...
+	 * * A data callback &ndash; a function which should return (through its callback) a suggestion data for the current
+	 * query string.
+	 *
+	 * # Creating an autocomplete instance
+	 *
+	 * Depending on your use case, put this code in the {@link CKEDITOR.pluginDefinition#init} callback of your
+	 * plugin or for example in the {@link CKEDITOR.editor#instanceReady} event listener.
+	 *
+	 * ```javascript
+	 *	// Called when the user types in the editor or moves the caret.
```
mlewand: This code is incomplete, as it won't work. Try to provide samples that requires as little work from the reader as possible. Most importantly, sample `itemsArray` should be available.

```js
// Example data feed.
var itemsArray = [ { id: 1, name: '@Andrew' }, { id: 2, name: '@Kate' } ];

// Called when the user types in the editor or moves the caret.
// The range represents the caret position.
function textTestCallback( range ) {
	// We don't want to autocomplete a non-empty selection.
	if ( !range.collapsed ) {
		return null;
	}

	// Use the textmatch plugin which does the tricky job of doing
	// a text search in the DOM. The matchCallback function should return
	// a matching fragment of the text.
	return CKEDITOR.plugins.textMatch.match( range, matchCallback );
}

// Returns a position of the matching text.
// It matches with text starting from the '@' character
// followed by spaces, up to the caret position.
function matchCallback( text, offset ) {
	// Get the text before the caret.
	var left = text.slice( 0, offset ),
		// Will look for an '@' character followed by word characters.
		match = left.match( /@\w*$/ );

	if ( !match ) {
		return null;
	}

	return { start: match.index, end: offset };
}

// Returns (through its callback) the suggestions for the current query.
// Note: the itemsArray variable is our example "database".
function dataCallback( query, range, callback ) {
	// Simple search.
	// Filter the entire items array so only the items that start
	// with the query remain.
	var suggestions = itemsArray.filter( function( item ) {
		return item.name.toLowerCase().indexOf( query.toLowerCase() ) === 0;
	} );

	// Note - the callback function can also be executed asynchronously
	// so dataCallback can do an XHR requests or use any other asynchronous API.
	callback( suggestions );
}

// Finally, instantiate the autocomplete class.
new CKEDITOR.plugins.autocomplete( editor, {
	textTestCallback: textTestCallback,
	dataCallback: dataCallback
} );
```

mlewand: Also it would be nice to mention that user should ensure she or he loads `textmatch` plugin too.

plugins/autocomplete/plugin.js
```diff
@@ -0,0 +1,1422 @@
...
+		 * Listeners registered by this autocomplete instance.
+		 *
+		 * @private
+		 */
+		this._listeners = [];
+
+		/**
+		 * Template of markup to be inserted as the autocomplete item gets committed.
+		 *
+		 * You can use {@link CKEDITOR.plugins.autocomplete.model#data data item} properties to customize the template.
```
mlewand: It should link to the `CKEDITOR.plugins.autocomplete.model.item` type directly.

plugins/autocomplete/plugin.js
```diff
@@ -0,0 +1,1422 @@
...
+	 *		// so dataCallback can do an XHR requests or use any other asynchronous API.
+	 *		callback( suggestions );
+	 *	}
+	 *
+	 * ```
+	 *
+	 * @method dataCallback
+	 * @param {String} query The query string that was accepted by the `textTestCallback`.
+	 * @param {CKEDITOR.dom.range} range The range in the DOM where the query text is.
+	 * @param {Function} callback The callback which should be executed with the data.
```
mlewand: How about

> which should be executed with the data

->

> which should be executed with the matched data

?

plugins/autocomplete/plugin.js
```diff
@@ -0,0 +1,1422 @@
...
+				this.view.open();
+				this.model.selectFirst();
+				this.view.updatePosition( this.model.range );
+			}
+		},
+
+		// LISTENERS ------------------
+
+		/**
+		 * The function registered by the {@link #attach} method as the
```
mlewand: This function should be marked as private.

Please do that for other listeners.

mlewand: It also should document what the method does instead of who is using it.

E.g. Function to be called when once content (?) has changed.

I started to wonder, shouldn't we rename it to `reposition`? 🤔 I think it would be a better name. Then again, it's used not only when content changes, but also for scrolls.

plugins/autocomplete/plugin.js
```diff
@@ -0,0 +1,1422 @@
...
+		 * {@link CKEDITOR.editor#change} event listener.
+		 */
+		onChange: function() {
+			if ( this.model.isActive ) {
+				this.view.updatePosition( this.model.range );
+			}
+		},
+
+		/**
+		 * The function registered by the {@link #attach} method as the
```
mlewand: Similarly to the previous one, i'd name it `modelChangeListener` or something like that.

plugins/autocomplete/plugin.js
```diff
@@ -0,0 +1,1422 @@
...
+		 * {@link CKEDITOR.plugins.autocomplete.view#click-item} event listener.
+		 *
+		 * @param {CKEDITOR.eventInfo} evt
+		 */
+		onItemClick: function( evt ) {
+			this.commit( evt.data );
+		},
+
+		/**
+		 * The function registered by the {@link #attach} method as the `keydown` event listener.
```
mlewand: Name of the method seems to be fine, but the description could be simplified. It should not mention where it is registered (there's a high chance someone will change it in the future and will miss this method).

plugins/autocomplete/plugin.js
```diff
@@ -0,0 +1,1422 @@
...
+	 * In order to use a different view, implement a new view class and override
+	 * the {@link CKEDITOR.plugins.autocomplete#getView} method.
+	 *
+	 * ```javascript
+	 *	myAutocomplete.prototype.getView = function() {
+	 *		return new myView( this.editor );
+	 *	};
+	 * ```
+	 *
+	 * You can also modify this class on the fly.
```
mlewand: It's not the class that is being modified here, but an instance. Slightly hacky snippet but I guess it's still acceptable.

plugins/autocomplete/plugin.js
```diff
@@ -0,0 +1,1422 @@
...
+
+		/**
+		 * Closes the panel.
+		 */
+		close: function() {
+			this.element.removeClass( 'cke_autocomplete_opened' );
+		},
+
+		/**
+		 * Creates and returns the view's main element.
```
mlewand: Should be marked as a private property.

plugins/autocomplete/plugin.js
```diff
@@ -0,0 +1,1422 @@
...
+	 * In order to create autocomplete instance use its {@link #constructor constructor}.
+	 *
+	 * @class CKEDITOR.plugins.autocomplete
+	 * @since 4.10.0
+	 * @constructor Creates the new instance of autocomplete and attaches it to the editor.
+	 *
+	 * In order to initialize the autocomplete feature it is enough to instantiate this class with
+	 * two required callbacks:
+	 *
+	 * * A text test callback &ndash; a function which should return a fragment of text (typed in the editor) that
```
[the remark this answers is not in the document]
jacekbogdanski: Looks good.

mlewand:
I'm not removing the branch yet, as other PRs are based on it.

PR #1923, opened 2018-04-25
Edge loses selection position when focused the first time

### LINKED ISSUE #504: Editor viewport is scrolled back to the top when focusing editor for the first time

**Bug**

[Original Report (14827)](http://dev.ckeditor.com/ticket/14827)

**In Microsoft Edge 14**

1. Go to http://ckeditor.com/demo.
2. Scroll editor viewport without focusing editor (e.g. using mouse wheel).
3. Click inside editor content to focus editor.

Optionally, simply do:

1. Go to http://ckeditor.com/demo
2. Click on the word _oven_ inside the editor

_This can also be reproduced in nightly.ckeditor.com_

### Expected result

The cursor to be placed where the mouse was pressed (the oven word) and the scroll position of the editor (if any) to be preserved

### Actual result

The cursor is placed at the beginning of the content and the scroll position (if any) is lost

## Other details

* Browser: Microsoft Edge 14
* OS: Windows 10
* CKEditor version: CKEditor 4.7
* Installed CKEditor plugins: default ones in CKEditor's demo

Comandeer:

Bug fix

### This PR contains

- [x] Unit tests
- [x] Manual tests

## What changes did you make?

It's just a dirty hack for Edge. I've added check in `domFix` to allow reselection in Edge only if the editor was previously focused. To check it I use `editor._.previousActive`.

The real source of the issue is buried deep in our selection. It seems that call to `selection.removeAllRanges` somehow prevents Edge from changing selection. The procedure for clicking the first time into editable area is:

* treat `focus` event as clicking at the beginning of the content
* then treat the click as a separate selection change.

Our `domFix` method interrupts this procedure and changes it to:

* treat `focus` event as clicking at the beginning of the content
* **add missing bogus `br`s and reselect the current selection**
* then treat the click as a separate selection change.

It works this way in IE. However Edge seems to stop after reselecting and does not change selection to the place of the click. But how it is connected with removing ranges – I don't really know.

As this issue is visible only on the first focus and researching our whole flow of focusing and selecting could take very long, I decided to apply some minimal hack instead.

Closes #504.

### DIFF
--- a/tests/core/editable/domfix3.js
+++ b/tests/core/editable/domfix3.js
@@ -1,0 +1,30 @@
+/* bender-tags: editor,edge */
+
+bender.test( {
+	// This test is as dirty as hack for Edge itself. To detect if reselection in domFix is not
+	// happening for the first focus, we stub selection's selectRanges to detect any
+	// selection after focus (#504).
+	'test first focus in Edge': function() {
+		if ( !CKEDITOR.env.edge ) {
+			return assert.ignore();
+		}
+
+		bender.editorBot.create( {
+			name: 'edgefirstfocus',
+			startupData: '<p>Lorem ipsum dolor sit amet</p>'
+		}, function( bot ) {
+			var editor = bot.editor,
+				stub = sinon.stub( CKEDITOR.dom.selection.prototype, 'selectRanges' );
+
+			setTimeout( function() {
+				resume( function() {
+					stub.restore();
+					assert.areSame( 0, stub.callCount, 'reselection count' );
+				} );
+			}, 150 );
+
+			editor.focus();
+			wait();
+		} );
+	}
+} );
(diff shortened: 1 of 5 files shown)


f1ames:
Rebased onto latest `major`.

tests/core/editable/domfix3.js
```diff
@@ -0,0 +1,30 @@
+/* bender-tags: editor,edge */
+
+bender.test( {
+	// This test is as dirty as hack for Edge itself. To detect if reselection in domFix is not
+	// happening for the first focus, we stub selection's selectRanges to detect any
+	// selection after focus (#504).
+	'test first focus in Edge': function() {
```
f1ames: This test checks some internals and doesn't check the result (correctly placed selection) for this fix. I assume creating the test which will focus the editor and then check selection position (and maybe scroll selection) was not plausible? Can you shed some light on this @Comandeer?

Comandeer: IIRC I just couldn't find any sensible way to focus the editor and not create selection at the beginning of the content at the same time. It seems to behave in such way only during user interaction. Due to that it's impossible to test selection position after first focus.

f1ames: That was my assumption. If such behaviour is caused by user interaction only we may go with what you proposed (in such situations we may also simply skip unit test, but since it's already done let's keep it) 👍

f1ames:
Rebased onto latest `major`.

PR #1976, opened 2018-05-15
Embed images pasted from Word in Safari

### LINKED ISSUE #1134: Integration PFW Image with Safari
Task for reminding that PFW Image should be tested with Safari, when upstream issue will be merged.

Task/bug
https://bugs.webkit.org/show_bug.cgi?id=124391

Check how PFW Image behave on Safari. Right now Safari doesn't allow on reading RTF clipboard in browser. Link above shows that soon this will change and proper implementation will be made on Safari side (current safari preview version still doesn't support RTF).
We should test how well PFW Image will integrate with Safari, and correct possible bugs if such appear.

## Other details

* Browser: Safari
* OS: MacOS
* CKEditor version: 4.8.0+
* Installed CKEditor plugins: Paste from word

msamsel:

Bug fix

yes

### This PR contains

- [x] Unit tests
- [x] Manual tests

## What changes did you make?

* Provide possibility to embed images pasted in safari browser
* Fix saves blobUrls detected during pasting images from Word and replacing them later asynchronously in editor's content.

Close #1134


plugins/ajax/plugin.js
```diff
@@ -65,14 +65,25 @@
 			return null;
 		}
 
-		function load( url, callback, getResponseFn ) {
+		function getCustomResponse( xhr ) {
```
Comandeer: Won't it be better to create one `getResponse` function that would return appropriate response instead of creating one function for every response type? Especially that now `ArrayBuffer` is supported only in `load` method and having one function would enable it everywhere with minimal amount of work.

If not, then at least change the function name to something more appropriate, e.g. `getResponseArrayBuffer`.

plugins/pastefromword/plugin.js
```diff
@@ -96,7 +96,18 @@
...
 
-						pfwEvtData.dataValue = CKEDITOR.cleanWord( pfwEvtData.dataValue, editor );
+						var filteredData = CKEDITOR.cleanWord( pfwEvtData.dataValue, editor );
+
+						if ( typeof filteredData === 'string' ) {
+							pfwEvtData.dataValue = filteredData;
+						} else if ( typeof filteredData === 'object' ) {
+							editor.fire( 'lockSnapshot' );
+							pfwEvtData.dataValue = filteredData.dataValue;
+							handleBlobs( filteredData.blobUrls, function() {
```
Comandeer: I'm wondering about time between `lockSnapshot` and `unlockSnapshot`. If user could write anything in this time, it could result in broken or at least weird snapshot.

plugins/pastefromword/plugin.js
```diff
@@ -7,7 +7,7 @@
 	/* global confirm */
 
 	CKEDITOR.plugins.add( 'pastefromword', {
-		requires: 'clipboard',
+		requires: 'clipboard,ajax',
```
Comandeer: Isn't it any other way to not require `ajax` for PfW? It looks a little bit strange. I wonder if `FileReader` would be an alternative here.

plugins/pastefromword/plugin.js
```diff
@@ -167,6 +178,40 @@
...
+				function removeDuplicates( arr ) {
+					return CKEDITOR.tools.array.filter( arr, function( item, index ) {
+						return index === CKEDITOR.tools.array.indexOf( arr, item );
+					} );
+				}
+
+				function replaceBlobUrlsInEditor( map ) {
+					for ( var blob in map ) {
+						var nodeList = editor.editable().find( 'img[src="' + blob + '"]' ).toArray();
+						CKEDITOR.tools.array.forEach( nodeList, function( element ) {
```
Comandeer: Amount of `CKEDITOR.tools` in `handleBlobs` is too high. Let's make some short alias, which will increase the readability.

msamsel:
> I can't get it working with Safari 11.1 (13605.1.33.1.2) or Safari Technology Preview Release 56 (Safari 11.2, WebKit 13606.1.17.2.2) and Microsoft Word 15.34 (170330). The image is not pasted at all.

I test it on Safari 11.1 (13605.1.33.1.4) and MS Word 15.41 (171205) MacOS HighSierra (10.13.4) and it is working completely fine. Maybe you have some strange picture which is not embed as blobUrl by Safari? Could you provide your docx file?


> We're also missing real fixtures for pasting images in Safari, generated from Word.

What do you have on your mind by "real fixtures". Do you want just to have soem real markup in test or you think about integrating it with generated tests? In case of 2nd option, it would require tremendous amount of work to adapt generated test to work with Safari fixtures:
- Fixture would contain bloburl which will be invalid and should be changed or mocked with every single test run, as blob url won't be available
- test should be fired completely asynchronously (it's not possible to read data as image from url synchrnously - only text can be done like that, and converting text to image isn't great idea). We should wait when editors content will be updated after asynchronous events and then compare its content. Right now PFW test check for `data` in event with high priority and cancel it before content is embed in editor. https://github.com/ckeditor/ckeditor-dev/blob/f4e106d5736e5d4211f7371a863d17dd46270e30/tests/plugins/pastefromword/generated/_helpers/promisePasteEvent.js#L23-L29

plugins/pastefromword/plugin.js
msamsel: @Comandeer, yes it's a drawback of this approach. Another option is to make editor somehow readOnly, but I'm not sure what would be better here. WDYT? Should editor be read only during this paste? Or maybe you have some other solutions which I don't see?

plugins/pastefromword/plugin.js
msamsel: I've considered using `FileTools` here. Unfortunately `FileTools` contains logic responsible for uploading files to the server, there is no methods helping downloading files from given URL, where `Ajax` has it and only requires to add new file type to support. That's it was chosen here. I could extend `FileTools`, with new functions, but this seems like new separate task, or hide completely XHR inside PFW and do not rely on `Ajax` plugin.
Using `Ajax` seems to be most convenient and time efficient for me.

@mlewand WDYT? Should we require Ajax in PFW or use some solution?

plugins/pastefromword/plugin.js
Comandeer: Making editor read-only sounds even worse. Let's leave it for now as it shouldn't be so critical issue.

Comandeer:
[Word document with image](https://github.com/ckeditor/ckeditor-dev/files/2044759/image.docx) – not working with Safari 11.1 (13605.1.33.1.4). I tried every single method of inserting image (pasting, via menu, d&d…), nothing works.

Comandeer:
Uh, after update of Word to 16.13.1 (180523) it started working.

core/tools.js
```diff
@@ -1596,6 +1596,64 @@
...
+		 * Currently supported file types: `image/png`, `image/jpeg`, `image/gif`.
+		 *
+		 * @since 4.10.0
+		 * @param {String} blobUrlSrc Address of blob which is going to be converted
+		 * @param {Function} callback Function execute when blob url will be converted.
+		 * @param {String} callback.dataUri data uri represent transformed blobUrl or empty string file type was unrecognized.
+		 */
+
+		convertBlobUrlToBase64: function( blobUrlSrc, callback ) {
+			CKEDITOR.ajax.load( blobUrlSrc, function( arrayBuffer ) {
```
mlewand: `ajax` plugin function can not be used in core, since it's not guaranteed it's going to be there. If anything the logic should go to the ajax plugin itself.

plugins/pastefromword/plugin.js
mlewand: @Comandeer what are the downsides of using `ajax` plugin for this case? Using FileReader just for sake of using it results with more code to worry about.

Comandeer:
Taking over this PR!

plugins/pastefromword/plugin.js
Comandeer: @mlewand it wasn't exactly a downside, I just felt that current use of ajax plugin is inappropriate, but I haven't spend much time on it as whole fix wasn't working for me. You're comment about using plugin in core explains my anxiety ;)

Comandeer:
I modified handling snapshots as the last version seemed to disable undo.

PR #1983, opened 2018-05-17
Added mentions feed throttling

### LINKED ISSUE #1972: Mentions plugin requests should be throttled

Feature request

Currently, given that `config.feed` is set to backend URL, our mentions plugin will issue a HTTP request for each typed character. That might be expensive.

For instance, imagine you type `@johnny`, you're issuing (depending on min chars) requests for `jo`, `joh`, `john` (etc.).

To mitigate this issue, requests should be throttled. By default I'd say somewhere around 70-120 ms each.

[`CKEDITOR.tools.eventsBuffer`](https://docs.ckeditor.com/ckeditor4/latest/api/CKEDITOR_tools.html#method-eventsBuffer) function provides handling for that.

Also throttling thershould should be configurable in mentions config object.

## Other details

* Browser: Any
* OS: Any
* CKEditor version: 4.10.0
* Installed CKEditor plugins: mentions

jacekbogdanski:

New feature

### This PR contains

- [x] Unit tests
- [x] Manual tests

Tests are available for autocomplete throttling.

## What changes did you make?

I added data request throttling for each feed type with default throttling time `200` ms
I did some testing on this issue and I think that proposed throttling thresholds in #1972 ticket (70-120ms) is too low for performance profit.

Also I tried to use [`CKEDITOR.tools.eventsBuffer`](https://docs.ckeditor.com/ckeditor4/latest/api/CKEDITOR_tools.html#method-eventsBuffer) for throttling implementation but this function doesn't allow to cancel buffered callback. It's always fired after provided timeout.

Closes #1972

### DIFF
--- a/plugins/mentions/plugin.js
+++ b/plugins/mentions/plugin.js
@@ -95,6 +95,14 @@
 		this.cache = config.cache !== undefined ? config.cache : true;
 
 		/**
+		 * See {@link CKEDITOR.plugins.mentions.configDefinition#throttle throttle}.
+		 *
+		 * @property {Boolean} [throttle]
+		 * @readonly
+		 */
+		this.throttle = config.throttle !== undefined ? config.throttle : 200;
+
+		/**
 		 * {@link CKEDITOR.plugins.autocomplete Autocomplete} instance used by mentions feature to implement autocompletion logic.
 		 *
 		 * @property {CKEDITOR.plugins.autocomplete}
@@ -105,7 +113,7 @@
 			dataCallback: getDataCallback( feed, this ),
 			itemTemplate: config.itemTemplate,
 			outputTemplate: config.outputTemplate,
-			throttle: 0
+			throttle: this.throttle
 		} );
 	}
 
@@ -393,6 +401,12 @@
 	 * See {@link CKEDITOR.plugins.autocomplete#outputTemplate}.
 	 *
 	 * @property {String} [outputTemplate]
+	 */
+
+	/**
+	 * See {@link CKEDITOR.plugins.autocomplete.configDefinition#throttle}.
+	 *
+	 * @property {String} [throttle=200]
 	 */
 
 	/**
--- a/plugins/mentions/plugin.js
+++ b/plugins/mentions/plugin.js
@@ -97,7 +97,7 @@
 		/**
 		 * See {@link CKEDITOR.plugins.mentions.configDefinition#throttle throttle}.
 		 *
-		 * @property {Boolean} [throttle]
+		 * @property {Number} [throttle]
 		 * @readonly
 		 */
 		this.throttle = config.throttle !== undefined ? config.throttle : 200;
@@ -406,7 +406,7 @@
 	/**
 	 * See {@link CKEDITOR.plugins.autocomplete.configDefinition#throttle}.
 	 *
-	 * @property {String} [throttle=200]
+	 * @property {Number} [throttle=200]
 	 */
 
 	/**
--- a/plugins/mentions/plugin.js
+++ b/plugins/mentions/plugin.js
@@ -95,7 +95,7 @@
 		this.cache = config.cache !== undefined ? config.cache : true;
 
 		/**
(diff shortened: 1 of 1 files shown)


jacekbogdanski:
I find out that there is `eventsBuffer.reset` function which allows you to clear scheduled timeout, so there is no need for additional function. Although I had to add an additional flag to `eventsBuffer.input` to postpone buffer execution.

There is one very important issue with `eventsBuffer` and I created separate `t/1972-dirty` branch for my changes because I think that `eventsBuffer` is not a good option for this implementation.

`eventsBuffer` allows you to pass parameters into callback only by binding `this` object into `scopeObj` parameter. The problem with this method is that in some cases it will produce a race condition between different callbacks. Slower callback which should be used will get an invalid query replaced by older one. You can test this issue with `test URL has no race condition` unit test (`tests/plugins/mentions`).

Of course, I could implement additional logic into `eventsBuffer` to handle this case but I think it will create much more complicated code than proposed in this PR.

Also, I think that `eventsBuffer` solves a different problem than we are fighting with. As I understand:

* `eventsBuffer` - allows you to execute a function during the given time once. So if we set our `eventsBuffer` to 200ms and type `@`, `@a`, `@an` separated by 10ms pauses during this 200ms a given callback will be fired once after exactly 200ms (not 220ms). Next, if we type `@ann` it will start new `200ms` time interval. For this 4 characters, we will get 400ms summary throttling time. 

* `throttle` - required to postpone data request by the given time after *each typed character*. Using the previous example it will execute the given callback by 10 * 2 + 200 + 200 = 420ms. If you type let's say `@` in 10ms and stop at `@a` the summary time will be 210ms.

In the result data request will be executed much more often when using `eventsBuffer` and I think that it will create some misunderstanding how this feature works (for end-users) because it will require some calculations to predict requests per throttle level.  

To get the second results using `eventsBuffer` I have to use in invalid way (IMO) by resetting previous callback before every `input` call. It brings us down to exactly the same code proposed with this PR.  Also there is still condition race issue and currently I don't have an idea how to solve this bug (except changing `eventsBuffer` again).

TBO I think I will have to overcomplicate this issue with `eventsBuffer`, where currently proposed code (except some simple function extraction) takes a couple of code lines to implement it, thus issue seems pretty simple:

https://github.com/ckeditor/ckeditor-dev/blob/17d1d67962ad6a4866da8b42a54d185af84d1012/plugins/mentions/plugin.js#L197-L210

against:
https://github.com/ckeditor/ckeditor-dev/blob/0c3fcc9bd3da76030f27cd46cd3c8a6459d6eb5b/plugins/mentions/plugin.js#L181-L203

Of course, I could be wrong and my implementation with `eventsBuffer` may be incorrect. I would really appreciate your opinion @mlewand if you see a better way to utilize `eventsBuffer` feature in this context.

mlewand:
You're right about the fact that we can't reuse `eventsBuffer` function to implement this cleanly. We need a throttling function that can accept, and pass arguments in the throttled function.

Also the way I see it I started to think that the throttling should actually be implemented in autocomplete plugin itself (or even on textwatcher level 🤔).

We _must not_ implement throttling logic directly in mentions/autocomplete or related plugin as this is a logic that should be _reusable_.

So as a solution let's make a `CKEDITOR.tools.throttle` function, that will return an object with the same API as the one returned by `eventsBuffer`. Also if there will be a need to, also implement `disable` function that  we were talking about before. This throttle function should be used in autocomplete plugin.

Note that the function should work the same way as eventBuffer does in terms of being called, so, say you have a throttling set to 100 ms and you have following calls:


call no. | time offset (relative to first call)
---------|----------
 1 | 0ms
 2 | 20ms
 3 | 40ms
 4 | 120ms

The throttled function should be executed thee times with following offsets:


 time offset |
---------- |
 0ms |
 100ms |
 200ms |

Throttling function should be developed as a separate issue (it deserves a separate API changelog entry), so this issue will be dependent on it.

mlewand:
This PR needs to be rebased.

jacekbogdanski:
I think there is no need for copy&paste unit and manual tests from autocomplete throttling. I'm just passing config property into autocomplete, so there is no logic inside mentions plugin.

PR #2028, opened 2018-05-29
Autocomplete dropdown position is now static when typing text

### LINKED ISSUE #1970: Autocomplete dropdown should not move while typing

Bug

Currently as you're typing the dropdown goes along with the text.

Instead, while typing, it should remain anchored to the initial position.

Of course, other things, like the scroll should still reposition the dropdown - as it does currently.

## Other details

* Browser: Any
* OS: Any
* CKEditor version: 4.10.0

jacekbogdanski:

Bug fix

### This PR contains

- [x] Unit tests (existing unit tests for view positioning)
- [x] Manual tests

## What changes did you make?

Refactored `autocomplete.view.getCaretRect` function to use the beginning of a typed word as a view starting position.

I tried to write some more integration test for text typing. However, I wasn't able to reproduce changing selection on typing. If you think we need additional integration test besides existing unit tests and manual test I will spend some more time on it. Any hints are welcome :)

Closes #1970

### DIFF
--- a/plugins/autocomplete/plugin.js
+++ b/plugins/autocomplete/plugin.js
@@ -908,8 +908,6 @@
 		 * @param {CKEDITOR.dom.range} range The range of text match.
 		 */
 		updatePosition: function( range ) {
-			range = range.clone();
-
 			this.setPosition( this.getViewPosition( range ) );
 		}
 	};
--- a/plugins/autocomplete/plugin.js
+++ b/plugins/autocomplete/plugin.js
@@ -672,21 +672,21 @@
 		},
 
 		/**
-		 * Returns the view position relative to the panel's offset parent.
+		 * Returns the view position based on a given `range`.
 		 *
 		 * Indicates the start position of the autocomplete dropdown.
 		 * The value returned by this function is passed to the {@link #setPosition} method
 		 * by the {@link #updatePosition} method.
 		 *
 		 * @param {CKEDITOR.dom.range} range The range of text match.
-		 * @returns {Object} Represents the position of the caret.
+		 * @returns {Object} Represents the position of the caret. The value is relative to the panel's offset parent.
 		 * @returns {Number} rect.left
 		 * @returns {Number} rect.top
 		 * @returns {Number} rect.bottom
 		 */
 		getViewPosition: function( range ) {
 			// Use the last rect so the view will be
-			// correctly positioned with a word splitted into few lines.
+			// correctly positioned with a word split into few lines.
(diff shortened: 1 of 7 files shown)


plugins/autocomplete/plugin.js
```diff
@@ -644,19 +644,20 @@
...
 		 * @returns {Object} Represents the position of the caret.
 		 * @returns {Number} rect.left
 		 * @returns {Number} rect.top
 		 * @returns {Number} rect.bottom
 		 */
-		getCaretRect: function() {
-			var caretClientRect = this.editor.getSelection()
-				.getRanges()[ 0 ]
-				.getClientRects()[ 0 ],
+		getViewStartPosition: function( range ) {
```
mlewand: The name could be simplified to `getViewPosition`.

plugins/autocomplete/plugin.js
```diff
@@ -873,8 +874,11 @@
 		 * {@link #setPosition} to move the panel to the best position close
 		 * to the caret.
 		 */
-		updatePosition: function() {
-			this.setPosition( this.getCaretRect() );
+		updatePosition: function( range ) {
```
mlewand: You didn't update the API docs.

tests/plugins/autocomplete/manual/viewstart.md
```diff
@@ -0,0 +1,14 @@
+@bender-tags: 4.10.0, bug, 1970
```
mlewand: This unit test throws a following exception:

> Uncaught ReferenceError: autocompleteUtils is not defined

plugins/autocomplete/plugin.js
```diff
@@ -644,19 +644,20 @@
...
 		 * @returns {Number} rect.left
 		 * @returns {Number} rect.top
 		 * @returns {Number} rect.bottom
 		 */
-		getCaretRect: function() {
-			var caretClientRect = this.editor.getSelection()
-				.getRanges()[ 0 ]
-				.getClientRects()[ 0 ],
+		getViewStartPosition: function( range ) {
+			var viewPositionRect = range.getClientRects()[ 0 ],
```
mlewand: First rect is good for a generic use case. However there could be a case where the matched word is split into few lines. In that case it's best to start autocomplete from the last range rectangle.

As an example think about autocomplete instance that accepts spaces, and shows autocomplete for "foo bar baz". Consider that you typed "foo bar" where both words are in different lines. Autocomplete should show up for the last rectangle.

the above "foo bar" case illustrated

PR #2041, opened 2018-05-31
Mentions dropdown is now activated regardless of styling

### LINKED ISSUE #2038: Mentions dropdown doesn't show up after activating styling
## Type of report

Bug

1. Open the editor with `mentions` plugin.
2. Change the font family/font style/activate bold or italic.
3. Press <kbd>@</kbd> to activate mentions dropdown.

### Expected result

The dropdown appears.

### Actual result

The dropdown doesn't appear. It will show up if you type something, then delete.

## Other details
**GIF**

bug_cke4

* Browser: Chrome, Safari, Opera
* OS: All platforms

jacekbogdanski:

Bug fix

### This PR contains

- [x] Unit tests
- [x] Manual tests

## What changes did you make?

Previously I thought this issue should be fixed inside `textmatch` plugin but I think that we shouldn't mess with pattern matching which is delegated to the user implementation. The issue exists because we are using filling char sequence for styles on some browsers (probably to create clickable placeholder). I handled this issue inside mentions pattern matching.

However I was thinking about moving code checking if a query is a part of word into a helper function which could be used by a users for their custom autocomplete implementation.

Closes #2038

### DIFF
--- a/plugins/textmatch/plugin.js
+++ b/plugins/textmatch/plugin.js
@@ -63,14 +63,23 @@
 	 */
 	CKEDITOR.plugins.textMatch.match = function( range, callback ) {
 		var textAndOffset = CKEDITOR.plugins.textMatch.getTextAndOffset( range ),
-			result = callback( textAndOffset.text, textAndOffset.offset );
+			fillingCharSequence = CKEDITOR.dom.selection.FILLING_CHAR_SEQUENCE,
+			fillingSequenceOffset = 0;
+
+		// Remove filling char sequence for clean query (#2038).
+		if ( textAndOffset.text.indexOf( fillingCharSequence ) == 0 ) {
+			textAndOffset.text = textAndOffset.text.replace( fillingCharSequence, '' );
+			textAndOffset.offset -= fillingSequenceOffset = 7;
+		}
+
+		var result = callback( textAndOffset.text, textAndOffset.offset );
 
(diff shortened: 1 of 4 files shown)


plugins/mentions/plugin.js
```diff
@@ -152,8 +152,12 @@
 			}
 
 			// Do not proceed if a query is a part of word.
-			var prevChar = text[ match.index - 1];
-			if ( prevChar !== undefined && !prevChar.match( /\s+/ ) ) {
+			var prev = text.substring( 0, match.index ),
+				endsWithSpace = prev.match( /^.*\s+$/ ),
```
Comandeer: Can't it be simplified to `/\s+$/`?

jacekbogdanski:
I also think that this fix is not perfect. In the best scenario, `match` function should return existing editor data - which means shouldn't be changed. However, I don't see a better solution if we would like to fix this issue for `mentions` and `emoji` in one place. I think that this function could fail in very edge cases where the user would like to somehow utilize filling char sequence but I'm not sure if this case could exist. 

I added additional docs with information that text does not reflect range offsets. If you think it's not enough we will probably have to fix the issue in`mentions` and `emoji` separately.

I also wonder if possible edge cases are not better than the issue with styles. Looks like we have to choose "lesser evil" ⚔️

PR #2043, opened 2018-06-01
Autocomplete view is hidden when outside of the viewable area

### LINKED ISSUE #1911: Hide autocomplete view if it's outside of the viewable area

Bug

1. Open [manual test](http://tests.ckeditor.test:1030/tests/plugins/autocomplete/manual/viewposition).
1. Focus the classic editor.
1. Start typing `@` to begin to autocomplete.
1. Scroll the editor editable to move autocomplete view outside of the viewable area.

### Expected result

Autocomplete view should be hidden when outside of the viewable area.

### Actual result

Autocomplete view sticks to the edges of the viewable area.

## Other details

* Browser: Any
* OS: Any
* CKEditor version: 4.10.0
* Installed CKEditor plugins: autocomplete

jacekbogdanski:

Bug fix

### This PR contains

- [x] Unit tests
- [x] Manual tests

## What changes did you make?

Simplified view positioning to hide the view when outside of the viewable area.

Closes #1911

### DIFF
--- a/plugins/autocomplete/plugin.js
+++ b/plugins/autocomplete/plugin.js
@@ -835,10 +835,34 @@
 				editorViewportRect = editable.isInline() ? editable.getClientRect( true ) : editor.window.getFrame().getClientRect( true );
 			}
 
+			// If the caret position is above the view - add dropdown top offset to simulate view clipping (#1911).
+			// +---------------------------------------------+
+			// |																						 |
+			// |     █ - caret position                      |
+			// |     +--------------+                        |
+			// |     |              |                        |
+			// +-----+==============+------------------------+
+			// |     |     view     |                        |
+			// |     +--------------+                        |
+			// |																						 |
+			// |       editor viewport                       |
+			// +---------------------------------------------+
 			if ( editorViewportRect.top > rect.bottom ) {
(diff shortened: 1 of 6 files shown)


mlewand:
I kind of feel that we should rather be clipping dropdown within the editor rather than completely removing it, see this edge case in proposed implementation:

the autocomplete panel hiding as soon as it gets even a single pixel outside of the viewport

The problem with current proposal is that as soon as you scroll the viewport one pixel below (or above) the dropdown "anchor" position, you get punished by hiding entire dropdown.

clipped dropdown mockup

Clipping would get us around this particular issue.

But I'd like to get some extra opinions here 

Just to be clear, this behavior will not exist in case of non-scrollable editors.

dkonopka:
Agree with @mlewand it looks much better if the dropdown hides instead of removing it instantly in case of "meeting" editor toolbar. 

OTOH we should check how it looks when dropdown is much longer 🤔 By default it has any `max-height` and `overflow-y: scroll` ?

jacekbogdanski:
In summary, the dropdown should scroll under editors toolbar. I will make appropriate changes and create some gif image when it's done.

jacekbogdanski:
viewclipping

jacekbogdanski:

Issue #2032, opened 2018-05-29
Use autocomplete output template in emoji plugin - refactoring


msamsel:
## Type of report

Task

## Provide description of the task

* Refactor code inside emoji responsible for output template. https://github.com/ckeditor/ckeditor-dev/pull/2006#discussion_r191233675
https://github.com/ckeditor/ckeditor-dev/blob/294c5d3da4153faedbfc429579d62648f87a9e7d/plugins/emoji/plugin.js#L32-L35

## Other details

* Browser: …
* OS: …
* CKEditor version: …
* Installed CKEditor plugins: …

mlewand:
Closed by #2071.

PR #2049, opened 2018-06-04
Asynchronous load of emoji list

### LINKED ISSUE #2036: Provide asynchronous emoji list loading
## Type of report

Task

Provide possibility to load asynchronously emoji list for emoji plugin.

## Other details

* Browser: …
* OS: …
* CKEditor version: …

msamsel:

New feature

### This PR contains

- [x] Unit tests
- [x] Manual tests

## What changes did you make?

- provide asynchronous load of emojis
- provide configuration option to load custom emoji list

Close #2036


plugins/emoji/plugin.js
```diff
@@ -158,3 +174,34 @@
  * @cfg {Number} [emoji_minChars = 2]
  * @member CKEDITOR.config
  */
+
+/**
+ * Address to json file containing emoji list. File is downloaded through {@link CKEDITOR.ajax#load} method
```
mlewand: JSON is an acronym should be uppercased.

msamsel:
I've rebase branch to contain change introduced in #2032

mlewand:
Issue reported by @Comandeer is valid. What's more imagine a case when CKEditor makes few snapshots before the `:emoji_tokens:` are replaced. That would mean that we'd have to change an entire snapshots history.

Instead let's make a tough decision and drop the feature which replaced valid emoji tokens in the content, at paste, drop, etc.

Ideally I'd like to keep this feature at `paste` event, however looking at the implementation, it's implemented using core text filter mechanics and it would require some refactoring to make it happen. We want to conclude it so **instead we'll remove the feature of replacing text tokens** with emoji entirely.

To sum it up:

* Keep the async emoji loading
* Remove parts of emoji that are replacing text tokens with real emojis on setData/insertText/paste/drop etc.

plugins/emoji/plugin.js
```diff
@@ -11,132 +11,96 @@
...
 
-			function matchCallback( text, offset ) {
-				var left = text.slice( 0, offset ),
-					match = left.match( new RegExp( ':\\S{' + charactersToStart + '}\\S*$' ) );
+				var emojiList = editor._.emoji.list,
+					charactersToStart = editor.config.emoji_minChars === undefined ? 2 : editor.config.emoji_minChars;
 
-				if ( !match ) {
-					return null;
+				if ( editor.status !== 'ready' ) {
```
Comandeer: Is this `if` needed now, when substitution is removed and initialisation just creates autocomplete instance?

plugins/emoji/plugin.js
```diff
@@ -11,132 +11,96 @@
...
 
-			function isEmojiAllowed( range ) {
-				var elementsPath,
-					editable = editor.editable();
-				if ( range ) {
-					elementsPath = new CKEDITOR.dom.elementPath( range.startContainer, editable );
-					return elementsPath.contains( forbiddenScope ) ? false : true;
-				} else {
-					return true;
+				function initPlugin() {
```
Comandeer: If we remove the `if` mentioned above, then this function could be inlined.

tests/plugins/emoji/customemojilist.js
```diff
@@ -0,0 +1,72 @@
...
+
+	bender.test( {
+		setUp: function() {
+			if ( CKEDITOR.env.ie && CKEDITOR.env.version < 9 ) {
+				assert.ignore();
+			}
+		},
+
+		// #2036
+		'test custom emoji list is loadd': function() {
```
Comandeer: Typo: loadd → load

tests/plugins/emoji/customemojilist.js
```diff
@@ -0,0 +1,72 @@
...
+			emojiTools.runAfterInstanceReady( editor, null, function( editor ) {
+				assert.areSame( 1, editor._.emoji.list.length );
+				objectAssert.areEqual( { id: ':star:', symbol: '⭐' }, editor._.emoji.list[ 0 ] );
+			} );
+		},
+
+		'test invalid emoji list': function() {
+			var editor = this.editors.classic2;
+			emojiTools.runAfterInstanceReady( editor, null, function( editor ) {
+				assert.isUndefined( editor._.emoji, 'There are created emoji private data, so emoji was loaded what is wrong for this case.' );
```
Comandeer: This doesn't seem like good assertion message.

plugins/emoji/plugin.js
msamsel: yes, it's required because of this bug: https://github.com/ckeditor/ckeditor-dev/issues/2114

Inlining this function cause random error, where plugin might be not initialised during loading stage (depend how fast json will be loaded). Error was clearly visible in Firefox browser.

plugins/emoji/plugin.js
msamsel: related to previous comment :)

msamsel:
Font injection for IE browser in divarea editor is extracted here: https://github.com/ckeditor/ckeditor-dev/issues/2121

msamsel:
I correct those tests. :)

PR #2060, opened 2018-06-07
Refactored duplicated throttle and eventsBuffer code

### LINKED ISSUE #2045: Reuse code in tools.throttle and tools.eventsBuffer
## Type of report

Task

In #1993 we added throttling function, however it introduced some copy'n'pasted code as mentioned in review comment https://github.com/ckeditor/ckeditor-dev/pull/1996#pullrequestreview-124055850.

The code should be reused.

## Other details

* Browser: Any
* OS: Any
* CKEditor version: 4.10.0

jacekbogdanski:

Task

## What changes did you make?

Refactored duplicated throttle and eventsBuffer code.

Closes #2045


mlewand:
Changed target branch to `master`. Rebased onto latest `master` branch.

jacekbogdanski:
I agree that it could be simplified and better designed. Your changes look good and I didn't have a better option here than just use your branch. I added additional docs fixes and justified tests for new buffers.

jacekbogdanski:
Looks ok, I refactored docs a little - I think it's easier to track docs when they are as close as possible to a declared function. However, with all these changes refactored buffers seems slightly complicated, mostly thus creating input function inside constructor which isn't common practice.

mlewand:
I agree that it got more complicated. I don't like it, but we have spent already enough time trying to simplify this code.

PR #2063, opened 2018-06-07
2061 default table width

bskibinski:

New feature

Yes: There was already a test checking the default width of a table. I've update the test so the default should be empty.

### This PR contains

- [x] Unit tests
- [ ] Manual tests

## What changes did you make?

I removed the "smart checking" of ckeditor window width, the table shouldn't have a default width, regardless of ckeditor window width. Now the default width when you insert a table is empty.

I would also accept a default of "100%" but I think it's more opinionated then just leaving it empty.

Now you can set "defaults" through CSS, and override the CSS with a custom value through the ckeditor.


bskibinski:
Where do I email the CLA? I didn't get an email because my email address isn't public i think.

bskibinski:
not sure why it's still awaiting CLA signature... I recieved the email and signed it digitally.

vokiel:
@bskibinski it was an callback issue which didn't fire after the sign-up. All fixed now.

mlewand:
@bskibinski just to give you an update: your idea sounds good, however before changing the default value I'd like to make sure that people that were happy with good old `500px` width can still easily revert previous value.

And instead of creating yet another config value for table's width alone, we decided to go with a more multi purpose feature to configure any dialog's input default value (#2277). We're already working on it and it looks like it might hit 4.11. Once that's done we'll move back to change the default width for the tables.

f1ames:
This PR still waits for #2277. Once it's done, we will get back to it.

jacekbogdanski:
Freezing due to https://github.com/ckeditor/ckeditor4/pull/2063#issuecomment-527096299

ivangretsky:
It seems like the generic solution is not going to come soon. Maybe just fix the table for now?

bskibinski:
I would also opt to just remove that 500px if those other fixes are taking this long (2 years). It's nice to refactor everything, but this is a realy simple fix, that I think everybody making/mainting websites in 2020 wants.

it doesn't have to slow down development on those other features.

You have to watch out you don't fall into this trap:
https://gfycat.com/infamousunluckyhagfish (malcolm in the middle lightbulb)

Issue #2062, opened 2018-06-07
Emoji list button


jacekbogdanski:
## Type of report

Feature request

## Preview:
Users interested in seeing how the feature looks today can preview it on [our Emoji feature branch preview](https://ckeditor4-preview.ckeditor.com/t/2062/plugins/emoji/samples/emoji.html).

## Provide description of the new feature

There should be an option to insert emoji from a button. Clicking the button would show full list of available emojis.

Because of the default emoji count, this feature requires some performance to make sure that the feature is responsive.

This feature could be implemented as a drop-down with some UI aiding usability.

![Emoji dropdown prototype](https://user-images.githubusercontent.com/5353898/41096829-1f29207e-6a56-11e8-8a4d-e19fbad99524.png)

## Browser Support

Same as [Emoji plugin](https://github.com/ckeditor/ckeditor-dev/issues/1746) itself.

## Features

Here are some words on the initial design:

#### Search

Search box should search based on the emoji ID or keyword (#2181).

If the search box is empty it just displays the initial view.

It is to be determined how the search results should be displayed - whether results should be grouped in categories or should it put it into one bunch.

#### Grouping

Emojis should be categorized. Categories should be shown as tabs, e.g. on the top of a dropdown.

The industry generally threads these tabs purely as navigation, meaning that clicking a given category does scroll to a given section, rather than filter the results.

#### Accessibility

There needs to be basic accessibility. We need to make sure that dropdown elements has correct labels and that the dropdown is operable using the keyboard alone.

### Nice To Have

The following features are optional, and may be skipped.

#### Initial Render

All the emojis should be displayed during the first render.

It is important to provide a very good performance here.

#### Skin Tone

Possibility to set the skin tone of an emoji.

#### Improved Configuration

It would be nice to allow for groups customization. At the minimum being able to hide groups.

Would be better if there's an option for adding custom categories to the mix - however this one would affect the design, as it would be very easy to break the tabs layout.

#### Turning Popular Matches into Emoji

Feature would turn some old-school style emotion expressions into an emoji, for instance `:)`, `<3` or `:D`. From the technical side of things I believe we could use text watcher/match combo for that.

mlewand:
Option that I'd like to see here is a small dorpdown that adds a search field and allows to look through entire emoji set. Once emoji is clidked it is inserted into a position where the editor selection is.

For sure it should be a modal dialog like the old one, but just a dropdown isntead.

wwalc:
[...]

PR #2069, opened 2018-06-11
Added items limit option for autocomplete plugin

### LINKED ISSUE #2030: Autocomplete should allow setting a limit of items in the dropdown

Feature request

It's super important feature in case someone uses a synchronous feed with an array of, say, 1k entries.

Rendering all of them is pretty extensive in terms of rendering time, especially given that after each subsequent character typed the list is recreated.

This feature should also be exposed by the mentions plugin.

## Other details

* Browser: Any
* OS: Any
* CKEditor version: 4.10.0
* Installed CKEditor plugins: autocomplete

jacekbogdanski:

New feature

### This PR contains

- [x] Unit tests
- [x] Manual tests

## What changes did you make?

Added additional `itemsLimit` configuration property allowing to limit the number of rendered items inside autocomplete dropdown.

Closes #2030, #2099

### DIFF
--- a/plugins/autocomplete/plugin.js
+++ b/plugins/autocomplete/plugin.js
@@ -526,6 +526,8 @@
 	};
 
 	/**
+	 * Class representing the autocomplete view.
+	 *
 	 * In order to use a different view, implement a new view class and override
 	 * the {@link CKEDITOR.plugins.autocomplete#getView} method.
 	 *
@@ -944,6 +946,8 @@
 	CKEDITOR.event.implementOn( View.prototype );
 
 	/**
+	 * Class representing the autocomplete model.
+	 *
 	 * In case you want to modify model behavior, check out the
 	 * {@link CKEDITOR.plugins.autocomplete.view} documentation. It contains
 	 * examples of how to easily override the default behavior.
--- a/plugins/autocomplete/plugin.js
+++ b/plugins/autocomplete/plugin.js
@@ -526,7 +526,10 @@
 	};
 
 	/**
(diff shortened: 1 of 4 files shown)


mlewand:
I have also pushed `t/2062_emoji` branch, but it resembles a patchwork as the code around was not reviewed and has quite some field for improvement.

jacekbogdanski:
I pushed `t/2030c` and `t/2030c_emoji` branches. Proposed implementation decouple view specific information (as a view model) from the data model. The implementation is based on *observer* pattern so every time when the model changes each registered view model is notified and could be updated (may differ depending on a use case). The changes fix the issue with `emoji` sample where view specific information modifies data model. This implementation is experimental and requires some work (also there is missing documentation) but gives an overview how it could look

I think we could create `BaseViewModel` object keeping the most basic operation like `publish`, `queryData`, and filtering functions. For more view specific information its behavior could be extended by inheritance or composition.

mlewand:
The release date is very soon, and we decided not to rush changes discussed starting with https://github.com/ckeditor/ckeditor-dev/pull/2069#pullrequestreview-128800235 comment. These changes are required by #2062 which will be added in 4.11.0, so they are not crucial yet, also we'd like to give solutions more time (or use cases) to prove the value.

Meanwhile we'll mark related Autocomplete API as private, and marked as a potential subject of change.

As for this ticket we'll continue the approach to temporary workaround the solution, so something like presented `t/2030` branch but corrected by fixing a case found in review.

I'll extract the refactoring into a separate issue.

PR #2078, opened 2018-06-12
Added following space option into autocomplete

### LINKED ISSUE #2008: Autocomplete should put the selection after output template

Bug

This plugin is based on not yet published Mentions feature branch `t/1703-b`.

1. Open [output template manual test](http://tests.ckeditor.test:1030/tests/plugins/autocomplete/manual/outputtemplate).
1. Focus the editor.
1. Type "@" to start autocompleteion.
1. Accept first entry.
1. Type " and " (note spaces).
1. Type "@" to start another autocompletion.
1. Accept first entry.

### Expected result

Text between the autocompleted values should not be bolded.

### Actual result

The text is bolded.

Reason is that the selection gets inside of the inserted element.

There are two solutinos to that:

* Put a selection outside that element (might be impossible on WebKits due to their normalization).
* Add an option to insert a space after the insertion. This option could be technically realized by adding a space to `outputTemplate` but the problem is if there is already a space after the text where you started autocompletion, it would just stupidly insert a space there resulting with two spaces.

## Other details

This issue is a follow up of https://github.com/ckeditor/ckeditor-dev/pull/1992#pullrequestreview-123272523.

* Browser: Any
* OS: Any
* CKEditor version: 4.10.0

jacekbogdanski:

New feature

### This PR contains

- [x] Unit tests
- [x] Manual tests

## What changes did you make?

The following space can be added at the end of the inserted word. This feature also recognizes existing space.

Closes #2008

### DIFF
--- a/plugins/autocomplete/plugin.js
+++ b/plugins/autocomplete/plugin.js
@@ -176,7 +176,7 @@
 		 *
 		 * @property {Boolean} [followingSpace]
 		 */
-		this.throttle = config.followingSpace !== undefined ? config.followingSpace : true;
+		this.followingSpace = config.followingSpace !== undefined ? config.followingSpace : true;
 
 		/**
 		 * The autocomplete view instance.
--- a/plugins/autocomplete/plugin.js
+++ b/plugins/autocomplete/plugin.js
@@ -172,7 +172,7 @@
 		this.throttle = config.throttle !== undefined ? config.throttle : 20;
 
 		/**
-		 * See {@link CKEDITOR.plugins.autocomplete.configDefinition#followingSpace}.
+		 * Indicates if a following space should be added after inserted match into an editor.
 		 *
 		 * @property {Boolean} [followingSpace]
 		 */
@@ -1350,22 +1350,6 @@
 		return editor.window.getFrame().getParent();
 	}
 
-	function insertSpaceAfterMatch( editor ) {
-		var selection = editor.getSelection();
-
-		var nextNode = selection.getRanges()[ 0 ].getNextNode( function( node ) {
-			return Boolean( node.type == CKEDITOR.NODE_TEXT && node.getText() );
-		} );
-
-		if ( nextNode && nextNode.getText().match( /^\s+/ ) ) {
-			var range = editor.createRange();
-			range.setStart( nextNode, 1 );
-			selection.selectRanges( [ range ] );
-		} else {
-			editor.insertHtml( '&nbsp;' );
-		}
-	}
-
 	function encodeItem( item ) {
 		return CKEDITOR.tools.array.reduce( CKEDITOR.tools.objectKeys( item ), function( cur, key ) {
 			cur[ key ] = CKEDITOR.tools.htmlEncode( item[ key ] );
@@ -1499,6 +1483,11 @@
 	 */
 
 	/**
+	 * @inheritdoc CKEDITOR.plugins.mentions#followingSpace
+	 * @property {Boolean} [followingSpace]
+	 */
+
+	/**
 	 * Abstract class describing a set of properties that can be used to produce more adequate suggestion data based on the matched query.
 	 *
 	 * @class CKEDITOR.plugins.autocomplete.matchInfo
@@ -1511,12 +1500,6 @@
 	 * {@link CKEDITOR.plugins.autocomplete.configDefinition#textTestCallback config.textTestCallback}.
 	 *
 	 * @property {String} query
-	 */
-
-	/**
-	 * Indicates if a following space should be added after inserted match into an editor.
-	 *
-	 * @property {Boolean} [followingSpace=true]
 	 */
 
 	/**
--- a/plugins/autocomplete/plugin.js
+++ b/plugins/autocomplete/plugin.js
@@ -174,6 +174,7 @@
 		/**
 		 * Indicates if a following space should be added after inserted match into an editor.
 		 *
+		 * @since 4.11.0
 		 * @property {Boolean} [followingSpace]
 		 */
 		this.followingSpace = config.followingSpace !== undefined ? config.followingSpace : true;
@@ -1484,6 +1485,7 @@
 
 	/**
 	 * @inheritdoc CKEDITOR.plugins.mentions#followingSpace
+	 * @since 4.11.0
 	 * @property {Boolean} [followingSpace]
 	 */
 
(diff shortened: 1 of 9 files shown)


jacekbogdanski:

tests/plugins/autocomplete/manual/followingspace.md
```diff
@@ -0,0 +1,32 @@
+@bender-tags: 4.10.0, feature, 2008
```
msamsel: Tag needs to be updated.

tests/plugins/autocomplete/autocomplete.js
```diff
@@ -563,6 +565,57 @@
...
+				ac = new CKEDITOR.plugins.autocomplete( editor, {
+					dataCallback: dataCallback,
+					textTestCallback: textTestCallback
+				} );
+
+			this.editorBots.standard.setHtmlWithSelection( '' );
+
+			editor.editable().fire( 'keyup', new CKEDITOR.dom.event( {} ) );
+
+			editable.fire( 'keydown', new CKEDITOR.dom.event( { keyCode: 13 } ) ); // ENTER
```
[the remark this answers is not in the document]
msamsel: Comment is for 2 lines higher line. I missclicked it.

jacekbogdanski:
Your discovery makes me wonder if we should enable `followingSpace` by default. This PR was aiming to autocomplete release previously, but currently, we will change autocomplete and mentions behavior which may break users output templates.

Because of that, I disabled `followingSpace` by default.

plugins/mentions/plugin.js
```diff
@@ -106,6 +106,13 @@
 		 */
 		this.cache = config.cache !== undefined ? config.cache : true;
 
+		/**
+		 * @inheritdoc CKEDITOR.plugins.autocomplete#followingSpace
+		 * @property {Boolean} [followingSpace]
```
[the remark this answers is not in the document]
jacekbogdanski: Actually, there is a default value described by brackets `[ ... ]` which means that this property is optional -> it's `false`.

tests/plugins/autocomplete/manual/followingspace.html
```diff
@@ -0,0 +1,37 @@
...
+					textTestCallback: autocompleteUtils.getTextTestCallback(),
+					dataCallback: autocompleteUtils.getDataCallback(),
+					followingSpace: true
+				} );
+			}
+		}
+	};
+
+	CKEDITOR.replace( 'classic', config );
+	CKEDITOR.inline( 'inline', config );
```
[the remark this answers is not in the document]
jacekbogdanski: https://github.com/ckeditor/ckeditor-dev/pull/2078#discussion_r222292789

jacekbogdanski:
It seems like it's easier to just remove doubled space than inserting a new one. The reason is that we have to move with selection outside output template, although it should be still placed inside completion range to keep preselected styles.

Unit tests are already written e.g. `test following space with output template`. I also added an additional manual test for the issue.

plugins/autocomplete/plugin.js
```diff
@@ -1344,6 +1362,20 @@
 		}, {} );
 	}
 
+	function removeLeadingSpace( editor ) {
+		var selection = editor.getSelection(),
+			nextNode = selection.getRanges()[ 0 ].getNextNode( function( node ) {
+			return Boolean( node.type == CKEDITOR.NODE_TEXT && node.getText() );
```
msamsel: This and next line seems to have wrong indentions.

plugins/autocomplete/plugin.js
```diff
@@ -1344,6 +1362,20 @@
...
+		} );
+
+		if ( nextNode && nextNode.getText().match( /^\s+/ ) ) {
+			var range = editor.createRange();
```
msamsel: There should be only one var statement at the beginning of the function.

plugins/mentions/plugin.js
```diff
@@ -106,6 +106,13 @@
 		 */
 		this.cache = config.cache !== undefined ? config.cache : true;
 
+		/**
```
msamsel: There is missing `@since` tag, which is not inherited.

plugins/autocomplete/plugin.js
jacekbogdanski: `var` statement in most cases is the most readable when it's as close as possible to the first usage - as it's a subjective matter, it still objective that in the above use case you will unnecessary create a range object if `if` statement fail.

jacekbogdanski:
> I see that you implement new configuration option for mention plugin, however there is no test which covers that change. I know it just pass vale to autocomplete, but in case that autocomplete will introduce some bug, then mention plugin might have chance to detect it.

Mentions behaves as a wrapper around autocomplete and with such approach we have to duplicate every test for mentions which is placed in autocomplete. It probably would be the best solution, however it creates a lot of code to maintain. I propose to keep 1:1 ratio between tests and features which is done in most our code base and also in mentions plugin (you can see how it has been done for `throttle` feature).

However, I think it would be nice to test if the values are really passed to autocomplete against breaking changes in constructor. I already added unit test for it.

plugins/autocomplete/plugin.js
msamsel: You don't have to assign value with declaration of this variable, so range don't have to be created in such case and you just assign proper value inside `if` statement.
As it's quite small function and there is hoisting, from my perspective it's better to declare it at the beginning of scope where it used. However I understand your concerns and let remain it for main review to decide wether current approach should be changed or not.

tests/plugins/mentions/mentions.js
```diff
@@ -462,6 +462,30 @@
...
+
+			this.createMentionsInstance( config );
+
+			spy.restore();
+
+			assert.isTrue( spy.calledWith( this.editor, sinon.match( compareConfig ) ) );
+
+			function compareConfig( val ) {
+				for ( var k in config ) {
+					assert.areEqual( config[ k ], val[ k ], 'Option: ' + k + ' should be passed to autocomplete.' );
```
msamsel: You can use `objectAssert.areEqual` or `objectAssert.areDeepEqual` methods to compare objects.

jacekbogdanski:
Actually, I can't, these two config objects are not equal. Autocomplete config contains also callbacks created internally inside mentions plugin.

Additionally I updated config validation with template options: https://github.com/ckeditor/ckeditor-dev/pull/2078/commits/3f8f90ed028b3db6584545766ccbfdb9f52a50f1

tests/plugins/mentions/mentions.js
msamsel: https://github.com/ckeditor/ckeditor-dev/pull/2078#issuecomment-448494237

PR #2094, opened 2018-06-15
Ability to configure split button in config.toolbar property

### LINKED ISSUE #2091: Ability to configure split button in config.toolbar property
## Type of report

Feature request

It would be nice to give the ability to create a split button from the configuration alone.

An example config would be:

```javascript
config.toolbar = [
    [ 'Source', '-', 'Bold', 'Italic', {
		type: 'splitbutton',
		name: 'JustifySplitButton',
		label: 'Justify', // Human readable label.
		items: [ 'JustifyLeft', 'JustifyRight', 'JustifyBlock', 'JustifyCenter' ],
		default: 'JustifyRight'
	} ]
];
```

`name`, `label` and `default` property should be optional, in case the `name` is skipped the name should be simply generated, like SplitButton1 etc.

engineering-this:

New feature

### This PR contains

- [X] Unit tests
- [X] Manual tests

## What changes did you make?

This feature is rather straight forward. I've selected base as `t/1679` so GH displays real diff for this branch.

Closes #2091

### DIFF
--- a/plugins/splitbutton/plugin.js
+++ b/plugins/splitbutton/plugin.js
@@ -140,8 +140,8 @@
 				 * @since 4.11.0
 				 * @constructor
 				 * @param {Object} definition The Split Button definition
-				 * @param {Object[]} definition.items An array of Split Button dropdown items
-				 * @param {Object} [definition.face] Item which will be Split Button face
+				 * @param {Object[]/String[]} definition.items An array of Split Button dropdown items
+				 * @param {Object/String} [definition.face] Item which will be Split Button face
 				 * @param {String} [definition.default] Item which will be Split Button face when there is no face defined and none of items is currently active
 				 */
 				$: function( definition ) {
--- a/plugins/splitbutton/plugin.js
+++ b/plugins/splitbutton/plugin.js
@@ -44,6 +44,10 @@
 			if ( typeof item === 'string' ) {
 				item = CKEDITOR.tools.clone( editor.ui.items[ item ] );
 
+				// It is possible to define CKEDITOR.ui.button instance without `icon` property defined.
+				// In such case icon name will fallback to `name` or `command`. However CKEDITOR.ui.menu doesn't have such logic for its items.
+				// We need to reverse this fallback to pass an icon name to the menu item.
+				// See comment (https://github.com/ckeditor/ckeditor-dev/pull/2094#discussion_r196684673).
 				if ( !item.icon ) {
 					item.icon = item.name || item.command;
 				}
--- a/plugins/splitbutton/plugin.js
+++ b/plugins/splitbutton/plugin.js
@@ -120,7 +120,7 @@
 				previousButton = properties.buttons[ item.id ];
 				defaultButton = properties.buttons[ item.id ];
 			} else {
-				if ( definition[ 'default' ] === item.name ) {
+				if ( definition[ 'default' ] && definition[ 'default' ].toLowerCase() === item.name ) {
 					defaultButton && defaultButton.hide();
 					defaultButton = properties.buttons[ item.id ];
 				} else {
--- a/plugins/splitbutton/plugin.js
+++ b/plugins/splitbutton/plugin.js
@@ -27,7 +27,7 @@
 		if ( face ) {
 			// Split Button can be defined with simpler definition via strings, needed by #2091
 			if ( typeof face === 'string' ) {
-				face = editor.ui.items[ face ];
+				face = CKEDITOR.tools.clone( editor.ui.items[ face ] );
 
 				if ( !face.icon ) {
 					face.icon = face.name || face.command;
@@ -46,7 +46,7 @@
 
 			// Split Button can be defined with simpler definition via strings, needed by #2091
 			if ( typeof item === 'string' ) {
-				item = editor.ui.items[ item ];
+				item = CKEDITOR.tools.clone( editor.ui.items[ item ] );
 
 				if ( !item.icon ) {
 					item.icon = item.name || item.command;
--- a/plugins/splitbutton/plugin.js
+++ b/plugins/splitbutton/plugin.js
@@ -111,10 +111,13 @@
 
 			editor.addFeature( properties.buttons[ item.id ] );
 
-			// If no default button was defined then first button will act as default.
-			if ( !defaultButton && !definition[ 'default' ] ) {
+			// Store first item as default. It will be overwritten by actual default if it exists and has correct value.
+			if ( !defaultButton ) {
 				defaultButton = properties.buttons[ item.id ];
 			} else if ( definition[ 'default' ] && definition[ 'default' ].toLowerCase() === item.name ) {
+				if ( defaultButton ) {
+					defaultButton.hide();
+				}
 				defaultButton = properties.buttons[ item.id ];
 			} else {
 				properties.buttons[ item.id ].hide();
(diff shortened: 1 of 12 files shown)


plugins/splitbutton/plugin.js
```diff
@@ -21,10 +21,19 @@
...
 			i,
 			item,
+			itemName,
 			previousButton,
 			defaultButton;
 
 		if ( face ) {
+			// Split Button can be defined with simpler definition via strings, needed by #2091
+			if ( typeof face === 'string' ) {
+				face = editor.ui.items[ face ];
```
mlewand: You assigning an object that you modify few lines below. That means you're modifying referenced icon.

We can not do that, you should clone it first and work on it clone instead.

tests/plugins/splitbutton/splitbutton.js
```diff
@@ -3,46 +3,108 @@
...
+
+						editor.ui.add( 'staticface', CKEDITOR.UI_SPLITBUTTON, {
+							face: {
+								command: 'bold',
+								icon: 'bold'
+							},
+							items: items
+						} );
+
+						editor.ui.add( 'customclick', CKEDITOR.UI_SPLITBUTTON, {
```
mlewand: Code registered as `customclick` handler is duplicated, extract it to common parts.

tests/plugins/splitbutton/splitbutton.js
```diff
@@ -3,46 +3,108 @@
...
+							face: {
+								icon: 'underline',
+								click: function() {
+									counter.faceClick++;
```
mlewand: The whole `counter` concept could be simplified by using sinon's stubs. Since you're using them in tests, you could just define stubs in test-scope or in context of `tests` object.

Then you'd want to reset it in `setup/teardown` function to make sure that test can not be affected by other unit tests.

engineering-this:
I've pushed to this branch commit form #2098.

plugins/splitbutton/plugin.js
```diff
@@ -25,6 +25,14 @@
 			defaultButton;
 
 		if ( face ) {
+			// Split Button can be defined with simpler definition via strings, needed by (#2091).
+			if ( typeof face === 'string' ) {
+				face = CKEDITOR.tools.clone( editor.ui.items[ face ] );
+
+				if ( !face.icon ) {
+					face.icon = face.name || face.command;
```
mlewand: What is the use case for this? If there is no icon in the original button it's fine to leave it without an icon.

plugins/splitbutton/plugin.js
```diff
@@ -36,6 +44,19 @@
 		for ( i in allItems ) {
 			item = allItems[ i ];
 
+			// Split Button can be defined with simpler definition via strings, needed by (#2091).
+			if ( typeof item === 'string' ) {
+				item = CKEDITOR.tools.clone( editor.ui.items[ item ] );
+
+				if ( !item.icon ) {
+					item.icon = item.name || item.command;
```
mlewand: Similar case as before.

plugins/splitbutton/plugin.js
```diff
@@ -94,12 +115,12 @@
 
 			editor.addFeature( properties.buttons[ item.id ] );
 
-			// First button as default. It might be overwritten by actual default button.
-			if ( !defaultButton && !item[ 'default' ] ) {
+			// If no default button was defined then first button will act as default.
```
mlewand: This logic could be now simplified. Here, in the loop, you could just leave `if ( definition[ 'default' ] && definition[ 'default' ] === item.name ) {` check over here.

Then the second `if` should go outside the loop, and that would check for conditions:
* whether `defaultButton` was not picked, with the logic above ☝️
* whether `definition[ 'default' ]` was not defined at all

Then assign the first item as a default button (if there was at least one item).

plugins/splitbutton/plugin.js
```diff
@@ -94,12 +115,12 @@
...
+			if ( !defaultButton && !definition[ 'default' ] ) {
 				previousButton = properties.buttons[ item.id ];
 				defaultButton = properties.buttons[ item.id ];
 			} else {
-				if ( item[ 'default' ] ) {
+				if ( definition[ 'default' ] && definition[ 'default' ].toLowerCase() === item.name ) {
```
mlewand: `definition.default` should be given as a button name, thus it should not be case-sensitive. We expect name `Bold`, `Image` etc.

plugins/splitbutton/plugin.js
engineering-this: Actually this one can be removed, but next case is required, explanations below.

plugins/splitbutton/plugin.js
engineering-this: That is how button plugin works. 
https://github.com/ckeditor/ckeditor-dev/blob/major/plugins/button/plugin.js#L262-L263

```
var name = this.name || this.command,
	iconPath = null,
	overridePath = this.icon;
iconName = name;
```

If there is no `buttonDefinition.icon` property, then icon will fallback to name or command. But `menu plugins` doesn't do that, so in order for menu item to have an icon we need to find that icon same way as `button plugin` does.

Example opening `~/workspace/ckeditor/ckeditor-dev/samples/index.html` and executing following in console:
```
CKEDITOR.instances.editor.ui.get( 'Bold' );
```
Returned object will not have any icon property. Actually when writing this I'm thinking if maybe we should change how button works, so it always have `icon` property when there is icon on button. WDYT @mlewand?

plugins/toolbar/plugin.js
```diff
@@ -535,6 +535,10 @@
 
 					// Ignore items that are configured to be removed.
 					if ( !removeButtons || CKEDITOR.tools.indexOf( removeButtons, name ) == -1 ) {
+						// Option to define UI element in `config.toolbar` (#2091).
+						if ( item.type && item.type in editor.ui._.handlers ) {
```
mlewand: This feature of supporting type property is nowhere to be found in the API docs. It needs to be reflected in API docs.

tests/plugins/splitbutton/manual/splitbutton.md
```diff
@@ -1,5 +1,5 @@
 @bender-ui: collapsed
-@bender-tags: feature, 4.10.0, 1679
+@bender-tags: feature, 4.11.0, 1679
```
mlewand: Technically this change is out of scope of this issue. It would be nice if you could extract this commit from history and move it to a proper feature branch - but it's not a big deal if it will remain here in case it's tough to do so.

plugins/splitbutton/plugin.js
mlewand: As for assigning the `icon` property during button creation - this sounds reasonable if only we started with such approach. Now there might be 3rd party integrations that rely on this property, and rely on it being a "falsy" value. So such change could break some solutions, we don't want to break things.

This rationale sounds reasonable. In this case it make sense to revert this fallback, thanks for clarifying.

Please just drop a short note, or a link to your comment here.

Also make sure to convert the code link above to a perma link, so something like that: https://github.com/ckeditor/ckeditor-dev/blob/6c0469b530cbc767946379e00b5aedcb01e2ba5a/plugins/button/plugin.js#L262-L263

engineering-this:
Rebased onto updated #1679 branch. See [comment](https://github.com/ckeditor/ckeditor-dev/pull/1766#issuecomment-405903831).
Also dropped commits updating version number, as it is now corrected in feature branch.

engineering-this:
Rebased, as parent branch was rebased also.

PR #2112, opened 2018-06-20
Updated autocomplete dataCallback API

### LINKED ISSUE #2108: Pass autocomplete instance to dataCallback from autocomplete's model#setQuery method
## Feature request

`dataCallback` function should contain `editor` instance name info:

https://github.com/ckeditor/ckeditor-dev/blob/adc195906e56169978f7f4b255ecfcae5d5f8c6a/plugins/autocomplete/plugin.js#L1339

While `dataCallback` usage on top of plugin's level it's hard to understand which editor instance it belongs to. As result it's not able to have 2+ editors with different autocomplete models on the same page.

jacekbogdanski:

New feature

### This PR contains

- [ ] Unit tests
- [ ] Manual tests

Tests are already available for `autocomplete` plugin.

## What changes did you make?

I updated dataCallback API to accept object instead of multiple parameters. Object exposes `autocomplete` instance.

Closes #2108

### DIFF
--- a/plugins/autocomplete/plugin.js
+++ b/plugins/autocomplete/plugin.js
@@ -368,7 +368,7 @@
 		 * initializing the autocomplete and can be overwritten in order to
 		 * return an instance of a different class than the default model.
 		 *
-		 * @param {Function} dataCallback See {@link CKEDITOR.plugins.autocomplete} arguments.
+		 * @param {Function} dataCallback See {@link CKEDITOR.plugins.autocomplete.configDefinition#dataCallback configDefinition.dataCallback}.
 		 * @returns {CKEDITOR.plugins.autocomplete.model} The model instance.
 		 */
 		getModel: function( dataCallback ) {
@@ -376,6 +376,7 @@
 
 			return new Model( function( matchInfo, callback ) {
 				return dataCallback.call( this, CKEDITOR.tools.extend( {
+					// Make sure autocomplete instance is available in the callback (#2108).
 					autocomplete: that
 				}, matchInfo ), callback );
 			} );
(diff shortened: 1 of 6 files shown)


plugins/autocomplete/plugin.js
```diff
@@ -1336,28 +1342,23 @@
...
 	 * ```
 	 *
 	 * @method dataCallback
-	 * @param {String} query The query string that was accepted by the `textTestCallback`.
-	 * @param {CKEDITOR.dom.range} range The range in the DOM where the query text is.
-	 * @param {Function} callback The callback which should be executed with the matched data.
-	 * @param {CKEDITOR.plugins.autocomplete.model.item[]} callback.data The suggestion data that should be
-	 * displayed in the autocomplete view for a given query. The data items should implement the
-	 * {@link CKEDITOR.plugins.autocomplete.model.item} interface.
+	 * @param {CKEDITOR.plugins.autocomplete.dataCallbackDefinition} dataCallbackDefinition
```
mlewand: `dataCallbackDefinition` does not sound very helpful. If you look at the following usage sample:

```js
// Returns (through its callback) the suggestions for the current query.
// Note: the itemsArray variable is our example "database".
function dataCallback( options ) {
    // Simple search.
    // Filter the entire items array so only the items that start
    // with the query remain.
    var suggestions = itemsArray.filter( function( item ) {
        return item.name.indexOf( options.query ) === 0;
    } );

    // Note - the callback function can also be executed asynchronously
    // so dataCallback can do an XHR requests or use any other asynchronous API.
    options.callback( suggestions );
}
```

What is now `dataCallbackDefinition` acts as object with information about match, based on which proper data should be returned. The only thing that does not fit there is that it contains a `callback` which is not related to match itself.

So how about extracting callback as a second argument, and rename `dataCallbackDefinition` to describe better what it conveys?

jacekbogdanski:
I rebased this PR into the latest upstream and fixed `dataCallback` for `mentions` and `emoji`.

PR #2119, opened 2018-06-20
Text pasted from Word preserves indentation in paragraphs

### LINKED ISSUE #651: Indentation is not preserved when pasting from Word
Bug

When I paste text from word, the paragraph indent is missing

### Expected result

Keep paragraph indent

### Actual result

No indent for paragraph

## Other details

* Browser: Chrome 59.0.3071.115
* OS: Windows 10
* CKEditor version: 4.7.1
* Installed CKEditor plugins: clipboard, pastefromword

engineering-this:

New features

### This PR contains

- [X] Unit tests
- [X] Manual tests

## What changes did you make?

Indentation in paragraphs is preserved as left margin.

Closes #651


engineering-this:
As said in https://github.com/ckeditor/ckeditor-dev/issues/651#issuecomment-398372621
We should consider supporting CSS property `text-indent`. Otherwise this fix would be incomplete, because it won't cover case such as posted by @Comandeer.

mlewand:
Is this result a regression compared to the original markup? I suppose so, because previously it did not include `margin` style at all, right?

What does Word provides as HTML in case like above? Can we somehow filter this case out?

engineering-this:
Example output from Word:
```
 p.MsoNormal, li.MsoNormal, div.MsoNormal{
(...)
	margin-top:0in;
	margin-right:0in;
	margin-bottom:8.0pt;
	margin-left:0in;
(...)
}
<p class=MsoNormal style='margin-top:12.0pt;margin-right:.5in;margin-bottom:
12.0pt;margin-left:.5in;line-height:normal'>P<o:p></o:p></p>
```

When we parse such Word output we copy styles from classes and apply it on element as inline, hence some styles might be overwritten. I realise we should more carefully parse styles from classes or consider not including margins from class.

engineering-this:
New changes:
1. I fixed way how margins were managed. Previously there could be shorthand and longhand margins defined on same element which produced some conflicts. Now all margins are converted to shorthand.
2. All zero margins are now filtered, but that required changing plenty of expected results in tests.
3. I've removed filtering 'text-indent' from PFW plugin, as it is ACF role to filter that out.

engineering-this:
It works for me with right margin.

screen shot 2018-07-12 at 17 06 12

```<p style="margin-bottom:8.0pt; margin-left:148.5pt; margin-right:202.5pt">Block of text with left and right margins block of text with left and right margins block of text with left and right margins block of text with left and right margins block of text with left and right margins block of text with left and right margins block of text with left and right margins block of text with left and right margins.</p>```

For usage of <kbd>TAB</kbd> I don't think it would be good idea to convert it into margins. If we want such feature, we should extract it to another ticket and start with proper discussion. There are plenty of cases to consider:
-tabulation in middle of paragraph,
-other types of tabulations, like center aligned,  decimal, etc.

Comandeer:
> For usage of TAB I don't think it would be good idea to convert it into margins. If we want such feature, we should extract it to another ticket and start with proper discussion.

Sounds reasonable 👍

engineering-this:
>Currently we also add margin-bottom to the element, yet it's not connected with text indentation at all. Probably vertical margins should be skipped for now, as they could interfere with line height styles.

In my opinion vertical margins should be removed by ACF not by the plugin.

PR #2136, opened 2018-06-22
Fixed failing tests for autocomplete and textwatcher

### LINKED ISSUE #2132: tests/plugins/autocomplete/autocomplete failing test
## Type of report

Bug

1. Open `tests/plugins/autocomplete/autocomplete` test.
1. Open `tests/plugins/emoji/customemojilist` test.
1. Open `tests/plugins/textwatcher/textwatcher` test.

### Expected result

Tests pass.

### Actual result

Tests fail.

## Other details

* Browser: All browsers
* OS: All platforms
* CKEditor version: 4.10.0
* Installed CKEditor plugins: full build

jacekbogdanski:

Tests fix.

## What changes did you make?

*  easy fix for `textwatcher` by adding command into an event so `undo` plugin won't broke
* replaced `getClientRect` in `autocomplete` with stub so view positioning in test no longer depends on toolbar size
* with full built triggering `keydown` triggers `afterComandExec` event ( by `tableselection` ) and deactivates `autocompete` model. I'm manually activating model for next `keydown` event so the tests pass. Not best solution but I think the easiest and tests are still valid.

Closes #2132 ( emoji already fixed by @msamsel )

### DIFF
--- a/tests/plugins/autocomplete/autocomplete.js
+++ b/tests/plugins/autocomplete/autocomplete.js
@@ -1,5 +1,5 @@
 /* bender-tags: editor */
-/* bender-ckeditor-plugins: autocomplete */
+/* bender-ckeditor-plugins: autocomplete, toolbar, basicstyles */
 
 ( function() {
 	'use strict';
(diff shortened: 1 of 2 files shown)


tests/plugins/autocomplete/autocomplete.js
```diff
@@ -232,6 +232,8 @@
 			editable.fire( 'keydown', new CKEDITOR.dom.event( { keyCode: 9 } ) ); // TAB
 			assert.areEqual( '', editor.getData(), 'Tab caused insertion' );
 
+			// Model may be deactivated by the previous keydown.
+			ac.model.isActive = true;
```
Comandeer: If the issue is caused by `tableselection` plugin, then it'll be better to just remove this plugin from test via `removePlugins` config option. That way we'll keep the test clean.

jacekbogdanski:
It looks like it was `tab` plugin. Nevertheless, I removed it from tests and seems to work.

PR #2142, opened 2018-06-22
Round offset to the nearest integer in tests

### LINKED ISSUE #2141: Failing tests/plugins/autocomplete/view
## Type of report

Failing test

1. Open http://tests.ckeditor.test:1030/tests/plugins/autocomplete/view

### Expected result

All tests pass

### Actual result

Some tests fail due to fact that Firefox responds with _too precise_ result (6.5 instead of 7):

## Other details

* Browser: Firefox
* Resolution: 1280x800
* Device pixel ratio: 2

Comandeer:

Failing test fix

## What changes did you make?

I added `Math.ceil` to some assertions as Firefox sometimes returns non-integers.

Closes #2141.

### DIFF
--- a/tests/plugins/autocomplete/view.js
+++ b/tests/plugins/autocomplete/view.js
@@ -88,9 +88,11 @@
 
 			var rect = getCaretRect( this.editors.inline, { top: 2, height: 3, left: 4 }, { y: 2, x: 4 } );
 
-			assert.areEqual( 7, rect.bottom );
-			assert.areEqual( 8, rect.left );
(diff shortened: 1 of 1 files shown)


mlewand:
The feature works correctly, the problem is with tests on FF. I suggest we ignore this for now and fix it correctly in next minor release.

mlewand:
Some more insights on failing test - it fails only when run from the dashboard. It passes if ran individually.

It depends on screen resolution, it works when `window.devicePixelRatio` is `1.0` but fails if `window.devicePixelRatio` is `2.5`.

PR #2179, opened 2018-06-27
Fixed failing test tests/plugins/widget/nestededitables on IE

### LINKED ISSUE #2143: Failing tests/plugins/widget/nestededitables
## Type of report

Bug

1. Open the `tests/plugins/widget/nestededitables` test.

### Expected result

Test passes.

### Actual result

Test fails randomly.

## Other details

* Browser: IE11
* OS: Windows 10
* CKEditor version: 4.10.0
* Installed CKEditor plugins: Full build

engineering-this:

Fix to failing test

## What changes did you make?

Assertions were fired with 210ms delay after focusing element. As blur happens with 200ms delay everything should be ok, but IE11 sometimes lags and fires our blur event later. I don't want to increase timeout, so I've moved assertions to `once( 'blur', ... )` listener. If listener isn't fired test still fails, because resume() isn't called.

Closes #2143

### DIFF
--- a/tests/plugins/widget/nestededitables.js
+++ b/tests/plugins/widget/nestededitables.js
@@ -979,13 +979,15 @@
 				CKEDITOR.document.getById( 'some_input' ).focus();
 
 				editor.once( 'blur', function() {
-					resume();
-					assert.isFalse( editor.focusManager.hasFocus, 'editor lost focus' );
-					assert.isFalse( !!widget.focusedEditable, 'editable is not focused' );
+					resume( function() {
+							assert.isFalse( editor.focusManager.hasFocus, 'editor lost focus' );
+							assert.isFalse( !!widget.focusedEditable, 'editable is not focused' );
 
-					editor.focus();
-					assert.isTrue( editor.focusManager.hasFocus, 'editor has focus again' );
-					assert.areSame( eFoo, widget.focusedEditable, 'editable is focused again' );
+							editor.focus();
+							assert.isTrue( editor.focusManager.hasFocus, 'editor has focus again' );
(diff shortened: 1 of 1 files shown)


engineering-this:
I will rebase it into latest `master` once `master` is up to date with `major`.

tests/plugins/widget/nestededitables.js
```diff
@@ -978,14 +978,19 @@
 
 				CKEDITOR.document.getById( 'some_input' ).focus();
 
-				wait( function() {
+				editor.once( 'blur', function() {
+					resume();
```
jacekbogdanski: Pass asserts as resume callback.

tests/plugins/widget/nestededitables.js
```diff
@@ -1101,15 +1106,20 @@
...
```
jacekbogdanski: Pass assert as resume callback.

jacekbogdanski:
@mlewand  TBO I'm only able to reproduce the issue with:

`test blurring nested editable when an editor is blurred`
`test subsequent nested editable focus causes selectionChange`

However, it makes sense to update each test based on `blur` event.

engineering-this:
1. Rebased in on the latest master.
2. Fixed `test focusing editor when focusing nested editable`.

Note: other tests with timeout set for 210ms are testing if blur doesn't happen during that time, so we can't listen for blur. Once we have regression they will become unstable on IE and instantly fail on other browsers.

engineering-this:
@jacekbogdanski I think you are right on that last one. I can't think of possible regressions that this wait could protect from.

PR #2197, opened 2018-07-02
Added autolink typing support

### LINKED ISSUE #1815: Autolink plugin typing support

Feature request

[Autolink plugin](https://ckeditor.com/cke4/addon/autolink) currently only works for pasted content. We could expand the plugin to handle turning URLs and [emails](https://github.com/ckeditor/ckeditor-dev/issues/1761) into links when typing text into editor.

1. Open CKEditor sample with enabled `autolink` plugin.
2. Type `http://example.com`

### Expected result

URL has been replaced with HTML link e.g. `<a href=“http://example.com”>http://example.com</a>`.

### Actual result

URL has been typed as a normal text.

jacekbogdanski:

New feature

### This PR contains

- [x] Unit tests
- [x] Manual tests

## What changes did you make?

I used `textmatch` plugin to provide `autolink` typing support. Pattern matching is a little complicated but it's required if we want to utilize the same URL/Email regex for both pastings from clipboard and typing.

Closes #1815


plugins/autolink/plugin.js
```diff
@@ -64,6 +115,55 @@
...
+
+	/**
+	 * Regex used by [Autolink](https://ckeditor.com/cke4/addon/autolink) plugin to match email adresses.
+	 *
+	 * @cfg {RegExp} [autolink_emailRegex]
+	 * @since 4.11.0
+	 * @member CKEDITOR.config
+	 */
+	CKEDITOR.config.autolink_emailRegex = /^[a-zA-Z0-9.!#$%&'*+\/=?^_`{|}~-]+@[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?(?:\.[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?)*$/g;
+	// Regex by (https://www.w3.org/TR/html5/forms.html#valid-e-mail-address).
```
Comandeer: Actually it's no longer in HTML 5.x, now it's only in [HTML LS](https://html.spec.whatwg.org/#e-mail-state-(type=email)).

jacekbogdanski:
I refactored code to expose `autolink_commitKeystrokes` configuration property and removed redundant keystrokes. With the current implementation, a user can decide which keystrokes use to accept link completion. Defaults are `enter` and `space`.

I also changed undo behavior - now it requires 3 undo steps to remove a link:
* undo commit keystroke if it creates undo step like `space` or `enter`
* undo link
* undo link text

plugins/autolink/plugin.js
```diff
@@ -6,45 +6,75 @@
...
+			editor.on( 'contentDom', function() {
+				var commitKeystrokes = editor.config.autolink_commitKeystrokes || CKEDITOR.config.autolink_commitKeystrokes;
+				editor.editable().on( 'keydown', function( evt ) {
+					if ( CKEDITOR.tools.indexOf( commitKeystrokes, evt.data.getKey() ) == -1 ) {
+						return;
+					}
 
-				// If we found "<" it means that most likely there's some tag and we don't want to touch it.
-				if ( data.indexOf( '<' ) > -1 ) {
+					var matched = CKEDITOR.plugins.textMatch.match( editor.getSelection().getRanges()[ 0 ], matchCallback );
```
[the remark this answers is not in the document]
Comandeer: It could look something like that:

```javascript
function autolink( editor, condition ) {
	var matched = CKEDITOR.plugins.textMatch.match( editor.getSelection().getRanges()[ 0 ], matchCallback );

		if ( !condition( matched ) ) {
			return;
		}

		insertLink( matched );
}

editor.on( 'paste', function( evt ) {
	// Pasted text may be encoded. Decode it so we will be able to compare its length with match.
	var text = CKEDITOR.tools.htmlDecode( evt.data.dataValue );

	// Attach afterPaste event here to avoid race condition between events.
	editor.once( 'afterPaste', function() {
		autolink( editor, function( matched ) {
			return matched && matched.text.length === text.length;
		} )
	} );
} );
[…]
editor.on( 'contentDom', function() {
	var commitKeystrokes = editor.config.autolink_commitKeystrokes || CKEDITOR.config.autolink_commitKeystrokes;
	editor.editable().on( 'keydown', function( evt ) {
		if ( CKEDITOR.tools.indexOf( commitKeystrokes, evt.data.getKey() ) == -1 ) {
			return;
		}
		
		autolink( editor, function( matched ) {
			return matched;
		} );
		// Handle event ASAP thus some plugins may change
		// editor selection or cancel keydown events e.g. wysiwygarea, enterkey.
	}, null, null, 0 );
} );
```

jacekbogdanski:
I extracted https://github.com/ckeditor/ckeditor-dev/pull/2197#pullrequestreview-136186428 issues into separate tickets so they will be easy to track and don't block this feature. See #2221 and #2220.

jacekbogdanski:
The unit test for continuous typing is disabled for WebKit thus it will result in a false negative. Although I kept manual test for each browser so we will have some safety net for the issue.

jacekbogdanski:
I added an additional manual test for filling char sequence Webkit issue. However, I wasn't able to add a unit test for this issue thus Webkit behaves differently when manually typing text inside the editor and when changing editors content using javascript. In manual testing Webkit doesn't extend anchors when typing, however inserting text using javascript changes anchors text node which gives invalid testing results.

The issue doesn't exist in normal feature path, it was caused by special treatment for Firefox and Edge where anchors are extended by typing, so I believe the manual test could be enough for this case.

tests/plugins/autolink/manual/continoustyping.html
```diff
@@ -0,0 +1,5 @@
+<div id="editor"></div>
+
+<script>
+	CKEDITOR.replace( 'editor' );
```
Comandeer: This test should be ignored also in IE as our plugin does not run there.

plugins/autolink/plugin.js
```diff
@@ -6,45 +6,76 @@
...
+			} );
+
+			// IE has its own link completion and we don't want to interfere with it.
+			if ( CKEDITOR.env.ie && !CKEDITOR.env.edge ) {
+				return;
+			}
+
+			editor.on( 'contentDom', function() {
+				var commitKeystrokes = editor.config.autolink_commitKeystrokes || CKEDITOR.config.autolink_commitKeystrokes;
+				editor.editable().on( 'keydown', function( evt ) {
```
Comandeer: We have dedicated event for key handling in editor, [`key`](https://docs.ckeditor.com/ckeditor4/latest/api/CKEDITOR_editor.html#event-key).

plugins/autolink/plugin.js
```diff
@@ -6,45 +6,76 @@
...
+					if ( CKEDITOR.tools.indexOf( commitKeystrokes, evt.data.getKey() ) == -1 ) {
+						return;
+					}
+
+					autolink( function( matched ) {
+						return matched;
+					} );
+
+					// Handle event ASAP thus some plugins may change
```
Comandeer: I'm not sure if we should interfere with other plugins in such way. If something is trying to stop keyboard event propagation, then it has some important reason for doing it.

jacekbogdanski:
I fixed most of the issues except Blink one which has been extracted into a separate issue. See #2239 and #2240 for more information.

Comandeer:
Rebased onto latest `major`.

Issue #2186, opened 2018-06-28
Persistent ui.richcombo selection


jsonn:
## Type of report

Feature request

## Provide description of the new feature

At the moment, richcombo will explicitly reset the value in `updateState` on selection changes etc. This is undesirable when using the richcombo widget as filter. One use case is a playbook of common phrases where the first richcombo widget can be used to select the category and the second one allows selecting the specific phrase. The state at the moment is also not friendly to monkey patching. Please introduce an option for preserving the (displayed) value.

PR #2206, opened 2018-07-05
Optionally keep richcombo value on selection

jsonn:

New feature as requested in #2186 

No. I don't know how to trigger the event directly from a test and the plugin testing is too sparse right to adjust existing tests.

### This PR contains

- [ ] Unit tests
- [ ] Manual tests

## What changes did you make?

The default behavior of richcombo is to reset value on events like mode or selection change happen. This is not desirable in every case, i.e. if the richcombo is a filter for another list. The new clearOnBlur can be set to false to inhibit the value reset part.

### DIFF
--- a/plugins/richcombo/plugin.js
+++ b/plugins/richcombo/plugin.js
@@ -74,7 +74,8 @@ CKEDITOR.plugins.add( 'richcombo', {
 				canGroup: false,
 				title: definition.label,
 				modes: { wysiwyg: 1 },
-				editorFocus: 1
+				editorFocus: 1,
+				clearOnBlur: true
 			} );
 
 			// We don't want the panel definition in this object.
@@ -169,7 +170,7 @@ CKEDITOR.plugins.add( 'richcombo', {
 					if ( editor.readOnly && !this.readOnly )
 						state = CKEDITOR.TRISTATE_DISABLED;
 
-					this.setState( state );
+					this.clearOnBlur && this.setState( state );
 					this.setValue( '' );
 
 					// Let plugin to disable button.


plugins/richcombo/plugin.js
```diff
@@ -74,7 +74,8 @@ CKEDITOR.plugins.add( 'richcombo', {
 				canGroup: false,
 				title: definition.label,
 				modes: { wysiwyg: 1 },
-				editorFocus: 1
+				editorFocus: 1,
+				clearOnBlur: true
```
mlewand: Blur is not a precise word here. I'm keen more towards something like `persistentState` that's more generic.

This is not blur that causes the state check, see https://github.com/jsonn/ckeditor-dev/blob/96f7ebeb8ddf1674c81642e64712b667c5a4a53b/plugins/richcombo/plugin.js#L182-L184

plugins/richcombo/plugin.js
```diff
@@ -169,7 +170,7 @@ CKEDITOR.plugins.add( 'richcombo', {
 					if ( editor.readOnly && !this.readOnly )
 						state = CKEDITOR.TRISTATE_DISABLED;
 
-					this.setState( state );
+					this.clearOnBlur && this.setState( state );
 					this.setValue( '' );
```
mlewand: To keep it consistent, the value should not be reset as well.

mlewand:

jsonn:
It doesn't really help closing an issue when the main question of how to write a test case for this kind of situation hasn't been answered...

PR #2211, opened 2018-07-09
Fixed: Emoji shows suggestion box when colon preceded with other character than white space

### LINKED ISSUE #2195: Emoji shouldn't show dropdown when inside a text
## Type of report

Bug

1. Open `http://tests.ckeditor.test:1030/tests/plugins/emoji/manual/emoji`
2. Type `foo:fa`.

### Expected result

Nothing happens.

### Actual result

Emoji dropdown shows up.

## Other details

In the current implementation `emoji` triggers dropdown inside sentence with `:` character. It's little awkward especially if you set `config.minChars = 0`. We handled this issue inside `mentions` plugin and I think it should be also handled in `emoji`.

* CKEditor version: 4.10.0
* Installed CKEditor plugins: emoji

engineering-this:

Bug fix

### This PR contains

- [X] Unit tests
- [X] Manual tests

## What changes did you make?

Added check for last character before colon. If it's not white space then don't show drop down.

Closes #2195

### DIFF
--- a/plugins/emoji/plugin.js
+++ b/plugins/emoji/plugin.js
@@ -67,14 +67,15 @@
 
 				function matchCallback( text, offset ) {
 					var left = text.slice( 0, offset ),
-						// Positive lookbehind for space, as emoji should be started with space or newline (#2195).
-						match = left.match( new RegExp( '(?<=\\s\|^):\\S{' + charactersToStart + '}\\S*$' ) );
+						// Emoji should be started with space or newline, but space shouldn't leak to output, hence it is in non captured group (#2195).
+						match = left.match( new RegExp( '(?:\\s\|^)(:\\S{' + charactersToStart + '}\\S*)$' ) );
 
 					if ( !match ) {
 						return null;
 					}
 
-					return { start: match.index, end: offset };
+					// In case of space preceding colon we need to return index of capturing grup.
+					return { start: text.indexOf( match[ 1 ] ), end: offset };
 				}
 
 				function dataCallback( matchInfo, callback ) {
--- a/plugins/emoji/plugin.js
+++ b/plugins/emoji/plugin.js
@@ -67,15 +67,10 @@
 
 				function matchCallback( text, offset ) {
 					var left = text.slice( 0, offset ),
-						match = left.match( new RegExp( ':\\S{' + charactersToStart + '}\\S*$' ) );
+						// Positive lookbehind for space, as emoji should be started with space or newline (#2195).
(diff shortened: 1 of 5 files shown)


msamsel:
I have also one more thought related to this fix. Current solution doesn't allow on typing one emoji after another. You need to additionally add space after emoji to insert new one.
On GitHub there is insert space after each emoji, which immediately allow on adding new on. Similar situation is on slack.
@mlewand, @Comandeer, WDYT? should we also include solution for this situation inside this fox, or maybe separate it as new issue?

plugins/emoji/plugin.js
```diff
@@ -73,6 +73,12 @@
 						return null;
 					}
 
+					// Do not proceed if a query is a part of word.
+					var prevChar = text[ match.index - 1];
+					if ( prevChar !== undefined && !prevChar.match( /\s+/ ) ) {
```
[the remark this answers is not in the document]
engineering-this: Proposed RegExp would push space to returned match in order to not push wrong query we would have to filter that space. Instead I've changed it to look behind condition.

jacekbogdanski:
@msamsel  see https://github.com/ckeditor/ckeditor-dev/pull/2078

plugins/emoji/plugin.js
```diff
@@ -67,7 +67,8 @@
 
 				function matchCallback( text, offset ) {
 					var left = text.slice( 0, offset ),
-						match = left.match( new RegExp( ':\\S{' + charactersToStart + '}\\S*$' ) );
+						// Positive lookbehind for space, as emoji should be started with space or newline (#2195).
+						match = left.match( new RegExp( '(?<=\\s\|^):\\S{' + charactersToStart + '}\\S*$' ) );
```
msamsel: Unfortunately lookbehind and lookahead RegExp are not widely supported, because of that tests are failing on IE11, Firefox, Safari.
You can try to use groups and/or non-capturing groups, but still it will require changing way how start value is set up.

tests/plugins/emoji/manual/spacerequired.md
```diff
@@ -0,0 +1,38 @@
...
+
+1. Place caret at the end of line.
+
+1. Type ':bu' without space before.
+
+1. Type ' :bu' with space.
+
+### Expected
+
+When there is no space:
```
msamsel: WDYT about simplifying it even more like this:

> ### Expected:
> * Nothing happens, when is no preceding space
> * Emoji suggestion box appears, when preceding space is present
> 
> ### Unexpected
> * Emoji suggestion box appears, when is no preceding space

PR #2214, opened 2018-07-11
Telephone link type support in link plugin

### LINKED ISSUE #2154: Telephone link type support in the Link plugin
## Type of report

Feature request

We should be able to select a telephone link type (`<a href="tel:`) in addition to the mail link type with the [Link plugin](https://ckeditor.com/cke4/addon/link).

engineering-this:

New feature

### This PR contains

- [X] Unit tests
- [X] Manual tests

## What changes did you make?

Added new dialog `input[type="tel"]`. And used in in link plugin for inputing phone number in new link type. Also added config options to set validation regex and alert message when validation fails.

Closes #2154


plugins/link/dialogs/link.js
```diff
@@ -983,6 +1014,21 @@
 			}
 		};
 	} );
+
+	function validateTelNumber() {
+		var editor = this.getDialog()._.editor,
+			regExp =  editor.config.linkTelNumberValidate_regExp,
+			msg = editor.config.linkTelNumberValidate_msg,
+			dialog = this.getDialog();
```
Comandeer: If we move this variable definition earlier, we could use it to get `editor`:

```javascript
editor = dialog._.editor
```

Microoptimisations – FTW!

tests/plugins/link/link.js
```diff
@@ -562,6 +573,43 @@
...
+		// (#2154)
+		'test telephone number link': function() {
+			var editor = this.editor,
+				bot = this.editorBot;
+
+			bot.setHtmlWithSelection( '^' );
+			bot.dialog( 'link', function( dialog ) {
+				var expectedNumber = '123456789',
+					input,
+					stub = this._stubs.alert = sinon.stub( window, 'alert' );
```
Comandeer: TBH creating `this._stubs` and whole `tearDown` logic seems a little bit like overkill. We got only this one stub, which could be restored with one line of code instead of current 7. Keeping whole stubbing code inside test itself also improves readability.

plugins/link/dialogs/link.js
```diff
@@ -983,6 +1014,28 @@
...
 		};
 	} );
+
+	function validateTelNumber() {
+		var dialog = this.getDialog(),
+			editor = dialog._.editor,
+			regExp =  editor.config.linkTelNumberValidate_regExp,
+			msg = editor.config.linkTelNumberValidate_msg,
+			linkLang = editor.lang.link,
+			func = CKEDITOR.dialog.validate.notEmpty( linkLang.noTel ),
```
Comandeer: `func` could be just inlined into `messageWhenEmpty` variable.

plugins/link/dialogs/link.js
```diff
@@ -983,6 +1014,28 @@
...
+			messageWhenEmpty = func.apply( this );;
```
Comandeer: Double semicolon.

plugins/link/dialogs/link.js
```diff
@@ -983,6 +1014,28 @@
...
+
+		if ( !dialog.getContentElement( 'info', 'linkType' ) || dialog.getValueOf( 'info', 'linkType' ) != 'tel' ) {
+			return true;
+		}
+
+		if ( typeof messageWhenEmpty === 'string' || !messageWhenEmpty ) {
```
Comandeer: Can't it be simplified to `message !== true`?

plugins/link/dialogs/link.js
```diff
@@ -983,6 +1014,28 @@
...
+			return messageWhenEmpty;
+		}
+
+		if ( regExp && !regExp.test( this.getValue() ) ) {
```
Comandeer: Note that we have built-in `CKEDITOR.dialog.validate.regex`, which could simplify the logic.

plugins/link/plugin.js
```diff
@@ -873,5 +880,25 @@
...
 		 * @member CKEDITOR.config
 		 */
+
+		/**
+		 * Optional JavaScript Regular Expression whenever telephone numbers in Link dialog should be validated.
+		 *
+		 *		config.linkTelNumberValidate_regExp = /^[0-9]{9}$/;
+		 *
+		 * @since 4.11.0
+		 * @cfg {RegExp} [linkTelNumberValidate_regExp]
```
Comandeer: I'm thinking about changing names of variables to `linkTelNumberRegExp` and `linkTelNumberMsg`. ATM only these two config variables in link plugin contain underscore, introducing some kind of inconsistency.

mlewand:
@engineering-this @Comandeer I don't think we made a good choice with the `config.linkTelNumberMsg`:

* The message about invalid phone format should be simply put in the lang files and provided by our plugin. By putting it in the config you put the hassle of localizing this message to the dev integrating CKE4, while it _should just work™_. Providing it in lang file provides localization, contributed by our amazing community out of the box, for everyone.
* Secondly I think the name does not fully explains the purpose. Name is `linkTelNumberMsg` and it does n't mention that it's a message in case of invalid number/wrong format.

Comandeer:
The reason behind putting it in config was the fact that validation will be custom – so the message also could be custom, to include information about the desired format of the telephone number.

mlewand:
@Comandeer That sounds like a good use case to remember about, though you can still define the default message in the lang file and allow for customization in case of need.

So the default lang entry would look pretty much like that:

```
en
invalidTelNumberMessage: 'Invalid phone number format.',
```

While you could optionally override it. It _could_ be with config, so you would have a `config.linkInvalidTelNumberMessage = 'Format should follow an XYZ pattern.'`.

I have emphasized _could_ because I'm not a big fan of adding a config option to customize the error message alone, instead I'd like to see some reusable way to adjust _any_ lang entry during the runtime.

TBH I don't know any clear right off the bat (would need to check the code), but you can always do that by listening to `instanceReady` event and overwriting any lang property that should be customized. However I know, it's pretty advanced as for such use case.

PR #2219, opened 2018-07-12
Improved pastefromword to retain cell borders

### LINKED ISSUE #1490: Copy/paste table from Excel does not retain border value
Feature request

1. In Excel, create a table ( 3 by 5) with thick border only in the outer cells. Add some values in the cells.
2. Copy the table and paste in CKEditor ( Standard or Full verison 4.8.0)
3. Table is pasted with cell values and cell background color, but the cell border is not retained correctly.

### Expected result
After pasting the table in CKEditor, it should retail the border properties.

### Actual result
Table is pasted with cell values and cell background color, but the cell border is not retained correctly.

## Other details

* Browser: IE 11, Chrome 63.0.3239
* OS: Windows 10
* CKEditor version: 4.8.0 (Standard and Full)
* Installed CKEditor plugins: All standard plugins came with package

image

### LINKED ISSUE #2924: Extract CKEDITOR.tools.style.parse.border as a separate Border type
## Type of report

Task

#1490 implementation requires additional border parsing helpers. It makes sense to wrap all border helpers into a new `CKEDITOR.tools.style.border` type including `CKEDITOR.tools.style.parse.border` function, so we will have a single, unified object for CSS border representation. It also means that `CKEDITOR.tools.style.parse.border` function should be deprecated in favor of new border type member.

jacekbogdanski:

New feature

### This PR contains

- [x] Unit tests
- [x] Manual tests

## What changes did you make?

It's early PR which requires some changes. Among others:
1. I didn't cover tests thus proposed changes kills 134 unit tests only for chrome. I will fix it and cover with tests after early acceptance.
1. I was only able to test this feature on IE11, Chrome, Firefox. However, it looks like Safari and IE8 are not supported by `pastefromword`. Not sure about IE10 and IE9.

Closes #1490, closes #2924


plugins/pastefromword/filter/default.js
```diff
@@ -354,41 +343,128 @@
...
+							CKEDITOR.tools.convertToPx( borderObj.width ) + 'px'
+							: borderObj.width;
+
+						return borderObj.width + ' ' +
+							borderObj.style + ' ' +
+							// Replace old CSS2 `windowtext` color with `black`.
+							borderObj.color.replace( 'windowtext', 'black' );
+					}
+
+					function getBorderStyle( modifier, borderType ) {
```
mlewand: CSS parsing code should (re)use [`CKEDITOR.tools.style.parse`](https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_tools_style_parse.html) methods. In case something is missing there, just add new feature there.

tests/plugins/pastefromword/manual/pastetable.md
```diff
@@ -0,0 +1,14 @@
+@bender-tags: bug, 4.11.0, 1490, pastefromword
+@bender-ui: collapsed
+@bender-ckeditor-plugins: wysiwygarea, toolbar, pastefromword
```
Comandeer: Table from Excel is pastes as plain text. There are several missing plugins (e.g. `table`).

plugins/pastefromword/filter/default.js
```diff
@@ -354,41 +343,128 @@
...
-					var styles = tools.parseCssText( element.attributes.style );
+					var styles = tools.parseCssText( element.attributes.style, true ),
+						borderStyles = styles.border ? CKEDITOR.tools.style.parse.border( styles.border ) : {};
 
-					for ( var style in styles ) {
-						var temp = styles[ style ];
-						delete styles[ style ];
-						styles[ style.toLowerCase() ] = temp;
-					}
+					for ( var i = 0; i < allowedBorderTypes.length; i++ ) {
```
Comandeer: You use `CKEDITOR.tools.array.forEach` just right after this loop. Won't be better to use also here?

plugins/pastefromword/filter/default.js
```diff
@@ -354,41 +343,128 @@
...
+
+						// Width may be set to some string value e.g. `inherit`.
+						borderObj.width = parseFloat( borderObj.width ) ?
+							CKEDITOR.tools.convertToPx( borderObj.width ) + 'px'
+							: borderObj.width;
+
+						return borderObj.width + ' ' +
+							borderObj.style + ' ' +
+							// Replace old CSS2 `windowtext` color with `black`.
+							borderObj.color.replace( 'windowtext', 'black' );
```
Comandeer: `windowtext` isn't actually equal to `black`, it's platform dependent and could be changed by changing OS's theme or other settings. As for now every browser is supporting it, so I'm not sure if it's necessary to replace it.

jacekbogdanski:
The feature doesn't work on Safari and IE8 thus the issue: https://github.com/ckeditor/ckeditor-dev/issues/2609

jacekbogdanski:
> Just 2 cents: make sure to add a unit test for this enhancement. Currently, it is not covered.
> The unit tests are missing.

Yes, I know that this PR is missing unit tests, I already wrote it in the PR description. No worries, I won't leave this code without test coverage. Although this implementation kills every test case which uses table (currently failing 132 unit tests only for chrome, some tests are written per browser so it may multiply fast) so I would really appropriate solid review before I start fixing unit tests. Small implementation change may require correcting all tests again and thus they are generated, it's a highly time-consuming task. Although I added a unit test for Chrome, so it will be easier to validate output HTML.

I'm wondering if https://github.com/ckeditor/ckeditor-dev/blob/015e1cc00b114d6a503421db9c0a35b71afe525e/plugins/pastefromword/plugin.js#L26-L27 ACF filter shouldn't be moved to the `table` plugin? As I know `pastefromword` should be a bit dumb in the case of allowed content inside the editor, so a user can decide which styles he wants to enable by importing correct plugin.

I'm not very fond of `CKEDITOR.tools.style.parse.splitBorderStyles` name and I think it may be misleading. Also, it doesn't play well with namespace when you look at the other function names. However, nothing else comes to my mind now. Maybe you have some idea about the name?

plugins/pastefromword/filter/default.js
mlewand: I'm wondering here, it's tricky to work with these values.

95% (ofc my gut feeling) people use the default values, so it would kinda make sense to convert it anyway. And it is not rare to see MS Word using this:

image

We could expose a dedicated method in `CKEDITOR.tools.style` to replace these magic values. Then again I thought about possibility to adjust these values or even disable this replacement. So I kind of intuitively thought of `CKEDITOR.config` but on the second thought I'm not a fan of bloating (already enormous) config with very specific options like that.

So eventually it could hold some sort of public mapping in `CKEDITOR.tools.style`. That will allow only for global customization, but we don't need more than that.

core/tools.js
```diff
@@ -1805,11 +1807,46 @@
...
+				 * If not set, property value will be splitted by spaces.
+				 * @param {String[]} split.return Splitted shorthand value.
+				 * @returns {Object}
+				 * @returns {Number} return.top Top value.
+				 * @returns {Number} return.right Right value.
+				 * @returns {Number} return.bottom Bottom value.
+				 * @returns {Number} return.left Left value.
+				 * @member CKEDITOR.tools.style.parse
+				 */
+				shorthand: function( value, split ) {
```
Comandeer: This name is a little bit misleading, because there are several types of shorthands in CSS (e.g. one in `border-radius` property).

plugins/pastefromword/filter/default.js
```diff
@@ -542,6 +528,24 @@
 			element.attributes.style = CKEDITOR.tools.writeCssText( styles );
 		},
 
+		removeStyles: function( element, styles ) {
+			CKEDITOR.tools.array.forEach( styles, function( style ) {
+				Style.setStyle( element, style, null );
+			} );
+		},
+
+		convertStyleToPx: function( element ) {
```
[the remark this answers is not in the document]
jacekbogdanski: Not sure if I understand, as I know `CKEDITOR.tools.convertToPx` does more or less what you said, so how this function is different?

https://github.com/ckeditor/ckeditor-dev/blob/015e1cc00b114d6a503421db9c0a35b71afe525e/plugins/pastefromword/filter/default.js#L537-L547

plugins/pastefromword/plugin.js
```diff
@@ -23,6 +23,8 @@
 				// Snapshots are done manually by editable.insertXXX methods.
 				canUndo: false,
 				async: true,
+				allowedContent: 'table{border-collapse};' +
```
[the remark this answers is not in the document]
jacekbogdanski: Yep, I already mentioned it in https://github.com/ckeditor/ckeditor-dev/pull/2219#issuecomment-443502713, so we agree on this one. However, not sure how far can we go with changing ACF of the `table` plugin for this feature.

plugins/pastefromword/filter/default.js
```diff
@@ -533,7 +519,7 @@
 				return;
 			}
 
-			if ( value === '' ) {
+			if ( value === '' || value == null ) {
```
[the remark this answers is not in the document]
jacekbogdanski: I will remove this change because it's not important for this PR, although I'm curious how could it remove styles with `0` value? In JS `0 == null` and `'0' == null` gives false logical value. I'm using double equal operator here to perform `null == undefined` coercion.

mlewand:
Just to add my 2 cents here, I like the proposal added by @Comandeer. Given that we're able to make it work 100% of the time and that it will not cause any side effects it would really improve the pasting experience.

However with this, this PR starts to do _too many things™_. We should extract this feature proposed by @Comandeer to a separate issue (as it is _feature request_) and fix the _bug_ here.

core/tools.js
```diff
@@ -1825,15 +1862,94 @@
...
+				 * @param {Object} [fallback.style] Style CSS style used in absence of `border-style` style.
+				 * @param {Object} [fallback.width] Width CSS style used in absence of `border-width` style.
+				 * @returns {Object}
+				 * @returns {String} return.border-top Border top style.
+				 * @returns {String} return.border-right Border right style.
+				 * @returns {String} return.border-bottom Border bottom style.
+				 * @returns {String} return.border-left Border left style.
+				 * @member CKEDITOR.tools.style.parse
+				 */
+				splitBorderStyles: function( styles, fallback ) {
```
mlewand: Naming: it's already in style namespace, so there's no reason to call it `splitBorderStyles` because it would be `CKEDITOR.tools.style.parse.splitBorderStyles` (note style twice).

`CKEDITOR.tools.style.parse.splitBorder` is definitely simpler.

mlewand:
I have pushed [`t/1490-oop`](https://github.com/ckeditor/ckeditor-dev/tree/t/1490-oop) branch, where I added simplified introduction of `BorderStyle`. This is a type that I initially wanted to be returned by members like https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_tools_style_parse.html#method-border

But we decided not to add it in order to simplify the code (we didn't need this at this point). Now as there is more than 1 method working on borders it make sense to unify it.

Note that on this branch I didn't expose the type (for sure the types should be exposed, as they're useful).

Also I'd work on simplifying the shorthand interaction. It didn't work well inside `splitBorderStyles` in terms of API.

I think that `BorderStyle` type could accept object { size, type, color } instead of positioned arguments, or it could also handle simple array in a predefined order. Just do whatever makes the client-code simpler.

Finally this part is garbage, and should be simplfieid: https://github.com/ckeditor/ckeditor-dev/blob/t/1490-oop/core/tools.js#L1939-L1941 This acutally is related to the API mentioned before.

Anyway this is just a proposal for possible direction. As a result function returns border object that expose border information in a convenient way (just `obj.color`, `obj.style` instead of having to parse it over and over). It also stringifies well, so the use case that you're using in this particular ticket.

jacekbogdanski:
At the moment `BorderStyle._normalize` a function is probably overcomplicated and should be inlined inside a constructor. However, I think it makes sense if we want to make `normalizeMap` public. WDYT?

plugins/pastefromword/filter/default.js
```diff
@@ -29,7 +29,8 @@
 			'v:group'
 		],
 		links = {},
-		inComment = 0;
+		inComment = 0,
+		convertToPxRegex = /\d+(\.\d+)?pt/g;
```
Comandeer: Why this regex is here, as it could be inlined inside `convertToPx` function?

plugins/pastefromword/filter/default.js
```diff
@@ -556,6 +563,21 @@
 			}
 		},
 
+		// Maps common attributes to styles.
+		mapCommonStyles: function( element ) {
+			return Style.mapStyles( element, {
+				vAlign: function( value ) {
+					Style.setStyle( element, 'vertical-align', value );
+				},
+				width: function( value ) {
```
Comandeer: What if `width` or `height` were set in percentages, not in pixels?

tests/plugins/pastefromword/manual/pastetable.html
```diff
@@ -0,0 +1,29 @@
...
+<label for="input">Paste clibpoard data</label>
+<input id="input" name="input" />
+<div id="result"></div>
+
+<script>
+	var input = document.getElementById( 'input' ),
+		result = document.getElementById( 'result' );
+
+	if ( !CKEDITOR.env.ie || CKEDITOR.env.version > 11 ) {
+		input.addEventListener( 'paste', function( evt ) {
```
Comandeer: We should be able to get raw data directly from editor's `paste` event, see [`CKEDITOR.plugins.clipboard.dataTransfer.getData`'s 2nd parameter](https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_plugins_clipboard_dataTransfer.html#method-getData).

core/tools.js
```diff
@@ -1805,11 +1808,46 @@
...
+				 * function split( value ) {
+				 *  	return value.match( /(foo|baz)/g ) || [ 'bar', 'quix' ];
+				 * }
+				 * ```
+				 *
+				 * @since 4.12.0
+				 * @param {String} value The shorthand property value.
+				 * @param {Function} [split] Function used to split `value`.
+				 * If not set, property value will be splitted by spaces.
+				 * @param {String[]} split.return Splitted shorthand value.
```
Comandeer: In generated docs it's marked as property of `split`, not the return value.

core/tools.js
```diff
@@ -1825,55 +1863,134 @@
...
+				 * @since 4.12.0
+				 * @param {Object} styles Border styles shorthand object.
+				 * @param {Object} [styles.border-color] Border color shorthand.
+				 * @param {Object} [styles.border-style] Border style shorthand.
+				 * @param {Object} [styles.border-width] Border width shorthand.
+				 * @param {Object} [fallback] Fallback object used to fill up missing style.
+				 * @param {Object} [fallback.color] Color CSS style used in absence of `border-color` style.
+				 * @param {Object} [fallback.style] Style CSS style used in absence of `border-style` style.
+				 * @param {Object} [fallback.width] Width CSS style used in absence of `border-width` style.
+				 * @returns {Object}
```
Comandeer: In fact it's more of `Object.<String, CKEDITOR.tools.style.borderStyle>`.

core/tools.js
```diff
@@ -2437,6 +2554,59 @@
...
+			color: [
+				[ /windowtext/g, 'black' ]
+			]
+		},
+
+		_normalize: function() {
+			for ( var propName in this._normalizeMap ) {
+				var val = this[ propName ];
+
+				if ( val ) {
```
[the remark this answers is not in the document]
Comandeer: OTOH: is this normalization border-only? IMO we should replace all `windowtext` occurrences in all pasted styles, not only in borders.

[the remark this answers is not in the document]
jacekbogdanski: `propName` is taken from `_normalizeMap` but `this` can contain different info. Suppose you have something like:
```

File: plugins/pastefromword/filter/default.js (hunk 10 above)
Comment (jacekbogdanski) [reply plugins/pastefromword/filter/default.js]: It supposed to make some performance difference where regex object is created only once. However I tested it a bit and it looks like there is no difference or it's marginal, so it doesn't make sense to keep it separated so far from usage for readability.

File: plugins/pastefromword/filter/default.js (hunk 11 above)
Comment (jacekbogdanski) [reply plugins/pastefromword/filter/default.js]: AFAIK World/Excel uses pixels for width/height data attributes, at least for images and tables. I'm not eager for premature implementation to keep the code simple. Do you have some verification that these attributes may have a different type?

File: tests/plugins/pastefromword/manual/pastetable.html (hunk 12 above)
Comment (jacekbogdanski) [reply tests/plugins/pastefromword/manual/pastetable.html]: Unfortunately, it still returns an empty string for IE. The input element is not a part of an editor, so it requires additional code to construct our dataTransfer object for the same results.

**Comment by jacekbogdanski** [comment]
I don't think we will get much if `BorderStyle` constructor functions will be prived. I think that making them public creates a unified namespace for border style helpers and in the future (or maybe now?), we may decide to deprecate `CKEDITOR.tools.style.parse.border` which is just a dummy alias now.

I'm not sure if we want to keep `borderStyle.parse.*` namespace. I'm not a big fan of long namespaces, however, it nicely describes contained functions.

File: plugins/pastefromword/filter/default.js (hunk 6 above)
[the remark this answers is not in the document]
Comment (Comandeer) [reply plugins/pastefromword/filter/default.js]: Difference in sizes between 50pt (bigger) and 50px (smaller)

Converting in this way we change the actual size of the text. To get the real size we need to get the size of pt in pixels. To do this we must use `getComputedStyle` on element, which has styles declared in pt. However it can be extracted to the separate issue.

File: plugins/pastefromword/filter/default.js (hunk 11 above)
Comment (Comandeer) [reply plugins/pastefromword/filter/default.js]: Ok, so let's wait for actual reports 👍

**Comment by jacekbogdanski** [comment]
Your renaming looks nice. Not sure if really need to keep `parse` namespace for such a small class. Actually, parsing methods are in fact object constructors. Maybe we could skip `parse` namespace and create something like:

* `CKEDITOR.tools.style.borderStyle` → `CKEDITOR.tools.style.border`
* `CKEDITOR.tools.style.borderStyle.parse.border` → `CKEDITOR.tools.style.border.fromCssRule`
* `CKEDITOR.tools.style.borderStyle.parse.splitBorder` → `CKEDITOR.tools.style.border.mapBorderCssRules`

Not sure about the last one. `splitBorder` looks for me a bit like splitting single rule e.g. `1px solid black` but it actually maps the whole border shorthands into separate border objects.

**Comment by jacekbogdanski** [comment]
After small chat with @Comandeer, we decided that the above naming will suit our needs, with one exception for the last one:

`CKEDITOR.tools.style.borderStyle.parse.splitBorder` -> `CKEDITOR.tools.border.splitCssValues`

File: core/tools.js [hunk 16]
```diff
@@ -1805,11 +1808,46 @@
...
+				 * // Logs: { top: 'bar', right: 'quix', bottom: 'bar', left: 'quix' }
+				 *
+				 * function split( value ) {
+				 *  	return value.match( /(foo|baz)/g ) || [ 'bar', 'quix' ];
+				 * }
+				 * ```
+				 *
+				 * @since 4.12.0
+				 * @param {String} value The shorthand property value.
+				 * @param {Function} [split] Function used to split `value`.
```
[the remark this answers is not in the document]
jacekbogdanski: It would look cleaner but we already have a case where `split` function uses a default value if match fails https://github.com/ckeditor/ckeditor-dev/blob/t/1490/core/tools.js#L1812

core/tools.js
```diff
@@ -1805,11 +1808,46 @@
...
+				 * If not set, property value will be splitted by spaces.
+				 * @param split.return String[] Splitted shorthand value.
+				 * @returns {Object}
```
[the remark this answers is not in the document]
jacekbogdanski: Actually, it should be `Object.<String, String>`. I incorrectly copied the above type from `CKEDITOR.tools.style.parse.margin` which is also incorrect. Will update both types.

core/tools.js
```diff
@@ -1805,11 +1808,46 @@
...
+				 * function split( value ) {
+				 *  	return value.match( /(foo|baz)/g ) || [ 'bar', 'quix' ];
+				 * }
+				 * ```
+				 *
+				 * @since 4.12.0
+				 * @param {String} value The shorthand property value.
+				 * @param {Function} [split] Function used to split `value`.
+				 * If not set, property value will be splitted by spaces.
+				 * @param split.return String[] Splitted shorthand value.
```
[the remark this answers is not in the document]
jacekbogdanski: It doesn't look perfect but we are already using this convention: https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_plugins_textMatch.html
It looks readable for me although incorrect.

I'm not sure if it's worth to provide additional own type for such simple callback - if I'm correct most of our callbacks are rather explained in docs than exists as a separate abstract type definition. What do you think about improving callback description rather providing dedicated type?

plugins/pastefromword/filter/default.js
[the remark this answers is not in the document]
jacekbogdanski: I'm not sure if I understand the issue. As you can see this function uses `CKEDITOR.tools.convertToPx` under the hood, so it won't give `50pt` -> `50px` but `50pt` -> `67px`. You can test it here: https://codepen.io/jacekbogdanski/pen/rProxW

What's wrong with that?

jacekbogdanski:
I fixed issues from the latest review. It looks for me that we are pretty close to the final form of the implementation. It's probably a good time to start working on unit tests. Please, let me know if you also share the same feeling - if so, I will follow up with unit tests and changelog updates.

JohnCSMoodys:
Do you have any update when 4.11.3 (that has the pastefromword retain cell border fix) will be available for download? We are using 4.8.0 right now  and are looking forward for the fix to the cell border issue.

Comandeer:
@JohnCSMoodys, this fix is not merged yet and not available in 4.11.3. It should be available in 4.12.0, which does not have release date set yet.

Comandeer:
Rebased onto latest `major`.

tests/plugins/pastefromword/generated/_fixtures/Meta_generator/Excel/expected.html
```diff
@@ -52,9 +52,9 @@
 						</span>
 					</span>
 				</span>
-		</td>
-		<td style="background-color:#ffc000; height:26.25pt; vertical-align:bottom; white-space:nowrap; width:48pt">
-				<span style="font-size:11pt">
+			</td>
+			<td style="background-color:#ffc000; border-bottom:1px solid black; border-left:1px solid black; border-right:1px solid black; border-top:1px solid black; vertical-align:bottom; white-space:nowrap; width:64px">
```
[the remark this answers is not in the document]
jacekbogdanski: It's not lost. It rather never been there:
https://github.com/ckeditor/ckeditor-dev/blob/f0e034d40432cdb0f8e14c20538be9fac1a78b36/tests/plugins/pastefromword/generated/_fixtures/Meta_generator/Excel/excel2016/datatransfer.html#L207-L218
But I see that in the test case `tr` ascendant element have duplicated `height`. I'm wondering if it's worth to copy this property into a all child cells? The single first cell should be enough to keep row heigt. WDYT?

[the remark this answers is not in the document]
Comandeer: I'm wondering if it can be broken by some styling involving e.g. `line-height`. In such case, if any other cell has height bigger than 35px, the whole row will be taller than intended. But it's rather an edge case.

I was just wondering, why the height disappeared.

Comandeer:
> Would you like to change very simple regex into more complicated border picking mechanism just to extract it into separate ticket and than restore current implementation? What will we gain by doing this?

More clean implementation and much more extensible mechanism. However if it's really that complicated, then probably it would be better to leave it as it is. We can always just create a followup ticket to improve it in the future.

> Do you mean border color normalization from `windowtext` to black? If so, I'm rather thinking that currently it's nicely hidden under Border type and works correctly, we just need a separate ticket to fix the rest of our styles with this mapping and in the meantime making normalization feature from it (as manageable mapping) / refactoring Border type to use it. WDYT?

Yup, I was thinking about such normalization feature.

> Good point, additionally performance for this operation should be much better than using a hidden element for that. However, it probably should be upstream to `CKEDITOR.tools.convertToPx` method than separate function doing the same.

Yes, upstream sounds like a good idea.

jacekbogdanski:
Extracted issues:
https://github.com/ckeditor/ckeditor-dev/issues/2923
https://github.com/ckeditor/ckeditor-dev/issues/3111
https://github.com/ckeditor/ckeditor-dev/issues/3112
https://github.com/ckeditor/ckeditor-dev/issues/3113

jacekbogdanski:
The issue with IE8 is caused by https://github.com/ckeditor/ckeditor-dev/issues/3120

Comandeer:
Rebased onto latest `major`.

PR #2229, opened 2018-07-16
Custom tags starting with protected element tag names are no longer removed when added to ACF

### LINKED ISSUE #988: Custom tags starting with embed are removed even when added to ACF.

Bug

1. Use below configuration:
```
CKEDITOR.dtd['embedded-expression'] = { '#' : 1, 'span' : 1, 'img' : 1 };
CKEDITOR.dtd.$inline['embedded-expression'] = 1;
CKEDITOR.dtd.body['embedded-expression'] = 1; 

var editor = CKEDITOR.replace( 'editor1', {
    extraAllowedContent : 'embedded-expression[*]{*}(*)'
});
```
2. Load sample page in browser and paste the following HTML in source mode `<embedded-expression data-test="test">test</embedded-expression>`
3. Switch to wysiwyg and back to source.

### Expected result

Tags stays

### Actual result

Tag gets removed.

## Other details

* Browser: Any
* OS: Any
* CKEditor version: N/A

It seems these lines are responsible for that result: https://github.com/ckeditor/ckeditor-dev/blob/a0dd80a/core/htmldataprocessor.js#L802-L806.

A simple workaround would be changing tag name so that it doesn't start with embed.

engineering-this:

Bug fix

### This PR contains

- [X] Unit tests
- [X] Manual tests

## What changes did you make?

Regex for filtering protected element tag names ( eg. 'embed' ) was previously filtering out custom tag names that were beginning with protected element tag ( eg. 'embedded' ). After fix it filters only exact tags from expression.

Closes #988


CHANGES.md
```diff
@@ -32,6 +32,7 @@ Fixed Issues:
 * [#1592](https://github.com/ckeditor/ckeditor-dev/issues/1592): Fixed: The [Image Base](https://ckeditor.com/cke4/addon/imagebase) plugin caption is not visible after paste.
 * [#620](https://github.com/ckeditor/ckeditor-dev/issues/620): Fixed: The [`config.forcePasteAsPlainText`](https://docs.ckeditor.com/ckeditor4/latest/api/CKEDITOR_config.html#cfg-forcePasteAsPlainText) option is not respected in internal and cross-editor pasting.
 * [#1467](https://github.com/ckeditor/ckeditor-dev/issues/1467): Fixed: The resizing cursor of the [Table Resize](https://ckeditor.com/cke4/addon/tableresize) plugin appearing in the middle of a merged cell.
+* [#988](https://github.com/ckeditor/ckeditor-dev/issues/988): Fixed: Custom tags starting with protected element tag names are filtered even when added to ACF.
```
jacekbogdanski: Move changelog entry into `4.10.1`.

tests/core/filter/customtag.js
```diff
@@ -0,0 +1,59 @@
...
+
+	bender.editor = {
+		config: {
+			extraAllowedContent: extraAllowedContent,
+			autoParagraph: false
+		}
+	};
+
+	bender.test( {
+		'test allowed custom tags': function() {
```
jacekbogdanski: Missing ticker reference.

tests/core/filter/customtag.js
```diff
@@ -0,0 +1,59 @@
...
+					'<htmlfoo data-foo="bar">htmlfoo</htmlfoo>' +
+					'<headfoo data-foo="bar">headfoo</headfoo>' +
+					'<bodyfoo data-foo="bar">bodyfoo</bodyfoo>' +
+					'<titlefoo data-foo="bar">titlefoo</titlefoo>';
+
+			bot.setData( data, function() {
+				assert.areEqual( data, editor.getData() );
+			} );
+		},
+		'test not allowed custom tags': function() {
```
jacekbogdanski: Missing ticket reference.

tests/core/filter/manual/customtag.md
```diff
@@ -0,0 +1,26 @@
+@bender-tags: bug, 988, 4.10.1
+@bender-ui: collapsed
+@bender-ckeditor-plugins: wysiwygarea, toolbar, sourcearea
+
+----
```
jacekbogdanski: Not sure if we need this separator. Didn't see it in our codebase.

tests/core/filter/manual/customtag.md
```diff
@@ -0,0 +1,26 @@
...
+
+	- objectfoo
+	- embedfoo
+	- paramfoo
+	- htmlfoo
+	- headfoo
+	- bodyfoo
+	- titlefoo
+
+Tags, has \[data-foo="bar"\] attribute, and text matching tag name.
```
jacekbogdanski: Looks better if you wrap it with code block, i.e. `[data-foo="bar"]`.

tests/core/htmldataprocessor/customtag.js
```diff
@@ -0,0 +1,88 @@
...
+				'<embed data-bar="foo"></embed>' +
+				'<param data-bar="foo"/>' +
+				'<html data-bar="foo">html</html>' +
+				'<head data-bar="foo">head</head>' +
+				'<body data-bar="foo">body</body>' +
+				'<title data-bar="foo">title</title>',
+
+			expected: '<object data-bar="foo">object</object>' +
+				'<embed data-bar="foo"></embed>' +
+				'<param data-bar="foo" />' +
```
jacekbogdanski: Not sure, but isn't `param` element [only allowed inside `object` element?](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/param).

tests/core/htmldataprocessor/customtag.js
```diff
@@ -0,0 +1,88 @@
...
+				'htmlheadbody' +
+				'<title data-bar="foo"></title>'
```
jacekbogdanski: Shouldn't `title` element be removed from editors content? It's `head` element and defines correct HTML document.

CHANGES.md
```diff
@@ -11,6 +11,7 @@ Fixed Issues:
 * [#2167](https://github.com/ckeditor/ckeditor-dev/issues/2167): Fixed: Matching in [Emoji](https://ckeditor.com/cke4/addon/emoji) plugin is not case insensitive.
 * [#1084](https://github.com/ckeditor/ckeditor-dev/issues/1084) Fixed: Using the "Automatic" option with [Color Button](https://ckeditor.com/cke4/addon/colorbutton) on a text with color already defined sets an invalid color value.
 * [#966](https://github.com/ckeditor/ckeditor-dev/issues/966): Fixed: Executing [`editor.destroy()`](https://docs.ckeditor.com/ckeditor4/latest/api/CKEDITOR_editor.html#destroy) during [file upload](https://docs.ckeditor.com/ckeditor4/latest/api/CKEDITOR_fileTools_uploadWidgetDefinition.html#onUploading) throws error. Thanks to [Maksim Makarevich](https://github.com/MaksimMakarevich)!
+* [#988](https://github.com/ckeditor/ckeditor-dev/issues/988): Fixed: Custom tags starting with protected element tag names are filtered even when added to ACF.
```
jacekbogdanski: Maybe something like:

ACF - enabled custom elements prefixed with `object`, `embed`, `param`, `html`, `body`, `head`, `title` are removed from editor's content.

`proctected element tag names` doesn't tell much.

tests/core/htmldataprocessor/customtag.js
```diff
@@ -0,0 +1,88 @@
...
+					'<htmlbar data-bar="foo">htmlbar</htmlbar>' +
+					'<headbar data-bar="foo">headbar</headbar>' +
+					'<bodybar data-bar="foo">bodybar</bodybar>' +
+					'<titlebar data-bar="foo">titlebar</titlebar>',
+
+			expected: 'objectbarembedbarparambarhtmlbarheadbarbodybartitlebar'
+		} ),
+
+		// (#988)
+		'test protected element names': assertHtmlDataProcessor( {
```
jacekbogdanski: The test fails on IE8.

tests/core/htmldataprocessor/customtag.js
jacekbogdanski: `title` is also removed on IE8 where the test fails.

engineering-this:
Unfortunately comments in code related to protecting elements, aren't very clear on what and why it is done. After reading code from start again it looks like I had wrong idea.

https://github.com/ckeditor/ckeditor-dev/blob/104d650639301ac993bcc66d349654dc4b9fe426/core/htmldataprocessor.js#L73-L75

What I understand now is that we protect elements, so browser doesn't break them when parsing, not because we want to remove them from editor.

I have removed `test protected elements`, as element names protection is tested in `/tests/core/htmldataprocessor/htmldataprocessor`. Breaking regex would result in many test cases failing in this test suite. Some changes to regex doesn't break tests, but possibly some cases doesn't need protection from browser.

It also looks like names unprotected early with following regex:
https://github.com/ckeditor/ckeditor-dev/blob/47a07b061363a494ac82fad9947b0c4129527160/core/htmldataprocessor.js#L804

Doesn't remove custom elements, so I will include in changelog only ones that are removed.

tests/core/htmldataprocessor/customtag.js
```diff
@@ -0,0 +1,88 @@
...
+
+		// (#988)
+		'test not allowed custom tags': assertHtmlDataProcessor( {
+			data: '<objectbar data-bar="foo">objectbar</objectbar>' +
+					'<embedbar data-bar="foo">embedbar</embedbar>' +
+					'<parambar data-bar="foo">parambar</parambar>' +
+					'<htmlbar data-bar="foo">htmlbar</htmlbar>' +
+					'<headbar data-bar="foo">headbar</headbar>' +
+					'<bodybar data-bar="foo">bodybar</bodybar>' +
+					'<titlebar data-bar="foo">titlebar</titlebar>',
```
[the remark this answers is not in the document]
engineering-this: This TC has different string.

tests/core/htmldataprocessor/customtag.js
```diff
@@ -0,0 +1,68 @@
...
+
+			expected: 'objectbarembedbarparambarhtmlbarheadbarbodybartitlebar'
+		} )
+	} );
+
+	function assertHtmlDataProcessor( config ) {
```
jacekbogdanski: What do you think about `assertDataFormat` function name?

tests/core/htmldataprocessor/customtag.js
```diff
@@ -0,0 +1,68 @@
...
+		'embed',
+		'param',
+		'html',
+		'head',
+		'body',
+		'title'
+	],
+	extraAllowedContent = '';
+
+	CKEDITOR.tools.array.forEach( elements, function( item ) {
```
mlewand: This could be easily replaced with a one liner, like:

`elements.join( 'foo[*]{*}(*);' ) + 'foo[*]{*}(*);' or `elements.concat( [ '' ] ).join( 'foo[*]{*}(*);' )`  (though I like second more because it has less repetition, conat `[ '' ]` thing might not be obvious in a first place.

From there `extraAllowedContent` var can be inlined directly into config object.

tests/core/htmldataprocessor/customtag.js
```diff
@@ -0,0 +1,68 @@
...
+			return assertDataFormat( {
+				data: data,
+				expected: data
+			} );
+		} )(),
+
+		// (#988)
+		'test not allowed custom tags': assertDataFormat( {
+			data: '<objectbar data-bar="foo">objectbar</objectbar>' +
+					'<embedbar data-bar="foo">embedbar</embedbar>' +
```
mlewand: Wrong indentation, should be one tab less.

tests/core/htmldataprocessor/manual/customtag.html
```diff
@@ -0,0 +1,32 @@
...
+			'param',
+			'html',
+			'head',
+			'body',
+			'title'
+		],
+		extraAllowedContent = '';
+
+	CKEDITOR.tools.array.forEach( elements, function( item ) {
+		var customTag = item + 'foo';
```
mlewand: Here, the same note as in `tests/core/htmldataprocessor/customtag.js`.

PR #2246, opened 2018-07-20
More advanced image plugin is now preferred in the case of conflict

### LINKED ISSUE #1791: Disable image and image2 plugin when Easy Image is present

Task

## Description

Currently it's possible to enable Easy Image alongside image and image2 plugins, resulting in strange editor's behaviour. Image plugin has code to prevent loading it when image2 is present. Adding similar code to both image and image2 plugin to prevent loading when EI is present seems like a sensible solution.

### LINKED ISSUE #2249: Plugins conflict detection helper
## Type of report

Feature request

Some plugins require additional conflict detection. This feature could be extracted into separate function like `CKEDITOR.tools.detectPluginsConflict`. You can read more about the issue here: #1791 and in the PR discussion: https://github.com/ckeditor/ckeditor-dev/pull/2246#pullrequestreview-139034733

jacekbogdanski:

Task

### This PR contains

- [x] Unit tests
- [x] Manual tests

## What changes did you make?

More advanced image plugin is now preferred in the case in conflict. I will also create a followup issue to add missing warning description into our docs.

Error code reference followup: https://github.com/ckeditor/ckeditor-docs/issues/179

Closes #1791, closes #2249


core/tools.js
```diff
@@ -2298,6 +2298,44 @@
...
+		 * @return {Boolean} Returns true, if there is some conflict. Returns false otherwise.
+		 */
+		detectPluginsConflict: function( editor, plugin, conflicted ) {
+			var conflictPlugins = [];
+
+			for ( var i = 0; i < conflicted.length; i++ ) {
+				var pluginName = conflicted[ i ];
+
+				if ( editor.plugins[ pluginName ] ) {
+					conflictPlugins.push( pluginName );
```
Comandeer: We don't need the array of all conflicting plugins. In fact we need only the last one. It could be simplified by reversing `conflicted` parameter (or reversing the loop) and returning as early as the first matched element.

core/tools.js
```diff
@@ -2298,6 +2298,44 @@
...
+		 * message.
+		 *
+		 * ```javascript
+		 * CKEDITOR.tools.detectPluginsConflict( editor, 'image', [ 'image2', 'easyimage' ] );
+		 * ```
+		 *
+		 * @param {CKEDITOR.editor} editor Editor searched for conflicting plugins.
+		 * @param {String} plugin Current plugin name.
+		 * @param {String[]} conflicted Names of plugins that are conflicted with a current plugin.
+		 * @return {Boolean} Returns true, if there is some conflict. Returns false otherwise.
```
Comandeer: `@since` is missing.

core/tools.js
Comandeer: OTOH – IMO it's more intuitive to have the option with the highest priority at the beginning, not the end of the array.

PR #2255, opened 2018-07-24
Widget definition's insert method provides commandData and editor instance

### LINKED ISSUE #2253: Widget definition insert method should include editor instance and command data
## Type of report

Feature request

As requested in the #235 PR, currently [widgetDefinition.insert](https://docs.ckeditor.com/ckeditor4/latest/api/CKEDITOR_plugins_widget_definition.html#property-insert) doesn't feature neither `editor` nor `commandData` passed to the command that caused the insert method call.

While `editor` issue could be mitigated, by storing `editor` variable as a local variable in the scope, and passing it later on, same can not happen for `commandData`.

Adding options object to the API that would provide `editor` and `commandData` will solve this problem.

mlewand:

New feature

### This PR contains

- [x] Unit tests
- [ ] Manual tests

No need for manual test as this is API only change.

## What changes did you make?

Simply added proper argument, and tests.

Follow-up of a #235.

Closes #2253.

### DIFF
--- a/plugins/widget/plugin.js
+++ b/plugins/widget/plugin.js
@@ -1902,15 +1902,17 @@
 		editor.addCommand( widgetDef.name, {
 			exec: function( editor, commandData ) {
 				var focused = editor.widgets.focused;
-				// If a widget of the same type is focused, start editing.
-				if ( focused && focused.name == widgetDef.name )
+				if ( focused && focused.name == widgetDef.name ) {
+					// If a widget of the same type is focused, start editing.
 					focused.edit();
-				// Otherwise...
-				// ... use insert method is was defined.
-				else if ( widgetDef.insert )
-					widgetDef.insert( editor );
-				// ... or create a brand-new widget from template.
-				else if ( widgetDef.template ) {
+				} else if ( widgetDef.insert ) {
+					// ... use insert method is was defined.
+					widgetDef.insert( {
+						editor: editor,
+						commandData: commandData
+					} );
+				} else if ( widgetDef.template ) {
+					// ... or create a brand-new widget from template.
 					var defaults = typeof widgetDef.defaults == 'function' ? widgetDef.defaults() : widgetDef.defaults,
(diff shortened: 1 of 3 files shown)


plugins/widget/plugin.js
```diff
@@ -4041,6 +4043,9 @@
  * * The widget element will be inserted into DOM.
  *
  * @property {Function} insert
+ * @param {Object} options Options object added in **4.11.0**.
+ * @param {CKEDITOR.editor} options.editor Editor where the widget is going to be inserted to.
+ * @param {Object} [options.commandData] Command data passed to the invoking command, if any.
```
jacekbogdanski: After docs built additional params are not visible in docs.

mlewand: @jacekbogdanski Interesting find, we have actually many of those:

https://github.com/ckeditor/ckeditor-dev/blob/8e2b356/plugins/embedbase/plugin.js#L618-L619
https://github.com/ckeditor/ckeditor-dev/blob/8e2b356/plugins/embedbase/plugin.js#L625-L626
https://github.com/ckeditor/ckeditor-dev/blob/8e2b356/plugins/uploadwidget/plugin.js#L300-L302
https://github.com/ckeditor/ckeditor-dev/blob/8e2b356/plugins/widget/plugin.js#L4227-L4229

Just to give a few. It was my presumption that it works correctly.

It might happen that this is a regression after switching to Umberto, I'll investigate this.

engineering-this: @mlewand Jaceks finding got my attention, and after super short research I found that:
https://github.com/senchalabs/jsduck/wiki/@param
`Documents parameters for methods and events.`
So @param can be defined on method, but not on property.

mlewand: It was actually working before, this is upstream issue, I have created an issue for that cksource/umberto#567.

jacekbogdanski:
LGTMI don't see more problems with this PR except for the one already extracted.

PR #2259, opened 2018-07-25
Improved image aspect ratio locker

### LINKED ISSUE #2254: Image ratio is not correctly locked for images with odd pixel values
## Type of report

Bug

1. Open editor sample with `image` plugin.
1. Make sure that `Lock ratio` option is locked.
1. Add image into an editor and change its ratio from even height into an odd height e.g from `1800x1200` to `500x333`. You can use http://via.placeholder.com/1800x1200 for this test.
1. Check if `Lock ratio` option is locked.

### Expected result

`Lock ratio` option should be locked.

### Actual result

`Lock ratio` option is not locked.

## Other details

* CKEditor version: 4.10.0
* Installed CKEditor plugins: image

jacekbogdanski:

Bug fix

### This PR contains

- [x] Unit tests
- [x] Manual tests

## What changes did you make?

I refactored `switchLockRatio` function and added tests.

Closes #2254

### DIFF
--- a/plugins/image/dialogs/image.js
+++ b/plugins/image/dialogs/image.js
@@ -118,26 +118,23 @@
 								height = dialog.getValueOf( 'info', 'txtHeight' ),
 								originalRatio = oImageOriginal.$.width / oImageOriginal.$.height,
 								thisRatio = width / height;
+
 							dialog.lockRatio = false; // Default: unlock ratio
 
-							if ( !width && !height )
+							if ( !width && !height ) {
 								dialog.lockRatio = true;
-							else if ( !isNaN( originalRatio ) && !isNaN( thisRatio ) ) {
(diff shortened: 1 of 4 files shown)


mlewand:
@Comandeer if possible, I'd stay away from introducing new stuff. Keep the compatibility first - that was meant to be a limited improvement, we're not interested in rewriting parts of this plugin.

jacekbogdanski:
@Comandeer it looks like somehow proposed implementation is less precise than `toFixed` function and gives better results. You can try https://github.com/ckeditor/ckeditor-dev/tree/t/2254-b - reproduce manual test steps but instead setting a width to `500px`, set it to `100px`. With new implementation, locker gets broken, although proposed implementation works fine. It's an edge case but possible to avoid.

PR #2274, opened 2018-07-31
Fixed Ctrl+A focusing inline editor if it was started and ended with a list

### LINKED ISSUE #1719: CTRL+A forces focus on editor starting and/or ending with a list in it's content

The strangest thing: In Chrome, using CTRL+A anywhere on the page will give an unexpected result while there is a list set at the beginning and/or end of the content in the editor. It sets the focus on the editor containing the list.

Since we are using multiple editors on one page, this results in a focus shift from the editor without list to the editor of which the content starts and/or ends in a list.

Only happens with divarea

Strange behaviour with list:
1. Add a list in the editordivarea (make sure the content starts or ends with a list)
2. Remove focus from the editor
3. Hit CTRL+A, see how focus turns to editor...
4. Remove focus from the editor
5. Hit CTRL+A, expected behaviour is back

### Expected result

Expected behaviour:
1. Type some tekst in an editor divarea
2. Remove focus from the editor
3. Hit CTRL+A

## Other details

* Browser: Chrome
* OS: Windows
* CKEditor version: 4.7.2
* Installed CKEditor plugins: divarea

mlewand:

Bug fix

### This PR contains

- [x] Unit tests
- [x] Manual tests

## What changes did you make?

The issue is described here: https://github.com/ckeditor/ckeditor-dev/pull/1773#discussion_r206509602

I had to adjust the initial pull request as commented above. Also tests were added.

Follow-up of #1773.

Closes #1719.

### DIFF
--- a/tests/plugins/widgetselection/keys.js
+++ b/tests/plugins/widgetselection/keys.js
@@ -5,10 +5,10 @@
 	'use strict';
 
 	bender.editors = {
-		editor1: {
+		classic: {
 			creator: 'replace'
 		},
-		editor2: {
+		divarea: {
 			creator: 'replace',
 			config: {
 				extraPlugins: 'divarea'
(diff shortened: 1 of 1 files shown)


plugins/widgetselection/plugin.js
```diff
@@ -22,10 +22,9 @@
 				editor.on( 'contentDom', function( evt ) {
 
 					var editor = evt.editor,
-						doc = editor.document,
 						editable = editor.editable();
 
-					editable.attachListener( doc, 'keydown', function( evt ) {
+					editable.attachListener( editable, 'keydown', function( evt ) {
```
jacekbogdanski: I would be nice to provide some simple explanation comment with the issue ticket reference for this change.

tests/plugins/widgetselection/keys.js
```diff
@@ -4,12 +4,36 @@
 ( function() {
 	'use strict';
 
-	bender.editor = true;
+	bender.editors = {
+		editor1: {
```
jacekbogdanski: The name doesn't tell much. Maybe `classic`?

tests/plugins/widgetselection/keys.js
```diff
@@ -4,12 +4,36 @@
...
+			creator: 'replace'
+		},
+		editor2: {
```
jacekbogdanski: What about `divarea`?

tests/plugins/widgetselection/keys.js
mlewand: Addressed in 84265bd798.

tests/plugins/widgetselection/keys.js
mlewand: Addressed in 84265bd798.

plugins/widgetselection/plugin.js
mlewand: TBH I don't think it is needed there, it's fairly logical that `ctrl/cmd+a` should be catched from within the editable. It's actually shame that we didn't catch this earlier, because in fact it didn't make sense in case of inline editors. 🙂

PR #2286, opened 2018-08-03
Renamed telephone into phone for link plugin

### LINKED ISSUE #2284: Rename phone link option label
## Type of report

Task

In #2154 we added a phone link type. That's all good, but it got labelled as "telephone", while more common way to say that is just "phone".

Let's simplify this before we ship this feature.

## Other details

* Browser: Any
* CKEditor version: 4.11.0

jacekbogdanski:

Task

## What changes did you make?

I refactored `Telephone` name into `Phone` for our lang files. I Also added some missing meta tags and changes docs + tests.

I see that we still have `Telephone` labels for `form` plugin. Not sure if we want to change it also.

Closes #2284


plugins/link/plugin.js
```diff
@@ -882,7 +882,7 @@
 		 */
 
 		/**
-		 * Optional JavaScript Regular Expression whenever telephone numbers in Link dialog should be validated.
+		 * Optional JavaScript Regular Expression whenever phone numbers in Link dialog should be validated.
 		 *
 		 *		config.linkTelNumberRegExp = /^[0-9]{9}$/;
```
Comandeer: After this change, config variables have names using the old name of the field. What about changing them to `linkPhoneNumberRegExp` and `linkPhoneNumberMsg` or – even shorter – `linkPhoneRegExp` and `linkPhoneMsg`?

jacekbogdanski: That makes sense, I will refactor variables in the following commits.

PR #2299, opened 2018-08-09
Improved IE 10 and IE 11 version detection

uzquiano:
In IE 10 and IE 11, there are cases observed where document.documentMode may be undefined and the user agent does not match what CKEditor expects for detection.  For example, in IE11, the user agent may be "Mozilla/5.0 (Windows NT 6.3); Trident/7.0; rv:11.0) like Gecko) and documentMode can be undefined.  In this case, CKEditor will error out on line 196 with:

"SCRIPT5007: Unable to get property '1' of undefined or null reference"

The code added here does a slower but more rigorous check of the user agent.  For IE10 or IE11, if documentMode is undefined, the version is extracted from the user agent and set properly.  This makes it possible for CKEditor 4.10 to support IE11 (at least in the cases that we tested).

### This PR contains

- [ ] Unit tests
- [ ] Manual tests

## What changes did you make?

*Give an overview…*

### DIFF
--- a/core/env.js
+++ b/core/env.js
@@ -193,7 +193,17 @@ if ( !CKEDITOR.env ) {
 			if ( edge ) {
 				version = parseFloat( edge[ 1 ] );
 			} else if ( env.quirks || !document.documentMode ) {
-				version = parseFloat( agent.match( /msie (\d+)/ )[ 1 ] );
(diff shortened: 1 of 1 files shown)


mlewand:
Can you provide us with exact steps needed to reproduce this case?

mlewand:

PR #1494, opened 2018-01-24
Selection is now preserved after images upload

jacekbogdanski:
## What is the purpose of this pull request?

New feature, bug fix

## Does your PR contain necessary tests?

All patches which change the editor code must include tests. You can always read more
on [PR testing](https://docs.ckeditor.com/ckeditor4/docs/#!/guide/dev_contributing_code-section-tests),
[how to set the testing environment](https://docs.ckeditor.com/ckeditor4/docs/#!/guide/dev_tests) and
[how to create tests](https://docs.ckeditor.com/ckeditor4/docs/#!/guide/dev_tests-section-creating-your-own-test)
in the official CKEditor documentation.

### This PR contains

- [x] Unit tests
- [x] Manual tests

## What changes did you make?

Added selection preserving for [Upload Image](https://ckeditor.com/cke4/addon/uploadimage) based on starting editor selection or selection change during upload.

Closes #1493, #1217

PR #2377, opened 2018-09-05
Fixed an issue causing error when pasting multiple images

### LINKED ISSUE #1217: Exception thrown when pasting multiple images into an empty editable

Bug

1. Open a new tab - https://codepen.io/mlewand/pen/RjyoPR.
2. Select all four images rendered in HTML preview.
3. Copy with `ctrl/cmd+c`.
4. Open https://sdk.ckeditor.com/samples/fileupload.html#uploading-dropped-and-pasted-images.
1. Open developer console.
5. Focus editor under the "Uploading Dropped and Pasted Images" header.
1. Clear editor contents.
6. Paste using `ctrl/cmd+v`.

### Expected result

Once images are uploaded no exceptions are thrown.

### Actual result

Following error is reported:

```
ckeditor.js:443 The given range isn't in document.
```

## Other details

* Browser: Chrome
* OS: Windows 10
* CKEditor version: 4.7.3

jacekbogdanski:

Bug fix

### This PR contains

- [x] Unit tests
- [x] Manual tests

## What changes did you make?

Removed forcing image selection for `uploadwidgets#replaceWith` function.

Closes #1217


tests/plugins/uploadwidget/manual/paste.md
```diff
@@ -0,0 +1,38 @@
+﻿@bender-tags: 4.10.2, bug, clipboard, widget, filetools, 1217
+@bender-ui: collapsed
+@bender-ckeditor-plugins: wysiwygarea, toolbar, uploadwidget, image2, uploadimage, uploadfile
```
engineering-this: I wonder if it's ok to skip `uploadwidget` plugin, as it is required by other listed plugins. Or is it better to keep it explicitly required like this. WDYT?

tests/plugins/uploadwidget/manual/paste.md
```diff
@@ -0,0 +1,38 @@
...
+@bender-include: _helpers/xhr.js
+
+## No selection
+
+1. Copy an image.
```
engineering-this: It would be nice to add here some image for copying. Especially for second test with multiple images.

tests/plugins/uploadwidget/manual/paste.md
```diff
@@ -0,0 +1,38 @@
...
+## Unexpected
+
+Image is selected.
+
+## Image preselection
+
+1. Copy an image.
+1. Focus the editor.
+1. Paste an image using `cmd/ctrl-v`.
+1. Focus pasted image during image upload.
```
engineering-this: This step is redundant, as image placeholder is focused after previous step.

tests/plugins/uploadwidget/uploadwidget.js
```diff
@@ -255,7 +255,7 @@
 			} );
 		},
 
-		'test replaceWith preserves selection placed before the widget': function() {
+		'test replaceWith preserves widget selection': function() {
```
engineering-this: This test passes on `next` branch, so your fix isn't tested.

tests/plugins/uploadwidget/manual/paste.md
jacekbogdanski: Good catch, it doesn't make sense to keep plugin if it's not required. Moreover, this test is missing proper `uploadwidget` tag.

tests/plugins/uploadwidget/manual/paste.md
jacekbogdanski: Agree, I will add some images. Note that they may look ugly thus required base64 format.

tests/plugins/uploadwidget/uploadwidget.js
jacekbogdanski: It's more like this test only validates happy path. I see your point, I will write some regression test.

tests/plugins/uploadwidget/manual/paste.html
```diff
@@ -0,0 +1,13 @@
...
+
+<div id="editor"></div>
+
+<script>
+	if ( CKEDITOR.env.ie && CKEDITOR.env.version < 10 ) {
```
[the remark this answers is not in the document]
jacekbogdanski: I couldn't get copying and pasting the image on mobile (at least on Android) to work so I decided to ignore the test.

jacekbogdanski:
I wonder how to fix this issue.

Currently, we are using [`editable.insertHtmlIntoRange`](https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_editable.html#method-insertHtmlIntoRange) function to insert image with `replaceWith` function. The advantage of this function is that it doesn't select selection nor ranges, which is important because a user can change selection during the upload process. However, it has a side effect - a selection is preserved always at the beginning of the last pasted image ( if a user didn't change it during upload ). As a result, we have the issue you requested changes for.

If we will start modyfying selection, we are back to #1494 problems where we cannot predict when selection changed by user interaction. We already decided that such complicated feature should be extracted into separate plugin.

There is another option, we could just simply use [`editable.insertHtml`](https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_editable.html#method-insertHtml) function, although this function will change selection despite of user interaction and preselect the last item.

PR #2385, opened 2018-09-06
Widget is accessible from dialog ok, show and hide events

### LINKED ISSUE #1044: Widget dialogs - provide an easy way to access edited widget in onOk / onShow
From : [https://dev.ckeditor.com/ticket/12374](https://dev.ckeditor.com/ticket/12374) Still not available in 4.7.3 :

While working on a test widget I found it quite hard to get a reference to a widget instance for which a dialog is launched.

Take the code snippet dialog (dialogs/codesnippet.js) and add there

`onOk: function( evt ) {
	console.log(evt);
	console.log(this);
}`
`onShow: function( evt ) {
	console.log(evt);
	console.log(this);
}`

See that there is no access to a widget object which is created on the fly by the widget system, in case of a new widget. As a result the usage of onOk / onShow as a replacement for commit/setup functions in each UI element is quite troublesome.

There is a workaround for it:

**onShow**

`onShow: function() {
	var widget = editor.widgets.focused;
}`

**onOk**

1. We need to add event handler in an init method executed while initializing a widget:
```
editor.widgets.add( 'xyz', {
	// ...
	init : function() {
		// ...
		// Pass the reference to this widget to the dialog. See "onOk" in the dialog definition, we needed widget there.
		this.on( 'dialog', function( evt ) {
			evt.data.widget = this;
		}, this );
	},
	// ...
}
```
2. Then this will work:
`onOk: function() {
	var widget = this.widget;
}`

In ideal world this could be simpler though.

jacekbogdanski:

New feature

### This PR contains

- [x] Unit tests
- [x] Manual tests

## What changes did you make?

Included widget into dialog events.

Closes #1044

### DIFF
--- a/plugins/widget/plugin.js
+++ b/plugins/widget/plugin.js
@@ -1187,10 +1187,12 @@
 					}
 				}, null, null, 0 );
 
-				dialog.once( 'hide', function() {
+				dialog.once( 'hide', function( evt ) {
+					evt.data.widget = that;
+
 					showListener.removeListener();
 					okListener.removeListener();
-				} );
+				}, null, null, 0 );
 			} );
 
 			return true;
--- a/plugins/widget/plugin.js
+++ b/plugins/widget/plugin.js
@@ -1158,11 +1158,13 @@
 				if ( that.fire( 'dialog', dialog ) === false )
 					return;
 
-				showListener = dialog.on( 'show', function() {
+				showListener = dialog.on( 'show', function( evt ) {
+					evt.data.widget = that;
 					dialog.setupContent( that );
-				} );
+				}, null, null, 0 );
 
-				okListener = dialog.on( 'ok', function() {
+				okListener = dialog.on( 'ok', function( evt ) {
+					evt.data.widget = that;
 					// Commit dialog's fields, but prevent from
 					// firing data event for every field. Fire only one,
 					// bulk event at the end.
@@ -1183,7 +1185,7 @@
 						that.fire( 'data', that.data );
 						that.editor.fire( 'saveSnapshot' );
 					}
-				} );
+				}, null, null, 0 );
 
 				dialog.once( 'hide', function() {
 					showListener.removeListener();
--- a/tests/plugins/widget/editing.js
+++ b/tests/plugins/widget/editing.js
@@ -55,6 +55,30 @@
 							]
 						};
 					} );
+
+					evt.editor.widgets.add( 'dialogtest1', {
+						template: '<div>bar</div>',
+						dialog: 'dialogtest1'
+					} );
+
+					CKEDITOR.dialog.add( 'dialogtest1', function() {
+						return {
+							title: 'DialogTest1',
+							contents: [
+								{
+									id: 'info',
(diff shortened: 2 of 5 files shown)


plugins/widget/plugin.js
```diff
@@ -1183,12 +1185,14 @@
 						that.fire( 'data', that.data );
 						that.editor.fire( 'saveSnapshot' );
 					}
-				} );
+				}, null, null, 0 );
```
Comandeer: Probably the priority of this listener shouldn't be modified as it basically changes the state of the editor. The simplest solution would be to introduce dedicated `onOk` handler to add appropriate data to the event.

tests/plugins/widget/manual/dialogevents.md
```diff
@@ -0,0 +1,24 @@
+@bender-tags: widget, dialog, feature, 1044
```
Comandeer: Please add version tag.

jacekbogdanski:
It makes much more sense to provide widget as dialog property. Although dialog definition is the same for each dialog instance, so widget has to be removed from the definition on 'hide' event. Nothing blows if it doesn't happen thus widget will be every time is overwritten. Although I think it may be cleaner to delete this property.

tests/plugins/widget/editing.js
```diff
@@ -335,6 +360,24 @@
 			} );
 		},
 
+		// (#1044)
+		'test opening and hiding dialog events': function() {
```
Comandeer: This test is not checking events themselves, but the fact if widget is passed to them. Therefore the name is misleading.

tests/plugins/widget/editing.js
```diff
@@ -335,6 +360,24 @@
...
+			var editor = this.editor;
+
+			this.editorBot.setData( '<p data-widget="dialogtest1" id="x">bar</p>', function() {
+				var widget = getWidgetById( editor, 'x' );
+
+				editor.once( 'dialogShow', function( evt ) {
+					evt.data.getButton( 'ok' ).click();
+					resume();
```
Comandeer: Moving assertions outside the test harms its readability. At the first glance this test looks like always failing one as there is no assertion in it. Please move assertions into the test itself.

tests/plugins/widget/editing.js
```diff
@@ -55,6 +55,31 @@
...
+							],
+							onShow: dialogEventListener,
+							onOk: dialogEventListener,
+							onHide: dialogEventListener
+						};
+					} );
+
+					function dialogEventListener( evt ) {
+						assert.isNotUndefined( this.widget, 'No widget passed for: ' + evt.name );
+						assert.areEqual( 'dialogtest1', this.widget.name, 'Invalid widget for: ' + evt.name );
```
Comandeer: If the assertions were inside the test, the whole listener would just check if `this.widget === widget`.

Comandeer:
Seems like there is a new, wider proposition for fixing integration between dialog and widgets – #2423. Let's wait for its outcome as it should also fix this issue.

PR #2390, opened 2018-09-07
Added possibility to customize dialog field default values

### LINKED ISSUE #2277: Configuration option for default dialog values
## Type of report

Feature request

I know that you can already freely customize dialog fields using API, by hooking to a proper event - but this requires some basic programming skills.

This feature requests aims to simplify default dialog fields customization, by allowing that through configuration object.

Let's talk about the formats, currently there are two proposals that come to my mind:

### Flat Object

```js
config.dialog_defaultValues = {
	// '<dialogName>.<tabId>.<fieldId>': '<value>'
	'link.info.protocol': 'https://',
	'table.info.txtWidth': '100%'
};
```

#### Pros

* Easier to read by humans, and take less space in the config file.

#### Cons

* Slightly harder for us to parse in the dialog plugin.

### Regular Object

```js
config.dialog_defaultValues = {
	link: {
		info: {
			protocol: 'https://'
		}
	},
	table: {
		info: {
			txtWidth: '100%'
		}
	}
}
```

#### Pros

* Easier to operate from API.

#### Cons

* It's more verbose, takes more space.
* Harder for non tech-savvy people, especially when they're using simple tools like notepad for adjusting the values.

### Future Extensibility

I can see that both formats could be further extended if there will be need for that.

For instance, assume that the default value should be different for a editor in RTL language, or that the value should differ in case there is a `xyz` plugin.

To satisfy this requirement the values could be provided as functions, like that:

```js
config.dialog_defaultValues = {
	'link.info.protocol': function( editor ) {
		return editor.lang.dir == 'ltr' ? 'foo' : 'bar';
	}
};
```

I'm just wondering wouldn't it be useful also to allow more customization, like mark the field as required or hide a given view, like so:

```js
config.dialog_defaultValues = {
	'link.info.protocol': 'https://', // Simple format.
	'table.info.txtWidth': { // Advanced format.
		placeholder: 'width',
		required: true,
		type: 'number'
	}
};
```

In which case the config property should get more generic name, that does not imply that it's about dialog values.

### Concerns

I can already see some problems lurking around the corner. In case of mentioned link protocol, there is a special case for RTL languages, which for instance gets `https://\u200E` instead of `https://`. So if we treat values literally it would not work in case of RTL langs.

### Developer Tools Plugin

As you might recall we have a [Developer Tools plugin](https://ckeditor.com/cke4/addon/devtools) - it's an old plugin, but it could be helpful for people to quickly identify how a particular dialog field could be addressed.

You can [check it in our old samples](http://cdn.ckeditor.com/4.10.0/full-all/samples/old/devtools/devtools.html).

jacekbogdanski:

New feature

### This PR contains

- [x] Unit tests
- [x] Manual tests

## What changes did you make?

Overwrite dialog definition with configured `config.dialog_defaultValues` on dialog initialization.

Closes #2277

### DIFF
--- a/plugins/dialog/plugin.js
+++ b/plugins/dialog/plugin.js
@@ -1610,7 +1610,6 @@ CKEDITOR.DIALOG_STATE_BUSY = 2;
 		 *
 		 * @since 4.11.0
 		 * @private
-		 *
 		 * @param {String} tabId
 		 * @param {CKEDITOR.ui.dialog.uiElement} element
 		 */
@@ -1636,10 +1635,8 @@ CKEDITOR.DIALOG_STATE_BUSY = 2;
 		 *
 		 * @since 4.11.0
 		 * @private
-		 *
 		 * @param {String} tabId
 		 * @param {String} elementName
-		 *
 		 * @returns {String} [defaultValue]
 		 */
 		_getFieldConfigDefaultValue: function( tabId, elementName ) {
--- a/plugins/dialog/plugin.js
+++ b/plugins/dialog/plugin.js
@@ -313,6 +313,36 @@ CKEDITOR.DIALOG_STATE_BUSY = 2;
 			dialog: this
 		}, editor ).definition;
 
+		var me = this,
+			contents = this.definition.contents;
+
+		// Overwrite definition with default values (#2277).
+		for ( var key in contents ) {
+			var contentObj = contents[ key ];
+			CKEDITOR.tools.array.forEach( contentObj.elements, function( element ) {
+				setDefaultValues( me, contentObj.id, element );
+			} );
+		}
+
+		function setDefaultValues( dialog, tabId, element ) {
+			var dialogId = dialog._.name;
+			if ( element.type === 'vbox' || element.type === 'hbox' ) {
+				return CKEDITOR.tools.array.forEach( element.children, function( child ) {
+					setDefaultValues( dialog, tabId, child );
+				} );
+			}
+
+			var defaultValue = getFieldConfig( dialog._.editor, dialogId, tabId, element.id );
+			if ( defaultValue ) {
+				element[ 'default' ] = defaultValue;
+			}
+		}
+
+		function getFieldConfig( editor, dialogId, tabId, fieldName ) {
+			var defaultValues = editor.config.dialog_defaultValues || CKEDITOR.config.dialog_defaultValues;
+			return defaultValues[ [ dialogId, tabId, fieldName ].join( '.' ) ];
+		}
+
 		// Cache tabs that should be removed.
 		if ( !( 'removeDialogTabs' in editor._ ) && editor.config.removeDialogTabs ) {
 			var removeContents = editor.config.removeDialogTabs.split( ';' );
@@ -366,8 +396,6 @@ CKEDITOR.DIALOG_STATE_BUSY = 2;
 					evt.data.hide = false;
 			} );
 		}
-
-		var me = this;
 
 		// Iterates over all items inside all content in the dialog, calling a
 		// function for each of them.
@@ -954,7 +982,6 @@ CKEDITOR.DIALOG_STATE_BUSY = 2;
 				this.fire( 'show', {} );
 
 				// Configure default values right after `show` event so they won't be modified by plugin (#2277).
-				setConfigurationDefaultValues( this );
 
 				this._.editor.fire( 'dialogShow', this );
 
@@ -967,28 +994,6 @@ CKEDITOR.DIALOG_STATE_BUSY = 2;
 				} );
 
 			}, 100, this );
-
-			function setConfigurationDefaultValues( dialog ) {
-				var defaultValues = CKEDITOR.config.dialog_defaultValues;
-
-				if ( !defaultValues ) {
-					return;
-				}
-
-				for ( var chain in defaultValues ) {
-					var props = chain.split( '.' ),
-						name = props[ 0 ],
-						page = props[ 1 ],
-						field = props[ 2 ];
-
-					if ( name === dialog._.name ) {
-						var contentElement = dialog.getContentElement( page, field );
-						if ( contentElement ) {
-							contentElement.setValue( defaultValues[ chain ] );
-						}
-					}
-				}
-			}
 		},
 
 		/**
--- a/plugins/dialog/plugin.js
+++ b/plugins/dialog/plugin.js
@@ -715,7 +715,7 @@ CKEDITOR.DIALOG_STATE_BUSY = 2;
 	}
 
 	function getFieldConfig( editor, dialogId, tabId, fieldName ) {
-		var defaultValues = editor.config.dialog_defaultValues || CKEDITOR.config.dialog_defaultValues;
+		var defaultValues = editor.config.dialog_defaultValues;
 
 		if ( !defaultValues ) {
 			return null;
--- a/plugins/dialog/plugin.js
+++ b/plugins/dialog/plugin.js
@@ -313,7 +313,7 @@ CKEDITOR.DIALOG_STATE_BUSY = 2;
 			dialog: this
 		}, editor ).definition;
 
-		// Overwrite definition with default config values (#2277).
+		// (#2277)
 		setConfigDefaultValues( this );
 
 		// Cache tabs that should be removed.
--- a/plugins/dialog/plugin.js
+++ b/plugins/dialog/plugin.js
@@ -3634,7 +3634,7 @@ CKEDITOR.plugins.add( 'dialog', {
  * ```
  *
  * @since 4.11.0
- * @cfg {Object} [dialog_defaultValues]
+ * @cfg {Object.<String, String>} [dialog_defaultValues]
  * @member CKEDITOR.config
  */
 
--- a/plugins/dialog/plugin.js
+++ b/plugins/dialog/plugin.js
@@ -3634,7 +3634,7 @@ CKEDITOR.plugins.add( 'dialog', {
  * ```
  *
  * @since 4.11.0
- * @cfg {String} [dialog_defaultValues]
+ * @cfg {Object} [dialog_defaultValues]
  * @member CKEDITOR.config
  */
 
--- a/plugins/dialog/plugin.js
+++ b/plugins/dialog/plugin.js
@@ -313,40 +313,8 @@ CKEDITOR.DIALOG_STATE_BUSY = 2;
 			dialog: this
 		}, editor ).definition;
 
-		var me = this,
-			contents = this.definition.contents;
-
-		// Overwrite definition with default values (#2277).
-		for ( var key in contents ) {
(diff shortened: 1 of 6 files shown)


plugins/dialog/plugin.js
```diff
@@ -300,6 +300,41 @@ CKEDITOR.DIALOG_STATE_BUSY = 2;
 			definition: definition
 		}, editor ).definition;
 
+		var me = this,
+			contents = this.definition.contents;
+
+		// Overwrite definition with default values (#2277).
+		for ( var key in contents ) {
```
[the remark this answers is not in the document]
mlewand: That is important especially that we're going to improve this feature in followup requests.

plugins/dialog/plugin.js
```diff
@@ -300,6 +300,41 @@ CKEDITOR.DIALOG_STATE_BUSY = 2;
...
+			}
+
+			var defaultValue = getFieldConfig( dialog._.editor, dialogId, tabId, element.id );
+			if ( defaultValue ) {
+				element[ 'default' ] = defaultValue;
+			}
+		}
+
+		function getFieldConfig( editor, dialogId, tabId, fieldName ) {
+			var defaultValues = editor.config.dialog_defaultValues || CKEDITOR.config.dialog_defaultValues;
```
mlewand: The global `CKEDITOR.config` should be considered here. Only per-instance config applies here.

Global config should only be considered when creating `config` for a editor instance.

Imagine that someone _don't want any_ custom default values to be applied for a particular instance. While the default `CKEDITOR.config.dialog_defaultValues` contained something. I'd do it that way:

```javascript
CKEDITOR.config.dialog_defaultValues = {
	// values here
};
// (…)
CKEDITOR.replace( 'my-editor', {
	dialog_defaultValues: false
} );
```

The above code would be handled incorrectly with this approach.

plugins/dialog/plugin.js
```diff
@@ -3440,6 +3474,31 @@ CKEDITOR.plugins.add( 'dialog', {
...
+ *
+ * ```javascript
+ *  config.dialog_defaultValues = {
+ *    'link.info.protocol': 'https://',
+ *    'table.info.txtWidth': '100%'
+ *  }
+ * ```
+ *
+ * @since 4.11.0
+ * @cfg {String} [dialog_defaultValues]
```
mlewand: By any means is that property a string.

tests/plugins/dialog/manual/defaultvalues.md
```diff
@@ -0,0 +1,19 @@
...
+@bender-ui: collapsed
+@bender-ckeditor-plugins: wysiwygarea, toolbar, link
+
+1. Click `link` button.
+1. Wait until dialog show up.
+1. Compare `Protocol` field with expected results.
+1. Click `Advanced` tab.
+1. Compare `Id` and `Name` fields with expected results.
+
+## Expected
```
mlewand: Please, restructure the test in a way that each expected is below relevant step. It makes testing easier.

tests/plugins/dialog/plugin.js
```diff
@@ -401,6 +432,22 @@
...
+		'test default values': function() {
+			var tc = this;
+			this.editor.openDialog( 'testDialog1', function( dialog ) {
+				tc.resume( function() {
+					wait( function() {
+						assert.areEqual( 'text1', dialog.getContentElement( 'info', 'text1' ).getValue(), 'text1 field has invalid value' );
+						assert.areEqual( 'text2', dialog.getContentElement( 'info', 'text2' ).getValue(), 'text2 field has invalid value' );
+
+						dialog.getButton( 'cancel' ).click();
+					}, 100 );
```
mlewand: Do we really need to do a nasty timeout here? 😢 

There's `editorBot.dialog` which would simplify that. It actually uses proper(ish) event to resume, though still it uses 0 ms timeout - but that's much better than that.

plugins/dialog/plugin.js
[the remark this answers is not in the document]
jacekbogdanski: The function constructor is indeed too complex :cry: I will refactor the code :+1:

plugins/dialog/plugin.js
```diff
@@ -300,6 +300,41 @@ CKEDITOR.DIALOG_STATE_BUSY = 2;
...
 		}, editor ).definition;
 
+		var me = this,
+			contents = this.definition.contents;
+
+		// Overwrite definition with default values (#2277).
+		for ( var key in contents ) {
+			var contentObj = contents[ key ];
+			CKEDITOR.tools.array.forEach( contentObj.elements, function( element ) {
+				setDefaultValues( me, contentObj.id, element );
```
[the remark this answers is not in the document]
jacekbogdanski: Nice, didn't catch up this argument earlier.

plugins/dialog/plugin.js
```diff
@@ -300,6 +300,41 @@ CKEDITOR.DIALOG_STATE_BUSY = 2;
 			definition: definition
 		}, editor ).definition;
 
+		var me = this,
+			contents = this.definition.contents;
+
+		// Overwrite definition with default values (#2277).
```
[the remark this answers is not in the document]
jacekbogdanski: Actually, I dropped this comment after refactoring thus it doesn't provide any value.

plugins/dialog/plugin.js
```diff
@@ -3440,6 +3478,31 @@ CKEDITOR.plugins.add( 'dialog', {
...
+ *
+ * ```javascript
+ *  config.dialog_defaultValues = {
+ *    'link.info.protocol': 'https://',
+ *    'table.info.txtWidth': '100%'
+ *  }
+ * ```
+ *
+ * @since 4.11.0
+ * @cfg {Object} [dialog_defaultValues]
```
mlewand: We agreed some time ago on using more detailed Object type notation. Here You'd want to say what type keys and values are.

At this point we support only `String:String` (custom types gonna be supported as a value in followup PRs). So that's like `{Object.<String, String>}`.

plugins/dialog/plugin.js
```diff
@@ -300,6 +300,9 @@ CKEDITOR.DIALOG_STATE_BUSY = 2;
 			definition: definition
 		}, editor ).definition;
 
+		// (#2277)
+		setConfigDefaultValues( this );
```
mlewand: This is a subjective thing, but I'd rather like to see this method be a private one, defined in prototype of CKEditor.dialog type. Then the function would contain helper functions just like you did here. The difference would be there would be much closer to the place where used, so that would make the code more concise. Now we end up with tens of helper functions one after the other, which might not be related to each other.

tests/plugins/dialog/plugin.js
```diff
@@ -26,9 +50,19 @@
...
+			dialog_defaultValues: {
+				'testDialog1.info.text1': 'text1',
+				'testDialog1.info.text2': 'text2'
+			}
+		}
+	};
 
 	bender.test( {
+		setUp: function() {
+			this.editor.addCommand( 'testDialog1', new CKEDITOR.dialogCommand( 'testDialog1' ) );
```
mlewand: Technically this is invalid construction. You're not supposed to add the same command multiple times, and this is what is happening here. `setUp` runs for each TC.

mlewand: Instead you could simply do that at the end of `instanceLoaded` event.

tests/plugins/dialog/plugin.js
```diff
@@ -401,6 +435,16 @@
 
 			editor.execCommand( 'testDialog6' );
 			wait();
+		},
+
+		// (#2277)
+		'test default values': function() {
+			this.editorBot.dialog( 'testDialog1', function( dialog ) {
+				assert.areEqual( 'text1', dialog.getContentElement( 'info', 'text1' ).getValue(), 'text1 field has invalid value' );
```
mlewand: Asser messages should end with a dot.

tests/plugins/dialog/plugin.js
```diff
@@ -401,6 +435,16 @@
...
+				assert.areEqual( 'text2', dialog.getContentElement( 'info', 'text2' ).getValue(), 'text2 field has invalid value' );
+
+				dialog.getButton( 'cancel' ).click();
```
mlewand: There are several issues for this way of closing a dialog:

1. It is called only if no assertions fails (meaning that if something fails, rest of the cases is messed).
1. It requires copy and pasting this code to each assertion.

You could do it better, by putting this kind of cleanup code into `tearDown` funciton of test suite. Call `CKEDITOR.dialog.getCurrent()` and if there is any then close it.

I have added one code where I intentionally copied the code from here, so you can do the extraction explained above.

plugins/dialog/plugin.js
mlewand: Make sure to do proper api docs for this method.

mlewand:
I have pushed unit test for missing feature in 9b821726b4.

jacekbogdanski:
I made sure to check if the feature works with checkbox - added additional manual and unit tests. I was thinking about refactoring whole `tests/plugins/dialog/plugin` test file to use `tearDown` for dialog cleanup, although I would have to refactor logic responsible for opening dialogs. Decided to not spend time on additional refactoring. However, maybe it's worth to touch it? WDYT?

jacekbogdanski:
I fixed smaller styling issues, although I will wait for the main issues to our meeting.

f1ames:
Tests doesn't fail on local env so might be related to https://github.com/ckeditor/ckeditor4/issues/3410.

f1ames:
Rebased again onto latest `major`, so we have Travis fixes included.

plugins/dialog/plugin.js
```diff
@@ -1580,6 +1587,71 @@ CKEDITOR.DIALOG_STATE_BUSY = 2;
 			}
 
 			return CKEDITOR.dialog.EDITING_MODE;
+		},
+
+		/*
```
f1ames:
```suggestion
		/**
```

jacekbogdanski:
Unfortunately, it looks like the issue is more complicated than we thought. Overall, the dialog plugin works as a singleton which is reset into default values every time when it's opened. It's a part of dialog lifecycle and this logic should probably not change, otherwise, we may introduce breaking changes.

However, with the current implementation, a dialog will reset into default values if the value is unset e.g. by deleting input content as it changes `dialogDefinition[ 'default' ]` property. You can reproduce the issue using `/tests/plugins/dialog/manual/defaultvalues` or by following https://github.com/ckeditor/ckeditor4/pull/2390#pullrequestreview-155836507 steps.

The most obvious option would be resetting dialog only when it's in `CKEDITOR.dialog.CREATION_MODE`, however, in that case, some plugins may relay on resetting logic even in editing mode which loops back to
>It's a part of dialog lifecycle and this logic should probably not change, otherwise, we may introduce breaking changes.

We may need to introduce some additional logic which would make sure that some fields shouldn't be reset into defaults as they have been explicitly changed.

PR #2396, opened 2018-09-10
Fixed: Scroll jumps when closing dialogs

### LINKED ISSUE #748: Scroll jump when closing dialogs (Chrome only)

 type: Bug 

https://dev.ckeditor.com/ticket/16996

1. Go to ​http://sdk.ckeditor.com/samples/inline.html
2. Go to one of the inline editors, then scroll until the top of the editor is hidden
3. Open an image dialog and click to "Cancel"
4. The scroll jumps when it gets the focus back

### Expected result

Using Edge on the same computers doesn't make the scroll jump.

### Actual result

a scroll jump to the top of the editable div

## Other details

* Browser: Chrome 60.0.3112.90
* OS: Win 10
* CKEditor version: 4.7.1

### LINKED ISSUE #895: "test scroll editable and focus" in tests/core/editable/misc fails in Chrome 61+

### LINKED ISSUE #1429: Unfocused editor scrolls to top after opening dialog

Bug

https://dev.ckeditor.com/ticket/14894

1. Go to https://sdk.ckeditor.com/samples/fullpreset.html.
2. Scroll editor without focusing.
3. Open dialog form toolbar (e.g. link dialog).

### Expected result

Dialog is opened, editor scroll position does not change.

### Actual result

Dialog is opened, editor scrolls to top.

## Other details

This issue was fixed in `4.7.0` (see https://dev.ckeditor.com/ticket/14894), but now even on 4.7.0 (http://cdn.ckeditor.com/4.7.0/full-all/samples/old/image2/image2.html) it does not work, so I assume it may be caused by some browser behaviour changes.

* Browser: Chrome
* OS: All
* CKEditor version: 4.7.0
* Installed CKEditor plugins: Tested on `full preset`

### LINKED ISSUE #2420: Add an option to editable.focus to prevent default scroll
## Type of report

Feature request

[Living Standard](https://html.spec.whatwg.org/multipage/interaction.html#dom-focus) specification describes an option to prevent scroll on focus:
```
element . focus([ { preventScroll: true } ])
```

Although after testing it works only in Chrome and Safari, so we will need a fallback for other browsers.

Adding such option would enable us to fix #748.

engineering-this:

Bug fix + new feature

### This PR contains

- [X] Unit tests
- [X] Manual tests

## What changes did you make?

New feature part:

- Both `editor.focus` and `editable.focus` have now optional param object with property `preventScroll`. If it's set to true focusing won't affect scroll.
- Using the native focus with that flag for Chrome and Safari.
- Other browsers uses simple fallback: storing scroll position of any scrollable element and restoring it after calling native focus.
- Edge: detecting of scrollable element is not possible until [14721015](https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/14721015/) is fixed. 

Bugfix part:

- When the dialog is opened (by calling dialog command) it focuses the editor with `{ preventScroll: true }`.
- When the `dialog.hide` is called it focuses the editor with same flag.

And about mobiles: they are ignored in manuals and I don't think we can handle this easily within this ticket.

Note: first case wasn't reported, but I noticed that opening dialog on blurred editor will also trigger focus with scroll, so added fix for that too.

Closes #748 
Closes #895
Closes #1429
Closes #2420


jacekbogdanski:
What about an additional flag like `editable.focus( scroll = true )`? I think it would be nice to have both options available.

engineering-this:
Thanks @jacekbogdanski for your input. I guess this is best option.

engineering-this:
Edited PR message to reflect reworked solution.

tests/plugins/dialog/manual/pagescroll.html
```diff
@@ -0,0 +1,8 @@
+<div style="min-height: 100vh"></div>
+<textarea id="editor"></textarea>
+<script>
+	if ( CKEDITOR.env.ie || bender.tools.env.mobile ) {
```
Comandeer: Is there any other reason to ignore manual tests in IEs than lack of support for `vh` unit? If no, then this issue could be fixed with using large `px` value for `height`.

core/editable.js
```diff
@@ -81,29 +96,72 @@
...
+										var scrollable = element,
+											scrollTop = element.$.scrollTop;
+
+										scrollables.push( {
+											element: scrollable,
+											scrollTop: scrollTop
+										} );
+
+										// IE and Edge scroll asynchronously.
+										if ( CKEDITOR.env.ie ) {
```
engineering-this: I could drop IIFE by moving this part to separe forEach loop. But I'm not sure if it's better to loop two times, or just use same loop, but with IIFE for function scoping.

engineering-this:
Fix now works with IE and Edge and they aren't ignored in tests any more.

It looks like `setActive` which was called on IE instead of `focus` was scrolling page even if our comments said it didn't.

tests/plugins/dialog/manual/pagescroll.html
Comandeer: Still ignored…

core/editable.js
```diff
@@ -69,7 +84,7 @@
 					// Restore focus on element which we cached (on selectionCheck) as previously active.
 					active = this.editor._.previousActive || this.getDocument().getActive();
 					if ( this.contains( active ) ) {
-						active.focus();
+						active.focus( focusOptions );
```
Comandeer: This invocation does not work as intended. `active` is an instance of `CKEDITOR.dom.element`, not an editable. In such case, the signature of the `focus` method is different. In fact passing this parameter causes `focus` to be called asynchronously, as is treated as `defer` attribute. Please see: https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_dom_element.html#method-focus

That creates the tricky situation when the new fix is not working with nested editables (e.g. captions in `image2` or `easyimage` plugins), as viewport is always scrolled to these elements when inline editable is used (it's not present in classic, `iframe` based editor, as in such situation the `iframe` itself is a viewport, so only the content of `iframe` is scrolled, not the whole page).

I don't see any really sensible solution to that. We can overload `CKEDITOR.dom.element#focus` method to treat the fist parameter as `defer` one if value other than object is passed and as `focusOptions` if object is passed (in the second case we could make `defer` one of options, as probably there will be cases when both of the behaviours will be needed). It'd result in moving most of the logic from `CKEDITOR.editable` to `CKEDITOR.dom.element`. After that changes `CKEDITOR.editable#focus` will be using `CKEDITOR.dom.element#focus`, only adding some new behaviour.

However this change is not trivial anymore and requires much more tests. WDYT @engineering-this, @mlewand?

core/editable.js
Comandeer: I'm for keeping IIFE.

engineering-this:
I moved logic to `CKEDITOR.dom.element.prototype.focus`. So far it looks to be working fine. I kept `focusOption` to overload element method. So now we have two possible options `preventScroll` and `defer`.

core/editor.js
```diff
@@ -1305,9 +1305,25 @@
...
+		 *
+		 * Example:
+		 * ```javascript
+		 * editor.focus( { preventScroll: true } ); // Focuses but prevents scrolling.
+		 * editor.focus( { preventScroll: false } ); // Focuses without preventing scroll.
+		 * editor.focus( { defer: false } ); // Focuses after 100ms.
+		 * editor.focus(); // Focuses without preventing scroll.
+		 * ```
+		 *
+		 * @param {Object} [focusOptions] Focus options since **4.11.0**.
```
Comandeer: We describe `focusOptions` in three different places, everywhere the same way.

It seems that we need… [DU-DU-DUUUUUUM](https://www.youtube.com/watch?v=UgwSSAzxcs8) dedicated type for it.

tests/core/editable/manual/focus.html
```diff
@@ -0,0 +1,38 @@
+<div class="buttons">
+	<button data-editor-name="inline" style="font-size:20px">Focus inline</button>
+	<button data-editor-name="classic" style="font-size:20px">Focus classic</button>
+</div>
+
+<div style="height:8000px"></div>
+
+<div id="inline" contenteditable="true">
+	<figure class="easyimage">
```
Comandeer: Using widget as the only content inside the editor makes testing if editor is focused more difficult. The most obvious way (start typing) is not working anymore.

core/editable.js
```diff
@@ -81,34 +99,13 @@
...
-						} else {
-							this.$.focus();
-						}
-					}
-				} catch ( e ) {
-					// IE throws unspecified error when focusing editable after closing dialog opened on nested editable.
-					if ( !CKEDITOR.env.ie )
-						throw e;
-				}
+				CKEDITOR.dom.element.prototype.focus.call( this, focusOptions );
```
Comandeer: IMO it's better (_more elegant_) to call this class base class method:

```javascript
CKEDITOR.editable.baseProto.focus.call( this, focusOptions );
```

core/dom/element.js
```diff
@@ -419,26 +419,135 @@ CKEDITOR.dom.element.clearMarkers = function( database, element, removeFromDatab
...
 		 * execution by 100 ms.
 		 */
 		focus: ( function() {
-			function exec() {
+			function exec( focusOptions ) {
 				// IE throws error if the element is not visible.
 				try {
-					this.$.focus();
-				} catch ( e ) {}
+					var isScrollLockFallbackNeeded = !CKEDITOR.env.webkit;
```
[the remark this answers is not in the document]
engineering-this: As we have talked, I'm implementing detection for `focusOptions` in test, so it doesn't affect editor.

core/editable.js
```diff
@@ -59,8 +59,26 @@
...
+			 * ```javascript
+			 * editable.focus( { preventScroll: true } ); // Focuses but prevents scrolling.
+			 * editable.focus( { preventScroll: false } ); // Focuses without preventing scroll.
+			 * editable.focus( { defer: false } ); // Focuses after 100ms.
+			 * editable.focus(); // Focuses without preventing scroll.
+			 * ```
+			 *
+			 * @param {Object} [focusOptions] Focus options since **4.11.0**.
+			 * @param {Boolean} [focusOptions.preventScroll] Whenever after focus scroll should be prevented.
+			 * @param {Boolean} [focusOptions.defer] Whether to asynchronously defer the
```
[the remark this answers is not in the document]
engineering-this: Actually I don't think we need it, but as this method calls element.focus with given focus options we would have to explicitly `delete focusOptions.defer`. Otherwise it would work as an undocumented feature. I'm for keeping it. OTOH as I'm moving `focusOptions` docs to dedicated type to keep it consistent - each focus method should allow same `focusOptios` type holding same properties.

core/dom/element.js
```diff
@@ -419,26 +419,135 @@ CKEDITOR.dom.element.clearMarkers = function( database, element, removeFromDatab
...
+											} );
+
+											// Clean listener to not break user scroll.
+											setTimeout( function() {
+												listener.removeListener();
+											}, 50 );
+										}
+									} )();
+								}
+								if ( element.getParent() ) {
```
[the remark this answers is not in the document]
engineering-this: Just to clarify:

>That's not true for CKEditor as we want to keep scrolling position inside the editable without preventing the scrolling of the viewport (see https://dev.ckeditor.com/ticket/14825 ).

I assume you have meant 'we want to keep _scroll_ position...'. In other words: always prevent scrolling inside editor.

core/editable.js
```diff
@@ -81,34 +99,13 @@
...
-					}
-				} catch ( e ) {
-					// IE throws unspecified error when focusing editable after closing dialog opened on nested editable.
-					if ( !CKEDITOR.env.ie )
-						throw e;
-				}
+				CKEDITOR.dom.element.prototype.focus.call( this, focusOptions );
 
-				// Remedy if Safari doens't applies focus properly. (https://dev.ckeditor.com/ticket/279)
+				// Remedy if Safari doesn't applies focus properly. (https://dev.ckeditor.com/ticket/279)
```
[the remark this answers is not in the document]
engineering-this: It comes from https://github.com/ckeditor/ckeditor-dev/commit/0f19031124d5e496471e93aca92b899fa89f9dc3 there were no tests included, but nothing looks broken after removing it, and editor is focused after closing floatpanel.

core/dom/element.js
```diff
@@ -419,28 +419,117 @@ CKEDITOR.dom.element.clearMarkers = function( database, element, removeFromDatab
...
+					if ( this instanceof CKEDITOR.editable ) {
+						isInline = this.isInline();
+					} else {
+						var editable = this.getAscendant( '.cke_editable' );
+						isInline = editable ? editable.isInline() : null;
+					}
+
+					if ( isScrollLockFallbackNeeded && focusOptions && focusOptions.preventScroll ) {
+						scrollables = storeScrollPosition( element );
+					} else if ( !isInline && CKEDITOR.env.chrome && ( !focusOptions || !focusOptions.preventScroll ) ) {
```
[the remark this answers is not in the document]
engineering-this: To fix this Chrome bug we need to restore scroll position on editable parent, so general approach to restore  scroll position only in focused element wouldn't work in that case.

engineering-this:
- Fixed case when editor scrolled in IE8 when focused and in Chrome for inline editors.
- Simplified manual test.
- Refactored code a bit as.

I didn't rebase it yet, I will do so, when `major` branch is ready for commits.

Comandeer:
Rebased onto latest `major`.

core/dom/element.js
```diff
@@ -419,28 +419,127 @@ CKEDITOR.dom.element.clearMarkers = function( database, element, removeFromDatab
...
+								element: element,
+								scrollTop: element.$.scrollTop
+							} );
+						}
+
+						if ( element === lastElement ) {
+							break;
+						}
+
+						if ( !element.getParent() ) {
```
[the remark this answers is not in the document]
engineering-this:
```
element = element.getParent();

if ( !element ) {
	element = element.getWindow().getFrame();
}
```
This wouldn't work, as you are overwritting element with null and then trying to access it's properties.

The second one is perfect tho. I was so afraid of this typeError I didn't realise the simplest solution 🤦‍♂️

core/dom/element.js
```diff
@@ -419,28 +419,127 @@ CKEDITOR.dom.element.clearMarkers = function( database, element, removeFromDatab
...
+
+				function restoreScrollPosition( scrollables ) {
+					forEach( scrollables, function( item ) {
+						item.element.$.scrollTop = item.scrollTop;
+					} );
+				}
+
+				function restoreScrollAndCleanScrollables( scrollables ) {
+					restoreScrollPosition( scrollables );
+					scrollables = [];
```
Comandeer: This line does not clean `scrollables` variable from outer scope, it's just change object referenced by _local_ `scrollables` variable. To really clean `scrollables`, `scrollables.length = 0` could be used.

engineering-this:
`tests/core/dom/element/focusoptions`

I fixed this test on Edge, although there is still small issue. When you refresh page with test it fails, but for some reason browser doesn't focus element when refreshing. I don't think we should worry, as test passes when run from dashboard, and when run solo. It's isn't supposed to detect any regressions.

engineering-this:
>I wonder about fixing the issue with flickering. Maybe temporal overflow: hidden for scrollable containers would do the trick?

@Comandeer This won't work. `overflow: hidden` hides the scrollbars and content, but doesn't prevent scrolling. Browser/JS can scroll content anyway.

See example:
https://codepen.io/anon/pen/BvmYmV

engineering-this:
It is possible to focus element without scroll and even scroll event happening, but this requires to change target elements positioning to `fixed` before focus, and restore positioning after focus.

However in Ege/IE scroll is asynchronous so restoring styles should be wrapped in `setTimeout`, and this leads to another visual flickering, now instead of scroll we have flying elements. To prevent that we should get elements actual position and size, then place next to it some dummy element that will take same space, and add positioning to our focused element so it is rendered in same place when it has fixed position.

But still there is scroll to an iframe, so same logic should be applied on it. 

One would say: why do we care about positioning of elements that out of view?
Actually they can be visible partially and page still would be scrolled on focus. Also some visible container might collapse when it's focusable child is changed to absolute positioning.

I can't think of other options, so I guess for now we will have to live with Edge/IE flickering.

hissy:
Is this issue still open?

f1ames:
Hello @hissy,
yes, this PR is still work in progress. However, we have shift our focus to similar dialog related fix which is https://github.com/ckeditor/ckeditor-dev/pull/2413 - it may solves some of issues mentioned here, but we have to check it. You may watch https://github.com/ckeditor/ckeditor-dev/pull/2413 to stay updated.

hissy:
Thank you, I'd like to check that issue too.

PR #2399, opened 2018-09-11
Added missing docs for CKEDITOR.htmlParser.filterRulesDefinition

### LINKED ISSUE #2387: filterRulesDefinition docs are empty
## Type of report

Task

[`filterRulesDefinition`](https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_htmlParser_filterRulesDefinition.html) docs are empty. We need to document its members.

engineering-this:

Task

### This PR contains

- [ ] Unit tests
- [ ] Manual tests

## What changes did you make?

Added missing docs for CKEDITOR.htmlParser.filterRulesDefinition and its properties.

Closes #2387


core/htmlparser/filter.js
```diff
@@ -402,6 +402,36 @@
 } )();
 
 /**
+ * This is an abstract class representing filter rules passed to {@link CKEDITOR.htmlParser.filter#constructor} as an argument. See {@link CKEDITOR.htmlParser.filter} for examples.
```
jacekbogdanski: I'm not sure if limiting this class to filter constructor is a good idea. This class may be used in different places - moreover, it's already used as an argument for [addRules](https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_htmlParser_filter.html#method-addRules) function. Also in the future, we may want to use it in other places which with the proposed documentation shape will require another refactoring.

Also, mentioning that this class is abstract and used in some other places is redundant. This information is already viewable from functions arguments and returned types.

Instead of linking `htmlParser.filter` for examples, I think that it would be nice to attach some example code of definition and refactor duplicated code from filter listing.

jacekbogdanski: I would extend this description with the information like:

What this class represents?
Why would I have to use it?
How I have to use it?

not implementation details.

core/htmlparser/filter.js
```diff
@@ -402,6 +402,140 @@
...
+ * Examples:
+ *
+ * ```javascript
+ * elementNames: [
+ * 		[ /^div$/, 'p' ], // Converts 'div' into 'p'.
+ * 		[ /^cke:?/, '' ] // Removes 'cke:' prefixes.
+ * ]
+ * ```
+ *
+ * Incorrect:
```
jacekbogdanski: Examples already describe correct rule regex, not sure if we need this section. It seems obvious to me that if you are matching string by regex, too small specificity will give you invalid results. WDYT?

core/htmlparser/filter.js
```diff
@@ -402,6 +402,140 @@
...
+ * 		[ 'p', 'div' ] // Converts every letter 'p' inside element name into 'div'
+ * 		// 'p' element will be converted into 'div', 'span' wil be converted into 'sdivan'.
+ * 		// To convert only 'p' elements use regex `/^p$/`.
+ * ]
+ * ```
+ *
+ */
+
+/**
+ * @property {String[][]/RegExp[][]} attributeNames An array of rules for data processing attribute names.
```
jacekbogdanski: The same as above.

core/htmlparser/filter.js
```diff
@@ -402,6 +402,140 @@
...
+ * 		[ /^data-custom$/, 'data-cke' ]
+ * 		// Converts attribute 'data-custom' into 'data-cke'.
+ * ]
+ * ```
+ *
+ */
+
+/**
+ * @property {Object.<String, Function>} elements An object containing pairs of keys representing
+ * element selectors and function for data processing selected elements. If function contains return statement element will be removed.
```
jacekbogdanski: I would split this long comment into two lines (after the dot).

Also, sencence `If function contains return statement element will be removed` is invalid. Every javascript function returns something (also without a return statement, which is `undefined`) so you should be more precise about return type. What if I write:
```javascript
function() {
  return undefined;
}
```
Will element still be removed?

jacekbogdanski: This `data processing` phrase is very mysterious to me. I really don't understand what it does and what my function should do. Something like `transformation function` is more understable for me than it, although it may be subjective. E.g.

`An object containing pairs of keys representing element selectors and transformation function.`

Also, is the element really selected inside the editor? What if I apply filter into document fragment?

core/htmlparser/filter.js
```diff
@@ -402,6 +402,140 @@
...
+ *
+ * ```javascript
+ * attributes: {
+ * 		'class': function( value, element ) {
+ * 			if ( element.name === 'div' ) {
+ * 				return value + ' cke_div' // Adds class 'cke_div' to every div element.
+ * 			}
+ * 		},
+ * 		id: function() {
+ * 			return false; // Removes every elements 'id' attribute.
```
jacekbogdanski: I would rephrase it into `Removes processed element id attribute`. `Every` word is so wide that it may introduce some misconceptions. It applies to the whole proposed documentation.

core/htmlparser/filter.js
```diff
@@ -402,6 +402,140 @@
...
+ */
+
+/**
+ * @property {Function} root Function for data processing root element.
+ *
+ * Examples:
+ *
+ * ```javascript
+ * root: function( element ) {
+ * 		element.children.push( someElement ) // Appends child to root element.
```
jacekbogdanski: Missing semicolon.

core/htmlparser/filter.js
```diff
@@ -402,6 +402,140 @@
...
+ * to be applied for transforming and filtering content upon data processing.
+ *
+ * Filter rules definition can be also added with {@link CKEDITOR.htmlParser.filter.addrules}.
+ *
  * @class CKEDITOR.htmlParser.filterRulesDefinition
  * @abstract
  */
+
+/**
+ * @property {String[][]/RegExp[][]} elementNames An array of rules for data processing element names.
```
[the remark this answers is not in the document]
jacekbogdanski: I'm sure that a new type would require some discussion on our team.

core/htmlparser/filter.js
```diff
@@ -402,6 +402,144 @@
...
+ * and functions used upon element filtering and transformation.
+ *
+ * A selector can be either element name or one of following: '^', '$'.
+ *
+ * '^' and '$' is to be applied on every filtered element. The first is applied before element specific filter,
+ * and second is applied after element specific filter.
+ *
+ * Function can contain return statement:
+ * - If `false` is returned the element is removed.
+ * - If an other element is returned it overwrites the original element.
```
jacekbogdanski: Actually, returning element always overwrites the original element, although new type element won't be processed anymore: https://github.com/ckeditor/ckeditor-dev/blob/master/core/htmlparser/filter.js#L347
What do you think about:

> Additionaly function can contain return statement:                                                                                                                                                                                    
> - `false` will remove whole element from the node tree.                                                                                                                                                                               
> - {@link CKEDITOR.dom.element element} will overwrite the original element.

core/htmlparser/filter.js
```diff
@@ -402,6 +402,144 @@
...
+ *
+ * Definition object represents rules as a set of properties with callback functions
+ * to be applied for transforming and filtering content upon data processing.
+ *
  * @class CKEDITOR.htmlParser.filterRulesDefinition
  * @abstract
  */
+
+/**
+ * @property {Array<String,RegExp>[]} elementNames An array of rules for element names transformation.
```
jacekbogdanski: It seems like umberto/jsduck incorrectly parsers this construction :thinking:

core/htmlparser/filter.js
```diff
@@ -402,6 +402,164 @@
...
+ *
+ * Definition object represents rules as a set of properties with callback functions
+ * to be applied for transforming and filtering content upon data processing.
+ *
  * @class CKEDITOR.htmlParser.filterRulesDefinition
  * @abstract
  */
+
+/**
+ * @property {CKEDITOR.htmlParser.nameFilterRule[]} elementNames An array of rules for element names transformation.
```
Comandeer: TBH I do not like how this custom type looks like. I would think about replacing it with `Array.<RexExp,String>` syntax (note the dot that magically fixes jsduck issues; it would be also consistent with `Object.<whatever>` syntax that we use elsewhere).

engineering-this: Yes it was my mistake, to skip the dot, yet it is an array of arrays, which is not correctly parsed by jsduck.
`Array.<RegExp,String>[]` -> `Array.&ltRegExp,String>[]`

core/htmlparser/nameFilterRule.js
```diff
@@ -0,0 +1,29 @@
...
+ */
+
+/**
+ * Abstract class describing the definition of {@link CKEDITOR.htmlParser.filterRulesDefinition} `elementNames` and `attributesNames` filtering rules.
+ *
+ * ```javascript
+ *  var rule = [ /^div$/, 'p' ];
+ * ```
+ *
+ * @class CKEDITOR.htmlParser.nameFilterRule
```
Comandeer: I'm wondering about the name of the class. It's used for elements's transformations, so maybe `CKEDITOR.htmlParse.transformRule`?

engineering-this: Every member of `CKEDITOR.htmlParser.filterRulesDefinition` class can used for some kind of transformation. Although only two uses this abstract class. `transformRule` sounds a bit misleading to me.
What do you think about `nameTransformRule`? These rules alters element names or attributes, but they don't touch values/content so this name would be a bit more precise. WDYT @Comandeer?

engineering-this:
- Added links to methods that uses `filterRulesDefinition`.
- Renamed `nameFilterRule` to `nameTransformRule`.

Issue #2438, opened 2018-09-25
Pasting empty list and converting it to different list type throws error


jswiderski:
## Type of report

Bug

## Provide detailed reproduction steps (if any)

1. Paste the below code in Source mode of [full preset sample](https://sdk.ckeditor.com/samples/fullpreset.html).
```
<p><br></p><ul></ul><ul><li><a href="https://powerusers.microsoft.com/plugins/common/feature/oauth2sso/sso_login_redirect?referer=https%3A%2F%2Fpowerusers.microsoft.com%2Ft5%2FPowerApps-Community%2Fct-p%2FPowerApps1">Register</a></li><li><br></li><li>·</li><li><br></li><li><a href="https://powerusers.microsoft.com/plugins/common/feature/oauth2sso/sso_login_redirect?referer=https%3A%2F%2Fpowerusers.microsoft.com%2Ft5%2FPowerApps-Community%2Fct-p%2FPowerApps1">Sign In</a></li><li><br></li><li>·</li><li><br></li><li><a href="https://powerusers.microsoft.com/t5/help/faqpage">Help</a></li></ul><p><br></p>
```
**NOTE:** Starting from CKEditor 4.8 it is enough to paste just the empty list `<ul></ul>` to get the error.

2. Switch to wysiwyg and click inside the editor to focus is
3. Press Ctrl+A to select all content
4. Click on numbered or bulleted list button 

### Expected result

Hard to say because Empty lists is not something standard:
1. Empty `li` element gets created in empty list and it gets converted to different list type in necessary.
2. Empty list gets removed
3. Nothing or `ul` gets converted to empty `ol` if numbered list was pressed

### Actual result

JavaScript error gets thrown:
**Message:** Cannot read property 'listNode' of null
**URI:** /ckeditor-dev/plugins/list/plugin.js
**Line:**  334

## Other details

* Browser: Any
* OS: Any
* CKEditor version: 4.0+
* Installed CKEditor plugins: List

mlewand:
Looks clearly like a duplicate of #2411 to me.

msamsel:
I have some prototype of solution for the #2411 and it's fix throwing error in this case too. We could merge those 2 issues together and move any further discussion to that issue.

Approach which I used is just preventing of throwing error. This empty `ul` will remain in markup as user want to paste it.

PR #2440, opened 2018-09-26
Changing list type wont throw error for empty list elements

### LINKED ISSUE #2411: Pasting empty nested list and converting it to different list type throws error
## Type of report

Bug

1. Go to https://sdk.ckeditor.com/samples/fullpreset.html.
2. Clear Editor contents and switch to source mode.
3. Paste below HTML
```
<ul>	
	<li>
	<div>test
	<ul>
	</ul>
	</div>
	</li>
	<li>test</li>
</ul>
```
In Chrome it is enough to use 

```
<ul>	
	<li>
	<div>test
	<ul>
	</ul>
	</div>
	</li>
</ul>
```
4. Switch to wysiwyg mode
5. Press `Ctrl+A` na dclick on Numbered  List button

### Expected result

No error gets thrown.

Perhaps as an alternative solution such list should be filtered by ACF.

### Actual result

JavaScript error gets thrown: 
**Message:** Cannot read property 'parent' of undefined
**Line:** 319
**URI:** `/ckeditor-dev/plugins/list/plugin.js
`
## Other details

* Browser: Any
* OS: Any
* CKEditor version: 4.0+
* Installed CKEditor plugins: List

----
The Original scenario was:
1. Open https://codepen.io/pratikgolchha/pen/wxGbRd in Edge
2. Copy everything form https://docs.microsoft.com/zh-cn/powerapps/maker/canvas-apps/controls/control-gallery and paste in the CKEditor
3. Select all contents and apply bulleted list and then numbered list
4. Check the console

This document is huge so maybe there is something extra what is causing the same issue (there shouldn't be but it may).

msamsel:

Bug fix

### This PR contains

- [x] Unit tests
- [x] Manual tests

## What changes did you make?

* Fix throwing errors in console, when markup with empty `ul` list was changed to different list type.

Close #2411

Related also to #2438

### DIFF
--- a/core/tools.js
+++ b/core/tools.js
@@ -1570,6 +1570,19 @@
 		},
 
 		/**
+		 * Counts childes of given {@link CKEDITOR.dom.element} omitting {@link CKEDITOR.dom.comment}.
+		 *
+		 * @since 4.10.2
+		 * @param {CKEDITOR.dom.element} node Element whos children will be counted.
+		 * @returns {Number} Amount of node's children without counting comments.
+		 */
+		childCountWithoutComments: function( node ) {
+			return this.array.filter( node.getChildren().toArray(), function( node ) {
+				return node.type !== CKEDITOR.NODE_COMMENT;
+			} ).length;
+		},
+
+		/**
 		 * A set of functions for operations on styles.
 		 *
 		 * @property {CKEDITOR.tools.style}
--- a/core/tools.js
+++ b/core/tools.js
@@ -1572,7 +1572,7 @@
 		/**
 		 * Counts childes of given {@link CKEDITOR.dom.element} omitting {@link CKEDITOR.dom.comment}.
 		 *
-		 * @since 4.10.2
+		 * @since 4.11.2
 		 * @param {CKEDITOR.dom.element} node Element whos children will be counted.
 		 * @returns {Number} Amount of node's children without counting comments.
 		 */
--- a/core/tools.js
+++ b/core/tools.js
@@ -1570,19 +1570,6 @@
 		},
 
 		/**
-		 * Counts childes of given {@link CKEDITOR.dom.element} omitting {@link CKEDITOR.dom.comment}.
-		 *
-		 * @since 4.11.2
-		 * @param {CKEDITOR.dom.element} node Element whos children will be counted.
-		 * @returns {Number} Amount of node's children without counting comments.
-		 */
-		childCountWithoutComments: function( node ) {
-			return this.array.filter( node.getChildren().toArray(), function( node ) {
-				return node.type !== CKEDITOR.NODE_COMMENT;
-			} ).length;
-		},
-
-		/**
 		 * A set of functions for operations on styles.
 		 *
 		 * @property {CKEDITOR.tools.style}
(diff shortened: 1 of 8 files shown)


tests/plugins/list/blockconversion.js
```diff
@@ -0,0 +1,57 @@
...
+	},
+	inline: {
+		name: 'inline',
+		creator: 'inline'
+	}
+};
+
+var tests = {
+	'test conversion of embedded unordered list and paragraphs to numbered list': function( editor, bot ) { // Strange test result for: 0, 8
+		var testesLength = 9;
```
mlewand: Subsequent `var` declarations/initialization should be contained within a single `var` statement.

mlewand: This issue appears later on, please correct each occurence.

tests/plugins/list/blockconversion.html
```diff
@@ -0,0 +1,468 @@
+<!--
+
+Strange output markup for tests:
```
mlewand: Avoid information duplication. You have already provided this info in `blockconversion.js`. 

`blockconversion.js` is where the logic of tests are so it is the best place to put this information. Here it's not needed, as it is only a _resource_ file for the logic.

tests/plugins/list/blockconversion.js
```diff
@@ -0,0 +1,57 @@
...
+		}
+	},
+	inline: {
+		name: 'inline',
+		creator: 'inline'
+	}
+};
+
+var tests = {
+	'test conversion of embedded unordered list and paragraphs to numbered list': function( editor, bot ) { // Strange test result for: 0, 8
```
mlewand: `Strange test result for: 0, 8` comment will look much better in a line below. Here it exceeds 120 chars which, if convenient, we like to avoid.

core/tools.js
```diff
@@ -1569,6 +1569,19 @@
...
 		},
 
+		/**
+		 * Counts childes of given {@link CKEDITOR.dom.element} omitting {@link CKEDITOR.dom.comment}.
+		 *
+		 * @since 4.11.2
+		 * @param {CKEDITOR.dom.element} node Element whos children will be counted.
+		 * @returns {Number} Amount of node's children without counting comments.
+		 */
+		childCountWithoutComments: function( node ) {
```
mlewand: We can't add a method like that, it doesn't belong here.

1. Logically it belongs to `CKEDITOR.dom.element` type - this is where it should be added.
1. It's too specific to your particular use case. Having a method like that would be excuse for getChildCountWithoutNodes, getChildCountWithoutElements etc. Instead [`getChildCount` method](https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_dom_element.html#method-getChildCount) should be modified to accept `options` param, that allows to specify `options.exclude` or `options.include`. So you'd call it like `elem.getChildCount( { exclude: [ CKEDITOR.NODE_COMMENT ] } );`.
1. Name is not aligned with existing [`getChildCount` method](https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_dom_element.html#method-getChildCount).

Finally there's a minor naming issue, `node` can not have children so the arg should be called `element` just like you described in `@param`'s description.

I should have pointed this out in previous review.

mlewand:
I have fixed the issue that I raised during the review by simply inlining the function. Ideally I'd like CKEDITOR.dom.element methods to get `include` / `exclude` options, but that would mean that this fix would need to wait for the major release.

PR #2457, opened 2018-10-04
Add input as focusable element inside panels

msamsel:

New feature

### This PR contains

- [x] Unit tests
- [x] Manual tests

## What changes did you make?

- Feature required by #2062, to have possibility to focus on input element inside panels.

### DIFF
--- a/plugins/panel/plugin.js
+++ b/plugins/panel/plugin.js
@@ -90,6 +90,17 @@
 		 * to this button.
 		 */
 		render: function( editor, output ) {
+			var data = {
+				editorId: editor.id,
+				id: this.id,
+				langCode: editor.langCode,
+				dir: editor.lang.dir,
+				cls: this.className,
+				frame: '',
+				env: CKEDITOR.env.cssClass,
+				'z-index': editor.config.baseFloatZIndex + 1
+			};
+
 			this.getHolderElement = function() {
 				var holder = this._.holder;
 
@@ -150,17 +161,6 @@
 				}
 
 				return holder;
-			};
-
-			var data = {
-				editorId: editor.id,
-				id: this.id,
-				langCode: editor.langCode,
-				dir: editor.lang.dir,
-				cls: this.className,
-				frame: '',
-				env: CKEDITOR.env.cssClass,
-				'z-index': editor.config.baseFloatZIndex + 1
 			};
 
 			if ( this.isFramed ) {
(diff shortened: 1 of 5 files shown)


mlewand:
@msamsel Please add reference to the issue that is closed by this PR.

plugins/panel/plugin.js
```diff
@@ -381,7 +381,7 @@
 					// Move forward.
 					case 'next':
 						var index = this._.focusIndex,
-							links = this.element.getElementsByTag( 'a' ),
+							links = this._.getItems(),
```
mlewand: These are no longer links. We should refer to this in a more generic way, like "focusables".

plugins/panel/plugin.js
```diff
@@ -361,7 +361,7 @@
 			 * @returns {*|CKEDITOR.dom.nodeList}
```
mlewand: This format is wrong. It will always return `CKEDITOR.dom.nodeList`, so please use this opportunity to adjust the API docs.

plugins/panel/plugin.js
mlewand: This problem is repeated in couple other places.

tests/plugins/panel/manual/focusableinput.md
```diff
@@ -0,0 +1,14 @@
+@bender-tags: 4.11.0, feature, 2453
+@bender-ui: collapsed
+@bender-ckeditor-plugins: wysiwygarea,toolbar,panel
```
mlewand: Please add space between the comma separated values just like you did for tags.

tests/plugins/panel/manual/focusableinput.md
```diff
@@ -0,0 +1,14 @@
...
+1. Click into rectangle with black border
+2. Press <kbd>TAB</kbd>
+3. Check if you can focus input element with keyboard. Currently focused elements are highlighted with red background.
+
+
+### Expected:
+* You can acess to `input` element with <kbd>TAB</kbd> key.
+
+### Unexpected:
+* Input element is noto focusable with keyboard.
```
mlewand: Typo, not focusable.

plugins/panel/plugin.js
```diff
@@ -297,8 +297,8 @@
 			markItem: function( index ) {
 				if ( index == -1 )
 					return;
-				var links = this.element.getElementsByTag( 'a' );
-				var item = links.getItem( this._.focusIndex = index );
+				var focusables = this._.getItems();
+				var item = focusables.getItem( this._.focusIndex = index );
```
mlewand: You could join these two subsequent `var` statements, as this is our code style since couple of years.

tests/plugins/panel/panel.js
```diff
@@ -23,6 +38,19 @@
 			src = src.substring( 5, src.length - 1 );
 
 			assert.areSame( '', src, 'Frame source should be empty.' );
+		},
+
+		// #2453
```
mlewand: More commonly we put issue references in parenthesis.

Issue #2403, opened 2018-09-12
Styling inline editor nested inside table causes styles leaking


markussc:
## Type of report

Bug

## Provide detailed reproduction steps (if any)

1. Open http://jsfiddle.net/r9bfgwsk.
1. Type foo inside a table cell.
1. Select whole text ( with text selection, not table selection).
1. Mark selected text bold.

### Expected result

The selected text is bold.

### Actual result

The table cell gets a grey background. The css-class "cke_table-faked-selection" is added to the table's td-element

## Other details

* Browser: Firefox 62.0
* OS: Windows 10
* CKEditor version: 4.10.0
* Installed CKEditor plugins: sharedspace,lite,undo

jacekbogdanski:
Hello,

I tried to reproduce your issue without success. Could you check if you can still reproduce it with clean CKEditor installation (without external plugins)? You could use https://ckeditor.com/latest/samples/ to check if it works correctly.

markussc:
It's a bit weird. When I download https://github.com/ckeditor/ckeditor-releases/archive/4.10.1/4.10.1.zip , the file ckeditor.js contains at lines 1056 / 1057:
`if(p.active&&c&&p.first.getAscendant("table").equals(c.getAscendant("table"))){g=e(p.first,c);if(!p.dirty&&1===g.length&&!a(f.data.getTarget()))return d(b,"mouseup"===f.name);p.dirty=!0;p.last=c;l(b,g)}}function h(a){var b=(a=a.editor||a.sender.editor)&&a.getSelection(),c=b&&b.getRanges()||[],e;if(b&&(d(a),b.isInTable()&&b.isFake)){1===c.length&&c[0]._getTableElement()&&c[0]._getTableElement().is("table")&&(e=c[0]._getTableElement());
e=q(b,e);a.fire("lockSnapshot");for(b=0;b<e.length;b++)e[b].addClass("cke_table-faked-selection");`

The interesting part is the last statement, i.e. adding class "cke_table-faked-selection". This will be added to all ascendant table elements. In my case also the surrounding table. I can fix it if prepending a check like:
`if (!e[b].parent().hasClass("cke_editable") {continue;}`
I don't find this piece of code in the src/ directory anywhere. Where does it come from in the released zip ?

jacekbogdanski:
As I understand your issue is about creating table selection when selecting whole text and applying style. If so, this issue is not limited to forms - exists for any element inside table cell:

[...]

PR #2459, opened 2018-10-04
fixed tableselection if ckeditor is inside table

markussc:

Bug fix for #2403 

No tests included.

### This PR contains

- [ ] Unit tests
- [ ] Manual tests

## What changes did you make?

Added check to only add fake-selection class to ckeditor-related table elements.


markussc:
Switched branch. I will not be able to add tests for this soon. I'd appreciate if somebody else could add them instead.

Comandeer:
Thanks for the PR, however this issue was fixed in 4.11.2 via #2476. We used a different approach, based on adding appropriate conditional directly into selection core.

Issue #2453, opened 2018-10-03
Add support for input element in keydown handler of panel


msamsel:
## Type of report

Feature request

## Provide description of the new feature

Add possibility to move focus over input elements of panel if poses proper class.
https://github.com/ckeditor/ckeditor-dev/blob/549456c24d4fdb173996616eb7afd9edd2c3b83d/plugins/panel/plugin.js#L384
https://github.com/ckeditor/ckeditor-dev/blob/549456c24d4fdb173996616eb7afd9edd2c3b83d/plugins/panel/plugin.js#L409
https://github.com/ckeditor/ckeditor-dev/blob/549456c24d4fdb173996616eb7afd9edd2c3b83d/plugins/panel/plugin.js#L437

Feature required to properly manage focus in Emoji dropdown menu, #2062

PR #2461, opened 2018-10-05
Introduce emoji dropdown

### LINKED ISSUE #2062: Emoji list button

### LINKED ISSUE #2181: Emoji plugin should support keywords
## Type of report

Feature request

Open sample with Emoji plugin.
Start typing `:smile`

### Expected

`slightly_smiling_face` emoji is being matched too

### Actual

`slightly_smiling_face` emoji is not matched

### Additional Details

[Unicode](https://unicode.org/emoji/charts/emoji-list.html) specifies keywords associated with each smiley, that increases the likelihood of matching given emoji. We should add a support for that too.

msamsel:

New feature

### This PR contains

- [x] Unit tests
- [x] Manual tests

## What changes did you make?

* Plugin require fix from #2453 to have possibility to focus on input element. **Branch is rebased on this change.**
* Add new float panel which displays list of emoji, with basic functionality
* Emoji dropdown also search using keywords not only names of given emoji

Closes #2062, closes #2181.

### DIFF
--- a/tests/plugins/emoji/dropdown.js
+++ b/tests/plugins/emoji/dropdown.js
@@ -10,6 +10,17 @@
 	bender.editor = true;
 
 	bender.test( {
+		_should: {
+			ignore: {
+				// (#2528).
+				'test click inserts emoji to editor and has proper focus': CKEDITOR.env.safari,
+				'test keyboard event should inserts emoji to editor and had proper focus': CKEDITOR.env.safari,
+				'test clicking into navigation list item does not throw an error': CKEDITOR.env.safari,
+				'test input is focused element when dropdown opens': CKEDITOR.env.safari,
+				'test navigation highlights proper section when scrolls': CKEDITOR.env.safari
+			}
+		},
+
 		setUp: function() {
 			if ( emojiTools.notSupportedEnvironment ) {
 				assert.ignore();


msamsel:
I think all issues should be solved now. I've also extract 2 separate issues after your feedback:
#2482, #2483

mlewand:
Some additional notes:

There are very few unit tests.

* There are no unit tests ensuring that the focus goes into the search bar first.
	There is a unit test called `test input is focused element when dropdown opens` but it doesn't fail when the filter input is not focused. You should test `document.activeElement` to verify that.
* There's no unit test for this issue where space key is not inserted into filter input.

---

Button should not create a separate group: https://github.com/ckeditor/ckeditor-dev/blob/f8bf8c2/plugins/emoji/plugin.js#L149 instead it should be used in the "insert" toolbar group, just like special characters button.

---

https://github.com/ckeditor/ckeditor-dev/blob/f8bf8c2/plugins/emoji/plugin.js#L134 - I'm not sure if this line is needed, there are couple of issues with it:

* Actually by default any command has [`editorFocus`](https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_command.html#property-editorFocus) set to `true` so **it already is focusing** the editor.
* It passes a boolean argument to the [editor.focus](https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_editor.html#method-focus) which is not supported/implemented.

---

I think that right and left arrow should not be catched in the filter input, Imagine you made a typo "Slghtly" instead of "Slightly" (missing "i") so you want to go back and insert a missing character. It's impossible now.

Mgsy:
## Steps to reproduce
1. Open the emoji dropdown.
1. Focus the search bar.
1. Type something.
1. Press <kbd>Arrow left</kbd>.

## Expected result
The caret moves to the left.

## Current result
The focus is moved to the list above the search bar.

Mgsy:
## Steps to reproduce
1. Open the emoji dropdown.
1. Change the emoji group three times (by clicking icons).

## Current result
The error is thrown.

```
SCRIPT5007: SCRIPT5007: Unable to get property 'data' of undefined or null reference
```

It's reproducible in **Edge 17.17134**.

Mgsy:
## Steps to reproduce
1. Focus the editor.
1. Open the emoji dropdown.

## Current result
The virtual keyboard blinks and the emoji dropdown jumps. There is no possibility to close the dropdown, sometimes the browser crashes.

Reproducible on **Huawei MediaPad M5/Android 8.0.0/Chrome**.

Mgsy:
### Mobile
## Steps to reproduce
1. Click on the emoji icon in the toolbar.

## Expected result
The emoji dropdown appears.

## Current result
The icon becomes selected and requires second tap to open it.

Reproducible on **iPad/iOS 12.0/Safari**

Mgsy:
### Mobile - iOS
## Steps to reproduce
1. Open the emoji dropdown.
1. Insert some emoji.

## Current result
The page has been scrolled to the bottom.

Reproducible on **iPad/iOS 12.0/Safari**

Mgsy:
## Steps to reproduce
1. Open the emoji dropdown.
1. Open the console (<kbd>F12</kbd>).
1. Close the emoji dropdown.
1. Open the emoji dropdown.
1. Put the caret in some word.

## Expected result
The emoji dropdown disappears.

## Current result
The emoji dropdown is still visible and there is no possibility to close it.

It occurs only in **IE11**.

msamsel:
> There's no unit test for this issue where space key is not inserted into filter input.

I'm not sure how such test should looks like. I couldn't find a way how to mark keydown event with space, which would insert space to input field. I'll appreciate all tips related to that :)

> Issue with [opened console in IE11](https://github.com/ckeditor/ckeditor-dev/pull/2461#issuecomment-432157112) seems to be super edge case, and I feel that best would be extract it as separate case

> [Scrolled page on iOS](https://github.com/ckeditor/ckeditor-dev/pull/2461#issuecomment-432143327) is case where I couldn't find solution, ans case seems to be not the highest priority. Emoji is displayed and possible to use. I suspect that scroll might be somehow related to hiding keyboard used in input search.

> [Android issues](https://github.com/ckeditor/ckeditor-dev/pull/2461#issuecomment-432135592) seems to be quite annoying and complicated. I consult this case, and situation seems to be really similar to case happen with dialogs. When keyboard show up, resize event happen. Floatpanel contains logic which hide panel on resize, and then open it again. This generates a loop, where opening emoji and focusing it, generate opening keyboard, what cause resize event and hiding panel, what close keyboard, and then opening it again. This case should be extracted as refactoring of Floatpanel.

plugins/emoji/plugin.js
```diff
@@ -79,26 +152,466 @@
...
+					group = getEmojiListGroup( groupName );
+
+				return '<section data-cke-emoji-group="' + groupName + '" ><h2 id="' + groupName + '">' + sectionName + '</h2><ul>' + group + '</ul></section>';
+			}
+
+			function getEmojiListGroup( groupName ) {
+				var emojiList = editor._.emoji.list,
+					emojiTpl = new CKEDITOR.template( '<li class="cke_emoji_item">' +
+					'<a draggable="false" data-cke-emoji-full-name="{id}" data-cke-emoji-name="{name}" data-cke-emoji-symbol="{symbol}" data-cke-emoji-group="{group}" ' +
+					'data-cke-emoji-keywords="{keywords}" title="{id}" href="#" _cke_focus="1">{symbol}</a>' +
```
mlewand: Title should contain a human-readable label of an emoji. This will improve accessibility.

plugins/emoji/plugin.js
```diff
@@ -79,26 +152,466 @@
...
+				var itemTemplate,
+					items,
+					svgUrl,
+					imgUrl;
+
+				if ( !CKEDITOR.env.ie ) {
+					svgUrl = CKEDITOR.getUrl( that.path + 'assets/icons-all.svg' );
+
+					itemTemplate = new CKEDITOR.template(
+						'<li class="cke_emoji-navigation_item" data-cke-emoji-group="{group}"><a href="{href}" draggable="false" _cke_focus="1"><svg viewBox="0 0 34 34"> <use href="' +
```
mlewand: `title` with human readable labels would be nice. TBH there might be cases where the icon won't ring a bell. Having the titles will be nice.

mlewand: For `<svg>` element you want to include it as a `<title>foo bar baz</title>` child.

plugins/emoji/plugin.js
```diff
@@ -79,26 +152,466 @@
...
+						var inputNumber;
+						return function() {
+							if ( inputNumber === undefined ) {
+								inputNumber = getItemIndex( blockObject._.getItems().toArray(), blockElement.findOne( 'input' ) );
+							}
+							blockObject._.markItem( inputNumber );
+						};
+					} )()
+				} );
+				return '<label class="cke_emoji-search"><img src="' + escapeString( loupeUrl ) +
```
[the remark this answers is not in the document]
mlewand: Why isn't the magnifier in the same sprite as other icons? It should be, less http requests = better.

plugins/emoji/plugin.js
```diff
@@ -79,26 +152,466 @@
...
+						listener = item.listener,
+						event = item.event;
+
+					CKEDITOR.tools.array.forEach( root.find( selector ).toArray(), function( node ) {
+						node.on( event, listener );
+					} );
+				} );
+			}
+
+			function escapeString( str ) {
```
mlewand: There is already a function for HTML encoding. Look in `CKEDITOR.tools` to reuse it.

plugins/emoji/plugin.js
```diff
@@ -79,26 +152,466 @@
...
+				if ( !CKEDITOR.env.ie ) {
+					svgUrl = CKEDITOR.getUrl( that.path + 'assets/icons-all.svg' );
+
+					itemTemplate = new CKEDITOR.template(
+						'<li class="cke_emoji-navigation_item" data-cke-emoji-group="{group}"><a href="{href}" draggable="false" _cke_focus="1"><svg viewBox="0 0 34 34"> <use href="' +
+						svgUrl +
+						'{href}"></use></svg></a></li>'
+					);
+
+					items = CKEDITOR.tools.array.reduce( GROUPS, function( acc, item ) {
```
mlewand: You are using `CKEDITOR.tools.array` quire a few times. It make sense to add var that makes a simpler alias, like `arrayTools`.

plugins/emoji/plugin.js
```diff
@@ -79,26 +152,466 @@
...
+							group: escapeString( item.name ),
+							href: escapeString( '#' + item.name.toLowerCase() )
+						} );
+					}, '' );
+				} else {
+					imgUrl = CKEDITOR.getUrl( that.path + 'assets/icons-all.png' );
+
+					itemTemplate = new CKEDITOR.template(
+						'<li class="cke_emoji-navigation_item" data-cke-emoji-group="{group}"><a href="{href}" draggable="false" _cke_focus="1">' +
+						'<span style="background-image:url(' + imgUrl + ');display:inline-block;width:21px;height:21px;background-size:105px 42px;' +
```
mlewand: There is a lot of styling going on here. JS is not for styling, all of this (except dynamical stuff) needs to be moved into proper stylesheets.

plugins/emoji/plugin.js
```diff
@@ -79,26 +152,466 @@
...
+					);
+
+					items = CKEDITOR.tools.array.reduce( GROUPS, function( acc, item ) {
+						return acc + itemTemplate.output( {
+							group: escapeString( item.name ),
+							href: escapeString( '#' + item.name.toLowerCase() )
+						} );
+					}, '' );
+				} else {
+					imgUrl = CKEDITOR.getUrl( that.path + 'assets/icons-all.png' );
```
mlewand: Sadly we don't use separators in file name. It's a pain, but we need to keep it consistent with the rest of the editor.

plugins/emoji/lang/en.js
```diff
@@ -0,0 +1,17 @@
+/*
+Copyright (c) 2003-2018, CKSource - Frederico Knabben. All rights reserved.
+For licensing, see LICENSE.md or https://ckeditor.com/legal/ckeditor-oss-license
+*/
+CKEDITOR.plugins.setLang( 'emoji', 'en', {
+	searchPlaceholder: 'Search emoji...',
```
mlewand: It should use a proper ellipsis utf character ("…").

plugins/emoji/plugin.js
```diff
@@ -79,26 +152,466 @@
...
+
+			if ( !editor.plugins.toolbar ) {
+				return;
+			}
+
+			editor.ui.addToolbarGroup( 'emoji', 'insert' );
+			// Name is responsible for icon name also.
+			editor.ui.add( 'emojiPanel', CKEDITOR.UI_PANELBUTTON, {
+				label: 'emoji',
+				title: 'Emoji List',
```
mlewand: This string is not localized.

plugins/emoji/plugin.js
```diff
@@ -79,26 +152,466 @@
...
+							positionY: item.position.y
+						} );
+					}, '' );
+				}
+
+				listeners.push( {
+					selector: 'nav li',
+					event: 'click',
+					listener: function( event ) {
+						var nodeArr = blockElement.find( 'nav li' ).toArray();
```
mlewand: Subsequent `var` statements.

mlewand:
Merged in d735c71dc - for some reason GH did not understand this fact 🤔.

PR #2479, opened 2018-10-15
Add Ctrl+K keystroke for link command

### LINKED ISSUE #2478: Add Ctrl+K keystroke for link command
## Type of report

Feature request

Currently link command is hidden under <kbd>Ctrl</kbd>+<kbd>L</kbd> keystroke, however [the industry standard is <kbd>Ctrl</kbd>+K</kbd>](https://github.com/ckeditor/editor-recommendations/issues/10#issuecomment-193686789). Probably it'd be better to add the new keystroke for our link plugin to align our behaviour with the rest of the world.

Comandeer:

New feature

### This PR contains

- [x] Unit tests
- [x] Manual tests

## What changes did you make?

I've added new keystroke for link command. Additionally I added tests to check if it's working alongside the old <kbd>Ctrl</kbd> + <kbd>L</kbd> one.

Closes #2478.

### DIFF
--- a/tests/plugins/link/link.js
+++ b/tests/plugins/link/link.js
@@ -617,7 +617,6 @@
 			} );
 		},
 
-		// (#2478)
 		'test Ctrl+L keystroke': function() {
 			var bot = this.editorBots.noValidation,
 				editor = bot.editor;
--- a/tests/plugins/link/link.js
+++ b/tests/plugins/link/link.js
@@ -595,50 +595,34 @@
 		} ),
 
 		// (#2478)
-		'test Ctrl+K keystroke': function() {
-			var bot = this.editorBots.noValidation,
-				editor = bot.editor;
+		'test Ctrl+K keystroke': assertKeystroke( 75 ),
 
-			bot.setData( '', function() {
-				editor.once( 'dialogShow', function( evt ) {
-					resume( function() {
-						var dialog = evt.data;
-
-						dialog.hide();
-						assert.areSame( 'link', dialog._.name );
-					} );
-				} );
-
-				editor.editable().fire( 'keydown', new CKEDITOR.dom.event( {
-					keyCode: 75,
-					ctrlKey: true
-				} ) );
-				wait();
-			} );
-		},
-
-		'test Ctrl+L keystroke': function() {
-			var bot = this.editorBots.noValidation,
-				editor = bot.editor;
-
-			bot.setData( '', function() {
-				editor.once( 'dialogShow', function( evt ) {
-					resume( function() {
-						var dialog = evt.data;
-
-						dialog.hide();
-						assert.areSame( 'link', dialog._.name );
-					} );
-				} );
-
-				editor.editable().fire( 'keydown', new CKEDITOR.dom.event( {
-					keyCode: 76,
-					ctrlKey: true
-				} ) );
-				wait();
-			} );
-		}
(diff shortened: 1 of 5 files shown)


tests/plugins/link/link.js
```diff
@@ -592,7 +592,53 @@
...
+
+				editor.editable().fire( 'keydown', new CKEDITOR.dom.event( {
+					keyCode: 75,
+					ctrlKey: true
+				} ) );
+				wait();
+			} );
+		},
+
+		// (#2478)
```
jacekbogdanski: This case is not a part of the issue so I would remove tag reference here.

tests/plugins/link/link.js
```diff
@@ -592,7 +592,53 @@
...
+		'test Ctrl+L keystroke': function() {
+			var bot = this.editorBots.noValidation,
+				editor = bot.editor;
+
+			bot.setData( '', function() {
+				editor.once( 'dialogShow', function( evt ) {
+					resume( function() {
+						var dialog = evt.data;
+
+						dialog.hide();
```
jacekbogdanski: Instead of hiding dialog here, you could move the whole teardown login into `tearDown` test function. Something like:

```javascript
  var dialog = CKEDITOR.dialog.getCurrent();
  if ( dialog ) {
    dialog.hide();
  }
```
It also applies to the second unit test.

mlewand:
I'm not sure about removing the hotkey, we'd need to have some data on how people really use the app to back it up.

E.g. personally I find `ctrl + l` much more intuitive. I also do not miss the native hotkey, as I use `alt + d` for focusing the address bar (more convenient to press).

Comandeer:
I'm also against removing the old keystroke. It was present from the very beginning and probably many people are accustomed to it.

tests/plugins/link/link.js
Comandeer: I'm not sure if it's necessary. Every other test in this file hides dialog on its own. Having two tests that are using `tearDown` for it would break consistency.

tests/plugins/link/link.js
Comandeer: I'd keep the reference and treat this one as prevention from regression (accidentally breaking the former keystroke).

tests/plugins/link/manual/keystroke.md
```diff
@@ -0,0 +1,18 @@
+@bender-tags: feature, link, 2478, 4.11.0
+@bender-ckeditor-plugins: link, toolbar, wysiwygarea
+@bender-ui: collapsed
+----
+
+1. Focus the editor.
+2. Press <kbd>Ctrl</kbd>/<kbd>Cmd</kbd> + <kbd>K</kbd>.
+
+	## Expected
```
jacekbogdanski: I'm not really sure about test construction. I'm more used to linear tests. Although it's subjective, so would like to see the opinion of another reviewer.

tests/plugins/link/link.js
jacekbogdanski: Not if you refactor other tests. However, following the rule "If it ain't broke, don't fix it", lets deal with important stuff, not test refactoring.

tests/plugins/link/link.js
mlewand: I'd second to what @jacekbogdanski noted, this case is not related. Regression protection (related to #2478) is provided by the `test Ctrl+K keystroke` test case.

PR #2487, opened 2018-10-16
Fixed issue with incorrectly escaped additional entities

### LINKED ISSUE #2448: Special entities in entities_additional may break escaping
## Type of report

Bug

1) Initialize CKE with `entities_additional` set to `'apos'`
2) Enter text `amp'` into the editor
3) Switch to source view

### Expected result

Source view shows `<p>amp&apos;</p>`

### Actual result

Source view shows `<p>&undefined;&undefined;&undefined;&apos;</p>`

## Other details

* Browser: Firefox, Chrome
* OS: Debian, Windows
* CKEditor version: 4.10.1
* Installed CKEditor plugins: Default plugins as in development version

I pinned this down to be caused by `plugins/entities/plugin.js::buildTable`. During the replacement of `entities` in line 63 any special entity will be removed from the entities list together with the following `,`. If there is no comma (because it is the last value) the resulting ends with a comma.

This causes `entities.split(...)` to create an array with empty string as last value while the `join` then creates a string ending with `&;` which will be replaced by `&amp;;` when reading inner HTML.
Due to this `a`, `m` and `p` will be interpreted as entities by CKE.

jacekbogdanski:

Bug fix

### This PR contains

- [x] Unit tests
- [x] Manual tests

## What changes did you make?

Simple comma leftover broke the whole `entities_additional` feature. I could modify regex instead of additional trailing coma removal, although it may look nasty.

Closes #2448


mlewand:
Rebased and adjusted that one minor thing in manual test.

PR #2489, opened 2018-10-17
Add option to allow non-merging deletion across blocks

mysticfall:

This PR adds an option `forceMergeBlocks` to allow deleting of content without merging blocks when the selection spans across two or more blocks when it's set to `false` (default to `true` which is the current behavior).

Popular word processors like Microsoft Word, for example, lets users to select texts across two adjacent blocks and delete the content without merging them, thus preserving their respective styles.

In comparison, CKEditor always merge the blocks if an expanded selection spans across them. It may not be a desired behavior, especially when the blocks have different styles (i.e. line-height) from each other.

### This PR contains

- [x] Unit tests
- [ ] Manual tests

## What changes did you make?

Adds `config.forceMergeBlocks` options which determines how the above mentioned behavior works. The default value leaves the current behavior as it is, as it's shown in one of the unit tests.

### DIFF
--- a/config.js
+++ b/config.js
@@ -11,7 +11,8 @@ CKEDITOR.editorConfig = function( config ) {
 	config.plugins =
 		'about,' +
 		'a11yhelp,' +
-		'basicstyles,' +
+		//'basicstyles,' +
+		'computedstyles,' +
 		'bidi,' +
 		'blockquote,' +
 		'clipboard,' +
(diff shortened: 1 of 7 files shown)


mysticfall:
Sorry, it looks like I better separate these commits from other non related changes needed only for my project. I'll open another PR with a proper branch later.

PR #2501, opened 2018-10-22
Block new line inside list item no longer breaks list structure

### LINKED ISSUE #2205: It is not possible to add new list items under an item containing block element
## Type of report

Bug

1. Open editor sample and clean editor contents - https://sdk.ckeditor.com/samples/fullpreset.html.
2. Insert list.
3. Type text in first list item e.g. "test" and press Enter.
4. Insert table in second list item, move below the table using arrows and press Enter.

Please see attached video to get a visual representation of what I’m describing.
[table.zip](https://github.com/ckeditor/ckeditor-dev/files/2166667/table.zip)

### Expected result

Third list item gets created.

### Actual result

Second item together with table is being removed.

## Other details

Problem with removing list item together with a table can be reproduced from CKEditor 4.4.3 however first problems started occurring in version 4.3.0 where it was no longer possible to create new list item but instead new lines were added to second item of the list.

* Browser: Any
* OS: Any
* CKEditor version: 4.4.3+

jacekbogdanski:

Bug fix

### This PR contains

- [x] Unit tests
- [x] Manual tests

## What changes did you make?

The fix could be as easy as skipping whole list moving inside `enterkey` plugin when block item is a placeholder (new line added for block elements, without it you cannot create new list item by an enter key). Although, it causes an issue where block tag style is copied into following list items, which is unwanted behavior in this special use case.

Be aware that this fix doesn't creating break empty lines inside list item. They still can be added using `shift+enter` keystroke which creates `br` tags in the same node and in the result are not touched by the fix.

Closes #2205


Comandeer:
Rebased onto latest `master`.

tests/plugins/enterkey/manual/blockplaceholder.md
```diff
@@ -0,0 +1,20 @@
+@bender-tags: 4.11.0, bug, 2205
```
Comandeer: Please update version tag.

jacekbogdanski:
If I'm right the issue can be easily resolved by dropping block element type.

PR #2502, opened 2018-10-22
Mentions plugin now matches diacritic characters

### LINKED ISSUE #2491: Mentions are not matching unicode/diacritic characters
## Type of report

Bug

1. Use CKEDITOR 4.10 with mentions plugin
2. Include mentions data with accents like 'Céline TEST'
3. Try to use mentions with data typing '@Cé'

### Expected result

Mentions propose entry 'Céline TEST'

### Actual result

No results

## Other details

It seems due to emphasis/accents 
I've tried with direct data in array or with function feed.

I think we would have same results trying with accents or not (@Cel or @Cél must return same results).
It can be solved with a new regexp parameter to apply to request and data ?

Thanks per advance !

* Browser: Any
* CKEditor version: 4.10
* Installed CKEditor plugins: Mentions

jacekbogdanski:

Bug

### This PR contains

- [x] Unit tests
- [x] Manual tests

## What changes did you make?

Fixed mentions regex to match diacritic characters.

Closes #2491

### DIFF
--- a/plugins/mentions/plugin.js
+++ b/plugins/mentions/plugin.js
@@ -142,7 +142,8 @@
 	};
 
 	function createPattern( marker, minChars ) {
-		var pattern = '\\' + marker + '\\w';
+		// Match also diacritic characters (#2491).
+		var pattern = '\\' + marker + '[_A-z0-9À-ž]';
 
 		if ( minChars ) {
 			pattern += '{' + minChars + ',}';
--- a/plugins/mentions/plugin.js
+++ b/plugins/mentions/plugin.js
@@ -143,7 +143,7 @@
 
 	function createPattern( marker, minChars ) {
 		// Match also diacritic characters (#2491).
-		var pattern = '\\' + marker + '[_A-z0-9À-ž]';
+		var pattern = '\\' + marker + '[_a-zA-Z0-9À-ž]';
 
 		if ( minChars ) {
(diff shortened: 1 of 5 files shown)


plugins/mentions/plugin.js
```diff
@@ -142,7 +142,8 @@
 	};
 
 	function createPattern( marker, minChars ) {
-		var pattern = '\\' + marker + '\\w';
+		// Match also diacritic characters (#2491).
+		var pattern = '\\' + marker + '[_A-z0-9À-ž]';
```
mlewand: Let's be more precise with the regexp, your regexp catches also chars like `[`, `^`, backtick etc. `a-zA-Z` is the way to go here.

tests/plugins/mentions/manual/diacritic.md
```diff
@@ -0,0 +1,16 @@
+@bender-tags: 4.11.0, 2491, bug, mentions
+@bender-ui: collapsed
+@bender-ckeditor-plugins: wysiwygarea, toolbar, basicstyles, mentions
+
+1. Focus the editor.
```
mlewand: Manual test should mention that you need to set up a Polish keyboard.

plugins/mentions/plugin.js
mlewand: Please add a unit test ensuring that phrases with [^ are not matched.

PR #2504, opened 2018-10-22
Added additional option to retrieve all keys

### LINKED ISSUE #2493: Add a way to fetch multiple keystrokes for a single command
## Type of report

Feature request

Some commands might have multiple hotkeys, we need to API that returns all the hotkeys.

There's already a [`editor.getCommandKeystroke`](https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_editor.html#method-getCommandKeystroke) but it supports only one returned keystroke.

An example case is link command. In 4.11.0 we added <kbd>ctrl+k</kbd> keystroke in addition to already existing <kbd>ctrl+l</kbd>.

What are the options?

* Introduce `asArray` (or `returnMultiple` or something like that) parameter that tells that the function should return an array. (We can't change currently returned value for backward compatibility)
* Introduce complimentary `editor.getCommandKeystrokes` (note plural form).

Looking at above I'm more keen towards additional argument. Adding an extra function will further bloat our docs.

jacekbogdanski:

New feature

### This PR contains

- [x] Unit tests
- [x] Manual tests

## What changes did you make?

Closes #2493


jacekbogdanski:
Actually, I think that it makes more sense to return an empty array instead of `null` when dealing with arrays. In that case statement: `returns all keystrokes` is valid. Also, it will be easier to use:
```javascript
var keystrokes = editor.getCommandKeystroke( 'command', true );

// With an array, we can skip this condition, but still able to check if an array is empty using `keystrokes.length`
if ( !keystrokes ) {
  result = keystrokes.map( doSth );
}
```

mlewand:
> Actually, I think that it makes more sense to return an empty array instead of `null` when (…)

I thought about the same. However TBH I have hard times writing it simply in the docs (I mean in `@returns` tag). Please revise it as it needs to be dead simple. 🙂

core/editor.js
```diff
@@ -1430,28 +1430,36 @@
 		 *
 		 * @since 4.6.0
 		 * @param {CKEDITOR.command/String} command The {@link CKEDITOR.command} instance or a string with the command name.
-		 * @returns {Number/null} The keystroke assigned to the provided command or `null` if there is no keystroke.
+		 * @param {Boolean} [all=false] If `true` function will return an array of the assigned keystrokes.
+		 * Available since 4.11.0.
+		 * @returns {Number/Number[]/null} Depending on the `all` parameter value:
+		 *
+		 * `false` (default): the first keystroke assigned to the provided command or `null` if there is no keystroke.
```
mlewand: This proposal looks really nice. I'll just remove info about `false` being a default value of `all` attribute as it is indicated by `[all=false]` and it should be the only source of truth for this information.

PR #2518, opened 2018-10-25
Link is now required by autolink plugin

### LINKED ISSUE #1824: Autolink plugin doesn't work without link plugin

Bug

1. Open CKEditor sample with enabled `autolink` plugin and not included `link` plugin.
2. Copy URL e.g. `http://google.com`.
3. Paste URL into an editor.

### Expected result

URL should be turned into link e.g. `<a href="http://google.com">google.com</a>`

### Actual result

URL is pasted as a text.

## Other details

`autolink` plugin should probably [require](https://docs.ckeditor.com/ckeditor4/latest/api/CKEDITOR_pluginDefinition.html#property-requires) `link` plugin by default.

* Browser: Any
* OS: Any
* CKEditor version: 4.8.0
* Installed CKEditor plugins: autolink

### LINKED ISSUE #2208: Autolink duplicated for IE
## Type of report

Bug

## Description

Internet Explorer has its own autolinking feature which works out of the box for both pasting and typing. There is no need to duplicate this behavior and also no easy way to properly [disable IE autolink feature](https://blogs.msdn.microsoft.com/ieinternals/2009/09/17/preventing-automatic-hyperlinking-in-contenteditable-html/).

## Other details

* Browser: IE
* OS: Windows
* CKEditor version: 4.10.0
* Installed CKEditor plugins: autolink

jacekbogdanski:

Bug fix

### This PR contains

- [x] Unit tests
- [x] Manual tests

## What changes did you make?

Updated autolink plugin to require link.

Closes #1824.
Closes #2208.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -5,7 +5,7 @@ CKEditor 4 Changelog
 
 Fixed Issues:
 
-* [#2208](https://github.com/ckeditor/ckeditor4/issues/2208): [IE] Fixed: [Autolink](https://ckeditor.com/cke4/addon/autolink) plugin duplicated with native browser implementation.
+* [#2208](https://github.com/ckeditor/ckeditor4/issues/2208): [IE] Fixed: [Autolink](https://ckeditor.com/cke4/addon/autolink) plugin duplicates the native browser implementation.
 * [#1824](https://github.com/ckeditor/ckeditor4/issues/1824): Fixed: [Autolink](https://ckeditor.com/cke4/addon/autolink) plugin does not work without [Link](https://ckeditor.com/cke4/addon/link) plugin.
 * [#4253](https://github.com/ckeditor/ckeditor4/issues/4253): Fixed: [Editor Placeholder](https://ckeditor.com/cke4/addon/editorplaceholder) plugin throws an error during editor initialization with [fullpage](https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_config.html#cfg-fullPage) enabled when there is no `body` tag in editor content.
 * [#4372](https://github.com/ckeditor/ckeditor4/issues/4372): Fixed: [Autogrow](https://ckeditor.com/cke4/addon/autogrow) plugin changes editor's width when used with absolute [`config.width`](https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_config.html#cfg-width) value.
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -5,6 +5,7 @@ CKEditor 4 Changelog
 
 Fixed Issues:
 
+* [#1824](https://github.com/ckeditor/ckeditor-dev/issues/1824): Fixed: [`Autolink`](https://ckeditor.com/cke4/addon/autolink) plugin doesn't work without [`Link`](https://ckeditor.com/cke4/addon/link) plugin.
 * [#4253](https://github.com/ckeditor/ckeditor4/issues/4253): Fixed: [Editor Placeholder](https://ckeditor.com/cke4/addon/editorplaceholder) plugin throws an error during editor initialization with [fullpage](https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_config.html#cfg-fullPage) enabled when there is no `body` tag in editor content.
 * [#4372](https://github.com/ckeditor/ckeditor4/issues/4372): Fixed: [Autogrow](https://ckeditor.com/cke4/addon/autogrow) plugin changes editor's width when used with absolute [`config.width`](https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_config.html#cfg-width) value.
 * [#4379](https://github.com/ckeditor/ckeditor4/issues/4379): [Edge] Fixed: Incorrect detection of [high contrast mode](https://ckeditor.com/docs/ckeditor4/latest/guide/dev_a11y.html#high-contrast-mode).
@@ -310,15 +311,16 @@ Fixed Issues:
 
 Fixed Issues:
 
-* [#1986](https://github.com/ckeditor/ckeditor-dev/issues/1986): Fixed: Cell Properties dialog from [Table Tools](https://ckeditor.com/cke4/addon/tabletools) plugin shows styles that are not allowed through [`config.allowedContent`](https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_config.html#cfg-allowedContent).
-* [#2403](https://github.com/ckeditor/ckeditor-dev/issues/2403): Fixed: Styling inline editor initialized inside a table with [Table Selection](https://ckeditor.com/cke4/addon/tabletools) plugin is causing style leaks.
-* [#2514](https://github.com/ckeditor/ckeditor-dev/issues/2403): Fixed: Pasting table data into inline editor initialized inside a table with [Table Selection](https://ckeditor.com/cke4/addon/tabletools) plugin inserts pasted content into a wrapping table.
-* [#2451](https://github.com/ckeditor/ckeditor-dev/issues/2451): Fixed: The [Remove Format](https://ckeditor.com/cke4/addon/removeformat) changes selection.
-* [#2546](https://github.com/ckeditor/ckeditor-dev/issues/2546): Fixed: Separator in toolbar moves when buttons are focused.
-* [#2506](https://github.com/ckeditor/ckeditor-dev/issues/2506): Fixed: [Enhanced Image](https://ckeditor.com/cke4/addon/image2) throws type error when empty figure tag with 'image' class is upcasted.
-* [#2650](https://github.com/ckeditor/ckeditor-dev/issues/2650): Fixed: [Table](https://ckeditor.com/cke4/addon/table) dialog validator fails when function `getValue` is defined in global scope.
-* [#2208](https://github.com/ckeditor/ckeditor-dev/issues/2208): [IE8-11] Fixed: [Autolink](https://ckeditor.com/cke4/addon/autolink) plugin duplicated with native browser implementation.
-* [#1824](https://github.com/ckeditor/ckeditor-dev/issues/1824): Fixed: [`Autolink`](https://ckeditor.com/cke4/addon/autolink) plugin doesn't work without [`Link`](https://ckeditor.com/cke4/addon/link) plugin.
+* [#2403](https://github.com/ckeditor/ckeditor4/issues/2403): Fixed: Styling inline editor initialized inside a table with the [Table Selection](https://ckeditor.com/cke4/addon/tableselection) plugin is causing style leaks.
+* [#2514](https://github.com/ckeditor/ckeditor4/issues/2403): Fixed: Pasting table data into inline editor initialized inside a table with the [Table Selection](https://ckeditor.com/cke4/addon/tableselection) plugin inserts pasted content into the wrapping table.
+* [#2451](https://github.com/ckeditor/ckeditor4/issues/2451): Fixed: The [Remove Format](https://ckeditor.com/cke4/addon/removeformat) plugin changes selection.
+* [#2546](https://github.com/ckeditor/ckeditor4/issues/2546): Fixed: The separator in the toolbar moves when buttons are focused.
+* [#2506](https://github.com/ckeditor/ckeditor4/issues/2506): Fixed: [Enhanced Image](https://ckeditor.com/cke4/addon/image2) throws a type error when an empty `<figure>` tag with an `image` class is upcasted.
+* [#2650](https://github.com/ckeditor/ckeditor4/issues/2650): Fixed: [Table](https://ckeditor.com/cke4/addon/table) dialog validator fails when the `getValue()` function is defined in the global scope.
+* [#2690](https://github.com/ckeditor/ckeditor4/issues/2690): Fixed: Decimal characters are removed from the inside of numbered lists when pasting content using the [Paste from Word](https://ckeditor.com/cke4/addon/pastefromword) plugin.
+* [#2205](https://github.com/ckeditor/ckeditor4/issues/2205): Fixed: It is not possible to add new list items under an item containing a block element.
+* [#2411](https://github.com/ckeditor/ckeditor4/issues/2411), [#2438](https://github.com/ckeditor/ckeditor4/issues/2438) Fixed: Apply numbered list option throws a console error for a specific markup.
+* [#2430](https://github.com/ckeditor/ckeditor4/issues/2430) Fixed: [Color Button](https://ckeditor.com/cke4/addon/colorbutton) and [List Block](https://ckeditor.com/cke4/addon/listblock) items are draggable.
 
 Other Changes:
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,6 +4,7 @@ CKEditor 4 Changelog
 ## CKEditor 4.15.2
 
 Fixed Issues:
+* [#2208](https://github.com/ckeditor/ckeditor4/issues/2208): [IE] Fixed: [Autolink](https://ckeditor.com/cke4/addon/autolink) plugin duplicated with native browser implementation.
 * [#1824](https://github.com/ckeditor/ckeditor4/issues/1824): Fixed: [Autolink](https://ckeditor.com/cke4/addon/autolink) plugin doesn't work without [Link](https://ckeditor.com/cke4/addon/link) plugin.
 * [#4253](https://github.com/ckeditor/ckeditor4/issues/4253): Fixed: [Editor Placeholder](https://ckeditor.com/cke4/addon/editorplaceholder) plugin throws an error during editor initialization with [fullpage](https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_config.html#cfg-fullPage) enabled when there is no `body` tag in editor content.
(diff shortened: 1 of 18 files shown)


plugins/autolink/plugin.js
```diff
@@ -9,12 +9,9 @@
 	var doubleQuoteRegex = /"/g;
 
 	CKEDITOR.plugins.add( 'autolink', {
-		requires: 'clipboard,textmatch',
+		requires: 'clipboard,textmatch,link',
 
 		init: function( editor ) {
-			var urlTemplate = new CKEDITOR.template( '<a href="{link}">{text}</a>' ),
```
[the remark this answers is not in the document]
jacekbogdanski: @Comandeer templates have been introduced because in the previous version `autolink` doesn't require link plugin which is a bug thus `autolink` doesn't make sense without `link`. With the `link` plugin, it's easy to create a new link using its API, so we can safely drop manual DOM creation by templates. I understand your point that templating **may** be useful in the future, but currently, we don't have a ticket for such a feature and I think it's not worth to keep bloated code "just in case".

How having two-step link creation improves readability if we can explicitly set link text and attribute instead of reading HTML template?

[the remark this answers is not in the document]
Comandeer: After second thought it makes sense… but it would be nice to have user-modifiable templates. Maybe some day… ;)

Comandeer (dismissed):
Unit tests are failing in IE8: screenshot 2018-11-14 at 16 25 12

jacekbogdanski:
The tests fail mostly because the whole IE8 uppercase element name (upstream), which is pretty easy to fix. Although, I'm rather thinking about ignoring this test for IE taking into account this issue: https://github.com/ckeditor/ckeditor-dev/issues/2208

I would prefer to create followup PR than bloating code with fixes.

jacekbogdanski:
@Comandeer I changed the base branch to `t/2208` - it contains fix disabling autolink plugin for IE8-11.

sculpt0r:
Please note that all `IE ignors` also fix #2208

sculpt0r:
Ready for the review :)

CHANGES.md
```diff
@@ -5,6 +5,7 @@ CKEditor 4 Changelog
 
 Fixed Issues:
 
+* [#1824](https://github.com/ckeditor/ckeditor-dev/issues/1824): Fixed: [`Autolink`](https://ckeditor.com/cke4/addon/autolink) plugin doesn't work without [`Link`](https://ckeditor.com/cke4/addon/link) plugin.
```
Dumluregn, line 8: In changelog we use backticks for code, but not for plugin names, so we should remove them from this entry. Also, repository name in the link needs to be updated (ckeditor-dev -> ckeditor4).

plugins/autolink/plugin.js
```diff
@@ -9,11 +9,14 @@
 	var doubleQuoteRegex = /"/g;
```
Dumluregn, line 9: This variable is only used in one place, so I don't see a reason why it needs to be detached from the scope it's necessary.

plugins/autolink/plugin.js
```diff
@@ -127,6 +121,10 @@
 				return query.match( editor.config.autolink_urlRegex ) ||
 					query.match( editor.config.autolink_emailRegex );
 			}
+		},
+
+		isSupportedEnvironment: function() {
```
Dumluregn, line 126: `isSupportedEnvironment` is a kind of plugin metadata, it should be placed right after the plugin `requires` property.

plugins/autolink/plugin.js
```diff
@@ -127,6 +121,10 @@
...
+			return !CKEDITOR.env.ie || CKEDITOR.env.edge;
```
Dumluregn, line 127: Since Edge has a new chromium face, it should support the plugin now.
BTW, in Edge Chromium `CKEDITOR.env.edge === false` 🙈 . That means it is meaningful in the old Edge only and while we could leave this condition to disable the plugin there, it may be misleading. So since we don't really bother much about edgehtml version of Edge (it's less important for us than e.g. IE11), we can remove this part of conditional statement for the sake of clarity.

tests/plugins/autolink/manual/continoustyping.md
```diff
@@ -1,6 +1,6 @@
 @bender-ui: collapsed
 @bender-tags: 4.11.0, feature, 1815
-@bender-ckeditor-plugins: clipboard, toolbar, wysiwygarea, link, autolink
+@bender-ckeditor-plugins: clipboard, toolbar, wysiwygarea, autolink
```
Dumluregn, line 3: The `link` plugin can also be removed from `customregex.md` and `sourcemode.md`.

tests/plugins/autolink/manual/link.md
```diff
@@ -0,0 +1,12 @@
+@bender-ui: collapsed
```
Dumluregn, line 1: Please update this manual test styling according to [our style guide](https://github.com/ckeditor/ckeditor4/blob/major/dev/docs/codestyle.md#31-manual-tests).

tests/plugins/autolink/autolink.js
```diff
@@ -41,6 +40,12 @@
 
 	bender.test( {
 
+		setUp: function() {
+			if ( !CKEDITOR.plugins.registered.autolink.isSupportedEnvironment() ) {
```
Dumluregn, line 44: There is a nice bender alias for this: `bender.tools.ignoreUnsupportedEnvironment( 'autolink' );` (used e.g. [here](https://github.com/ckeditor/ckeditor4/blob/major/tests/plugins/autocomplete/autocomplete.js#L41)). This should be replaced also in all manual tests.

plugins/autolink/plugin.js
```diff
@@ -9,11 +9,14 @@
...
 	CKEDITOR.plugins.add( 'autolink', {
-		requires: 'clipboard,textmatch',
+		requires: 'clipboard,textmatch,link',
 
 		init: function( editor ) {
-			var urlTemplate = new CKEDITOR.template( '<a href="{link}">{text}</a>' ),
-				emailTemplate = new CKEDITOR.template( '<a href="mailto:{link}">{text}</a>' );
+
+			// (#2208)
+			if ( !this.isSupportedEnvironment() ) {
```
Dumluregn, line 17: With this early return added we should remove lines 39-42 as they duplicate the same logic.

Dumluregn (changes requested):
Works and looks good overall, however in the last two years since this PR was created we developed some new tools and changed some conventions, so the code needs a bit of fresh air (nothing major I think).

plugins/autolink/plugin.js
jacekbogdanski: @Dumluregn please, take a look at this statement. It tells that the feature is supported on all modern browsers and Edge HTML. We shouldn't break the feature on the browser where it works correctly, even if it's the older version.

Dumluregn: Ok, so this comment can be ignored 👍

Dumluregn:
Does this PR also close #2208?

sculpt0r:
> Does this PR also close #2208?

I wasn't sure at the beginning, but `yes`. So I update PR description and improve tests.

sculpt0r:
I think how to test not loading plugin...
The best-case scenario will be if there is already a built-in system that detects this situation.
I found, that there was already discussed: https://github.com/ckeditor/ckeditor4/issues/2692, https://github.com/ckeditor/ckeditor4/issues/3216

Also, I found, that @jacekbogdanski already tried it: https://github.com/ckeditor/ckeditor4/pull/2694/files#diff-12f7a7dd3054291d44fad4154f938122a36b5e400d619d374d23f168da41a9caR25

But I am not convinced, that we should add `_loaded` to production code only in one case, just to satisfy tests.
What is more, we already have this situation with `easyimage`. Plugins `return` from their initialization at the very beginning if there are loaded in `isSupportedEnvironment`.
We don't have such test for `easyimage`. So maybe we should rethink https://github.com/ckeditor/ckeditor4/issues/3216, add `_loaded` flag just for tests or don't test this at all. WDYT @Dumluregn ?

tests/plugins/autolink/manual/link.md
```diff
@@ -0,0 +1,12 @@
...
+@bender-tags: 4.11.0, bug, 1824
```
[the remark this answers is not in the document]
sculpt0r: Finally decided to leave it bumped to `4.11.2`: https://github.com/ckeditor/ckeditor4/pull/2518/commits/285699a71b4885e1a2af9b1e652f56703be6dfc7

CHANGES.md
sculpt0r: Removed backticks and update repo address: https://github.com/ckeditor/ckeditor4/pull/2518/commits/9b9cb89b5693a11c71747e3ff7c34f0fbfc5bc43

Also added an entry about fixing #2208 issue.

plugins/autolink/plugin.js
sculpt0r: Since the regular expression isn't complicated here, I removed variable: https://github.com/ckeditor/ckeditor4/pull/2518/commits/1e6a75833fa6a310f880161d14028617839641c6

plugins/autolink/plugin.js
sculpt0r: Moved `isSupportedEnvironment` to the top of the file:  https://github.com/ckeditor/ckeditor4/pull/2518/commits/b6fd74b101e83c6b360cdb5c5ad7be256632522c

tests/plugins/autolink/manual/continoustyping.md
sculpt0r: Removed `link` from plugins loaded directly by bender: https://github.com/ckeditor/ckeditor4/pull/2518/commits/fcca3bc70aed2cfb91f41de1a0f044c3a2b1c5aa

tests/plugins/autolink/manual/link.md
sculpt0r: Adjust style to newest guides: https://github.com/ckeditor/ckeditor4/pull/2518/commits/94363bad1d3ce4ed51b6c261bc8988cb4d265347

Also, I reformat all manual tests: https://github.com/ckeditor/ckeditor4/pull/2518/commits/a57e0137534c60559ce5e85da8c5b76fd84127eb and https://github.com/ckeditor/ckeditor4/pull/2518/commits/ef3e02ddb8a46a3ed3b4c1889d4693ed24fdceb4

tests/plugins/autolink/autolink.js
sculpt0r: Updated: https://github.com/ckeditor/ckeditor4/pull/2518/commits/1102077485b8c4f75ceeb337f143fce444c88c87 and https://github.com/ckeditor/ckeditor4/pull/2518/commits/8bea34024c8ca1039d85b0743ac6dfc943c1dacc and https://github.com/ckeditor/ckeditor4/pull/2518/commits/14ae36c2daf708739665e3b322c53b797dd61745

plugins/autolink/plugin.js
sculpt0r: Duplicated logic removed: https://github.com/ckeditor/ckeditor4/pull/2518/commits/246502b71b98ea5d4fec093d0044256dcf611956

sculpt0r:
@Dumluregn - ready for next round of review :)

Dumluregn:
As for the [comment concerning IE testing](https://github.com/ckeditor/ckeditor4/pull/2518#issuecomment-744183980) - I also think that adding an additional property just for this case isn't desirable. I think that the fact that we are using the `isSupportedEnvironment` in both plugin code and in tests (since it's under the hood of `bender.tools.ignoreUnsupportedEnvironment`) is testing itself, and as you mentioned such an approach seems to be used in other plugins with limited browser support.

I was thinking about some test comparing the behaviour of autolinking in IE without plugin and in modern browsers with the plugin on to detect when it is really turned off in IE, but the produced HTML seems to be the same in the cases I've tested (by pasting the content with link and by typing), so it doesn't seem to be worth digging.

CHANGES.md
```diff
@@ -4,7 +4,8 @@ CKEditor 4 Changelog
 ## CKEditor 4.15.2
 
 Fixed Issues:
-
```
Dumluregn, line 7: This empty line should stay here for nicer formatting.

plugins/autolink/plugin.js
```diff
@@ -73,38 +73,29 @@
...
+				value = value.match( CKEDITOR.config.autolink_urlRegex ) ? value : 'mailto:' + value;
 
-					if ( attributes.removed.length ) {
-						link.removeAttributes( attributes.removed );
-					}
+				link.setText( text );
+				link.setAttribute( 'href', value );
 
-					link.removeAttribute( 'data-cke-saved-href' );
+				var linkData = CKEDITOR.plugins.link.parseLinkAttributes( editor, link ),
```
Dumluregn, line 85: It isn't absolutely necessary, but let's add a ticket reference above this line as it isn't present anywhere yet.

tests/plugins/autolink/manual/autolink.html
```diff
@@ -24,5 +24,11 @@ <h3>URLs that shouldn't be turned into links</h3>
 </textarea>
 
 <script>
-	CKEDITOR.replace( 'editor1' );
+	CKEDITOR.replace( 'editor1', {
+		on: {
```
Dumluregn, line 28: This can be done easier - you can just put `bender.tools.ignoreUnsupportedEnvironment( 'autolink' );` right after the `<script>` tag. Same for all manuals 🙂

tests/plugins/autolink/manual/link.md
```diff
@@ -0,0 +1,10 @@
+@bender-ui: collapsed
+@bender-tags: 4.11.2, bug, 1824
```
Dumluregn, line 2: This test is brand new - could you explain why you decided to go with `4.11.2` tag instead of `4.15.2`?

tests/plugins/autolink/manual/link.md
```diff
@@ -0,0 +1,10 @@
...
+@bender-ckeditor-plugins: toolbar, wysiwygarea, autolink
+
+1. Open the editor.
```
Dumluregn, line 5: This step isn't necessary i guess 🙂

tests/plugins/autolink/manual/autolink.md
```diff
@@ -1,10 +1,9 @@
 @bender-ui: collapsed
 @bender-tags: 4.7.0, bug, trac14769
-@bender-ckeditor-plugins: clipboard, toolbar, wysiwygarea, elementspath, link, autolink
+@bender-ckeditor-plugins: clipboard, toolbar, wysiwygarea, elementspath, autolink
 
-## Test Scenario
+1. Paste each URL to the editor without trailing spaces.
 
-Paste each URL to the editor and check whether it's turned into a link.
+**Expected** Pasted URL turns into a link.
```
Dumluregn, line 7: Well, I've fallen into my own trap. They say "don't touch old tests!".  And damn, aren't they right.

With the current instructions test fails, as there are URLs that shouldn't turn into links. So here we are adding ourselves work with rephrasing the test which is totally unnecessary as it will not be run in the testing phase anyway (we'll only run manual tests with bender tag in line with the tested release). So as far as I appreciate your effort to improve also this old test, please restore the original phrasing (but you can of course leave `1.` at the beginning).

CHANGES.md
sculpt0r: Added empty line.

CHANGES.md
```diff
@@ -4,7 +4,8 @@ CKEditor 4 Changelog
...
+* [#2208](https://github.com/ckeditor/ckeditor4/issues/2208): [IE] Fixed: [Autolink](https://ckeditor.com/cke4/addon/autolink) plugin duplicated with native browser implementation.
+* [#1824](https://github.com/ckeditor/ckeditor4/issues/1824): Fixed: [Autolink](https://ckeditor.com/cke4/addon/autolink) plugin doesn't work without [Link](https://ckeditor.com/cke4/addon/link) plugin.
```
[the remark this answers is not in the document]
sculpt0r: Replace `doesn't` with `does not`.

plugins/autolink/plugin.js
sculpt0r: I added issue reference.

tests/plugins/autolink/manual/autolink.html
sculpt0r: I simplified it.

tests/plugins/autolink/manual/link.md
sculpt0r: First: I read Comandeer comment about bumping and I thought that it should be bumped to the newest version. Then for some reason... I thought that if Jacek already bumped it - It should be ok like it was after his changes. 

I didn't think that it was 2 years ago and since it wasn't merged then, I should use the current version - so `4.15.2`.

Sorry for the confusion...

I change this to `4.15.2` again :)

tests/plugins/autolink/manual/link.md
sculpt0r: I removed first step.

tests/plugins/autolink/manual/autolink.md
sculpt0r: I restored the previous sense of test description.

sculpt0r:
Ready for another round.

Dumluregn:
Rebased onto latest `master`.

Dumluregn (approved):
LGTM! 🎉 

Tests:
* CI is red because of the `exportPdf` plugin test.
* Chrome: 👍🏻 
* FF: 👍🏻 
* Edge: 👍🏻 
* Safari: 👍🏻 
* Opera: 👍🏻 
* IE 8-11: 👍🏻  (all ignored)

PR #2520, opened 2018-10-26
Accessibility Helper dialog supports multiple command keystrokes

### LINKED ISSUE #2519: Multiple keyboard shortcuts (for a single command) should be displayed in a11yhelp
## Type of report

Bug

Recently we added a new keystroke to link command so it now handles both `ctrl + l` and `ctrl + l`. This is not reflected by a11yhelp dialog as it uses only a single keystroke.

1. Open page with CKE4 including link and a11yhelp plugins.
2. Focus the editor.
3. Press alt+0.
3. Scroll to "Link command" section.

### Expected result

The line should read "Press Ctrl+K / Ctrl+L".

### Actual result

It lists only a single keystroke: "Press Ctrl+K".

image

## Other details

* Browser: Any
* OS: Any
* CKEditor version: 4.11.0

jacekbogdanski:

Bug fix

### This PR contains

- [ ] Unit tests
- [x] Manual tests

## What changes did you make?

Included all command keystrokes if available.

I wonder if we need a unit test for proposed change. It would probably require searching for correct text by regex inside dialog HTML. WDYT?

Closes #2519


mlewand:
> I wonder if we need a unit test for proposed change. It would probably require searching for correct text by regex inside dialog HTML. WDYT?

I think this is a no-brainer so we can skip this one. But yea, basically your approach would be a nice way to test it.

PR #2581, opened 2018-11-06
Docs: corrected wrong link.

engineering-this:
Fixed wrong link in docs.


mlewand:
Actually I'll include it still into 4.11.0 as there's going to be one more push to the repo today and this change is safe.

PR #2628, opened 2018-11-28
Update tr lang file in templates plugin

emr:


mlewand:
Thank you very much for contribution! We're using [Transifex](https://www.transifex.com/ckeditor/ckeditor/) for managing language strings. Please send your contribution there, so that we can pull it into upcoming release.

Thanks!

PR #2651, opened 2018-12-05
Fix for: image2 throws an exception when empty figure with class 'image' is upcasted

### LINKED ISSUE #2506: image2: throws an exception if set a content with an empty figure tag
## Type of report

Bug

1.  Set extraPlugin as `config.extraPlugins = 'image2'`;
2.  Set a value for ckeditor as  `<figure class="image"></figure>`

image2 plugin throws an exception if ckeditor's content contains empty "figure" tag (or if a "figure" does not contains nor "img" nor "a" tag). 

```
<div id="editor">
    <figure class="image"></figure>
</div>

```

image2  expects that "figure" must always contains "img" tag or "a" tag that contains "img".
`image = el.getFirst( 'img' ) || el.getFirst( 'a' ).getFirst( 'img' );`
The issue can be reloved by adding additional checking:
`image = el.getFirst( 'img' ) || (el.getFirst( 'a' ) && el.getFirst( 'a' ).getFirst( 'img' ));`

### Expected result

Exception should not be thrown.

### Actual result

`Uncaught TypeError: Cannot read property 'getFirst' of null`  exception is being thrown.

## Other details

* Browser: Google Chrome 70, FireFox 62
* OS: Windows 10
* CKEditor version: ckeditor_4.10.1
* Installed CKEditor plugins: image2

engineering-this:

Bug Fix

### This PR contains

- [x] Unit tests
- [x] Manual tests

## What changes did you make?

Added an extra check in upcasting method, so it won't throw type error.

Closes #2506

### DIFF
--- a/plugins/image2/plugin.js
+++ b/plugins/image2/plugin.js
@@ -942,7 +942,8 @@
 
 			// No center wrapper has been found.
 			else if ( name == 'figure' && el.hasClass( captionedClass ) ) {
-				image = el.getFirst( 'img' ) || el.getFirst( 'a' ).getFirst( 'img' );
+				var anchor = el.getFirst( 'a' );
+				image = el.getFirst( 'img' ) || ( anchor && anchor.getFirst( 'img' ) );
 
 				// Upcast linked image like <a><img/></a>.
 			} else if ( isLinkedOrStandaloneImage( el ) ) {
--- a/plugins/image2/plugin.js
+++ b/plugins/image2/plugin.js
@@ -942,8 +942,10 @@
 
 			// No center wrapper has been found.
 			else if ( name == 'figure' && el.hasClass( captionedClass ) ) {
-				var anchor = el.getFirst( 'a' );
-				image = el.getFirst( 'img' ) || ( anchor && anchor.getFirst( 'img' ) );
+				image = el.find( function( child ) {
+					return child.name === 'img' &&
+						CKEDITOR.tools.array.indexOf( [ 'figure', 'a' ], child.parent.name ) !== -1;
+				}, true )[ 0 ];
 
 				// Upcast linked image like <a><img/></a>.
 			} else if ( isLinkedOrStandaloneImage( el ) ) {
--- a/tests/plugins/image2/manual/upcast.md
+++ b/tests/plugins/image2/manual/upcast.md
@@ -1,12 +1,9 @@
 @bender-tags: 4.11.2, bug, 2506
 @bender-ui: collapsed
-@bender-ckeditor-plugins: wysiwygarea, toolbar, image2, sourcearea
+@bender-ckeditor-plugins: wysiwygarea, toolbar, image2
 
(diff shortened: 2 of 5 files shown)


tests/plugins/image2/upcast.js
```diff
@@ -37,7 +37,7 @@
 				name: 'enterP',
 				data: '<p style="text-align:center">' +
 					'<img id="w1" src="_assets/foo.png" alt="foo" />' +
-				'</p>'
+					'</p>'
```
engineering-this: Just hit that autoformatting keystrong, although that's correct indentation so I'm keeping that.

Comandeer: Actually it's not correct, as nested `img` tag is on the same level as `p` now.

plugins/image2/plugin.js
```diff
@@ -942,7 +942,8 @@
 
 			// No center wrapper has been found.
 			else if ( name == 'figure' && el.hasClass( captionedClass ) ) {
-				image = el.getFirst( 'img' ) || el.getFirst( 'a' ).getFirst( 'img' );
+				var anchor = el.getFirst( 'a' );
+				image = el.getFirst( 'img' ) || ( anchor && anchor.getFirst( 'img' ) );
```
Comandeer: I wonder if [`find`](https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_htmlParser_element.html#method-find) wouldn't be more readable here:

```javascript
image = el.find( function( child ) {
	return child.name === 'img' &&
		CKEDITOR.tools.array.indexOf( [ 'figure', 'a' ], child.parent.name ) !== -1;
}, true )[ 0 ];
```

The issue I have with [`getFirst`](https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_htmlParser_element.html#method-getFirst) is the fact that it works in totally counter-intuitive way, changing behaviour when parameter is passed (bold's mine):

> Gets this element's first child. If condition is given, this method returns the first child **which satisfies that condition**.

Maybe the version with `find` is more verbose, yet also more explicit and easier to understand – without the need to consult the docs.

tests/plugins/image2/manual/upcast.md
```diff
@@ -0,0 +1,17 @@
+@bender-tags: 4.11.2, bug, 2506
+@bender-ui: collapsed
+@bender-ckeditor-plugins: wysiwygarea, toolbar, image2, sourcearea
+
+1. Open developer console.
```
Comandeer: Probably testing procedure could be simplified, e.g. by adding two buttons, which will be inserting empty `figure` and `figure > a` elements.

tests/plugins/image2/upcast.js
engineering-this: You are right. It's not correct as html markup it would be for regular string, however original one wasn't correct either, as `'</p>'` shouldn't be on same level as `data` property.

PR #2652, opened 2018-12-05
Fix for: table dialog validator fails when function getValue is defined in global scope

### LINKED ISSUE #2650: Validator fails when having functions defined in global scope
## Type of report

Bug

1. Open e.g. ckeditor-dev/samples/old/replacebycode.html sample and add below code above `replace` method.

```
window.getValue = function(){
	return false;
}
```
You can also attach external JavaScript file with below code inside 
```
function getValue(){
	return false;
}
```
2. Load sample in a browser and open table dialog
3. Click ok

### Expected result

Table gets inserted.

### Actual result

Table doesn't get inserted because the validator fails, claiming that row number must greater than 0 even though there is a default value of 3 present in the field.

## Other details

* Browser: Any
* OS: Any
* CKEditor version: tested in latest CKEditor 4.11.1

It seems that regex function in https://github.com/ckeditor/ckeditor-dev/blob/major/plugins/dialog/plugin.js#L3138 references window object.

I have only managed to reproduce this issue with Table dialog but other dialogs could be affected as well.

Sample JS File with global function: [testfile.zip](https://github.com/ckeditor/ckeditor-dev/files/2648566/testfile.zip)

engineering-this:

Bug fix

### This PR contains

- [X] Unit tests
- [X] Manual tests

## What changes did you make?

As `dialog.validate` methods are defined to be used with proper context I've updated `validatorNum ` helper in table dialog.

Closes #2650


engineering-this:
Removed `validatorNum`, it seems to work ok.

PR #2658, opened 2018-12-06
Fix for: Failing tests /tests/core/tools/absoluterectposition

### LINKED ISSUE #2566: Failing tests /tests/core/tools/absoluterectposition
## Type of report

Failing tests

Both tests from `/tests/core/tools/absoluterectposition` fails on iOS when run from dashboard. Works fine when tests are executed directly.

## Other details

* Browser: iOS
* CKEditor version: 4.11.0

engineering-this:

Fix for failing tests

### This PR contains

- [ ] Unit tests
- [ ] Manual tests

## What changes did you make?

Mocked `window.getScrollPosition` so test gives consistent results.

Closes #2566

### DIFF
--- a/tests/core/tools/absoluterectposition.js
+++ b/tests/core/tools/absoluterectposition.js
@@ -3,12 +3,23 @@
 ( function() {
 	'use strict';
 
-	var tests = {
+	var scrollPositionMock,
+		windowMock,
+		tests = {
 		setUp: function() {
+			var window = CKEDITOR.document.getWindow();
+
+			windowMock = sinon.stub( CKEDITOR.document, 'getWindow' );
+			windowMock.returns( window );
+
 			CKEDITOR.document.getBody().setStyle( 'height', '5000px' );
+
+			scrollPositionMock = sinon.stub( window, 'getScrollPosition' );
+			scrollPositionMock.returns( { x: 0, y: 0 } );
 		},
 		tearDown: function() {
-			window.scrollBy( 0, -CKEDITOR.document.getWindow().getScrollPosition().y );
+			scrollPositionMock.restore();
+			windowMock.restore();
 		},
 		'test absolute rect inline': function() {
 			var rect = {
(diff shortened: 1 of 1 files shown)


tests/core/tools/absoluterectposition.js
```diff
@@ -3,12 +3,23 @@
...
+		tests = {
 		setUp: function() {
+			var window = CKEDITOR.document.getWindow();
+
+			windowMock = sinon.stub( CKEDITOR.document, 'getWindow' );
+			windowMock.returns( window );
+
 			CKEDITOR.document.getBody().setStyle( 'height', '5000px' );
+
+			scrollPositionMock = sinon.stub( window, 'getScrollPosition' );
```
mlewand: You could simplify it by stubbing getScrollPosition directly in our window wrapper's prototype, like so:

> scrollPositionMock = sinon.stub( CKEDITOR.dom.window.prototype, 'getScrollPosition' );

tests/core/tools/absoluterectposition.js
```diff
@@ -32,7 +43,7 @@
 
 			assertRect( absoluteRect, expected );
 
-			window.scrollBy( 0, 100 );
+			scrollPositionMock.returns( { x: 0, y: 100 } );
```
mlewand: Since you're explicitly setting the return value of a stub for each test case, I'd favor the entire mock creation to happen in the test case, and not in `setUp` method.

So you assign it like `scrollPositionMock = sinon.stub( window, 'getScrollPosition' ).returns( { x: 0, y: 100 } )`.

Restoring in `tearDown` is perfectly fine, but you'd have to adjust it for case if `scrollPosition` is `null` and also nullify it after it has been restored.

Reason for this is that someone adding a new test case won't be surprised by garbage `undefined` being returned by `getScrollPosition`.

tests/core/tools/absoluterectposition.js
```diff
@@ -3,14 +3,20 @@
 ( function() {
 	'use strict';
 
-	var tests = {
+	var scrollPositionMock,
+		tests = {
 		setUp: function() {
```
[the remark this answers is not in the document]
mlewand: We can not make exception like that. We need to follow agreed code style. I have fixed that in 1d286993f.

PR #2667, opened 2018-12-11
Exposed CKEDITOR.ui.balloonToolbar.reposition

### LINKED ISSUE #1496: Balloon Toolbar context should expose a way to reposition the context

Feature request

While working on [Easy Image plugin](https://github.com/ckeditor/ckeditor-dev/issues/932) we had a requirement to reposition the balloon under in a certain scenario.

Here the case was that the image was loaded asynchronously, and the default balloon toolbar matching happened too early (when the image size was not yet known) thus the balloon was mispositioned.

In turn we used Balloon Toolbar internal methods to do that, however there should be a nice method to force refresh of a toolbar.

engineering-this:

New Feature

### This PR contains

- [x] Unit tests
- [x] Manual tests

## What changes did you make?

I've added method to reposition balloon toolbar.

Closes #1496

### DIFF
--- a/plugins/balloontoolbar/plugin.js
+++ b/plugins/balloontoolbar/plugin.js
@@ -125,6 +125,16 @@
 	};
 
 	/**
+	 * Repositions the balloon toolbar, pointing to the previously attached `element`.
+	 *
+	 * @since 4.12.0
+	 * @member CKEDITOR.ui.balloonToolbar
+	 */
+	CKEDITOR.ui.balloonToolbar.prototype.reposition = function() {
+		this._view.reposition();
+	};
+
+	/**
 	 * Adds an item to the balloon toolbar.
 	 *
 	 * @param {String} name The menu item name.
@@ -665,9 +675,7 @@
 				this._detachListeners();
 
 				function attachListener() {
-					this.attach( this._pointedElement, {
-						focusElement: false
-					} );
+					this.reposition( { focusElement: false } );
 				}
 
 				this._listeners.push( this.editor.on( 'change', attachListener, this ) );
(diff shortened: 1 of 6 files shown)


tests/plugins/balloontoolbar/positioning.js
```diff
@@ -160,6 +161,28 @@
...
+					options = { foo: 'bar' },
+					spy;
+
+				balloonToolbar.attach( markerElement );
+
+				spy = sinon.spy( balloonToolbar, 'attach' );
+
+				balloonToolbar.reposition( options );
+
+				assert.areSame( markerElement, spy.args[ 0 ][ 0 ] );
```
Comandeer: Probably the element is the same, yet tests fail due to the fact that element's wrappers are different. [`equals`](https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_dom_element.html#method-equals) should be used here.

tests/plugins/balloontoolbar/manual/reposition.html
```diff
@@ -0,0 +1,60 @@
...
+					bold: new CKEDITOR.ui.button( {
+						label: 'test',
+						command: 'bold'
+					} )
+				} );
+
+				panel.attach( square );
+
+				CKEDITOR.tools.array.forEach( CKEDITOR.document.find( '.control' ).toArray(), function( input ) {
+					input.on( 'change', function() {
```
Comandeer: In newer browsers we could use `input` instead of `change`, to dynamically move square when interacting with slider.

plugins/balloontoolbar/plugin.js
```diff
@@ -665,9 +675,7 @@
 				this._detachListeners();
 
 				function attachListener() {
-					this.attach( this._pointedElement, {
-						focusElement: false
-					} );
+					this.reposition( { focusElement: false } );
```
Comandeer: You pass parameter to the `reposition` method here, yet it does not accept any parameters. IMO it should, as it should pass options to the `attach`.

tests/plugins/balloontoolbar/positioning.js
```diff
@@ -160,6 +161,28 @@
...
+				assert.areSame( options, spy.args[ 0 ][ 1 ] );
```
Comandeer: This statement also can fail as ATM `reposition` does not support options.

plugins/balloontoolbar/plugin.js
engineering-this: I wonder if reposition should really accept options, as `balloonToolbar.attach` doesn't.

PR #2682, opened 2018-12-17
Remove redundant code from emoji plugin

### LINKED ISSUE #2592: Remove redundant code in Emoji listeners / CSS code
## Type of report

Task

This is a cleaning up after #2587.

There are two issues in code that should be corrected, but we want to avoid extra changes minutes before the release.

* Two functions do the same thing, fetching the `event.data.getTarget().getAscendant( 'li', true );`, `activeElement` should be simply passed. Also given current implementation [this check](https://github.com/ckeditor/ckeditor-dev/blob/65f37687dca04667470e72427b0d839c054ec93d/plugins/emoji/plugin.js#L514) seems to be redundant.
* Since we're using `a[href]` again, there's [no longer a need to explicitly define `cursor: pointer`](https://github.com/ckeditor/ckeditor-dev/blob/major/plugins/emoji/skins/default.css#L159).

## Other details

* Browser: Any
* OS: Any
* CKEditor version: 4.11.1

msamsel:

Task.

### This PR contains

- [ ] Unit tests
- [x] Manual tests

## What changes did you make?

- remove unnecessary css for emoji
- pass element to function instead of reading it twice

Close #2592

### DIFF
--- a/tests/plugins/emoji/manual/dropdown/pointercursorinnavigation.md
+++ b/tests/plugins/emoji/manual/dropdown/pointercursorinnavigation.md
@@ -1,9 +1,7 @@
-@bender-tags: 4.11.2, bug, emoji, 2592
+@bender-tags: 4.11.3, bug, emoji, 2592
 @bender-ckeditor-plugins: wysiwygarea, toolbar, elementspath, sourcearea, emoji, clipboard, undo, stylescombo, format
 @bender-ui: collapsed
 @bender-include: ../../_helpers/tools.js
-
-## Test Case 1:
 
 1. Open emoji dropdown.
 2. Move cursor over group navigation at top of dropdown.
@@ -12,10 +10,3 @@ Cursor is changed to pointer, which indicates that element is a link.
 3. Move cursor over emoji in main section.
 ### Expected:
 Cursor is changed to pointer, which indicates that element is a link.
-
-## Test Case 2:
-
-1. Open emoji dropdown.
-2. Click into navigation group.
-### Expected:
-Emoji list is scrolled to proper position. First element of clicked group is selected.
--- a/tests/plugins/emoji/manual/dropdown/pointercursorinnavigation.md
+++ b/tests/plugins/emoji/manual/dropdown/pointercursorinnavigation.md
@@ -5,8 +5,10 @@
 
 1. Open emoji dropdown.
 2. Move cursor over group navigation at top of dropdown.
-### Expected:
-Cursor is changed to pointer, which indicates that element is a link.
-3. Move cursor over emoji in main section.
-### Expected:
-Cursor is changed to pointer, which indicates that element is a link.
+
+	### Expected
+	Cursor is changed to pointer, which indicates that element is a link.
+3. Move cursor over an emoji in the main section.
+
+	### Expected
+	Cursor is changed to pointer, which indicates that element is a link.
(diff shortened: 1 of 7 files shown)


tests/plugins/emoji/manual/dropdown/smallrefactoring.md
```diff
@@ -0,0 +1,21 @@
+@bender-tags: 4.11.2, bug, emoji, 2592
```
engineering-this: I'd rename this test.
Tests are supposed to check if code works, not if it looks nice, so refactoring doesn't sound like something that should be mentioned in test name. I understand that these cases are unrelated and it's hard to find a generic name that fits both, so maybe consider splitting them for separate test cases? It will be easier to give names to these cases as separate tests.

Also test steps from the first case can't be done on mobiles, so you should either add proper information in a description, or separate cases so the first can be ignored on mobiles.

CHANGES.md
```diff
@@ -10,6 +10,7 @@ Fixed Issues:
 * [#2514](https://github.com/ckeditor/ckeditor-dev/issues/2403): Fixed: Pasting table data into inline editor initialized inside a table with [Table Selection](https://ckeditor.com/cke4/addon/tabletools) plugin inserts pasted content into a wrapping table.
 * [#2451](https://github.com/ckeditor/ckeditor-dev/issues/2451): Fixed: The [Remove Format](https://ckeditor.com/cke4/addon/removeformat) changes selection.
 * [#2546](https://github.com/ckeditor/ckeditor-dev/issues/2546): Fixed: Separator in toolbar moves when buttons are focused.
+* [#2592](https://github.com/ckeditor/ckeditor-dev/issues/2592): Fixed: Remove redundant code in [Emoji](https://ckeditor.com/cke4/addon/emoji) plugin.
```
mlewand: That was a task, so it make sense to put it in other changes section. It brings no value in terms of fixing issues for the end user.

tests/plugins/emoji/manual/dropdown/smallrefactoring.md
mlewand: For sure the test name is really bad, _smallrefactoring_ tells you nothing about the test. Similarly to `thatOneThing.md` or `simplecase.md`. Always avoid names like that.

As mentioned by @engineering-this giving it a proper name would be tricky. Generic test will also be problematic to match this behavior. Idea for making two micro manual tests sounds really appealing.

tests/plugins/emoji/manual/dropdown/pointercursorinnavigation.md
```diff
@@ -1,21 +1,12 @@
...
 3. Move cursor over emoji in main section.
 ### Expected:
 Cursor is changed to pointer, which indicates that element is a link.
-
-## Test Case 2:
-
-1. Open emoji dropdown.
-2. Click into navigation group.
-### Expected:
-Emoji list is scrolled to proper position. First element of clicked group is selected.
```
engineering-this: Commit message says it should be splitted in two, but I see only one test.
Am I missing something, or this part is just removed?

msamsel: I've add tag `4.11.3` to old test which covers this behaviour. (`tests/plugins/emoji/manual/dropdown/scrollpagebug.md`). I feel that is better to reuse already existing test, rather than write new one.
However I see that test had invalid description so I add more precise info to expected and unexpected value.

tests/plugins/emoji/manual/dropdown/scrollpagebug.md
```diff
@@ -1,11 +1,11 @@
...
 @bender-include: ../../_helpers/tools.js
 
 1. Open emoji dropdown.
 2. Click into emoji group.
-### Expected
-Page is not scrolled down.
 ### Unexpected
-Page scrolls down.
+Page is not scrolled down.
+### Expected
```
mlewand: Actually with your change the manual test has first unexpected value and then expected. That's an uncommon notation in manual tests.

I'll flip the order during the review.

mlewand:
Rebased onto latest `major`.

PR #2683, opened 2018-12-17
Added CKEDITOR.config.image2_maxSize

### LINKED ISSUE #2048: Image2 could have limiter for resizing image to specified maximum
## Type of report

Feature request

Originally reported here: https://dev.ckeditor.com/ticket/12701

Image2 could have an extra setting or field in dialog for defining maximum resize size. If doing that inside the dialog, we could have a checkbox (Limit resizer to current image size) or 2 fields for entering max width/height manually. We could also have a configuration option with predefined values e.g. `image_size`, `image_size+20`, `200x300`.

### LINKED ISSUE #2672: image2: resizing to minimum size doesn't update dialogs values
## Type of report

Bug

1. Open test `http://localhost:1030/tests/plugins/image2/manual/resizer`
2. Scroll to last editor.
3. Resize image to minimum size.
4. Double click on image to open dialog.

### Expected result

- Dialogs fields match image properties.

### Actual result

- `width` and `height` are unset.

## Other details

Look at code:
https://github.com/ckeditor/ckeditor-dev/blob/7c937e875c35b9bd333fcc5de9c0910a2584ad34/plugins/image2/plugin.js#L1317-L1322

This is the part of `mousemove` handler. When mouse moves decreasing image size, attributes of img element are changed, but `widget.data` is updated in `mouseup` handler only if `updateData` is set to true. So in this scenario image size is changed, but the second one of dimensions is lower than allowed `updateData` is overwritten with `false` and thus data is not updated, and dialog shows previews value.

* Browser: …
* OS: …
* CKEditor version: …

### LINKED ISSUE #2714: waitForImage should call callback if image is already loaded
## Type of report

Task

Test tool `waitForImage` calls callback on `load` event, but when image is already loaded this event will never happen, and some browsers might load images earlier than others.

https://github.com/ckeditor/ckeditor-dev/blob/master/tests/plugins/uploadfile/_helpers/waitForImage.js

I propose to add if branch:

```
else if ( image.$.complete )  {
    callback()
}
```

## Other details

* Browser: …
* OS: …
* CKEditor version: …

engineering-this:

New feature + Bugfix

### This PR contains

- [X] Unit tests
- [X] Manual tests

## What changes did you make?

- Added new config option `CKEDITOR.config.image2_maxSize` that limits maximum size that image can be resized with resizer.
- Fixed resizer `mousemove` listener to update widget data with last correct value which is same as one set on element.

Closes #2048 
Closes #2672
Closes #2714

### DIFF
--- a/plugins/image2/plugin.js
+++ b/plugins/image2/plugin.js
@@ -1160,6 +1160,15 @@
 		resizer.on( 'mousedown', function( evt ) {
 			var image = widget.parts.image,
 
+				// Don't update attributes if less than 15.
+				// This is to prevent images to visually disappear.
+				min = {
+					width: 15,
+					height: 15
+				},
+
+				max = getMaxSize(),
+
 				// "factor" can be either 1 or -1. I.e.: For right-aligned images, we need to
 				// subtract the difference to get proper width, etc. Without "factor",
 				// resizer starts working the opposite way.
@@ -1315,13 +1324,9 @@
 					}
 				}
 
-				// Don't update attributes if less than 10.
-				// This is to prevent images to visually disappear.
-				if ( newWidth >= 15 && newHeight >= 15 ) {
-					image.setAttributes( { width: newWidth, height: newHeight } );
-					updateData = true;
-				} else {
-					updateData = false;
+				if ( isAllowedSize() ) {
+					updateData = { width: newWidth, height: newHeight };
+					image.setAttributes( updateData );
 				}
 			}
 
@@ -1338,7 +1343,7 @@
 				resizer.removeClass( 'cke_image_resizing' );
 
 				if ( updateData ) {
-					widget.setData( { width: newWidth, height: newHeight } );
+					widget.setData( updateData );
 
 					// Save another undo snapshot: after resizing.
 					editor.fire( 'saveSnapshot' );
@@ -1346,6 +1351,27 @@
 
 				// Don't update data twice or more.
 				updateData = false;
+			}
+
+			function getMaxSize() {
+				var maxSize = editor.config.image2_maxSize,
+					natural;
+
+				if ( !maxSize ) {
+					return null;
+				}
+
+				maxSize = CKEDITOR.tools.copy( maxSize );
+				natural = CKEDITOR.plugins.image2.getNatural( image );
+
+				maxSize.width = Math.max( maxSize.width === 'naturalWidth' ? natural.width : maxSize.width, min.width );
+				maxSize.height = Math.max( maxSize.height === 'naturalHeight' ? natural.height : maxSize.height, min.width );
+
+				return maxSize;
+			}
+
+			function isAllowedSize() {
+				return newWidth >= min.width && newWidth <= max.width && newHeight >= min.height && newHeight <= max.height;
 			}
 		} );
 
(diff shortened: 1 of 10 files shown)


tests/plugins/image2/resizehandler.js
```diff
@@ -0,0 +1,145 @@
...
+	};
+
+	bender.test( bender.tools.createTestsForEditors( CKEDITOR.tools.objectKeys( bender.editors ), tests ) );
+
+	function testResize( options ) {
+		return function( editor, bot ) {
+			bot.setData( '<img src="%BASE_PATH%/_assets/logo.png">', function() {
+				var image = editor.editable().findOne( 'img' );
+
+				if ( CKEDITOR.env.chrome ) {
```
Comandeer: Is this branch necessary? For the sake of the simplicity, Chrome can also wait for image.

engineering-this: Actually at this point Chrome unlike other browsers has image loaded, and `waitForImage` triggers on `load` event which won't happen in Chrome (#2714).

plugins/image2/plugin.js
```diff
@@ -1157,6 +1157,15 @@
 		resizer.on( 'mousedown', function( evt ) {
 			var image = widget.parts.image,
 
+				// Don't update attributes if less than 15.
+				// This is to prevent images to visually disappear.
+				min = {
```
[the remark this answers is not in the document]
engineering-this: I don't think it's important to allow setting min size for resizer. Also it wasn't mentioned in original issue.

tests/plugins/image2/manual/dialogafterresize.md
```diff
@@ -0,0 +1,15 @@
+@bender-tags: 4.12.0, feature, bug, 2672
```
Comandeer: After splitting test into two separate ones, declaring them as `feature` & `bug` is no longer needed – one is for feature and one is for bug.

tests/plugins/image2/manual/dialogafterresize.md
```diff
@@ -0,0 +1,15 @@
...
+@bender-ui: collapsed
+@bender-ckeditor-plugins: wysiwygarea, toolbar, image2, sourcearea
+
+1. Use resize handler to shrink image to minimum size.
+
+1. Open dialog for this image.
+
+## Expected:
+
+Dialog fields are populated with correct width/height.
```
Comandeer: It's not clear, which values are correct and which are not.

tests/plugins/image2/resizehandler.js
```diff
@@ -0,0 +1,138 @@
...
+				image2_maxSize: {
+					width: 'natural',
+					height: 'natural'
+				}
+			}
+		}
+	};
+
+	var tests = {
+		//2672
```
Comandeer: Our convention for such comments is `// (#<issue number>)`.

tests/plugins/uploadfile/uploadfile.js
```diff
@@ -226,5 +226,21 @@ bender.test( {
 		loader.abort();
 
 		assert.areSame( 'abort', loader.status, 'Loader status' );
+	},
+
+	// #2714
+	'test waitForImage when image is loaded': function() {
```
Comandeer: This test should not be here, as it's connected to the test helper, not to the plugin itself. In fact helpers usually don't have tests.

engineering-this:
- Updated manual test.
- `waitForImage` unit test extracted to separate file.
- Made `waitForImage` callback call asynchronous when image is loaded early.

PR #2689, opened 2018-12-18
Improve order of emoji in autocomplete

### LINKED ISSUE #2527: Filtered emoji order should be improved
## Type of report

Bug

After fixing #2181 the sorting order of emoji has changed. Previously it was better.

1. Open editor with Emoji plugin.
2. Type `:smiling`.

### Expected result

Emojis whose id starts with smiling should appear first.

Old behavior was:

Previous (4.10.1) emoji sorting

### Actual result

Current emoji sorting order

## Other details

* Browser: Any
* OS: Any
* CKEditor version: 4.11.0

msamsel:

Bug fix

### This PR contains

- [x] Unit tests
- [x] Manual tests

## What changes did you make?

* Add sort algorithm, which prioritize emoji which starts from searched string

Close #2527

### DIFF
--- a/plugins/emoji/plugin.js
+++ b/plugins/emoji/plugin.js
@@ -630,17 +630,14 @@
 							// Comparing lowercased strings, because emoji should be case insensitive (#2167).
 							return item.id.toLowerCase().indexOf( emojiName ) !== -1;
 						} ).sort( function( a, b ) {
-							var aIndex = a.id.indexOf( emojiName ),
-								bIndex = b.id.indexOf( emojiName );
+							var aIndex = a.id.substr( 1 ).indexOf( emojiName ),
+								bIndex = b.id.substr( 1 ).indexOf( emojiName );
 
-							if ( aIndex === bIndex ) {
-								if ( a.id > b.id ) {
-									return 1;
-								} else {
-									return -1;
-								}
+							// True when left and right side are oposite.
+							if ( aIndex === 0 ^ bIndex === 0 ) {
+								return aIndex === 0 ? -1 : 1;
 							} else {
-								return aIndex - bIndex;
+								return a.id > b.id ? 1 : -1;
 							}
 						} );
 					callback( data );
--- a/plugins/emoji/plugin.js
+++ b/plugins/emoji/plugin.js
@@ -627,15 +627,14 @@
 				function dataCallback( matchInfo, callback ) {
 					var emojiName = matchInfo.query.substr( 1 ).toLowerCase(),
 						data = arrTools.filter( emojiList, function( item ) {
-							// Comparing lowercased strings, because emoji should be case insensitive (#2167).
+							// Comparing lowercase strings, because emoji should be case insensitive (#2167).
 							return item.id.toLowerCase().indexOf( emojiName ) !== -1;
 						} ).sort( function( a, b ) {
-							var aIndex = a.id.substr( 1 ).indexOf( emojiName ),
-								bIndex = b.id.substr( 1 ).indexOf( emojiName );
+							var aStartsWithEmojiName = !a.id.substr( 1 ).indexOf( emojiName ),
+								bStartsWithEmojiName = !b.id.substr( 1 ).indexOf( emojiName );
 
-							// True when left and right side are oposite.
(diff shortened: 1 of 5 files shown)


msamsel:
> Or go even further and return only matches than are made of words which starts with given string.

As far as I know we don't want to go in that direction. There should be displayed match which exist inside emoji name.

About mentioned cases there always will be some benefits and drawbacks of each approach, which will be more suitable to some cases and less to others.

Option 1:
1. Emoji which starts from given phrase sort at the beginning. (If there are multiple match, then sort them alphabetically)
2. All other emojis sort alphabetically below. (Don't care if match was last 2-3 letters or not)

Options 2 (current approach):
1. Emoji which has searched phrased closer to the beginning are sorted with higher priority. (If there are multiple match, them sort them alphabetically)

Options 3:
1. Emoji which starts from given phrase are sorted alphabetically. (If there are multiple match, then sort it according to occurrence in input file)
2. All other emoji sort according to its occurrence in input file

Maybe some other mixed of proposed solution, but I feel like every of this approach might not be suitable to every possible case. And you always will be able to find some edge case, which will look odd.

msamsel:
1. Proposed solution does not work as intend. Take a look on screenshot for emoji autocomplete on your branch:
screen shot 2019-01-15 at 15 18 26
I also compare performance of executing `dataCallback` function between my and your approach, and result are the same (21ms for first search and 8-9ms for the same search later on. I've made 5-10 repeats for each.)
screen shot 2019-01-15 at 15 47 00screen shot 2019-01-15 at 15 43 47
I'm not convinced to use approach which divide id into separate words. Let's stick to the solution which was originally proposed and which is source of this change. I've made 2 groups here 1st group which starts from given string, 2nd group which contain given string. Both groups sort alphabetically inside them.

I've made a change in a moment to reflect this solution.

Separating words and searching inside them wasn't scope of reported issue and I feel like such change should be report as feature request and discussed separately.

plugins/emoji/plugin.js
```diff
@@ -629,6 +629,16 @@
 						data = arrTools.filter( emojiList, function( item ) {
 							// Comparing lowercased strings, because emoji should be case insensitive (#2167).
 							return item.id.toLowerCase().indexOf( emojiName ) !== -1;
+						} ).sort( function( a, b ) {
+							var aIndex = a.id.substr( 1 ).indexOf( emojiName ),
+								bIndex = b.id.substr( 1 ).indexOf( emojiName );
+
+							// True when left and right side are oposite.
```
mlewand:
```suggestion
							// True when left and right side are opposite.
```

mlewand: Still this comment will not be precise. This condition is only true if *only one* of sorted items begins with emojiName.

I'd understand: "left and right side are opposite" like if one has the match at the beginning, and the other at the end.

plugins/emoji/plugin.js
```diff
@@ -629,6 +629,16 @@
...
+							if ( aIndex === 0 ^ bIndex === 0 ) {
```
mlewand: This could be simplified down to `Boolean( aIndex ) != Boolean( bIndex )`. Reader will not have to google for bit operators and refresh the memory. 🙂

tests/plugins/emoji/basic.js
```diff
@@ -183,6 +183,19 @@
 				assert.areSame( 1, autocomplete.model.data.length, 'Emoji result contains more than one result' );
 				objectAssert.areEqual( { id: ':collision:', symbol: '💥' }, { id: autocomplete.model.data[ 0 ].id, symbol: autocomplete.model.data[ 0 ].symbol }, 'Emoji result contains wrong result' );
 			} );
+		},
+
+		// (#2527)
+		'test emoji autocomplete is sorted in proper order': function( editor, bot ) {
```
mlewand: I'd actually like to check the entire data set. Here you're checking just the first value.

I'll modify this unit test as a part of review.

plugins/emoji/plugin.js
msamsel: I've move even further and set boolean value of variable instead of preserving its index. Index is not used anywhere. Now sort condition should looks much more neat.

PR #2705, opened 2018-12-28
Added CKEDITOR.tools.array.find

### LINKED ISSUE #2700: Add CKEDITOR.tools.array.find method
## Type of report

Feature request

This method could simplify some of our code, it could be used by #2698

```
CKEDITOR.tools.array.find( array, fn, [ thisArg ] )
```

Specs: https://www.ecma-international.org/ecma-262/6.0/#sec-array.prototype.find

engineering-this:

New feature

### This PR contains

- [X] Unit tests
- [X] Manual tests

## What changes did you make?

Added new array tool: `CKEDITOR.tools.array.find`

Closes #2700

### DIFF
--- a/core/tools.js
+++ b/core/tools.js
@@ -1951,6 +1951,40 @@
 			},
 
 			/**
+			 * Returns the first element in the array for which the given callback `fn` returns true.
+			 * If there is no such element `undefined` is returned.
+			 *
+			 * ```javascript
+			 * var array = [ 'foo', 'bar', 'baz', 'faz' ];
+			 *
+			 * CKEDITOR.tools.array.find( array, function( item ) {
+			 *     return item.indexOf( 'az' ) > -1;
+			 * } ); // returns 'baz'.
+			 *
+			 * ```
+			 *
+			 * @param {Array} array An array to be iterated over.
+			 * @param {Function} fn A function with the signature `(a, index, array) -> b`.
+			 * @param [thisArg=undefined] The context object for `fn`.
+			 * @return {*}
+			 * @member CKEDITOR.tools.array
+			 * @since 4.12.0
+			 */
+			find: function( array, fn, thisArg ) {
+				var length = array.length,
+					i = 0;
+
+				while ( i < length ) {
+					if ( fn.call( thisArg, array[ i ], i, array ) ) {
+						return array[ i ];
+					}
+					i++;
+				}
+
+				return undefined;
+			},
+
+			/**
 			 * Iterates over every element in the `array`.
 			 *
 			 * @param {Array} array An array to be iterated over.
--- a/core/tools.js
+++ b/core/tools.js
@@ -1951,22 +1951,23 @@
 			},
 
 			/**
-			 * Returns the first element in the array for which the given callback `fn` returns true.
-			 * If there is no such element `undefined` is returned.
+			 * Returns the first element in the array for which the given callback `fn` returns `true`.
 			 *
 			 * ```javascript
-			 * var array = [ 'foo', 'bar', 'baz', 'faz' ];
+			 * var array = [ 1, 2, 3, 4 ];
 			 *
 			 * CKEDITOR.tools.array.find( array, function( item ) {
-			 *     return item.indexOf( 'az' ) > -1;
-			 * } ); // returns 'baz'.
-			 *
+			 *     return item > 2;
+			 * } ); // returns 3.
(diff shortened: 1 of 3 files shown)


core/tools.js
```diff
@@ -1950,6 +1950,40 @@
...
+			 * var array = [ 'foo', 'bar', 'baz', 'faz' ];
+			 *
+			 * CKEDITOR.tools.array.find( array, function( item ) {
+			 *     return item.indexOf( 'az' ) > -1;
+			 * } ); // returns 'baz'.
+			 *
+			 * ```
+			 *
+			 * @param {Array} array An array to be iterated over.
+			 * @param {Function} fn A function with the signature `(a, index, array) -> b`.
```
mlewand: JSDuck allows you to explicitly specify callback (`fn` in this case) signature. You don't have to do it in a _custom_ way.

See https://github.com/senchalabs/jsduck/wiki/%40param#callback-parameters for more info.

core/tools.js
```diff
@@ -1950,6 +1950,40 @@
...
+			 * @param [thisArg=undefined] The context object for `fn`.
+			 * @return {*}
```
mlewand: We use `@return*s*`

core/tools.js
```diff
@@ -1950,6 +1950,40 @@
 				return ret;
 			},
 
+			/**
+			 * Returns the first element in the array for which the given callback `fn` returns true.
+			 * If there is no such element `undefined` is returned.
```
mlewand: Actually I'd see this part in `@returns` part as otherwise it is empty, which is not optimal.

So it would be something like:
```
* @returns {*} First matched value, `undefined` otherwise.
```

tests/core/tools/array.js
```diff
@@ -151,8 +151,35 @@
...
+				results.push( item );
+
+				assert.areSame( arr.indexOf( item ), index, 'Index argument should match item index.' );
+
+				return false;
+			} );
 
+			assert.isUndefined( ret, 'Undefined should be returned.' );
+
+			assert.isTrue( CKEDITOR.tools.objectCompare( arr, results ), 'Each array item should be iterated.' );
```
mlewand: There's a dedicated `arrayAssert` object that contains better asserting methods for what you're trying to do here.

tests/core/tools/array.js
```diff
@@ -151,8 +151,35 @@
...
+
+			ret = this.array.find( arr, function( item, index ) {
+				results.push( item );
+
+				assert.areSame( arr.indexOf( item ), index, 'Index argument should match item index.' );
+
+				return false;
+			} );
 
+			assert.isUndefined( ret, 'Undefined should be returned.' );
```
mlewand: `Undefined should be returned` it is not a good practice to put expected value if you use proper assertion.

In this case you used the proper assertion, `assert.isUndefined()` which already tells exactly what value is expected. If assertion fails, there is an information that `undefined` was expected. And this should be **the only** source of truth.

Duplicating this information in assertion msg you add a risk that someone will change the method, but will leave msg unchanged.

Simply make it value agnostic, but keep the information what has failed, like: "Proper value is returned" or even simpler as we do in many other places "Returned value".

tests/core/tools/array.js
```diff
@@ -151,8 +151,35 @@
...
+
+		'test array.find': function() {
+			var arr = [ 'foo', 'bar', 'baz', 1, 2, 3 ],
+				results = [],
+				ret;
+
+			ret = this.array.find( arr, function( item, index ) {
+				results.push( item );
+
+				assert.areSame( arr.indexOf( item ), index, 'Index argument should match item index.' );
```
mlewand: Will fail on IE8.

tests/core/tools/array.js
```diff
@@ -151,8 +151,35 @@
 				assert.areSame( 0, index );
 				arrayAssert.itemsAreSame( [ 1234 ], array );
 			} );
-		}
+		},
+
+		'test array.find': function() {
```
mlewand: Technically you're doing two test cases in one. Please, split it into two clearly defined test cases.

tests/core/tools/manual/find.html
```diff
@@ -0,0 +1,27 @@
+<style>
```
mlewand: This is purely API addition and should not be tested manually, as it's possible to test it 100% using automatic testing.

PR #2719, opened 2019-01-03
Added htmlParser.element.findOne

### LINKED ISSUE #2698: Add htmlParser.element.findOne method
## Type of report

Task

There are some situations, in which we want to get only the first element that matches all criteria. We can use [`find`](https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_htmlParser_element.html#method-find) and fetch only the first element from the returned array, but DOM counterpart of `htmlParser.element` API has dedicated method for it, [`findOne`](https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_dom_element.html#method-findOne). It'd be nice to introduce it also to `htmlParser`.

engineering-this:

New feature

### This PR contains

- [x] Unit tests
- [ ] Manual tests

## What changes did you make?

- Added `CKEDITOR.htmlParser.element.findOne`
- Refactored Paste From Word filter function `removeSymbolText`.

Closes #2698.


tests/core/htmlparser/element.js
```diff
@@ -16,6 +16,14 @@ function write( el ) {
 	return writer.getHtml();
 }
 
+var testHtml = '<body>' +
+		'<p>Foo' +
+			'<div>Baz' +
```
Comandeer: This HTML is incorrect: `div` can't be nested in `p` tag.

core/htmlparser/element.js
```diff
@@ -470,6 +470,38 @@ CKEDITOR.htmlParser.cssStyle = function() {
...
+		 * } ); // Will return first child which is a span or a strong.
+		 * ```
+		 *
+		 * @param {String/Function} criteria Tag name or evaluator function.
+		 * @param {CKEDITOR.htmlParser.node} criteria.child The currently iterated child.
+		 * @returns {CKEDITOR.htmlParser.node} First matched child, `undefined` otherwise.
+		 */
+		findOne: function( criteria ) {
+			var findMethod = typeof criteria === 'function' ? criteria : findByName,
+				nestedMatch,
```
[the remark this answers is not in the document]
engineering-this: Yes, but I'm using it inside next declaration, so moving it to the end would break linter rule which requires variable declaration before it is used.

core/htmlparser/element.js
```diff
@@ -470,6 +470,38 @@ CKEDITOR.htmlParser.cssStyle = function() {
...
+				match = CKEDITOR.tools.array.find( this.children, function( child ) {
+					if ( findMethod( child ) ) {
+						return true;
+					} else if ( child.children && child.findOne ) {
```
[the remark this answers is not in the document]
engineering-this: We need `else if` branch, because child can be `CKEDITOR.htmlParser.text` which doesn't have property `children` or method `findOne`.

engineering-this:
- Changed returned type to `null` when no match is found.
- Extracted `findOne` tests to separate file, so they are much better readable with html string right above TCs. Also reformatted fixture, so each node is in new line, except for elements that contain only one text.
- Corrected changelog, and other small things.

plugins/pastefromword/filter/default.js
```diff
@@ -1172,25 +1172,21 @@
...
-					node.value = node.value.replace( symbol, '' );
-
-					if ( node.parent.getHtml().match( /^(\s|&nbsp;)*$/ ) ) {
-						removed = node.parent !== element ? node.parent : null;
-					} else if ( !node.value ) {
-						removed = node;
-					}
+				node = element.findOne( function( node ) {
+					return node.value && node.value.indexOf( symbol ) > -1;
+				}, true ),
```
Comandeer: Code style issue – there should be one more indentation level.

plugins/pastefromword/filter/default.js
```diff
@@ -1172,25 +1172,21 @@
 
 		removeSymbolText: function( element ) { // ...from a list element.
 			var symbol = element.attributes[ 'cke-symbol' ],
-				removed,
-				matched;
-
-			element.forEach( function( node ) {
-				// Since symbol may contains special characters we use `indexOf` (instead of RegExp) which is sufficient (#877).
-				// Stop after after first replaced string (#2690).
```
Comandeer: Why these comments are removed? They still seem to be accurate.

tests/core/htmlparser/elementfindone.js
```diff
@@ -0,0 +1,44 @@
+/* bender-tags: editor */
+
+( function() {
+	'use strict';
+
+	var testElement = CKEDITOR.htmlParser.fragment.fromHtml(
```
Comandeer: I wonder if it wouldn't be more readable to move the HTML into separate file, `elementfindone.html`.

tests/core/htmlparser/elementfindone.js
```diff
@@ -0,0 +1,44 @@
...
+			assert.areSame( testElement.children[ 0 ].children[ 1 ], testElement.findOne( 'p', true ) );
+		},
+
+		'test findOne by function': function() {
+			assert.areSame( testElement.children[ 1 ],
+				testElement.findOne( function( element ) {
+					return element.attributes && element.attributes[ 'class' ] === 'faz';
+				} )
+			);
+			assert.areSame( testElement.children[ 0 ].children[ 1 ].children[ 1 ],
```
Comandeer: In such tests it'd be nice to add some fail message to know, which elements are not equal.

plugins/pastefromword/filter/default.js
engineering-this: Adding indentation level would make closing bracket higher indentation than opening, shouldn't they be the same?

Comandeer: It's a very similar case to objects in such lists and they are indented with that extra tab: https://github.com/ckeditor/ckeditor-dev/blob/1f365357c043a74e69ad21f32c829929d8610106/plugins/tableselection/plugin.js#L893-L901

plugins/pastefromword/filter/default.js
```diff
@@ -1172,25 +1172,23 @@
...
-				matched;
+				// Find the first element which contains symbol to be replaced (#2690).
+				node = element.findOne( function( node ) {
+					return node.value && node.value.indexOf( symbol ) > -1;
+				}, true ),
+				parent;
 
-			element.forEach( function( node ) {
+			if ( node ) {
 				// Since symbol may contains special characters we use `indexOf` (instead of RegExp) which is sufficient (#877).
```
Comandeer: `indexOf` was moved up, so the comment also should.

core/htmlparser/element.js
```diff
@@ -470,6 +470,42 @@ CKEDITOR.htmlParser.cssStyle = function() {
...
+		 * ```javascript
+		 * element.findOne( function( child ) {
+		 *     return child.name === 'span' || child.name === 'strong';
+		 * } ); // Will return first child which is a span or a strong.
+		 * ```
+		 *
+		 * @param {String/Function} criteria Tag name or evaluator function.
+		 * @param {CKEDITOR.htmlParser.node} criteria.child The currently iterated child.
+		 * @param {Boolean} [recursive=false] If set to `true` will iterate all descendants, otherwise only direct children.
+		 * @returns {CKEDITOR.htmlParser.node} First matched child, `null` otherwise.
```
Comandeer: It should be marked as nullable type.

core/htmlparser/element.js
```diff
@@ -470,6 +470,42 @@ CKEDITOR.htmlParser.cssStyle = function() {
...
+		 */
+		findOne: function( criteria, recursive ) {
+			var findMethod = typeof criteria === 'function' ? criteria : function( item ) {
```
Comandeer: By inlining I meant "inline it directly into `CKEDITOR.tools.array.find` callback":

```javascript
var isMatching = typeof criteria === 'function' ? criteria( child ) : item.name === criteria;
```

IMO such version is still readable and dedicated function `findMethod` is not necessary in this case.

engineering-this:
Fixed failing test. Fail message was wrongly placed as `findOne` param instead of `assert`, as a result `findOne` treated message as `recursive` param, and `assert` didn't have any message to display.

engineering-this:
IE cuts some of text nodes. To make test constant regardless of environment I've moved tested html code back into js file and removed html file.

engineering-this:
These tests fails on major too -> #2806

PR #2723, opened 2019-01-04
Included event support in menu plugin and firing hide event when menu is hidden

### LINKED ISSUE #2307: Fire an event when context menu is being closed.
## Type of report

Feature request

When context menu is being opened the [menuShow](https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_editor.html#event-menuShow) event is being fired [here](https://github.com/ckeditor/ckeditor-dev/blob/master/plugins/menu/plugin.js#L415). We could also fire an event when menu is being closed.

engineering-this:

New feature

### This PR contains

- [x] Unit tests
- [x] Manual tests

## What changes did you make?

- Added support for event methods `fire`, `on`, `once`, to menu plugin.
- Added `hide` event firer inside `menu._.onHide`.
- Added missing docs for `menu.onHide` method.

Closes #2307


plugins/menu/plugin.js
```diff
@@ -597,6 +600,39 @@ CKEDITOR.plugins.add( 'menu', {
...
  */
+
+/**
+ * Optional callback function that is called when menu is hidden.
+ *
+ * @member CKEDITOR.menu
+ * @method onHide
+ */
+
+/**
```
Comandeer: Instead of documenting it this way, it would be better to mark `CKEDITOR.menu` class with `@mixins CKEDITOR.event`.

tests/plugins/contextmenu/manual/hide.html
```diff
@@ -0,0 +1,24 @@
...
+		width: 50px;
+		height: 50px;
+		background: blue;
+		margin: 5px;
+	}
+</style>
+<textarea id="editor"></textarea>
+<div class="marker"></div>
+<script>
+	if ( bender.tools.env.mobile ) {
```
Comandeer: Wouldn't it be possible to make this test work also on mobile devices by using `menubutton` plugin instead of `contextmenu` one?

It would also fix the issue with the fact that clicking the right button inside editor (especially empty one) can also show the native menu instead of desired one

tests/plugins/menu/menu.js
```diff
@@ -41,6 +41,19 @@
 				assert.isTrue( menuItemEl.hasClass( 'cke_menubutton' ), 'check ui type class name' );
 				assert.isTrue( menuItemEl.hasClass( 'cke_menubutton__custom_menuitem' ), 'check named ui type class name' );
 				assert.isTrue( menuItemEl.hasClass( customCls ), 'check ui item custom class name' );
+
+				menu.hide();
+			} );
+		},
+
+		'test menu hide event': function() {
```
Comandeer: Missing ticket number.

CHANGES.md
```diff
@@ -8,6 +8,7 @@ API Changes:
...
 * [#2021](https://github.com/ckeditor/ckeditor-dev/issues/2021): Add [`CKEDITOR.dom.documentFragment.find`](https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_dom_documentFragment.html#method-find) and [`CKEDITOR.dom.documentFragment.findOne`](https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_dom_documentFragment.html#method-findOne) methods.
 * [#2700](https://github.com/ckeditor/ckeditor-dev/issues/2700): Added the [`CKEDITOR.tools.array.find`](https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_tools_array.html#method-find) function.
+* [#2307](https://github.com/ckeditor/ckeditor-dev/issues/2307): Added event support for [Menu](https://ckeditor.com/cke4/addon/menu) plugin by exposing methods: [`CKEDITOR.menu.fire`](https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_menu.html#method-fire), [`CKEDITOR.menu.on`](https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_menu.html#method-on) and [`CKEDITOR.menu.once`](https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_menu.html#method-once). Additionally menu instances will fire `hide` event when hidden.
```
Comandeer: IMO it should mention implementing `CKEDITOR.event` mixin on top of `CKEDITOR.menu` instead of listing methods.

engineering-this:
>Seems to work, yet I'm thinking about extending the feature a bit and introduce also show event for the menu instance and deprecate menuShow event for editor. WDYT?

This would be nice, but we can't do this. Actual `menu` instance is private property of `menubutton`. And it is created upon first click on button.

Before clicking on button:
```
CKEDITOR.instances.editor.ui.get( 'Language' )._.menu -> undefined
```

After clicking on button:
```
CKEDITOR.instances.editor.ui.get( 'Language' )._.menu -> menu instance
```

tests/plugins/contextmenu/manual/hide.html
engineering-this: I'll just add button to open context menu on mobiles.

mlewand:
> Seems to work, yet I'm thinking about extending the feature a bit and introduce also show event for the menu instance and deprecate menuShow event for editor. WDYT?

I'd love to have this but the problem is getting the menu instance at runtime, just like @engineering-this mentioned. The thing is that it is lazy-loaded and there's no convenient way to get it.

In order to achieve @Comandeer we'd need to add a proper getter, but this is something we should extract to the separate issue.

tests/plugins/contextmenu/manual/hide.html
```diff
@@ -0,0 +1,27 @@
+<style>
+	.marker {
+		width: 50px;
+		height: 50px;
+		background: blue;
+		margin: 5px;
+	}
+</style>
+<textarea id="editor"><div style="width:100%;height:200px;"></div></textarea>
```
Comandeer: What is the purpose of the empty `div` inside editor? It renders unnecessary border on click in IE 8.

plugins/menu/plugin.js
```diff
@@ -597,6 +604,21 @@ CKEDITOR.plugins.add( 'menu', {
...
+ * Fired when a menu is hidden.
+ *
+ * @event hide
```
Comandeer: It should be marked as `@since 4.12.0`.

tests/plugins/contextmenu/manual/hide.html
engineering-this: It fills the editor, so it is possible to open context menu.

engineering-this:
>The manual test for menubutton is still missing. We can move the current manual test for contentmenu into menu and add additional steps for menu button.

When trying to create such nice unifying test for both cases I realised our API is still flawed.

We cannot access instance of menu before it is created, hence we can't bind the listener. `menuShow` event is passing only the `panel` in data and we cannot access menu from listener...

https://github.com/ckeditor/ckeditor-dev/blob/3533f36cdd7554b0f1e3fb71f9d8e81a962a4992/plugins/menu/plugin.js#L414-L415

I wonder if we could push second item to data which would be the menu instance? Or is it better to add menu property to panel instance if it's created by menu? WDYT @Comandeer?

engineering-this:
- I've added `menu` property on `panel` instance.
- Unit test for that change.
- Reworked manual test.

jacekbogdanski:
The second option about refactoring menu plugin to expose `show` and `hide` events seems the cleanest and reduces editor object pollution. However, as you already pointed out, it looks like a big feature. If it doesn't give us any functional nor performance improvement I'd rather vote for the 3rd option. If it ain't broke, don't fix it. The code already exists so we would avoid deprecating API and also avoid bugs which may occur during refactoring.

engineering-this:
New changes:
- Initialize `menu` if possible.
- Expose `menuButton.getMenu`.
- Fire `show` event when menu is shown.
- Updated tests.

Issue #2732, opened 2019-01-07
Cannot open cell properties in standard editor


msamsel:
## Type of report

Bug

## Provide detailed reproduction steps (if any)

1. Open sample of build `standard` version of CKE4 (4.11.2) (b7960bbabb73c42c09718a49287618b05281eb99)
2. Add table to editor's content
3. Open `cell properties` from context menu

### Expected result

Cell properties dialog is opened

### Actual result

There is error in console:
`Uncaught TypeError: Cannot read property 'type' of null`

## Other details

* Browser: All
* OS: All
* CKEditor version: 4.11.2
* Installed CKEditor plugins: …

mlewand:
The code that caused this regression was reverted in #2735. Related issue #1986 has been reopened.

PR #2086, opened 2018-06-14
Fixed: cell properties dialog shows styles that are not allowed through allowedContent

engineering-this:
## What is the purpose of this pull request?

Bug fix

## Does your PR contain necessary tests?

All patches which change the editor code must include tests. You can always read more
on [PR testing](https://docs.ckeditor.com/ckeditor4/docs/#!/guide/dev_contributing_code-section-tests),
[how to set the testing environment](https://docs.ckeditor.com/ckeditor4/docs/#!/guide/dev_tests) and
[how to create tests](https://docs.ckeditor.com/ckeditor4/docs/#!/guide/dev_tests-section-creating-your-own-test)
in the official CKEditor documentation.

### This PR contains

- [x] Unit tests
- [x] Manual tests

## What changes did you make?

I've added `requiredContent` for each option in table cell properties dialog.

Closes #1986

PR #2746, opened 2019-01-11
Restore reverted optimizations of table cell allowed content

### LINKED ISSUE #1986: Cell properties shows style elements that are not allowed through allowedContent
## Type of report

Bug

This is a followup to a previous issue https://github.com/ckeditor/ckeditor-dev/issues/1014

Which is based on an issue in Drupal: https://www.drupal.org/project/drupal/issues/2785483

1. Assume the following allowedContent: 
```
config.allowedContent = {
		td: {
			attributes: 'colspan rowspan',
			styles: false
		}
	}
```
2. Add a table and open the 'Cell properties' dialog

### Expected result

* The 'Cell properties' dialog shows form elements for 'Cell Type', 'Rows Span', 'Columns Span'

### Actual result

* The 'Cell properties' dialog still shows form elements for 'Word Wrap', 'Horizontal Alignment', 'Vertical Alignment', 'Background Color', 'Border Color'

## Other details

* Browser: All
* OS: All
* CKEditor version: -
* Installed CKEditor plugins: tabletools

engineering-this:

Bug fix

### This PR contains

- [x] Unit tests
- [x] Manual tests

## What changes did you make?

Changes from #2086 were reverted by #2735 because of nasty regression #2732.

Problem comes from here:
https://github.com/ckeditor/ckeditor-dev/blob/abd2465fd64ae9e9f84c275210f73d2fec54c219/plugins/tabletools/dialogs/tableCell.js#L317-L334
And repeated in [L364-382](https://github.com/ckeditor/ckeditor-dev/blob/abd2465fd64ae9e9f84c275210f73d2fec54c219/plugins/tabletools/dialogs/tableCell.js#L364-L382)

Originally I didn't want to move part of definition to conditional variable, as it would make reading whole definition harder, so I inlined which leaded to item being `null` when `colorDialog` plugin is absent. Resulting in an error when `filebrowser` plugin tried to access property of `null`.

To fix the error I wrapped linked above conditional with IIFE, so it returns one or two elements, but none of them is `null`.

New commits starts from: `5d1c1ac`

Closes #1986

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -11,7 +11,7 @@ Fixed Issues:
 * [#1191](https://github.com/ckeditor/ckeditor-dev/issues/1191): Fixed: Items in the [elements path](https://ckeditor.com/cke4/addon/elementspath) are draggable.
 * [#2292](https://github.com/ckeditor/ckeditor-dev/issues/2292): Fixed: Dropping list with link on editor's margin cause console error and removing dragged text from editor.
 * [#2756](https://github.com/ckeditor/ckeditor-dev/issues/2756): Fixed: The [Auto Link](https://ckeditor.com/cke4/addon/autolink) plugin causes an error when typing in [Source Editing Mode](https://ckeditor.com/docs/ckeditor4/latest/guide/dev_sourcearea.html).
-* [#1986](https://github.com/ckeditor/ckeditor-dev/issues/1986): Fixed: cell properties dialog shows styles that are not allowed through allowedContent.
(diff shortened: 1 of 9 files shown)


engineering-this:
>Maybe we can add additional check inside filebrowse plugin, to check if element is set before trying to access its properties?

[CKEDITOR.dialog.definition.content](https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_dialog_definition_content.html#property-elements) should be defined as an array of `uiElements`, `null` or `undefined` aren't allowed types,  so there is no need for extra check.

I've updated test so it will assert regardless of error thrown by filebrowser. And added another TC for regression.

PR #2772, opened 2019-01-21
Fix Travis CI

### LINKED ISSUE #2796: Fix Travis CI
## Type of report

Bug

1. Look at travis results

### Expected result

Travis is working properly

### Actual result

Travis breaks during building

## Other details

* Browser: …
* OS: …
* CKEditor version: …

msamsel:

Bug

### This PR contains

- [ ] Unit tests
- [ ] Manual tests

## What changes did you make?

* Update travis config
* Add updated dependencies of bender, which are necessary for Travis
* Add new config variable `isTravis` to detect if current test is run on Travis
* Update paths in some test to be proper relative paths
* Ignore few tests, which fails on Travis
* Include new bender tools, which detects if tests are run on build version of CKEditor

## Additional info
Tests on Firefox tends to fail randomly, usually it's related to problem with `wait()` and `resume()` methods. There are sometimes issues with image tests, where some attributes are not refresh fast enough. It looks like firefox error, where some test cases checks assertion too fast or some other optimization happens. Resetting build job usually helps, but it tends to be a little annoying.
In really tough situation, there is possibility to debug travis job https://docs.travis-ci.com/user/running-build-in-debug-mode/#legacy-repositories.

Close #2796

### DIFF
--- a/tests/_benderjs/ckeditor/static/tools.js
+++ b/tests/_benderjs/ckeditor/static/tools.js
@@ -91,12 +91,8 @@
 			return bender.tools.fixHtml( html, stripLineBreaks );
 		},
 
-		isBuild: function() {
-			return CKEDITOR.revision !== '%REV%';
-		},
-
 		env: {
-			/*
+			/**
 			 * Tells whether current environment is running on a mobile browser.
 			 *
 			 * It's different from deprecated {@link CKEDITOR.env.mobile} in a way that we are just
@@ -104,15 +100,20 @@
 			 */
 			mobile: CKEDITOR.env.iOS || navigator.userAgent.toLowerCase().indexOf( 'android' ) !== -1,
 
-			/*
+			/**
 			 * Whether current OS is a Linux environment.
 			 */
 			linux: navigator.userAgent.toLowerCase().indexOf( 'linux' ) !== -1,
 
-			/*
+			/**
 			 * Whether current environment is Opera browser.
 			 */
-			opera: navigator.userAgent.toLowerCase().indexOf( ' opr/' ) !== -1
+			opera: navigator.userAgent.toLowerCase().indexOf( ' opr/' ) !== -1,
+
+			/**
+			 * Whether current environment is run as build version of CKEditor.
+			 */
+			isBuild: CKEDITOR.revision !== '%REV%'
 		},
 
 		fixHtml: function( html, stripLineBreaks, toLowerCase ) {
--- a/tests/_benderjs/ckeditor/static/tools.js
+++ b/tests/_benderjs/ckeditor/static/tools.js
@@ -91,6 +91,10 @@
 			return bender.tools.fixHtml( html, stripLineBreaks );
 		},
 
+		isBuild: function() {
+			return CKEDITOR.revision !== '%REV%';
+		},
+
 		env: {
 			/*
 			 * Tells whether current environment is running on a mobile browser.
--- a/tests/core/dom/range/getclientrects.js
+++ b/tests/core/dom/range/getclientrects.js
@@ -12,14 +12,14 @@
 			this.playground = doc.getById( 'playground' );
 		},
 
-		_should: {
-			ignore: {
-				'test only element selection': bender.config.isTravis && CKEDITOR.revision !== '%REV%',
-				'test last element selection': bender.config.isTravis && CKEDITOR.revision !== '%REV%',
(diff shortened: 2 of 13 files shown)


tests/core/dom/range/getclientrects.js
```diff
@@ -12,6 +12,16 @@
 			this.playground = doc.getById( 'playground' );
 		},
 
+		_should: {
+			ignore: {
+				// Tests randomly fails on FF in Travis
+				'test only element selection': bender.config.isTravis && CKEDITOR.env.gecko,
```
mlewand: You could actually extract `bender.config.isTravis && CKEDITOR.env.gecko` to a variable.

tests/_benderjs/ckeditor/static/tools.js
```diff
@@ -91,6 +91,10 @@
 			return bender.tools.fixHtml( html, stripLineBreaks );
 		},
 
+		isBuild: function() {
```
mlewand: There is a much better namespace for this function: `bender.tools.env` namsepace. We should avoid throwing everything into `bender.tools` as we'd quickly have too much members there.

To better integrate with existing code over there, you can see that there are just properties. We don't expect build type (dev/build) to change between the tests, so it could be a property just like `bender.tools.env.mobile`.

tests/plugins/balloonpanel/_helpers/tools.js
```diff
@@ -119,6 +119,23 @@ var balloonTestsTools = {
 				break;
 		}
 		return pos;
+	},
+
+	removeDottsFromUrl: function( expectedPath ) {
+		var parts = expectedPath.split( '/' ),
```
mlewand: The implementation would be much simpler if you'd use `CKEDITOR.tools.array.filter`.

tests/plugins/balloonpanel/cssloading.js
```diff
@@ -17,13 +17,13 @@
...
 				// built version, such trick must be used (#1251).
-				'test loading css with path': CKEDITOR.revision !== '%REV%'
+				'test loading css with path': bender.tools.isBuild()
 			}
 		},
 
 		'test loading css with path': function() {
-			assert.areSame( 'kama,/apps/ckeditor/skins/kama/', CKEDITOR.skinName, 'config skinName should be with path' );
-			sinon.assert.calledWith( spy, balloonTestsTools.getDocumentOrigin() + '/apps/ckeditor/plugins/balloonpanel/skins/kama/balloonpanel.css' );
+			assert.areSame( 'kama,%BASE_PATH%../../apps/ckeditor/skins/kama/', CKEDITOR.skinName, 'config skinName should be with path' );
```
mlewand: Please add a short note of explanation to why `removeDottsFromUrl` function is used, and add an issue reference ticket related to this PR.

tests/plugins/balloontoolbar/positioning.js
```diff
@@ -92,8 +92,9 @@
 		},
 
 		'test panel - out of view - hcenter top': function( editor ) {
-			if ( editor.name == 'divarea' ) {
-				// divarea tests are failing, it's an upstream issue from balloonpanel (#1064).
+			if ( editor.name == 'divarea' || ( bender.config.isTravis && bender.tools.isBuild() ) ) {
+				// divarea tests are failing, it's an upstream issue from balloonpanel (#1064).ga
```
mlewand:
```
// divarea tests are failing, it's an upstream issue from balloonpanel (#1064).ga
```

What is this "ga" doing here? 👮

tests/plugins/balloonpanel/_helpers/tools.js
msamsel: I'm not sure if implementation would be simpler. Please notice that I need to remove also folder names in the path not only dots.
So when there is path like: `/one/two/three/../../four`, I need to receive `/one/four`. This is why I look on array from right to left and preserve information how many addtional folders have to be removed.
To achieve similar result I would need to revers array. Then have filter function in clouser, which will preserve information how many additoianl folders need to be removed. And then revers array again. I don't feel like using `CKEDITOR.tools.array.filter` will simplify thi for loop.

tests/plugins/balloontoolbar/positioning.js
msamsel: It's my shortcut from 'git add'. I probably write it here instead of terminal :(

mlewand:
I'll just rebase it onto latest `master` branch.

PR #2776, opened 2019-01-23
Fix for: Editor with enterMode set to ENTER_DIV alters content on paste

### LINKED ISSUE #2751: Editor with enterMode set to ENTER_DIV alters content on paste
## Type of report

Bug

## Provide detailed reproduction steps (if any). 

1. Open SDK sample: https://ckeditor.com/docs/ckeditor4/latest/examples/enterkey.html
2. Pick `Create a new <div>` from select under `When Enter is pressed:`
3. Focus editor at the end of content, press <kbd>Enter</kbd> and type some text.
4. Select all.
5. Copy.
6. Paste

### Expected result

Editor content looks like:
```
<div>This is some sample text.</div>

<div>New line.</div>
```

### Actual result

There is an extra div wrapping whole editor content:

```
<div>
<div>This is some sample text.</div>

<div>New line.</div>
</div>
```

## Other details

* Browser: …
* OS: …
* CKEditor version: …

engineering-this:

Bug fix

### This PR contains

- [x] Unit tests
- [x] Manual tests

## What changes did you make?
When a `div` is to be inserted in another empty `div` I've forced triggering logic which is normally trigerred when element is not allowed in selection.

Closes #2751

### DIFF
--- a/core/editable.js
+++ b/core/editable.js
@@ -1652,6 +1652,10 @@
 
 			prepareRangeToDataInsertion( that );
 
+			if ( editor.getData() === '' && editor.enterMode === CKEDITOR.ENTER_DIV ) {
+				clearEditable( editable, range );
+			}
+
 			// DATA PROCESSING
 
 			// Select range and stop execution.
@@ -1666,6 +1670,12 @@
 			// Set final range position and clean up.
 
 			cleanupAfterInsertion( that );
+		}
+
+		function clearEditable( editable, range ) {
+			editable.getFirst().remove();
+			range.setStartAt( editable, CKEDITOR.POSITION_AFTER_START );
+			range.collapse( true );
 		}
 
 		// Prepare range to its data deletion.
@@ -1889,8 +1899,7 @@
 
 					// Find the first ancestor that can contain current node.
 					// This one won't be split.
-					while ( insertionContainer && ( !DTD[ insertionContainer.getName() ][ nodeData.name ] ||
-						preventInsertingDiv( nodeData.name, insertionContainer, that.editor ) ) ) {
+					while ( insertionContainer && !DTD[ insertionContainer.getName() ][ nodeData.name ] ) {
 						if ( insertionContainer.equals( blockLimit ) ) {
 							insertionContainer = null;
 							break;
@@ -2056,7 +2065,7 @@
 						continue;
 					}
 
-					allowed = !!allowedNames[ nodeName ] && !preventInsertingDiv( nodeName, startContainer, that.editor );
+					allowed = !!allowedNames[ nodeName ];
 
 					// Mark <brs data-cke-eol="1"> at the beginning and at the end.
 					if ( nodeName == 'br' && node.data( 'cke-eol' ) && ( !nodeIndex || nodeIndex == nodesCount - 1 ) ) {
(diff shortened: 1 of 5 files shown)


core/editable.js
```diff
@@ -2081,6 +2082,17 @@
 			return nodesData;
 		}
 
+		// Prevent inserting div into div when Editor is in ENTER_DIV mode, and target element is an empty div.
+		function preventInsertingDiv( nodeName, element, editor ) {
```
Comandeer: The second parameter can use some better name, e.g. `container`. Now `nodeName` and `element` seem to be about the same element.

tests/core/editable/enterdiv.js
```diff
@@ -0,0 +1,28 @@
+/* bender-tags: editor */
+
+bender.editor = {
+	config: {
+		enterMode: CKEDITOR.ENTER_DIV
+	}
+};
+
+var tests = {
+	// 2751
```
Comandeer: Our convention for tests is `// (#<issue number>)`.

engineering-this:
> I'm wondering if just removing `body > div` tag when it's the only element and it's empty, wouldn't be a better solution.

Simplified solution by doing as above.

engineering-this:
That's an issue with selection. In Chrome and Safari using native `SelectAll` produces selection as following:
```<div><div>{This is some sample text.</div><div>New line.}</div></div>```
While it should be:
```[<div><div>This is some sample text.</div><div>New line.</div></div>]```

Extracted to #2801

mlewand:
Guys, just FYI:

`<div><div>{This is some sample text.</div><div>New line.}</div></div>`

Is a perfectly valid selection, and should not be a condition to any bug.

engineering-this:
@mlewand I've thought this through. Merging divs doesn't sound like good option to me. User can have `div` as a style container where he wants to paste content. If we limit merging divs only when they doesn't have any extra attributes, then there will be following case:

1. User has content:
`<div class="red-border"><div class="hightlight">Foo</div></div>`
1. SelectAll.
1. Copy.
1. Paste.

Content will be wrapped by duplicated divs.

I don't think we can safely cover every possible case by merging divs. 

However I have another idea for this case. I'll skip details for now, if it works I will push updates.

@Comandeer what's your opinion on that?

engineering-this:
The more I dig into it the more I think we should close original issue as an upstream and report to Chromium.

There are a lot of issues because of Chrome/Safari selection.

## First case:
1. Open editor sample, clear it.
1. Apply `special container` style from styles combo.
1. Type 'foo'.
1. Select all.
1. Copy.
1. Press 'New Page' button.
1. Paste.

### Expected
Pasted content is the same.
screenshot 2019-02-12 at 17 40 12

### Actual
Special container style is missing.
screenshot 2019-02-12 at 17 40 23
Because of such selection on paste dataTransfer `text/html` data is `foo`.

## Second case:
Repeat each step, but after step 3. press <kbd>Enter</kbd>.

### Expected
screenshot 2019-02-12 at 17 40 41

### Actual
screenshot 2019-02-12 at 17 41 05
The first line is preserved with styling, second is missing.

## Additional case
When using editor with `divarea` plugin, and `selectAll` button from toolbar pasted content is always the same as before paste, because created selection starts and ends at editable instead of text.

Every above case results in expected behaviour on Firefox and Edge.

engineering-this:
Reported it on chromium and team confirmed that they can reproduce it:
https://bugs.chromium.org/p/chromium/issues/detail?id=931285

engineering-this:
The original case which is reproduced also on Firefox is covered by this PR. Other cases which are present only on Chrome/Safari are upstream issues.

I've updated manual test, to reflect that.

Comandeer:
Rebased onto latest `major`.

engineering-this:
Webkit issue:
https://bugs.webkit.org/show_bug.cgi?id=198603

engineering-this:
When copy at step 3. actual is the same as your expected.
When copy at step 1. actual matches yours, but that's what should be pasted, because that is what is in the clipboard.

@Comandeer can you confirm when to copy?

Comandeer:
Yes, it should be copied at 3. And yes, seems that it works perfectly 🤔 Honestly, I don't know why I thought back then that something's not right 😅 sorry.

engineering-this:
IE8 and 9 have same buggy selection as Chrome and Safari:
`<div><div>[This is some sample text.</div><div>New line.]</div></div>`, so I'm excluding it from cases which would fail.

PR #2797, opened 2019-02-06
Enhance error messages for creators

### LINKED ISSUE #2748: Throw friendly error for invalid/missing id's on editor creation
## Type of report

Feature request

Given the code below:
```html
<textarea id="editor1"></textarea>

<script>
    CKEDITOR.replace('editor2');
</script>
```

The developer intended to target `editor1`, but by mistake targeted `editor2`, which doesn't exist in the DOM.

Doing this causes an error. 
```
Uncaught TypeError: Cannot read property 'getEditor' of undefined
```
Without looking at the CKEditor source, it's not obvious what's wrong.

The Feature request is to throw a friendly error for invalid/missing id's.

Solution:

```diff
diff --git a/core/creators/inline.js b/core/creators/inline.js
index 61f88b2..d48af34 100644
--- a/core/creators/inline.js
+++ b/core/creators/inline.js
@@ -31,8 +31,15 @@
                if ( !CKEDITOR.env.isCompatible )
                        return null;

+               var elementArg = element;
+
                element = CKEDITOR.dom.element.get( element );

+               // Throw error on missing target element.
+               if ( !element ) {
+                       throw 'The provided element, "' + elementArg + '", is missing from the DOM.';
+               }
+
                // Avoid multiple inline editor instances on the same element.
                if ( element.getEditor() )
                        throw 'The editor instance "' + element.getEditor().name + '" is already attached to the provided element.';
diff --git a/core/creators/themedui.js b/core/creators/themedui.js
index 3b1b00b..814b815 100644
--- a/core/creators/themedui.js
+++ b/core/creators/themedui.js
@@ -323,8 +323,15 @@ CKEDITOR.replaceClass = 'ckeditor';
                if ( !CKEDITOR.env.isCompatible )
                        return null;

+               var elementArg = element;
+
                element = CKEDITOR.dom.element.get( element );

+               // Throw error on missing target element.
+               if ( !element ) {
+                       throw 'The provided element, "' + elementArg + '", is missing from the DOM.';
+               }
+
                // Avoid multiple inline editor instances on the same element.
                if ( element.getEditor() )
                        throw 'The editor instance "' + element.getEditor().name + '" is already attached to the provided element.';
```

Comandeer:

New feature

### This PR contains

- [x] Unit tests
- [ ] Manual tests

## What changes did you make?

I've replaced `throw` statements in our creators with calls to `CKEDITOR.error` with two new error codes:

* `editor-incorrect-element` – logged when user tries to attach editor to nonexistent element,
* `editor-element-conflict` – logged when user tries to attach editor to the element with already attached one.

Additionally I moved the logic for getting element for editor outside creators, into `core/ckeditor.js` under the name `CKEDITOR.getEditorElement`. However I'm not entirely sure if these are the right place and right name.

Finally I had to refactor some tests that were expecting errors to be thrown.

Closes #2748.

### DIFF
--- a/core/ckeditor.js
+++ b/core/ckeditor.js
@@ -34,6 +34,35 @@ CKEDITOR.instances = {};
 CKEDITOR.document = new CKEDITOR.dom.document( document );
 
 /**
+ * Gets the element from DOM and checks if the editor can be instantiated on it.
+ * This function is available for internal use only.
+ *
+ * @private
+ * @since 4.12.0
+ */
+CKEDITOR.getEditorElement = function( elementOrId ) {
+	var element;
+
+	if ( !CKEDITOR.env.isCompatible ) {
+		return null;
+	}
+
+	element = CKEDITOR.dom.element.get( elementOrId );
+
+	// Throw error on missing target element.
+	if ( !element ) {
+		throw 'The provided element, "' + elementOrId + '", is missing from the DOM.';
+	}
+
+	// Avoid multiple inline editor instances on the same element.
+	if ( element.getEditor() ) {
+		throw 'The editor instance "' + element.getEditor().name + '" is already attached to the provided element.';
+	}
+
+	return element;
+};
+
+/**
  * Adds an editor instance to the global {@link CKEDITOR} object. This function
  * is available for internal use mainly.
  *
--- a/core/ckeditor.js
+++ b/core/ckeditor.js
@@ -51,12 +51,20 @@ CKEDITOR.getEditorElement = function( elementOrId ) {
 
 	// Throw error on missing target element.
 	if ( !element ) {
-		throw 'The provided element, "' + elementOrId + '", is missing from the DOM.';
+		CKEDITOR.error( 'editor-incorrect-element', {
+			element: elementOrId
+		} );
+
+		return;
 	}
 
 	// Avoid multiple inline editor instances on the same element.
 	if ( element.getEditor() ) {
-		throw 'The editor instance "' + element.getEditor().name + '" is already attached to the provided element.';
+		CKEDITOR.error( 'editor-element-conflict', {
+			editorName: element.getEditor().name
+		} );
+
+		return;
 	}
 
 	return element;
--- a/core/ckeditor.js
+++ b/core/ckeditor.js
@@ -39,6 +39,7 @@ CKEDITOR.document = new CKEDITOR.dom.document( document );
  *
  * @private
  * @since 4.12.0
+ * @returns {CKEDITOR.dom.element/null}
  */
 CKEDITOR.getEditorElement = function( elementOrId ) {
 	var element;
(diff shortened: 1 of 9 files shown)


core/ckeditor.js
```diff
@@ -33,6 +33,44 @@ CKEDITOR.instances = {};
...
+/**
+ * Gets the element from DOM and checks if the editor can be instantiated on it.
+ * This function is available for internal use only.
+ *
+ * @private
+ * @since 4.12.0
+ * @returns {CKEDITOR.dom.element/null}
+ */
+CKEDITOR.getEditorElement = function( elementOrId ) {
+	var element;
```
mlewand: You should just declare this `var` together with initialization few lines below:

```
element = CKEDITOR.dom.element.get( elementOrId );
```

core/ckeditor.js
```diff
@@ -33,6 +33,44 @@ CKEDITOR.instances = {};
...
 
+/**
+ * Gets the element from DOM and checks if the editor can be instantiated on it.
+ * This function is available for internal use only.
+ *
+ * @private
+ * @since 4.12.0
+ * @returns {CKEDITOR.dom.element/null}
+ */
+CKEDITOR.getEditorElement = function( elementOrId ) {
```
mlewand: You mentioned that you had mixed feelings regarding that place and it's well justified - as it simply doesn't belong here.

We should be very careful when it comes down to adding members directly in the `CKEDITOR` namespace (even though you proposed adding it as a private member). The `CKEDITOR.editor` type seems to be _more_ suitable place for this logic.

Ideally we could put it in the constructor, right? because both creators use that, but that means we'd change a very popular API part - we shouldn't do that unless we really, really must.

With that in mind I think that easiest thing would be to put as a static method in `CKEDITOR.editor` type. Again it should have restricted visibility, as this is a function that is not really needed except for creators.

tests/core/creators/duplicates.js
```diff
@@ -7,35 +7,24 @@
...
-				try {
-					CKEDITOR.replace( 'editor' );
-					fail = true;
-				} catch ( e ) {}
+				CKEDITOR[ creator ]( element );
 
-				assert.isFalse( fail, 'Expected error not thrown.' );
+				spy.restore();
+				assert.areSame( 1, spy.callCount, 'Error was thrown' );
+				assert.areSame( 'editor-element-conflict', spy.args[ 0 ][ 0 ], 'Appropriate error code was used' );
```
mlewand: Please, include assertion for `args[ 0 ][ 1 ]` so that it has valid `editorName`. Actually please use `calledWithExactly()` api, in conjuction with custom matcher for `editorName` val it will be nice and readable.

mlewand: Same thing for `element` property in the other `editor-incorrect-element` error.

PR #2811, opened 2019-02-14
Fix for focus issues in nested editables caused by copyformatting plugin

### LINKED ISSUE #2470: Widget nested editable triggers blur event on focus in Firefox
## Type of report

Bug

1. Open https://jsfiddle.net/7t4wvg06/6/ using **Firefox**
2. Open the JavaScript console (Developer Tools)
3. Click over "foo" to focus the widget (inside the CKEditor's editing area)
4. Type something in the widget
5. Click below "foo" to blur the widget (i.e. outside of the widget)

### Expected result

Step 3 should log

    editable focus

and step 5 should log

    editable blur

### Actual result

Step 3 logs

    editable focus
    editable blur

and step 5 doesn't log anything.

## Other details

* Browser: Firefox 61.0.1
* OS: Ubuntu 17.10
* CKEditor version: both 4.10.1 and 4.10.0
* Installed CKEditor plugins: **copyformatting**

It doesn't reproduce on Chrome or if you remove the ``copyformatting`` plugin.

My investigation shows that this is a regression caused by 3a9f058fb5a6dee84444254199fe70252182e90b which was committed in CKEditor 4.10.0 . The change for Firefox is that

    editorFocus: false,

has become

    editorFocus: true,

so the ``applyFormatting`` command requires the editor to be focused but in my case I don't call this command explicitly so I don't understand how it affects the widget.

### LINKED ISSUE #2655: Nested editable can't be focused under certain circumstances
## Type of report

Bug

1. Open https://jsfiddle.net/w08o1reL/11/ in **Chrome**.
2. Click over "foo". The nested editable is focused as a result.
3. Click just above the widget (above the black border of the widget). The nested editable looses the focus as a result.
4. Click again over "foo".

### Expected result

The nested editable is focused.

### Actual result

The nested editable gets the focus for a fraction of a second but then it looses it.

## Other details

* Browser: Chrome
* OS: Any
* CKEditor version: 4.11.1
* Installed CKEditor plugins: wysiwygarea, widget, **copyformatting**

Note that:
* the issue doesn't reproduce if I remove the **copyformatting** plugin. This makes me think that this issue is related to #2470
* the issue doesn't reproduce if I replace the **padding** on the body element with **margin**. The default CKEditor content CSS [is using margin](https://github.com/ckeditor/ckeditor-dev/blob/4.11.1/contents.css#L19) so you may not be affected by this issue if you don't use **fullPage** or if you use **fullPage** and you don't have padding on the body tag.
* the issue doesn't reproduce on Firefox

Comandeer:

Bug fix

### This PR contains

- [x] Unit tests
- [x] Manual tests

## What changes did you make?

The issue is connected with the fact that every `mouseup` event on editable was connected with executing `applyFormat` command, which required focusing the editor. It obviously was stealing focus from the editable. The simplest solution was to exec `applyFormat` only if there is really some formatting to be applied.

Closes #2655.
Closes #2470.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -9,6 +9,8 @@ Fixed Issues:
 * [#2527](https://github.com/ckeditor/ckeditor-dev/issues/2527): Fixed: [Emoji](https://ckeditor.com/cke4/addon/emoji) autocomplete order now prioritize emojis with name started from used string.
 * [#2572](https://github.com/ckeditor/ckeditor-dev/issues/2572): Fixed: Icons in [Emoji](https://ckeditor.com/cke4/addon/emoji) dropdown navigation groups are not centered.
 * [#1191](https://github.com/ckeditor/ckeditor-dev/issues/1191): Fixed: Elements in elements path are draggable.
+* [#2470](https://github.com/ckeditor/ckeditor-dev/issues/2470): [Firefox] Fixed: Widget's nested editable blurred upon focus.
+* [#2655](https://github.com/ckeditor/ckeditor-dev/issues/2655): [Chrome, Safari] Fixed: Widget's nested editable can't be focused under certain circumstances.
 
 ## CKEditor 4.11.2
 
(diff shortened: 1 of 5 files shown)


Comandeer:
Closed in favor of #2792.

PR #2819, opened 2019-02-18
Fix for failing test tests/plugins/balloontoolbar/positioning

### LINKED ISSUE #1076: Balloon Toolbar tests are failing in small browser window
Bug

1. Open unit test http://tests.ckeditor.test:1030/tests/plugins/balloontoolbar/positioning.
2. Resize window to height ~300-400px.
3. Run tests.

### Expected result

All test should be passed.

### Actual result

Teat are failing on `top align`.

## Other details

* Browser: Windows, MacOS
* OS: all
* CKEditor version: 4.7.3
* Installed CKEditor plugins: inlinetoolbar

engineering-this:

Failing tests fix

### This PR contains

- [ ] Unit tests
- [ ] Manual tests

## What changes did you make?

- Changed some values being rounded too early, corrected ternary operator which rounded values for any browsers instead of only IE.
- iOS has returns wrong rect values for `iframe` see https://github.com/ckeditor/ckeditor-dev/issues/1076#issuecomment-464793119. As a workaround I've used iframe parent to get proper rect. Iframe has `width:100%;height:100%;` styles, so rects matches.
- Updated tests to reflect above change.

Closes #1076

### DIFF
--- a/plugins/balloonpanel/plugin.js
+++ b/plugins/balloonpanel/plugin.js
@@ -455,7 +455,7 @@
 
 				if ( !isInline && CKEDITOR.env.safari ) {
 					// Overwrite frame with editor iframe closest parent, because iframe has wrong rect values in mobile Safari (#1076).
-					frame = this.editor.container.findOne( '.cke_contents' );
+					frame = frame.getParent();
 				}
 
 				var panelWidth = this.getWidth(),
--- a/plugins/balloonpanel/plugin.js
+++ b/plugins/balloonpanel/plugin.js
@@ -453,6 +453,11 @@
 				editable = this.editor.editable();
 				isInline = editable.isInline();
 
+				if ( !isInline && CKEDITOR.env.safari ) {
+					// Overwrite frame with editor iframe closest parent, because iframe has wrong rect values in mobile Safari.
+					frame = this.editor.container.findOne( '.cke_contents' );
+				}
+
 				var panelWidth = this.getWidth(),
 					panelHeight = this.getHeight(),
 
--- a/plugins/balloonpanel/plugin.js
+++ b/plugins/balloonpanel/plugin.js
@@ -454,7 +454,7 @@
 				isInline = editable.isInline();
 
 				if ( !isInline && CKEDITOR.env.safari ) {
-					// Overwrite frame with editor iframe closest parent, because iframe has wrong rect values in mobile Safari.
+					// Overwrite frame with editor iframe closest parent, because iframe has wrong rect values in mobile Safari (#1076).
 					frame = this.editor.container.findOne( '.cke_contents' );
 				}
 
--- a/tests/plugins/balloonpanel/positioning.js
+++ b/tests/plugins/balloonpanel/positioning.js
@@ -211,8 +211,7 @@
 							{ height: 15, width: 25, left: 130, bottom: 759.34375, right: 140, top: 744.34375 },
 							{ height: 15, width: 25, left: 145, bottom: 759.34375, right: 155, top: 744.34375 }
 						]
-					]
-				};
+					] };
 				this._testBalloonPanelPosition( ranges, 92.5, 422, true );
 			},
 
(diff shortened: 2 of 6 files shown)


f1ames:
Rebased on latest `master`.

tests/plugins/balloonpanel/positioning.js
```diff
@@ -80,19 +80,18 @@
...
-						}
-
-						return ret;
-					};
+					var editor = this.editor,
+						rect = { height: 300, width: 300, left: 1, bottom: 643, right: 301, top: 343 },
+						frame = editor.window.getFrame(),
+						container = frame.getParent();
+
+					sinon.stub( editor.window, 'getFrame' ).returns( frame );
```
f1ames: 🤔 First there is

```js
frame = editor.window.getFrame();
```

and then

```js
sinon.stub( editor.window, 'getFrame' ).returns( frame );
```

Is it to make `editor.window.getFrame()` function always return the result of this particular `editor.window.getFrame()` call? Would the result be different when the original function is called in specific tests?

tests/plugins/balloonpanel/positioning.js
```diff
@@ -80,19 +80,18 @@
...
+					sinon.stub( CKEDITOR.env.safari ? container : frame, 'getClientRect' ).returns( rect );
+
+					if ( CKEDITOR.env.safari ) {
+						// Override container because iframe has wrong rect values in mobile Safari (#1076).
```
f1ames: Again, comment mentions mobile Safari while the `if` condition is for both desktop and mobile ones.

plugins/balloonpanel/plugin.js
```diff
@@ -453,6 +453,11 @@
 				editable = this.editor.editable();
 				isInline = editable.isInline();
 
+				if ( !isInline && CKEDITOR.env.safari ) {
+					// Overwrite frame with editor iframe closest parent, because iframe has wrong rect values in mobile Safari (#1076).
```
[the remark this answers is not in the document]
engineering-this: https://github.com/ckeditor/ckeditor-dev/blob/master/core/env.js#L101-L110

`CKEDITOR.env.mobile` is labeled as deprecated and we shouldn't use it. There is no guaranteed check for mobile browser. Allowing this code on desktop Safari doesn't seem to break anything, so I decided to just stay with simple check for browser.

tests/plugins/balloonpanel/positioning.js
engineering-this: I need to alter `frame.getWindow`, see [line 93](https://github.com/ckeditor/ckeditor-dev/pull/2819/files/cb96f15b1abbdcec465efa634201c1842dc071f6#diff-955b377084efa16b3c9bcbde47dadd09R93).
[`window.getFrame()`](https://github.com/ckeditor/ckeditor-dev/blob/master/core/dom/window.js#L91-L94) returns new object with each call, so I need to have access to returned object to override it's `getWindow`.

tests/plugins/balloonpanel/positioning.js
engineering-this: Same reason as https://github.com/ckeditor/ckeditor-dev/pull/2819#discussion_r268246828

engineering-this:
>Is it doable to add manual test covering this particular case?

There are possible cases when balloon toolbar is wrongly positioned on iOS Safari, so I think I will be able to reproduce it in manual test.

tests/plugins/balloonpanel/positioning.js
engineering-this: Actually I can simplify it by calling stub on `CKEDITOR.dom.window.prototype.getViewPaneSize`.

engineering-this:
Apart from resolving code related comments I've added manual test for iOS case. I didn't limit test to any env, as it is very short test, and it should pass on every browser.

f1ames:

PR #2878, opened 2019-03-01
Replace deprecated 'arguments.callee'

### LINKED ISSUE #2741: Please remove all arguments.callee
## Type of report

Task

Like the bug #331, please remove all arguments.callee calls, There are five reaming in:
> adapters/jquery.js
> core/tools.js
> plugins/adobeair/plugin.js
> plugins/dialog/plugin.js
> plugins/find/dialogs/find.js

This prevents CKEditor to work in the strict mode.

Dumluregn:

Replacement of deprecated code.

Tests already in place.

## What changes did you make?

Replaced 'arguments.callee' in 5 files, mostly using named function expressions.

Closes #2741.

### DIFF
--- a/adapters/jquery.js
+++ b/adapters/jquery.js
@@ -157,10 +157,10 @@
 					editor.on( 'instanceReady', function( evt ) {
 						var editor = evt.editor;
 
-						setTimeout( function() {
+						setTimeout( function delayEditor() {
 							// Delay bit more if editor is still not ready.
 							if ( !editor.element ) {
-								setTimeout( arguments.callee, 100 );
+								setTimeout( delayEditor, 100 );
 								return;
 							}
 
--- a/adapters/jquery.js
+++ b/adapters/jquery.js
@@ -265,10 +265,10 @@
 				} else {
 					// Editor is already during creation process, bind our code to the event.
 					editor.once( 'instanceReady', function() {
-						setTimeout( function integrateWithjQuery() {
+						setTimeout( function waitForEditor() {
 							// Delay bit more if editor is still not ready.
 							if ( !editor.element ) {
-								setTimeout( integrateWithjQuery, 100 );
+								setTimeout( waitForEditor, 100 );
 								return;
 							}
 
--- a/adapters/jquery.js
+++ b/adapters/jquery.js
@@ -265,10 +265,11 @@
 				} else {
 					// Editor is already during creation process, bind our code to the event.
 					editor.once( 'instanceReady', function() {
-						setTimeout( function() {
+						setTimeout( function integrateWithjQuery() {
 							// Delay bit more if editor is still not ready.
 							if ( !editor.element ) {
-								setTimeout( arguments.callee, 100 );
+								// Replaced arguments.callee (#2741)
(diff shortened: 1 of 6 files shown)


adapters/jquery.js
```diff
@@ -265,10 +265,11 @@
 				} else {
 					// Editor is already during creation process, bind our code to the event.
 					editor.once( 'instanceReady', function() {
-						setTimeout( function() {
+						setTimeout( function integrateWithjQuery() {
```
f1ames: I think it should be more like `waitForEditor` or `waitForEditorElement` since this function just checks if `editor.element` is set.

plugins/adobeair/plugin.js
```diff
@@ -151,11 +151,12 @@
 					var panel = ui._.panel._.panel,
 						holder;
 
-					( function() {
+					( function addListeners() {
```
f1ames: Same here, this function waits for panel to be loaded, so something like `waitForPanel`, `waitForPanelToBeLoaded` could be used. Ofc, if panel is loaded the `convertInlineHandlers()` function is called which adds/manipulate listeners, but this is already different function.

plugins/dialog/plugin.js
```diff
@@ -2210,7 +2210,8 @@ CKEDITOR.DIALOG_STATE_BUSY = 2;
 			// So we need to invent a really funny way to make it work.
 			var myScrollHandler = function() {
 					scrollFunc();
-					arguments.callee.prevScrollHandler.apply( this, arguments );
+					// Replaced arguments.callee (#2741)
+					myScrollHandler.prevScrollHandler.apply( this, arguments );
```
f1ames: This code is executed only on IE < 7 (not supported) or in quirks mode (`env.ie6Compat = version < 7 || env.quirks;`) 😬 Please check manually (because we don't run unit tests in quirks mode) if it doesn't throw any errors in quirks mode. It should be executed when dialog is opened and window scrolled.

f1ames: I have already checked it and it works fine, but you may take another look.

f1ames:
Rebased on latest `major`.

adapters/jquery.js
```diff
@@ -157,10 +157,10 @@
 					editor.on( 'instanceReady', function( evt ) {
 						var editor = evt.editor;
 
-						setTimeout( function() {
+						setTimeout( function delayEditor() {
 							// Delay bit more if editor is still not ready.
 							if ( !editor.element ) {
-								setTimeout( arguments.callee, 100 );
+								setTimeout( delayEditor, 100 );
```
f1ames:
```suggestion
								setTimeout( waitForEditor, 100 );
```

Since this code does the same thing as the one already refactored, you may go with the same naming so I suggest `waitForEditor` here too.

adapters/jquery.js
```diff
@@ -157,10 +157,10 @@
 					editor.on( 'instanceReady', function( evt ) {
 						var editor = evt.editor;
 
-						setTimeout( function() {
+						setTimeout( function delayEditor() {
```
f1ames:
```suggestion
						setTimeout( function waitForEditor() {
```

PR #2934, opened 2019-03-07
Show current color in the Color Dialog

### LINKED ISSUE #2639: Show current color in the Color Dialog
## Type of report

Feature request.

1. Apply color to text from Color Dialog 
2. Re open the Color Dialog and already applied color is not set in hex code input box.

### Expected result

Color Dialog should show the already applied color value in the hex code input box

### Actual result

You can only apply the color to text but cannot retrieve the color value.

## Other details
2018-11-30_17h49_36

* Browser: CHROME
* OS: WINDOWS
* CKEditor version: latest
* Installed CKEditor plugins: colorbutton

Edit:

Consider https://github.com/ckeditor/ckeditor-dev/issues/2639#issuecomment-445239796

Dumluregn:

New feature

### This PR contains

- [x] Unit tests
- [x] Manual tests

## What changes did you make?

I added a new function to colordialog plugin (not changing dialog definition as it is used in other plugins). Right now it sets selection color as selected and shows its #code in color dialog. It does not change highlight or select color on palette yet, but besides that it works as intended. The code needs a lot of refactoring too. I just want to know if my proposal logic is valid and I can continue or I should find another way.

Closes #2639.


f1ames:
> The code needs a lot of refactoring too. I just want to know if my proposal logic is valid and I can continue or I should find another way.

Remember that you can always create Draft PR for such cases.

---

Please see `t/2639-2` branch (there is basically one commit - https://github.com/ckeditor/ckeditor-dev/commit/055ea2979160111ff561d447ffcf38c595325674). I prepared some quick PoC, you can see how you can reuse color detected on opening color dropdown without duplicating detection logic and how to pass it to color dialog via `show` event.

> not changing dialog definition as it is used in other plugins

Did you encountered issues with other plugins, what approach was problematic?

Dumluregn:
> Remember that you can always create Draft PR for such cases.

I thought reviewers do not look at drafts, sorry, should've ask about it before.

> Please see `t/2639-2` branch (there is basically one commit - [055ea29](https://github.com/ckeditor/ckeditor-dev/commit/055ea2979160111ff561d447ffcf38c595325674)). I prepared some quick PoC, you can see how you can reuse color detected on opening color dropdown without duplicating detection logic and how to pass it to color dialog via `show` event.

This is obviously a **much** better way to do this. Great to know we have such an awesome API for events - I didn't know I can edit event before passing it to another listener, lesson learned, thanks!

> > not changing dialog definition as it is used in other plugins
> 
> Did you encountered issues with other plugins, what approach was problematic?

I found out `colordialog` is used in several other plugins and wanted to avoid regression errors, but with event editing logic and line:
```
if ( evt.data.colordialog && evt.data.colordialog.selectedColor )
```
there should be no issues. I will use your PoC and come up with a better solution.

Dumluregn:
PR is ready for review.

plugins/colordialog/plugin.js
```diff
@@ -38,13 +39,22 @@ CKEDITOR.plugins.colordialog = {
 				}
 				callback.call( scope, color );
 			};
+			onShow = function( evt ) {
+				if ( options && options.selectionColor ) {
+					evt.data.colordialog = evt.data.colordialog || {};
```
Comandeer: I don't see a need to add additional namespace. I'd pass all the data directly into `evt.data`.

plugins/colordialog/plugin.js
```diff
@@ -38,13 +39,22 @@ CKEDITOR.plugins.colordialog = {
...
+					evt.data.colordialog.selectionColor = options.selectionColor;
```
Comandeer: Won't it be more futureproof to just do:

```javascript
evt.data = options;
```

?

tests/plugins/colordialog/manual/showcurrentcolor.md
```diff
@@ -0,0 +1,16 @@
+@bender-tags: feature, 4.12.0, 2639
+@bender-ui: collapsed
+@bender-ckeditor-plugins: wysiwygarea, toolbar, colorbutton, colordialog
+
+**Scenario 1:**
+
+1. Select text in editor.
+2. Click `Text Colour` button and `More Colours` button. (**Expected:** Color of text is by default focused, highlighted, selected and its RGB code is displayed.)
```
Comandeer: Expected are usually inserted in different way, as subheading inside step:

```markdown
2. Click `Text Colour` button and `More Colours` button.
	### Expected
	Color of text is by default focused, highlighted, selected and its RGB code is displayed.
```

plugins/colordialog/dialogs/colordialog.js
```diff
@@ -276,6 +276,22 @@ CKEDITOR.dialog.add( 'colordialog', function( editor ) {
...
 		minWidth: 360,
 		minHeight: 220,
+		onShow: function( evt ) {
+			if ( evt.data.colordialog && evt.data.colordialog.selectionColor ) {
+				var selectionColor = evt.data.colordialog.selectionColor;
+				var colors = evt.sender.parts.contents.getElementsByTag( 'td' );
+
+				dialog.getContentElement( 'picker', 'selectedColor' ).setValue( selectionColor );
+
+				for ( var i = 0; i < colors.count(); i++ ) {
```
Comandeer: You can use [`toArray`](https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_dom_nodeList.html#method-toArray) on `colors` and then use `CKEDITOR.tools.array.forEach` here, for iteration.

plugins/colordialog/dialogs/colordialog.js
```diff
@@ -276,6 +276,22 @@ CKEDITOR.dialog.add( 'colordialog', function( editor ) {
...
+					var paletteItem = colors.getItem( i );
+					var itemColor = CKEDITOR.tools.convertRgbToHex( paletteItem.getStyle( 'background-color' ) );
+					if ( ( selectionColor ) == itemColor ) {
```
Comandeer:
```suggestion
					if ( selectionColor === itemColor ) {
```

plugins/colordialog/dialogs/colordialog.js
```diff
@@ -276,6 +276,22 @@ CKEDITOR.dialog.add( 'colordialog', function( editor ) {
 		title: lang.title,
 		minWidth: 360,
 		minHeight: 220,
+		onShow: function( evt ) {
+			if ( evt.data.colordialog && evt.data.colordialog.selectionColor ) {
```
Comandeer: Reversing this conditional to create early return would allow us to flatten at least one indentation level:

```javascript
if ( !evt.data.colordialog || ! evt.data.colordialog.selectionColor ) {
	return;
}

// Rest of the code.
```

plugins/colordialog/dialogs/colordialog.js
```diff
@@ -276,6 +276,22 @@ CKEDITOR.dialog.add( 'colordialog', function( editor ) {
...
+				var selectionColor = evt.data.colordialog.selectionColor;
+				var colors = evt.sender.parts.contents.getElementsByTag( 'td' );
```
Comandeer: Please join these declarations to form one:

```javascript
var selectionColor = evt.data.colordialog.selectionColor,
	colors = evt.sender.parts.contents.getElementsByTag( 'td' );
```

Comandeer: Additionally, instead of `evt.sender` you can use `this` (as dialog's events are usually called in dialog's context).

plugins/colorbutton/plugin.js
```diff
@@ -86,6 +86,7 @@ CKEDITOR.plugins.add( 'colorbutton', {
 		function addButton( name, type, title, order, options ) {
 			var style = new CKEDITOR.style( config[ 'colorButton_' + type + 'Style' ] ),
 				colorBoxId = CKEDITOR.tools.getNextId() + '_colorBox',
+				colorData = {},
```
Comandeer: Instead of creating `colorData` variable, I'd think about extracting `onOpen` logic into separate function, `getColorFromSelection( editor )`. This function can be later used inside `clickFn` to pass data about color into dialog. IMO it would be more elegant this way.

Dumluregn: Although I agree that would be a bit cleaner way of doing this unfortunately I found it too difficult to extract `onopen` logic from its current scope. Therefore I suggest to leave the current solution as it is still not that ugly.

Dumluregn:
> I'm not sure if we should show default font color (not applied with color button) inside the dialog (colorbutton shows it as "automatic"). I think that we should show only the ones actually applied by the plugin. IMO showing default styles (not only colors, but also font families and sizes etc.) should be extracted to separate issue.

Do you suggest focusing color in the top left corner and leaving `selected color` box empty if `default` one is chosen (as it is currently on `master` branch)?

> I'm also not sure if we should consider text in whole selection (however it seems to be implemented this way in colorbutton). Most of our plugins use info from the beginning of the selection only (see #745).

I didn't change logic here, so if it should be changed I think we should create a separate task to avoid overlooking this later.

> There are no unit tests for this changes. It should be easy to use bot.dialog API to simulate opening of the dialog and checking what color is selected.

You're right, I will add unit test here.

> I also don't understand why tests/plugins/colordialog/colordialog.js was renamed?

I changed it because the name `addinghash.js` seemed more descriptive about what is tested here to me. If the tests for this issue are added to the same file the former name will be still proper and I will restore it.

> Sometimes focus is inconsistent with the value in the input

It happens in case of choosing color from outside a palette, i.e. by typing it's hex code. We discussed this case with @f1ames and decided it's better to focus default palette color than not set focus and highlight at all.

Comandeer:
> Do you suggest focusing color in the top left corner and leaving selected color box empty if default one is chosen (as it is currently on master branch)?

Yes, that's exactly what I'm proposing.

f1ames:
@Dumluregn please fix failing test (see CI):

image

tests/plugins/colordialog/_helpers/tools.js
```diff
@@ -0,0 +1,40 @@
...
+		resume( function() {
+			var dialog = evt.data,
+				selectedColor = dialog.getValueOf( 'picker', 'selectedColor' );
+			dialog.getButton( 'ok' ).click();
+
+			assert.areSame( expectedColor, selectedColor );
+		} );
+	} );
+
+	toolbarButton.click( editor );
```
Comandeer: IMO the more elegant way is to invoke appropriate editor's command, instead of simulating click on toolbar's button.

tests/plugins/colordialog/_helpers/tools.js
```diff
@@ -0,0 +1,40 @@
...
+	openColorDialog( toolbarButton );
+}
+
+function openColorDialog( button ) {
```
Comandeer: Can't `bot.dialog` be used for it?

tests/plugins/colordialog/manual/showcurrentcolor.md
```diff
@@ -0,0 +1,29 @@
+@bender-tags: feature, 4.12.0, 2639
+@bender-ui: collapsed
+@bender-ckeditor-plugins: wysiwygarea, toolbar, colorbutton, colordialog
+
+**Scenario 1:**
```
Comandeer: This scenario is not true, as in such case no color is selected and the first color in the dialog is focused.

tests/plugins/colordialog/omitdefaults.js
```diff
@@ -0,0 +1,23 @@
...
+/* bender-ckeditor-plugins: colordialog,wysiwygarea,toolbar,colorbutton */
+/* bender-include: _helpers/tools.js */
+/* global assertColorAtDialogShow */
+
+( function() {
+	'use strict';
+
+	bender.editor = true;
+
+	bender.test( {
```
Comandeer: I'd move this tests to `colordialog/colordialog.js`.

tests/plugins/colordialog/manual/showcurrentcolor.md
```diff
@@ -0,0 +1,29 @@
...
+
+3. Select different color and click `OK` button.
+    #### Expected
+    Text color has been changed. Text is selected.
+
+4. Click `Text Colour` button and `More Colours` button again.
+    #### Expected
+    New color is focused, highlighted, selected and its RGB code is displayed.
+
+**Scenario 2:**
```
Comandeer: This scenario does not work in IE8, where the color of selection beginning is selected inside the dialog. Additionally this scenario is not covered by unit tests.

tests/plugins/colordialog/_helpers/tools.js
Dumluregn: Some logic is nested inside functions which are not exposed outside plugin. To use editor's command `editor.getColorFromDialog()`I would have to either include significant part of plugin code in tests or explicitly pass the color that should be set (and then checked) in the dialog. Therefore I chose to simulate button click, which seems to be the best way to test if feature works correctly.

tests/plugins/colordialog/_helpers/tools.js
Dumluregn: Same as above, using `bot.dialog()` doesn't trigger some logic responsible for reading the color from selection. I even wrote such tests [here](https://github.com/ckeditor/ckeditor-dev/pull/2934/commits/b8e67ad706269cf1631138959cc573874d1b3f81) but later realised that I have to pass color by myself so it's basically simplified and poor version of previous tests and deleted them as redundant [here](https://github.com/ckeditor/ckeditor-dev/pull/2934/commits/77a929f211f5e50d2c34516273a5489806b9b122).

Comandeer:
Rebased onto latest `major`.

Dumluregn:
Ok, so there are two questions about colordialog look.

The first one is directly connected with this PR. What should be the colors of UI elements on dialog show if you selected text with inconsistent color, e.g. 
`<h1>Hello <span style="color:#3399cc">world</span>!</h1>`?
1. Default dialog look, i.e. focus and highlight set on the top left color (black), no selected color, no color code. The problem here is that focus is on black and selected color is white (actually it isn't but it looks so) so it's a little inconsistent.
2. Hands-off: no focus, no highlight, no selected color, no color code. Bad thing is that focus stays in the background (i.e. in the editor), so the accessibility is impaired.
3. Default dialog as in 1. but with the word `none` (or sth similar) in code field. This would be more readable for user but the color input is quite narrow and in some languages the equivalent of `none` may turn out to be too wide to fit in.

Of course if you have different ideas please give them.

The second question is about colordialog layout and do we want to change it (in the other PR then). 
I think right now the right column of dialog is a bit messy. The RGB code below the highlighted color looks like it belonged to the selected color. The highlight and selected color boxes should be the same size, and there should be a bit of space between. I'd propose something like this:

image

We could go even further and try to add some sample text (like `lorem ipsum`) in dialog that would change its color with either highlight or selected color, because the box is not always a good indicator of how the text will look.

WDYT?

Dumluregn:
As a reference - for inconsistent text color gdocs is selecting top left color (i.e. red).
Screenshot 2019-06-07 at 11 56 47

engineering-this:
>Default dialog look, i.e. focus and highlight set on the top left color (black), no selected color, no color code. The problem here is that focus is on black and selected color is white (actually it isn't but it looks so) so it's a little inconsistent.

I'm for this. To solve problem with selected color I'd consider just hiding it from dialog when no single color can be detected. Alternatively replace hash with `No selected color`, like in option 3., but with full sentence.

jacekbogdanski:
> The second question is about colordialog layout and do we want to change it (in the other PR then).
I think right now the right column of dialog is a bit messy. The RGB code below the highlighted color looks like it belonged to the selected color. The highlight and selected color boxes should be the same size, and there should be a bit of space between. I'd propose something like this:

Current UI looks better for me then centered inputs:
image

I also prefer leaving default color selection as it is, ie top left selection and empty `selected color` rectangles. It just feels natural/standard.

>Default dialog as in 1. but with the word none (or sth similar) in code field. This would be more readable for user but the color input is quite narrow and in some languages the equivalent of none may turn out to be too wide to fit in.

>Alternatively replace hash with No selected color, like in option 3., but with full sentence.

That would be very short :D :-1:

f1ames:
> 1. Default dialog look, i.e. focus and highlight set on the top left color (black), no selected color, no color code. The problem here is that focus is on black and selected color is white (actually it isn't but it looks so) so it's a little inconsistent.

I'm in favor of this option. The behavior is now consistent if no color or multiple colors are selected (so the cases when color cannot be detected). As @jacekbogdanski it feels like standard behaviour across many applications (and GDocs confirms that).

---

When it comes to improving color dialog UI maybe some inputs resizing and additional spacing could do the job:

image

Highlight hash could also be in the input like element (or just disabled input) to make looks more consistent, but I'm not sure if we should spend any time on polishing it. I would just report a ticket and see if there is any community feedback (also we may check if there are any related issue from the past) - @Dumluregn would you take care of it?

CHANGES.md
```diff
@@ -1,12 +1,13 @@
 ﻿CKEditor 4 Changelog
 ====================
 
-## CKEditor 4.12
+## CKEditor 4.12.0
```
Comandeer: We always omit 0 for major version in changelogs.

Dumluregn:
Yeah, that git history was odd indeed. I fixed it, rebased onto the latest `major` and deleted the last commit 'fixing' the editor version number in the changelog. We should be ready to go now.

PR #2938, opened 2019-03-07
GH template improvements

jacekbogdanski:
Small adjustments to remove duplication.


f1ames:
I have added default labels and change template names so they are ordered like:

1. Bug
1. Feature
1. Docs
1. Security
1. Task

Can you take a look @jacekbogdanski?

jacekbogdanski:
I'm wondering about pushing security even lower, thus it's actually warning rather than a functional button. WDYT?

f1ames:
@jacekbogdanski I was thinking about it, but tasks are usually only reported by us and security issues are also reported by community so I think it should be higher.

jacekbogdanski:
Makes sense and looks fine :+1:

PR #2955, opened 2019-03-13
Add method to disable tableselection feature for the given element

### LINKED ISSUE #2945: Tableselection plugin should expose function to ignore table element
## Type of report

Feature request

Currently, `tableselection` plugin works for any table in the editor. However, there may be some use cases when the table is a part of a different widget and shouldn't be handled by `tableselection` plugin.
I propose some simple API which could allow ignoring table (or its cells) when it's a child of passed element. E.g.

```javascript
CKEDITOR.plugins.tableselection.addIgnoredElement( widget.element );

// Removal function is needed to cleanup element and prevent 
// memory leaks e.g. when widget is destroyed.
CKEDITOR.plugins.tableselection.removeIgnoredElement( widget.element );
```

jacekbogdanski:

New feature

### This PR contains

- [x] Unit tests
- [x] Manual tests

**Note** Keyboard listeners don't need this improvement thus they handle faked table selection only which should be already disabled for the given element.

Closes #2945

### DIFF
--- a/core/selection.js
+++ b/core/selection.js
@@ -2018,6 +2018,7 @@
 
 			// Handle special case - fake selection of table cells.
 			if ( editor && editor.plugins.tableselection &&
+				// (#2945)
 				!editor.plugins.tableselection.isElementIgnored( ranges[ 0 ] && ranges[ 0 ].startContainer ) &&
 				CKEDITOR.plugins.tableselection.isSupportedEnvironment &&
 				isTableSelection( ranges ) && !isSelectingTable
--- a/core/selection.js
+++ b/core/selection.js
@@ -2018,6 +2018,7 @@
 
 			// Handle special case - fake selection of table cells.
 			if ( editor && editor.plugins.tableselection &&
+				!editor.plugins.tableselection.isElementIgnored( ranges[ 0 ] && ranges[ 0 ].startContainer ) &&
 				CKEDITOR.plugins.tableselection.isSupportedEnvironment &&
 				isTableSelection( ranges ) && !isSelectingTable
 			) {
--- a/core/selection.js
+++ b/core/selection.js
@@ -2018,10 +2018,10 @@
 
 			// Handle special case - fake selection of table cells.
 			if ( editor && editor.plugins.tableselection &&
-				// (#2945)
-				!editor.plugins.tableselection.isElementIgnored( ranges[ 0 ] && ranges[ 0 ].startContainer ) &&
 				CKEDITOR.plugins.tableselection.isSupportedEnvironment &&
-				isTableSelection( ranges ) && !isSelectingTable
+				isTableSelection( ranges ) &&
+				!isSelectingTable &&
+				!ranges[ 0 ]._getTableElement( { table: 1 } ).data( 'cke-tableselection-ignored' )
 			) {
 				performFakeTableSelection.call( this, ranges );
 				return;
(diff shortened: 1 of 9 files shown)


core/selection.js
```diff
@@ -2018,6 +2018,8 @@
 
 			// Handle special case - fake selection of table cells.
 			if ( editor && editor.plugins.tableselection &&
+				// (#2945)
+				!editor.plugins.tableselection.isElementIgnored( ranges[ 0 ] && ranges[ 0 ].startContainer ) &&
```
Comandeer: Shouldn't it be checked after all other checks? Additionally it would be probably better to invoke it on `range._getTableElement`, as whole table selection can be hooked directly in editable, so `ranges[ 0 ].startContainer` would be pointing to `body`/`div`.

plugins/tableselection/plugin.js
```diff
@@ -1122,12 +1181,28 @@
 			CKEDITOR.document.appendStyleSheet( this.path + 'styles/tableselection.css' );
 		},
 
+		isElementIgnored: function( target ) {
```
Comandeer: Why two of methods are located in `CKEDITOR.plugins.tableselection` and this one is inside `editor.plugins.tableselection`? I'd move them all to one place (preferably `CKEDITOR.plugins.tableselection`).

plugins/tableselection/plugin.js
```diff
@@ -1122,12 +1181,28 @@
...
+			if ( !target ) {
+				return false;
+			}
+
+			for ( var i = 0; i < this._ignoredElements.length; i++ ) {
+				if ( this._ignoredElements[ i ].contains( target ) ) {
```
Comandeer: This conditional does not work when table is compared to itself (as our implementation of `contains` does not allow passing the second parameter to the native `contains`). Probably it could be replaced with:

```suggestion
				if ( this._ignoredElements[ i ].contains( target ) || this._ignoredElements[ i ].equals( target ) ) {
```

plugins/tableselection/plugin.js
jacekbogdanski: The reason is keeping ignored elements as a part of an editor instance. Although, your idea about using custom data attribute instead of keeping reference into the element seems much better for me so we avoid an issue with accidental memory leaks and it handles other cases like already mentioned DOM unload.

plugins/tableselection/plugin.js
jacekbogdanski: This function should also work with the container keeping a table, something like:
```html
<div id="tableContainer">

jacekbogdanski:
You are right, I somehow missed a part where range container is not pointing directly on the selected element. Also, your idea about using attributes instead of DOM references seems much better.

Comandeer:
Rebased onto latest `major`.

jacekbogdanski:
I'm not sure what do you mean :thinking: 

Isn't this sentence:
> `[data-cke-*]` attributes are by default stripped from source mode output (as they're treated as internal ones), so there won't be any pollution.

denying fixing:

>However such attribute is lost during switching back to WYSIWYG mode, which enables table selection for the table.

As you wrote, `data-cke-tableselection-ignored` attribute will be stripped from the HTML as a part of internal attributes. How would you like to recover them moving back to `wysiwyg` mode? It's the reason why I proposed that this case could be delegated to the developer. From my perspective, this attribute will be used mostly for custom widgets using table element as a part of its content. In that case, the attribute should be a part of `downcast/upcast` methods.

Comandeer:
Rebased onto latest `major`.

PR #2960, opened 2019-03-15
Add unit tests covering headers manipulation in table plugin

### LINKED ISSUE #2882: Add unit tests covering headers manipulation in table plugin
## Type of report

Task

The `table` plugin doesn't have any tests covering headers manipulation (see https://github.com/ckeditor/ckeditor-dev/pull/1460#pullrequestreview-210111902). It would be good to add such test so we can prevent bugs like #1397 or #2881.

Dumluregn:

Task

### This PR contains

- [x] Unit tests
- Manual tests are not necessary

## What changes did you make?

Test coverage for changing table headers. Includes 64 test cases (table sizes: [1x1, 1x2, 2x1, 2x2] x input headers: [none, first column, first row, both] x output headers: [none, first column, first row, both]).

Closes #2882.


engineering-this:
>If that's because of some bug, then please report this bug, and these tests should be included along with the bug fix - we don't want to merge failing tests into our master branch.

After rethinking that - you can still add here such tests, but use `tests._should.ignore.${testCaseName}=true` to ignore them for now, with proper comment linking to the issue. You can see [here](https://github.com/ckeditor/ckeditor-dev/blob/master/tests/plugins/emoji/dropdown.js#L17-L26) how it works.

tests/plugins/table/headers1row1col.js
```diff
@@ -0,0 +1,75 @@
...
+			var bot = bender.editorBots.editor;
+			bot.setHtmlWithSelection( bender.tools.getValueAsHtml( input ).replace( ';', '' ) );
+
+			bot.dialog( 'tableProperties', function( dialog ) {
+				dialog.setValueOf( 'info', 'selHeaders', headerType );
+
+				dialog.fire( 'ok' );
+				dialog.hide();
+
+				var exp = bender.tools.getValueAsHtml( expected ).replace( ';', '' );
```
engineering-this: I think it's fine to overwrite original `expected` value with new one. So declaring new var `exp` could be omitted.

```suggestion
				expected = bender.tools.getValueAsHtml( expected ).replace( ';', '' );
```

tests/plugins/table/headers1row1col.js
```diff
@@ -0,0 +1,75 @@
...
+
+				assert.beautified.html( exp.replace( '^', '' ),
```
engineering-this: Follow up.
```suggestion
				assert.beautified.html( expected.replace( '^', '' ),
```

tests/plugins/table/headers1row1col.js
```diff
@@ -0,0 +1,75 @@
...
+		'1 row, 1 col, both -> none': assertHeadersCorrectnesssAfterManipulation( 'header-both', 'header-none', 'none' ),
+
+		'1 row, 1 col, both -> col': assertHeadersCorrectnesssAfterManipulation( 'header-both', 'header-col', 'col' ),
+
+		'1 row, 1 col, both -> row': assertHeadersCorrectnesssAfterManipulation( 'header-both', 'header-row', 'row' ),
+
+		'1 row, 1 col, both -> both': assertHeadersCorrectnesssAfterManipulation( 'header-both', 'header-both', 'both' )
+	} );
+
+	function assertHeadersCorrectnesssAfterManipulation( input, expected, headerType ) {
```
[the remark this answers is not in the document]
Dumluregn: I think looping through tests would be uncomfortable as there are some failing tests that you may want to run separately, so I stayed with your first proposal and extracted just testing function.

f1ames:

tests/plugins/table/_helpers/headersCorrectness.js
```diff
@@ -0,0 +1,22 @@
+/* exported assertHeadersCorrectnesssAfterManipulation */
+
+function assertHeadersCorrectnesssAfterManipulation( input, expected ) {
+	'use strict';
+
+	return function() {
+		var bot = bender.editorBots.editor;
+		bot.setHtmlWithSelection( bender.tools.getValueAsHtml( input ).replace( ';', '' ) );
```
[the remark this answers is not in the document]
f1ames: Ok, I see it fails if you remove another `replace` from line `16`. There is a dedicated method for comparing HTML which normalizes it before comparison, `assert.isInnerHtmlMatching( ... )` or the one you have used `assert.beautified.html`.

So when setting HTML `.replace( ';', '' )` is not needed because browser will format it in it's own way. And for asserting you can use `assert.beautified.html` as now but you need to pass proper configuration so styles are normalized (see [benderjs/benderjs-yui-beautified](https://github.com/benderjs/benderjs-yui-beautified/blob/a3ca835bb06c6dca6241c02112f21c1b55daa1f0/lib/client.js#L81) and [_benderjs/ckeditor/static/tools.js](https://github.com/ckeditor/ckeditor-dev/blob/c6969047a007e10d9bb2907ba11d0f25de1295d9/tests/_benderjs/ckeditor/static/tools.js#L286)).

tests/plugins/table/_helpers/headersCorrectness.js
```diff
@@ -0,0 +1,22 @@
+/* exported assertHeadersCorrectnesssAfterManipulation */
+
+function assertHeadersCorrectnesssAfterManipulation( input, expected ) {
```
f1ames: I would rename it to `assertHeadersAfterManipulation` (btw, there is triple `s` in the name now), because `assert` on it's own indicates checking if something is correct.

f1ames: Also I see this method accepts only 2 parameters while in the most tests (3/4 files) you pass 3. The fact it works is that 2nd and 3rd parameter are always the same:D

So you may just use 2 params version. Or make it more explicit (and more readable), e.g. `assertHeadersAfterManipulation( inputHtmlId, expectedHtmlId, headers )`

tests/plugins/table/headers1row1col.html
```diff
@@ -0,0 +1,40 @@
...
+		<thead>
+			<tr>
+				<th scope="col">^cell 1.1</th>
+			</tr>
+		</thead>
+		<tbody>
+		</tbody>
+	</table>
+</textarea>
+<textarea id="both">
```
f1ames: Isn't `row` and `both` html the same? If so there is no need to duplicate it (and also no need to duplicate test cases).

tests/plugins/table/headers1row2col.html
```diff
@@ -0,0 +1,44 @@
...
+			<tr>
+				<th scope="col">^cell 1.1</th>
+				<th scope="col">cell 1.2</th>
+			</tr>
+		</thead>
+		<tbody>
+		</tbody>
+	</table>
+</textarea>
+<textarea id="both">
```
f1ames: Isn't `row` and `both` html the same? If so there is no need to duplicate it (and also no need to duplicate test cases).

tests/plugins/table/_helpers/headersCorrectness.js
f1ames: > I would rename it to `assertHeadersAfterManipulation`...

I had some enlightenment moment and I think the function name should be something like `testHeadersManipulation`. This function basically creates and runs entire test instead of only asserting headers so it should be named accordingly.

tests/plugins/table/headers1row1col.html
engineering-this: Maybe to make clear that one fixture is used for both `row` and `both` cases, `id="both"` could be changed to `class="row both"`, so when looking for element in TC selector would match actual searched case. Otherwise it might be a bit inconvenient if fixture named `both` is uses for `row` case.

Dumluregn: I did as @engineering-this suggested - I changed `id` attribute to `class` and reused table code clearly indicating to which test cases it belongs.

tests/plugins/table/headers1row2col.html
Dumluregn: I did as @engineering-this suggested - I changed `id` attribute to `class` and reused table code clearly indicating to which test cases it belongs.

tests/plugins/table/_helpers/headersCorrectness.js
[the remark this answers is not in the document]
Dumluregn: That is one cool API, thanks!

tests/plugins/table/headers1row2col.html
```diff
@@ -0,0 +1,32 @@
+<textarea class="none">
```
f1ames: I noticed `getValue()` in the `testHeadersManipulation()` function and started to wonder why you used it instead of `getHtml()`. I think you may just use `<div>`s here and in the `testHeadersManipulation()` function use `getHtml()` instead of `getValue()`.

f1ames: So please update all the test files.

PR #2967, opened 2019-03-20
Fix undefined var on remove listener

jackwickham:
Fix a bug causing the error `me is undefined` when calling `removeListener` on the result of `CKEDITOR.event.prototype.on()` when the same function has previously been attached.

A (contrived) example error is
```js
function(editor) {
    this.listen = function() {
        this.listener = editor.on( 'change', this.respond, this);
    }

    this.respond = function(e) {
        console.log("Changed", e); // This will only be called once, regardless of removing the listener, because attaching the same function multiple times has the same effect as adding it once.
        if (this.listener) {
            this.listener.removeListener(); // this will fail because this.listener is from the second call
            this.listener = null;
        }
    }

    this.listen();
    this.listen();
}
```

Yes

### This PR contains

- [x] Unit tests
- [ ] Manual tests

## What changes did you make?
Move the assignment `var me = this` higher in `CKEDITOR.event.prototype.on` (out of an if statement) to ensure that it is always populated, to prevent an error.

I also added tests to cover both that this error does not occur (`test_duplicateListeners_removeListener_2`), and additional tests to verify that when a function is added as a listener multiple times, the function only actually gets attached once.


jackwickham:
I can create an issue for this PR if needed?

f1ames:
@jackwickham thanks for the PR!

> I can create an issue for this PR if needed?

If you could add one (or check it it exists already) it would be great. We will try to take a look on this PR soon.

Issue #2484, opened 2018-10-16
Empty span tag removed automatically


kite57:
This is the bug. I am facing strange Issue in ckeditor, following is the scenario

1. I have added table with content
2. Following is the html of Table

```
<tr>
   <td colspan="2">
	   <span style="font-size:12px;"><b>Final Diagnose :</b></span>
    </td>
    <td colspan="2">
 		<span style="font-size:12px;"><b>[Final_Diagnose]</b></span>
    </td>
</tr>
<tr>
     <td colspan="4"><span style="font-size:12px;"></span></td>
</tr>

```
3.When I am saving this html, blank span tag removed and only `<td>` tag remains, so that when i am setting font size to whole html(using select all and assign font-size), and if i add anything in span tag then font size will be applied but if i did not add any text content than span tag with font size attribute removed.

After Saving Html Code

```
<tr>
   <td colspan="2">
	   <span style="font-size:12px;"><b>Final Diagnose :</b></span>
    </td>
    <td colspan="2">
 		<span style="font-size:12px;"><b>[Final_Diagnose]</b></span>
    </td>
</tr>
<tr>
     <td colspan="4"></td>
</tr>

```

### Expected result

Empty Span Tag with attributes should not be  removed,So when i add any content in that span, it will take saved font size

### Actual result

Empty Span Tag with attributes removed, so when i add some content, default font size applied, it is not taking saved font size

## Other details

* Browser:Chrome(Latest Version)
* OS:(Windows7)
* CKEditor version:3.6

msamsel:
This behaviour is working as designed. Empty spans are considered as invisible and are removed from content to not pollute outputted HTML. You should add `&nbsp;` inside  such span to preserve it.
To modify this behaviour you might want to look closer to [CKEDITOR.dtd](https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_dtd.html#property-S-removeEmpty) class.

PR #2987, opened 2019-03-27
 Fix for: `editor.insertHtml` and `editor.insertElement` pollutes editable with empty spans

### LINKED ISSUE #2813: `editor.insertHtml` pollutes editable with empty spans
## Type of report

Bug

1. Open https://codepen.io/anon/pen/QYZZOG
2. Create selection in editor as following: `Foo {Bar}`.
3. Press button `insert` below editor.

### Expected result

Selected part is removed. And there is no empty element left.

### Actual result

Empty span is left behind editor which is visible by red border.

## Other details

These empty elements can't be focused (checked Chrome, Safari, Firefox), and they are removed on `getData`. They should be removed right away on `insertHtml`, because when styled they might be visible in editor.

https://github.com/ckeditor/ckeditor-dev/blob/002ec74de01c50d48e1ba2101785126a56407eeb/core/editable.js#L1719-L1723
☝️ This is where they ale splitted upon insertions, and when splitting leaves empty elements it's best place to clean them up.

engineering-this:

Bug fix

### This PR contains

- [X] Unit tests
- [X] Manual tests

## What changes did you make?

Inside `prepareRangeToDataInsertion` I've added check that will remove created nodes for which `isEmptyInlineRemoveable` is true.

Closes #2813


core/editable.js
```diff
@@ -1717,14 +1718,24 @@
 			// Split inline elements so HTML will be inserted with its own styles.
 			path = range.startPath();
 			if ( ( node = path.contains( isInline, false, 1 ) ) ) {
-				range.splitElement( node );
+				clone = range.splitElement( node );
```
jacekbogdanski: A bit misleading thus this element is not a clone but the late element after the split. Maybe rename it into `endNode`?

core/editable.js
```diff
@@ -1717,14 +1718,24 @@
...
 				that.inlineStylesRoot = node;
 				that.inlineStylesPeak = path.lastElement;
 			}
 
 			// Record inline merging candidates for later cleanup in place.
 			bm = range.createBookmark();
 
+			// When called by insertHtml remove empty element created after splitting (#2813).
+			if ( isHtml && clone && clone.isEmptyInlineRemoveable() ) {
```
jacekbogdanski: Would be nice to extract both `span` node removals into a separate function to avoid duplication.

tests/core/editable/inserthtmlintospan.js
```diff
@@ -0,0 +1,33 @@
+/* bender-tags: editor,insertion */
```
jacekbogdanski: Do we really need a separate file for these tests? It looks like they could be placed inside `tests/core/editable/insertion.js`. WDYT?

tests/core/editable/inserthtmlintospan.js
```diff
@@ -0,0 +1,33 @@
...
+		'test insertHtml when selection covers span': assertInsertHtml( '<span>{foobar}</span>', '<div>div</div>' )
+	} );
+
+	function assertInsertHtml( initial, expected ) {
+		return function() {
+			var editor = this.editorBot.editor;
+
+			bender.tools.selection.setWithHtml( editor, initial );
+
+			editor.insertHtml( '<div>div</div>' );
```
jacekbogdanski: I'm missing tests for `editor.insertHtmlIntoRange` which also uses `prepareRangeToDataInsertion` private function.

engineering-this:
>I'm also wondering if this PR shouldn't be a part of major release, because it may cause some breaking changes if someone gets around spans removal and using them as a part of some feature. WDYT?

It is possible that someone is somehow preserving empty spans (#2484). However this fix affects elements that were created or altered by removing selected content from selection containers thus it shouldn't affect empty spans that were in editor before insertion.

Also after looking at [element.isEmptyInlineRemovable](https://github.com/ckeditor/ckeditor-dev/blob/master/core/dom/element.js#L1010-L1011) it looks like changing `CKEDITOR.dtd.$removeEmpty[ elementName ]` to `false` should be sufficient to make desired element safe from being removed when empty.

tests/core/editable/inserthtmlintospan.js
engineering-this: I've added such tests. Small note, it's a method of `editable` 😉

engineering-this:
>I'm finding unit tests inside tests/core/editable/insert.js hightly unreadable. I see that you tried to save some LOC but it shouldn't be done in expense of readibility. As an example, the first test checking editor.insertHtml could be as simple as:

I can rework test, so each test case is explicitly declared, however I'd prefer to use one general test callback which will test each case based on `options` param like we do in many test files.

engineering-this:
>It looks like editor.insertHtml doesn't work correctly with block elements and your tests exploit this issue:

Extracted to #3042

engineering-this:
>It also would be great to add tests verifying that empty edge spans are preserved for both inserted html and editor content. E.g. you could check if div html into <span>foo{baz} </span> preserves the space. We will be sure that your changes don't touch user content.

I've added such test case, but I need to mention that when you press <kbd>space</kbd> in editor, when selection is at the end of any element, than `&nbsp;` will be inserted, because otherwise space won't be rendered by browser, so `&nbsp;` entity is used in tc.

tests/core/editable/insertion.js
```diff
@@ -159,7 +159,79 @@
...
+
+	function addInsertionTests( tests ) {
+		CKEDITOR.tools.array.forEach( tests, addInsertionTest );
+	}
+
+	function addInsertionTest( options ) {
+		var methods = [ 'insertHtml', 'insertHtmlIntoRange', 'insertElement', 'insertElementIntoRange', 'insertText' ];
+
+		CKEDITOR.tools.array.forEach( methods, function( methodName ) {
+			tests[ options.name.replace( 'insert', methodName ) ] = function() {
```
jacekbogdanski: Small suggestion to compose a test method name rather than replace `insert` string:

` tests[ 'test ' + methodName + ' ' + options.name ] = ...`

seems less hacky for me and more resistant for incorrect usage e.g. if we decide to add new test without `insert` prefix.

tests/core/editable/insertion.js
```diff
@@ -159,7 +159,79 @@
...
+					assertInsertText( editor, options.initial.replace( /{.*}/, 'text' ), 'text' );
+				} else {
+					assertInsertionMethod( editor, options.expected, methodName, options.data  );
+				}
+			};
+		} );
+	}
+
+	function assertInsertionMethod( editor, expected, methodName, html ) {
+		var useRange = methodName.indexOf( 'Range' ) >= 0,
```
jacekbogdanski: Maybe you could make it even more readable by using a `switch` statement instead of string sniffing directly from `addInsertionTests`?

tests/core/editable/insertion.js
```diff
@@ -159,7 +159,79 @@
...
+			initial: '<span>foo{bar}&nbsp;</span>',
+			data: '<div>div</div>',
+			expected: '<span>foo</span><div>div</div><span>&nbsp;</span>'
+		}
+	] );
+
+	bender.test( tests );
+
+	function addInsertionTests( tests ) {
+		CKEDITOR.tools.array.forEach( tests, addInsertionTest );
```
jacekbogdanski: `tests` variable already exists in this file. I know that function is very small, nevertheless, maybe you could rename it to avoid using the same name? Hoisting makes name covering a bit unreadable.

tests/core/editable/insertion.js
```diff
@@ -159,7 +159,79 @@
...
 
 			assert.isInnerHtmlMatching( '<p id="p1">foo</p><p>bam^@</p><p>bar@</p>',
-				bender.tools.selection.getWithHtml( editor ), { compareSelection: true, normalizeSelection: true } );
+				tools.selection.getWithHtml( editor ), { compareSelection: true, normalizeSelection: true } );
 		}
-	} );
+	};
+
+	// Add tests for (#2813).
+	addInsertionTests( [
```
jacekbogdanski: Small duplication here. I would suggest deciding between removing the comment (except tag reference) or inlining function.

CHANGES.md
```diff
@@ -12,6 +12,7 @@ Fixed Issues:
 
 * [#2672](https://github.com/ckeditor/ckeditor-dev/issues/2672): Fixed: When resizing [Enhanced Image](https://ckeditor.com/cke4/addon/image2) to minimum size with a resizer the image dialog doesn't show actual values.
 * [#1478](https://github.com/ckeditor/ckeditor-dev/issues/1478): Fixed: Custom colors added to [Color Button](https://ckeditor.com/cke4/addon/colorbutton) via [`config.colorButton_colors`](https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_config.html#cfg-colorButton_colors) in form label/code don't work correctly.
+* [#2813](https://github.com/ckeditor/ckeditor-dev/issues/2813): Fixed: [editor.insertHtml](https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_editor.html#method-insertHtml) and [editor.insertelement](https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_editor.html#method-insertElement) pollutes editable with empty spans.
```
jacekbogdanski: PR also covers `insertElementIntoRange` and `insertHtmlIntoRange`. Maybe instead of listing method names, we could make it more general, like:

`Fixed: Editor HTML insertion methods pollute the content with empty spans.`?

engineering-this:
Apparently space character ' ' should be used instead of `&nbsp;` on IE. This change result in multiple tests failing in IEs. I've extracted it to separate ticket #3061, as it seems to be unrelated with the issue we are fixing here. Failing test are not ignored on IE.

tests/core/editable/insertion.js
```diff
@@ -4,7 +4,8 @@
 	'use strict';
 
 	var doc = CKEDITOR.document,
-		tools = bender.tools;
+		tools = bender.tools,
+		space = CKEDITOR.env.ie ? ' ' : '&nbsp;';
```
jacekbogdanski: You are declaring space here for IE issue, but at the same time IE tests are ignored and the test is working on Edge without this modification.

tests/core/editable/insertion.js
```diff
@@ -159,7 +160,96 @@
...
+		}, {
+			name: 'when selection covers span',
+			initial: '<span>{foobar}</span>',
+			data: '<div>div</div>',
+			expected: '<div>div</div>'
+		}, {
+			name: 'when selection is in the middle of span',
+			initial: '<span>foo{bar}' + space + '</span>',
+			data: '<div>div</div>',
+			expected: '<span>foo</span><div>div</div><span>' + space + '</span>',
```
jacekbogdanski: It would be nice to add test covering the case when selection is preceded with space (`<span> {foo}</span>`) because you are removing it at https://github.com/ckeditor/ckeditor-dev/blob/2669504728055ff89038f8bf87828cb6925868ae/core/editable.js#L468-L469

tests/core/editable/insertion.js
engineering-this: I think this value should be here, as this is used for the IE, however because of other bug test is ignored. Once #3061 is resolved test can be just unignored.

engineering-this:
>But the tests are ignored on IE 🤔

For now only tests affected by #3061 are ignored on IE.

core/editable.js
```diff
@@ -2273,7 +2288,7 @@
 				wrapper = doc.createText( '{cke-peak}' ),
 				limit = that.inlineStylesRoot.getParent();
 
-			while ( !element.equals( limit ) ) {
+			while ( element && !element.equals( limit ) ) {
```
jacekbogdanski: What is the reason for this change? If it's required, could you cover it with tests?

engineering-this: Without this conditional sometimes I had an error when clicking randomly clicking in editor and then using one of test buttons. I couldn't reproduce some weird selection which caused the issue.

engineering-this:
>You didn't refer to this:
>
> >It would be nice to add a test case when you are inserting an empty element.

😳

tests/core/editable/manual/insert.md
```diff
@@ -0,0 +1,18 @@
...
+
+1. After each case use `reset editor` button.
+
+1. Select word `bar`.
+
+1. Press button.
+
+## Expected
+- `Insert text` button: Inserted word `text` replaces selected word inside border.
+- Any other button : there is one red border on the left of inserted `div`.
```
jacekbogdanski: Not true for `insertElement/insertElementInotRange` functions where `div` is placed below rectangle. I would suggest rewording manual test a bit to make it more general:

```
## For each insert button

1. After each case use `reset editor` button.

1. Select the word `bar`.

1. Press button.

## Expected
- `Insert text` button: `text` replaces selected word inside the rectangle
- Any other button: `div` replaces selected word outside the rectangle

## Unexpected
Clicked button causes the creation of an additional rectangle

**NOTE:** Insert HTML functions inserts `div` element as a plain text. Caused by [#3042](https://github.com/ckeditor/ckeditor-dev/issues/3042) issue.
```

In that case, #3042 issue won't invalidate the manual test.

core/editable.js
jacekbogdanski: That doesn't look like a good reason to change this code without tests. I would suggest trying to reproduce the issue or if it's no longer reproducible, remove the condition. It looks like an element doesn't have a parent at some point which means it's not attached to the DOM. It may be caused by a different bug and hiding it will make harder to find a real problem in the future.

Are you able to reproduce the issue without your changes? Please, try to find a real problem, we cannot leave it untested.

engineering-this: I see your point. I can't reproduce mentioned issue, so I'm re moving the condition.

engineering-this:
Rebased onto latest `major` branch.

CHANGES.md
```diff
@@ -12,6 +12,7 @@ Fixed Issues:
 
 * [#2672](https://github.com/ckeditor/ckeditor-dev/issues/2672): Fixed: When resizing [Enhanced Image](https://ckeditor.com/cke4/addon/image2) to minimum size with a resizer the image dialog doesn't show actual values.
 * [#1478](https://github.com/ckeditor/ckeditor-dev/issues/1478): Fixed: Custom colors added to [Color Button](https://ckeditor.com/cke4/addon/colorbutton) via [`config.colorButton_colors`](https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_config.html#cfg-colorButton_colors) in form label/code don't work correctly.
+* [#2813](https://github.com/ckeditor/ckeditor-dev/issues/2813): Fixed: Editor HTML insertion methods pollute the content with empty spans.
```
f1ames: Could you list fixed methods here (the public API) so anyone can see which were affected by this fix?

tests/core/editable/insertion.js
```diff
@@ -193,11 +193,21 @@
...
 			expected: '<span>foo</span><div></div>',
 			ignore: CKEDITOR.env.ie && CKEDITOR.env.version < 9 // IE8 fills empty element with `&nbsp;`, so we can skip this test.
+		}, {
+			name: 'when selection covers nested span',
+			initial: '<p><span>{foobar}</span></p>',
+			data: '<div>div</div>',
+			expected: '<div>div</div>'
+		}, {
+			name: 'when selection covers span with empty spans on boundaries',
+			initial: '<span></span><span>{foobar}</span><span></span>',
```
engineering-this: This test passes for me, but I'm not sure about it. Empty spans have no purpose.

- They are removed on `setData`, so going to source and back to wysiwyg makes them disappear.
- You can't place caret in empty span.
- https://github.com/ckeditor/ckeditor-dev/blob/cb1dd2945a20ba481b5f57560fb0bb96ae2d739c/core/dom/range.js#L724-L726

f1ames: This test also passes for me :D I have added it because empty `spans` are removed by the mechanism introduced in this PRa. So its purpose is to make sure that empty spans not added during splitting are not removed. But if you think it's redundant feel free to remove it.

The failing test (mentioned in https://github.com/ckeditor/ckeditor-dev/pull/2987#pullrequestreview-240620712) is the one above - `when selection covers nested span`.

engineering-this: It seems to work on all browsers, so I'm keeping it.

engineering-this:
The reason test case added by @f1ames failed is that normalization function used a regex which cut all the paragraphs. I've corrected it, now test passes. The reason for cuttion paragraphs is to make sure all environments and all tested functions has same output. Changes affects only inline elements, so it's safe to get rid of empty paragraphs.

>Do you think it would also make sense to add more complex scenarios to unit tests?
I don't think anything beyond what we are testing here might be affected.

f1ames:

engineering-this:
It looks like an issue with selection.
When there is only one word in span, double click will create given range:
```
start: text at 0
end: li after span
```

This case produces above issue. Same behaviour can be achieved in Chrome when triple clicking. I think it will be much easier to fix it with #3118 than updating range.deleteContents with new logic.

f1ames:
> It looks like an issue with selection.
When there is only one word in span, double click will create given range:
> ```
> start: text at 0
> end: li after span
> ```
> 
> This case produces above issue. Same behaviour can be achieved in Chrome when triple clicking. I think it will be much easier to fix it with #3118 than updating range.deleteContents with new logic.

Makes sense👍 Could you create an issue for this case and add a link to manual test so anyone who will encounter it sees it's covered. cc @engineering-this

engineering-this:
Extracted issue #3161

f1ames:
Rebase on the latest `major`.

engineering-this:
I improved regex for normalising output. It will remove empty paragraphs, also added rule that paragraph has to be at the end of string. This will keep us safe from regression when unexpected paragraph appears in the middle of editor content.

f1ames:

engineering-this:
Although it looks similar to the original bug it's caused by something else, so I'm for extracting it as a follow-up issue.

f1ames:
> Although it looks similar to the original bug it's caused by something else, so I'm for extracting it as a follow-up issue.

I have extracted it to https://github.com/ckeditor/ckeditor-dev/issues/3177. Since it is present from 4.5 (or even older), there is no need to fix it unless really necessary and we should be careful to make sure nothing breaks since it touches multiple block elements.

PR #2988, opened 2019-03-27
Promise support

### LINKED ISSUE #2962: Promise support
## Type of report

Feature request

It seems to be a good time to provide [Promise](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise) support for CKEditor4. It's already required and somehow implemented in our testing environment (see https://github.com/ckeditor/ckeditor-dev/pull/2752) although it would be nice to have dedicated `CKEDITOR.tools.promise` as a wrapper around native `Promise` object which will also handle polyfilling for IE. We should discuss polyfill library options - it doesn't make sense to spend time reinventing the wheel again.

Some requirements / things to discuss :
1. Use native `Promise` everytime when it's possible.
1. Load polyfill library asynchronously only for the browsers with no Promise support when needed.
1. Choose polyfill library with the same API as [Promise](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise).
1. Hide `Promise` under `CKEDITOR.tools.promise` namespace so we won't overwrite native objects.

jacekbogdanski:

### This PR contains

- [x] Unit tests
- [ ] Manual tests

## What changes did you make?

After some research, it looks like [ES6-Promise](https://github.com/stefanpenner/es6-promise) polyfill should suit our needs. It's pretty lightweight (6.17 KB / 2.4 KB gzipped) and seems to have the same API as native Promise. It's also recommended by [MDN Promise docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise#See_also).

Closes #2962


jacekbogdanski:
I realized the issue here - the whole `core` folder is minified by the builder, so there is no way to add the polyfill conditionally in that place. Instead, I created `vendor` folder for such 3rd party libs. It seems to be included in a built version and additionally minified, so we can use the full version for debugging.

I still would like to keep `core/promise.js` file instead of inlining this code inside `core/tools.js`, although in fact, it's tiny now.

Please, see https://github.com/ckeditor/ckeditor-docs/pull/242 for the error code (you may wanna review this one also).

engineering-this:
>Please, see ckeditor/ckeditor-docs#242 for the error code (you may wanna review this one also).

I think it should be reviewed after promise is merged.

jacekbogdanski:
I'm wondering how to fix the test for built version https://github.com/ckeditor/ckeditor-dev/pull/2988#pullrequestreview-227118520 - maybe it's a hint for poorly written API? 

tl;dr `core/promise.js` module is loaded when bootstrapping CKEditor and contains logic responsible for fetching promise polyfill (very simple, yet still requires testing). I'm trying to verify the sad path when there is no promise polyfill available at `vendors/promise.js`:

Options which comes to my mind:

* Enable unit test only for dev version
* Refactor promise loading into a separate function, something like `CKEDITOR.promiseLoader.load`, so it can be easily tested, although adds mostly unused API and also pollutes `CKEDITOR` namespace
* Resign from `CKEDITOR.error` message - at some point editor will blow with missing `CKEDITOR.tools.promise` type

Honestly, I'm feeling like I did some overengineering here and there is no reason for an explicit error message. WDYT?

engineering-this:
Error message is helpful. I'd ignore test under condition: `!CKEDITOR.loader`. It's unlikely that message will work correctly in dev and not in build. Additionally we can test it manually by deleting polyfill file.

jacekbogdanski:
> Additionally we can test it manually by deleting polyfill file.

engineering-this:
Missing an error message doesn't really break things, so I'd just check it before merging PR.

tests/_benderjs/ckeditor/static/tools.js
```diff
@@ -1258,11 +1223,11 @@
 					return function() {
 						var promise = test.apply( this );
 
-						if ( Q.isPromise( promise ) ) {
+						if ( promise ) {
 							promise.then( function() {
 									resume();
 								} )
-								.fail( function( err ) {
+								[ 'catch' ]( function( err ) {
```
f1ames: Why `[ 'catch ]` and not `.catch`? Does it break anything?

tests/core/tools/promise.js
```diff
@@ -0,0 +1,60 @@
...
+		},
+
+		tearDown: function() {
+			window.Promise = this.Promise;
+		},
+
+		// (#2962)
+		'test promise initialization': function() {
+			if ( window.Promise ) {
+				assert.areSame( window.Promise, CKEDITOR.tools.promise, 'Native Promise should be enabled' );
```
f1ames: Would it make sense to check

```js
assert.isUndefined( window.ES6Promise, 'Polyfill Promise should be loaded' );
```

here too?

core/promise.js
```diff
@@ -0,0 +1,64 @@
...
+	 *				timestamp = ( new Date() ).getTime();
+	 *			} catch ( e ) {
+	 *				reject( e );
+	 *			}
+	 *			resolve( timestamp );
+	 *		}, 5000 );
+	 *	} );
+	 * ```
+	 *
+	 * @param {Function} resolver
```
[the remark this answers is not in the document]
f1ames: And there are more functions available:

image

Should we list them all here?

tests/_benderjs/ckeditor/static/tools.js
engineering-this: https://github.com/lahmatiy/es6-promise-polyfill#usage-in-ie9

`catch` is reserved in IEs, I was thinking about adding an alias, so no `[ '' ]` characters are needed, apparently I didn't ask this in comment before.

f1ames: Hmmm, does it mean that every code we create will need to use `[ 'catch' ]` to support IE8? 😭 I think alias could be useful here, instead we are sentenced to using `[ 'catch' ]`. That's bad 😬

LICENSE.md
```diff
@@ -61,7 +61,7 @@ The following libraries are included only in the development version of CKEditor
 * CKBuilder - Copyright (c) 2012-2019, CKSource - Frederico Knabben.
 * CKLangTool - Copyright (c) 2012-2019, CKSource - Frederico Knabben.
 * Optimist - Copyright 2010 James Halliday (mail@substack.net).
-* Q - Copyright 2009–2014 Kristopher Michael Kowal.
+* ES6Promise - Copyright (c) 2014 Yehuda Katz, Tom Dale, Stefan Penner and contributors.
```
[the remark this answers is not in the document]
jacekbogdanski: Thanks, I just ignored this line :D

core/promise.js
```diff
@@ -0,0 +1,64 @@
...
+				} );
+			}
+		} );
+	}
+
+	/**
+	 * Alias for [Promise](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise)
+	 * object representing asynchronous operation.
+	 *
+	 * Uses [ES6-Promise](https://github.com/stefanpenner/es6-promise) as a polyfill.
```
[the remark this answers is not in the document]
jacekbogdanski: AFAIK it's all IE's, but we are not sniffing browsers, so I would rather leave it to the user, eventually link browser support. Something like 
> Visit [Promise Browser Compatibility](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise#Browser_compatibility) to learn more.

It's always easier to maintain such reference than listing browsers by hand.

core/promise.js
```diff
@@ -0,0 +1,64 @@
...
+	 * Note that the polyfill won't be loaded if a browser supports native Promise object.
+	 *
+	 * See [MDN Promise documentation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise) for
```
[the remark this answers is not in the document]
jacekbogdanski: I would actually link both.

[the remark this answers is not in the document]
jacekbogdanski: Promise docs are already linked so it makes sense.

core/promise.js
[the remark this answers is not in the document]
jacekbogdanski: > Isn't it just a Promise object returned (with resolve and reject functions)? So it should be * @param {Object} promise or * @param {Promise} promise if we are thinking about introducing promise type to docs (not sure TBH.)

You confuse `Promise.reject`/`Promise.resolve` static methods with resolver function. It's void return type function which allows you to promisify async operation. I suspect that you are talking about constructor return type, but in that case, we don't have a choice to leave current `CKEDITOR.tools.promise` type.

tests/_benderjs/ckeditor/static/tools.js
jacekbogdanski: How often do you need `catch` closure? For simple grep on our whole source code, I don't see it's used extensively - at the end, Javascript is not Python :wink:

Creating an alias for `catch` method would break the whole concept of making this implementation an alias for native `Promise` type. It would significantly complicate this implementation because if we start playing with changing native type, we should return wrapper inheriting/mapping this type.

Also, we will lose something I like to call ubiquitous API (referencing to BDD) where you just need to know the name of the type, and you don't need to read docs - you already know the API because it's the standard one.

tests/core/tools/promise.js
jacekbogdanski: We can, to make sure polyfill is not loaded.

jacekbogdanski:
>I wonder if we could move ES6Promise js file from vendor dir to adapters dir. Well, this is not strictly an adapter, but now we have adapters with one jquery file and vendor dir with one promise.js file thinking (Btw. we cannot rename adapters to not break backward compatibility).

If you are just worried about the additional folder, don't be - there is no filesystem limit we can exceed with that :smile: To be more serious - you already wrote that it's not adapter and in the future, we may have additional vendor libs with the same problem. It will be easier to keep 2 separate namespaces for that than always close eye for "not strictly adapter" argument. Although it's not something I want to fight for, if you still would like to move all vendor libs to `adapter` folder, it's fine for me. But the name may be misleading for the users.

>When it comes to unit test, there are two test promise initialization and test missing promise polyfill. Do you think it would make sense to add e.g. one more test to assert how CKEDITOR.tools.promise behaves? It will be testing native code / polyfill so I'm not sure. OTOH, if this test will start failing it means something changed in the implementation (not sure if it's probable).

We are not testing the native Array object nor other types. I would make no exception here and keep it simple.

>OTOH, if this test will start failing it means something changed in the implementation (not sure if it's probable).

The Internet would fall with such breaking native API change :smile: I think we will be fine.

>Is the content of this folder also minified during build?

It is.

engineering-this:
Something just caught my eye, so I'm posting it here, so we don't forget it :)

https://github.com/ckeditor/ckeditor-dev/blob/08ea0051cedd75c0bbb55d88d47fe9e05cddd1f5/gruntfile.js#L22

When this PR is merged line above can be removed.

f1ames:
One more thing, one unit test seems to be failing on IE8, can you check it @jacekbogdanski?

Screenshot 2019-05-20 at 16 28 36

tests/_benderjs/ckeditor/static/tools.js
jacekbogdanski: I was rather talking about `try/catch` statement. `Promise.prototype.catch` is mostly syntactic sugar over this statement with `reject` resolver support. It mostly exists in our test code base.

> OTOH, I would add some information to CKEDITOR.tools.promise docs, that if targeting browsers < IE9, [ 'catch' ] instead of .catch should be used. WDYT?

Yep, it's actually not obvious and may save someone googling time.

jacekbogdanski:
That's weird, I have no issue with this test on IE8 for both dev and built versions. @engineering-this could you check if the test fails for you also? Please, make sure to clean cache.

f1ames:
> That's weird, I have no issue with this test on IE8 for both dev and built versions.

I have rechecked and now it's green. I suspect it was caused by the low connection quality when I was running tests on the virtual machine.

f1ames:

PR #3005, opened 2019-03-29
Added support for dragging block widgets into tables

### LINKED ISSUE #1648: Impossible to drag and drop widget into empty table cell

Bug

Nothing found

1. Open editor that has widget and table plugins enabled.
e.g. http://tests.ckeditor.dev:1030/tests/plugins/widget/manual/block
2. Create empty table.
3. Try to drag widget into any cell.

### Expected result

Widget is dragged into table cell.

### Actual result

Widget is not dragged into cell.

### Additional info.

It is possible to drag widget into call that contains any block element <p>, <div>, etc.

## Other details

* Browser: any

engineering-this:

New feature

### This PR contains

- [X] Unit tests
- [X] Manual tests

## What changes did you make?

To allow fully dragging widgets into tables plenty of things had to be solved.

Widget can be dropped before and after block elements, this had to be changed, so it can be dropped into tables as well.

`lineutils` plugin exposes few classes which has plenty of methods for finding elements, calculating position of places where line can be placed, storing them, sorting them, creating line and placing it.

Unfortunately it was designed to work only only with one dimension `y` axis. As table is two dimensional grid, cells in one row were calculated to have same distance from mouse position, so it was possible to drag only into first cell. See https://github.com/ckeditor/ckeditor-dev/issues/1648#issuecomment-477148369 .

Fixing that within `lineutils` was possible in one of two ways:
- Breaking changes: basically rewrite half of plugin.
- Overload methods resulting in overly complicated code.

I decided to not do any of above. So I added new private function that calculates actual (2d) distance of relations (places where line can be placed) and sorts them based on that distance. This function is now used instead of `CKEDITOR.plugins.lineutils.Locator.sort`.

That fixed case of dropping into empty table cell, but in case of cell with text widget could be dropped only before text, not after. Additionally line would appear in the middle of cell - over the text.

My first idea to workaround this was to create range containing the text, and use `range.getClientRects()`, however all logic to place lines was based on actual elements, just passing rect wasn't enough. I've dropped that idea and decided to place collapsed temporary paragraphs instead. They doesn't take any place so no flickering occurs. Paragraphs are removed on `mouseup`, so they exist only during widget dnd and shouldn't leak.

The last issue was that line didn't appear in cells with inline elements. To not add much extra logic I've allowed creation of fake paragraphs next to inline elements too.

Closes #1648

### DIFF
--- a/plugins/widget/plugin.js
+++ b/plugins/widget/plugin.js
@@ -3506,30 +3506,31 @@
 			cell.append( fakeParagraph, before );
 		}
 
-		/**
-		 * Sorts relations by actual distance.
-		 *
-		 * @param {Object} relations {@link CKEDITOR.plugins.lineutils.finder#relations}
-		 * @param {Object} locations {@link CKEDITOR.plugins.lineutils.locator}
-		 * @param {Object} coordinates
-		 * @param {Number} coordinates.x horizontal coordinate
-		 * @param {Number} coordinates.y vertical coordinate
-		 * @returns [Array] Sorted array representation of relations {@link CKEDITOR.plugins.lineutils.finder#relations}
-		 */
+		// Sorts the relations by the actual distance.
+		//
+		// @param {Object} relations {@link CKEDITOR.plugins.lineutils.finder#relations}
+		// @param {Object} locations {@link CKEDITOR.plugins.lineutils.locator}
+		// @param {Object} coordinates
+		// @param {Number} coordinates.x horizontal coordinate
+		// @param {Number} coordinates.y vertical coordinate
+		// @returns [Array] Sorted array representation of relations {@link CKEDITOR.plugins.lineutils.finder#relations}
 		function sortByDistance( relations, locations, coordinates ) {
 			var x = coordinates.x,
 				y = coordinates.y,
 				sorted = [];
 
 			for ( var id in locations ) {
+				// The calculations are split for horizontal and vertical value, then the Pythagorean Theorem is used to calculate the actual distance.
 				var relation = relations[ id ],
 					rect = relation.elementRect,
 					type = relation.type,
 					verticalDistance, horizontalDistance, isStored, distanceBefore, distanceAfter;
 
 				if ( relation.type === CKEDITOR.LINEUTILS_INSIDE ) {
+					// Calculate the distance for the elements top and the bottom, and store the lower value.
 					horizontalDistance = getDistance( y, rect.top, rect.bottom );
 				} else {
+					// Calculate the distance for the elements before and after positions. Store the lower value, and it's type.
 					rect = relation.element.getParent().getClientRect();
 
 					distanceBefore = getDistance( y, locations[ id ][ 1 ] );
@@ -3550,12 +3551,13 @@
 					}
 				}
 
+				// Calculate the distance for the elements left and the right, and store the lowest.
 				verticalDistance = getDistance( x, rect.left, rect.right );
 
 				relation = {
 					uid: id,
 					type: type,
-					dist: Math.sqrt( Math.pow( verticalDistance, 2 ) + Math.pow( horizontalDistance, 2 ) )
+					dist: Math.sqrt( Math.pow( verticalDistance, 2 ) + Math.pow( horizontalDistance, 2 ) ) // Pythagorean Theorem
 				};
 
 				isStored = !!CKEDITOR.tools.array.find( sorted, function( item, index ) {
--- a/plugins/widget/plugin.js
+++ b/plugins/widget/plugin.js
@@ -2627,7 +2627,7 @@
 
 							// Allow line only on one side of fake paragraph.
 							if ( el.hasClass( 'cke_fake-paragraph' ) ) {
-								return el.getAttribute( 'data-cke-fake-paragraph' );
+								return el.data( 'cke-fake-paragraph' );
 							}
 
 							return isCell ? CKEDITOR.LINEUTILS_INSIDE : CKEDITOR.LINEUTILS_BEFORE | CKEDITOR.LINEUTILS_AFTER;
--- a/plugins/widget/plugin.js
+++ b/plugins/widget/plugin.js
@@ -2580,10 +2580,15 @@
 					lookups: {
 						// Element is block but not list item and not in nested editable.
 						'default': function( el ) {
+							var ret;
+							if ( el.getName() in { td: 1, th: 1 } ) {
+								ret = CKEDITOR.LINEUTILS_INSIDE;
+							}
+
 							if ( el.is( CKEDITOR.dtd.$listItem ) )
 								return;
 
-							if ( !el.is( CKEDITOR.dtd.$block ) )
+							if ( !ret && !el.is( CKEDITOR.dtd.$block ) )
 								return;
 
 							// Allow drop line inside, but never before or after nested editable (https://dev.ckeditor.com/ticket/12006).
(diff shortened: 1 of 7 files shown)


engineering-this:
>Maybe it would make sense to keep some kind of bookmarks for table cells by default? Of course, they would need to be downcasted properly (without bookmarks). In that case table logic would be extracted into table plugin. WDYT?

I'm afraid this will overcomplicate current solution. I can imagine multiple possible edge cases and potential conflicts, eg. dragging inline widgets, magicline, also these bookmarks should be prevented from selection and modification, from being replaced by paste, and more. Some tests might be affected, also bookmarks shouldn't be created if there is no `widget` plugin loaded. Also it's possible to not use drag handler at all and just cut paste, in this case we would be still keeping these bookmarks in memory even tho they aren't needed for this particular user.

If we stay with current solution: usually drag handler isn't clicked many times in short time and if someone decides to break editor by doing so then probably throttle would be best option.

jacekbogdanski:
> If we stay with current solution: usually drag handler isn't clicked many times in short time and if someone decides to break editor by doing so then probably throttle would be best option.

It takes 200ms to execute 'cell paragraphing` on chrome with 2000x8 table on the pretty solid machine. 
On Edge/IE11 it takes forever to drag and drop (like forever x2 because it's already slow).

engineering-this:
>It looks like on IE browsers there is additional drop line which causes dropping widget outside of the table:

The reason for this is that fake sel container was used as table sibling to create line. Fix for that is adding `sib.isVisible()` to conditional:

https://github.com/ckeditor/ckeditor-dev/blob/t/1648/plugins/lineutils/plugin.js#L487

engineering-this:
To deal with significant performance drop in big tables I've added new variable `limit`, for now it is set to 100, but we can adjust it. If count of all cells in editor is bigger than limit fake paragraphs won't be added all at once, but instead they will be added only for hovered cell.

plugins/widget/plugin.js
```diff
@@ -3399,19 +3418,34 @@
...
+			var cells = editable.find( 'td' ).toArray().concat( editable.find( 'th' ).toArray() );
+			addFakeParagraphs( cells );
+		}
+
 		// Harvest all possible relations and display some closest.
 		var relations = finder.greedySearch(),
-
 			buffer = CKEDITOR.tools.eventsBuffer( 50, function() {
+				if ( cellCount > limit && target.getName() in { td:1, th:1 } ) {
+					storeFakeParagraph( addFakeParagraph( target ) );
```
jacekbogdanski: This two functions + two execution has only one purpose, faking drop spot. Maybe you could merge it into one function with additional helpers if needed?

plugins/widget/plugin.js
```diff
@@ -3426,12 +3460,134 @@
...
+		 * Sorts relations by actual distance.
+		 *
+		 * @param {Object} relations {@link CKEDITOR.plugins.lineutils.finder#relations}
+		 * @param {Object} locations {@link CKEDITOR.plugins.lineutils.locator}
+		 * @param {Object} coordinates
+		 * @param {Number} coordinates.x horizontal coordinate
+		 * @param {Number} coordinates.y vertical coordinate
+		 * @returns [Array] Sorted array representation of relations {@link CKEDITOR.plugins.lineutils.finder#relations}
+		 */
+		function sortByDistance( relations, locations, coordinates ) {
```
jacekbogdanski: This function is a bit complicated (or I'm just scared of math). Maybe you could refactor it a bit or add some comments?

engineering-this: I'll add some comments.

engineering-this:
>Is there any reason why you would like to keep old behavior with 100 cells limit? It looks like it only adds complexity to the code.

Line appears at closest relation to mouse position. So it can appear when cell is not hovered, but mouse is near it. That's why I kept the original logic for smaller tables. However after comparing two versions it looks like there is very little difference, so we can simplify code a lot by removing original logic, and pool paragraphs.

plugins/widget/plugin.js
```diff
@@ -3426,12 +3453,134 @@
...
+		}
+
+		// Sorts the relations by the actual distance.
+		//
+		// @param {Object} relations {@link CKEDITOR.plugins.lineutils.finder#relations}
+		// @param {Object} locations {@link CKEDITOR.plugins.lineutils.locator}
+		// @param {Object} coordinates
+		// @param {Number} coordinates.x horizontal coordinate
+		// @param {Number} coordinates.y vertical coordinate
+		// @returns [Array] Sorted array representation of relations {@link CKEDITOR.plugins.lineutils.finder#relations}
```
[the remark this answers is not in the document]
engineering-this: Comment blocks are used exclusively for API-docs, no reason generate any docs for private functions.

tests/plugins/widget/dragintotable.js
```diff
@@ -0,0 +1,158 @@
...
+					editable.fire( 'mousemove', {
+						$: {
+							clientX: coordinates.x,
+							clientY: coordinates.y
+						},
+						getTarget: function() {
+							return element;
+						}
+					} );
+				}, null, null, 9999 );
```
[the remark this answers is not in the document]
engineering-this: What's wrong about it? If we use real life priority, then possible future listener might be executed after test and possible regression wouldn't be detected.

engineering-this:
>Tests are failing for IE randomly when running from the dashboard:
>It also occurred for FF once so I believe there is something wrong with it.

I can't reproduce it on IE11 neither on Firefox. Is it possible that you accidently clicked, thus changed focus? This would break TC.

engineering-this:
Refactored a lot, I hope readability is much better now.

plugins/widget/plugin.js
```diff
@@ -3392,6 +3409,8 @@
 		widget.dragHandlerContainer = container;
 	}
 
+	var fakeParagraphs = [];
```
jacekbogdanski: Maybe renaming it into `dropLineRelation` or something like that would make more sense? And also refactoring it into an object:

```javascript
var dropLinePosition = {
    before: { position: CKEDITOR.LINEUTILS_BEFORE, element: beforeElement }
    after: { position: CKEDITOR.LINEUTILS_AFTER, element, afterElement }
}
```
you could use it as:

```javascript
moveDropLine( cell, dropLinePosition );
```
Also, I wouldn't clean up this object. You can reuse paragraphs as long as you want without the need to destroy/rebuilt them. Although it may require some testing if they can be easily garbage collected when detached from the DOM. However, it may be micro-optimization not worth the try.

engineering-this: >Although it may require some testing if they can be easily garbage collected when detached from the DOM.

They won't be garbage collected when they are referenced in a variable which is accessible forever. We can't change it without heavy refactoring to whole plugin.

engineering-this: `dropLinePosition`, `moveDropLine ` - I thought about better naming that `fakeParagraph` and didn't come up with anything. However your proposal doesn't sound right to me. I wouldn't rename `placeFakeParagraph` into `moveDropLine`.  This function doesn't really do anything to drop line, it appends temporary elements that are to be used for finding possible drop line positions, but still they might not be used for placing drop line when they are not the closests elements to mouse position.

engineering-this:
I think that moving new sorting logic into `lineutils` plugin will require significant work with a few benefits.

```
locator.sort( 100, ... ) // default
locator.sort( { x: 100, y: 100 }, ... ) // extended
```

[lineutils.locator](https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_plugins_lineutils_locator.html) class is designed to calculate, store and sort locations based on Y coordinates. To make your example work each method should be reworked to work with 2-d coordinates. So first locator should be created with:

```
locator = locate( relations, 2d = true );
```

IMO we can skip this part of recatoring. We might go back to it when there is demand for such feature.

PR #3059, opened 2019-04-16
Fix for: Image in table cell has an empty URL field when edited from context menu opened by right-click

### LINKED ISSUE #2235: Image in table cell has an empty URL field when edited from context menu opened by right-click
## Type of report

Bug

1. Create a new table in ckeditor.
2. Put a single image in one of the cells.
3. Right click the image to open context menu.
4. Choose "Image Properties" from context menu to open the "Image Properties" dialog.

### Expected result

The fields of the "Image Properties" dialog should be populated with the corresponding properties of the image that was right-clicked.

### Actual result

The fields of the "Image Properties" dialog remain empty.

## Other details
Browsers: Current versions of Chrome/Chromium and Firefox under Linux and Windows 7/10.
CKEditor versions tested: 4.10.0 and 4.9.2
The issue only occurs if "Image Properties" was called via context menu and the image is the sole content of a table cell. Double clicking the image or selecting it and calling the dialog via tool bar operates as expected.
A possible cause of this bug maybe located in the selection range handling of the function isTableSelection (in core/selection.js)

Dumluregn:

Bug

### This PR contains

As this was a specific issue and required opening context menu by right-click (keyboard shortcuts were working fine before fix) there is just a short 15 seconds manual test included.

- [x] Unit tests
- [x] Manual tests

## What changes did you make?

I made `Table Selection` treat images from `Image` plugin the same way it operates on widgets, etc from `Enhanced Image` plugin (i.e. an early return in function checking if the table is selected).

Closes #2235.

### DIFF
--- a/core/selection.js
+++ b/core/selection.js
@@ -27,6 +27,13 @@
 		if ( isWidget( ranges[ 0 ].getEnclosedNode() ) ) {
 			return false;
 		}
+		// Also it's not table selection when image is selected. (#2235).
+		if ( ranges[ 0 ].getEnclosedNode() && ranges[ 0 ].getEnclosedNode().type == CKEDITOR.NODE_ELEMENT ) {
+			if ( ranges[ 0 ].getEnclosedNode().is( 'img' ) ) {
+				return false;
+			}
+		}
+
 		var node,
 			i;
 
--- a/core/selection.js
+++ b/core/selection.js
@@ -28,10 +28,10 @@
 			return false;
 		}
 		// Also it's not table selection when image is selected (#2235).
-		if ( ranges[ 0 ].getEnclosedNode() && ranges[ 0 ].getEnclosedNode().type == CKEDITOR.NODE_ELEMENT ) {
-			if ( ranges[ 0 ].getEnclosedNode().is( 'img' ) ) {
-				return false;
-			}
+		if ( ranges[ 0 ].getEnclosedNode() &&
+			ranges[ 0 ].getEnclosedNode().type == CKEDITOR.NODE_ELEMENT &&
+			ranges[ 0 ].getEnclosedNode().is( 'img' ) ) {
+			return false;
 		}
 
 		var node,
--- a/core/selection.js
+++ b/core/selection.js
@@ -27,7 +27,7 @@
 		if ( isWidget( ranges[ 0 ].getEnclosedNode() ) ) {
 			return false;
 		}
-		// Also it's not table selection when image is selected. (#2235).
+		// Also it's not table selection when image is selected (#2235).
 		if ( ranges[ 0 ].getEnclosedNode() && ranges[ 0 ].getEnclosedNode().type == CKEDITOR.NODE_ELEMENT ) {
 			if ( ranges[ 0 ].getEnclosedNode().is( 'img' ) ) {
 				return false;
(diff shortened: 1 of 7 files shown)


f1ames:

tests/plugins/tableselection/manual/integrations/image/nestedimage.html
```diff
@@ -0,0 +1,22 @@
+<div id="editor">
+	<table border="1" cellpadding="1" cellspacing="1" style="width:500px">
```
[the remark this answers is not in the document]
f1ames: And `undo` plugin so undo integration can be tested too?

core/selection.js
```diff
@@ -23,10 +23,7 @@
 		if ( ranges.length === 0 ) {
 			return false;
 		}
-		// It's not table selection when selected node is a widget (#1027).
-		if ( isWidget( ranges[ 0 ].getEnclosedNode() ) ) {
-			return false;
-		}
+
```
f1ames: This change is not related to this fix. We try to avoid changing things unrelated to the current fix in the same PR. I know it doesn't align with the proposed solution and it may seem tempting to move it together with proposed fix, however:

* This code is related to different fix which is similar but required a different approach (especially that it is related to widgets), so the assumption that things will work the same as for this fix may be incorrect.
* The solution was reviewed, accepted and tested already (some time ago) so changing it will require repeating the whole procedure from the previous issue perspective.

tests/plugins/tableselection/manual/integrations/image/nestedimage.md
```diff
@@ -0,0 +1,27 @@
...
+
+**Expected result:**
+
+URL field should be filled and image should be visible in the preview.
+
+**Unexpected result:**
+
+URL field is empty or image isn't visible in the preview.
+
+3. Select first row of the outer table and delete it.
```
f1ames: This steps are hardly visible in the scenario. Maybe you could just create two separate scenarios here, e.g:

```md
## Context menu for image in table

### Scenario 1

1. ...

Expected...
Unexpected...

### Scenario 2

1. ...

Expected...
Unexpected...

```

Dumluregn:
About unit test - it was impossible to exactly mimic the scenario (using context menu or dialog didn't work), so instead test is checking if after selecting image the cell becomes fake-selected or not. That was the reason of incorrect behaviour before, so as long as the test passes this patch should work fine.

f1ames:

Dumluregn:
@f1ames - thanks for bringing my attention to IE11+ and Edge. The issue was caused because those browsers has been marking selection as fake even when selection didn't exceed the original cell. I added an additional selection reset and now everything should work fine.

f1ames:
@Dumluregn please take a look at CI, there is quite significant number of unit tests failing for editor build version (both Chrome and Firefox).

On the dev version only few tests are run with `tableselection` plugin (ones where it is explicitly added). On the build one, where `full-all` build is used, all default plugins (which means `tableselection` too) are present in most of the tests. And it seems that after recent changes, `tableselection` doesn't play well with the rest of the editor 😅

Dumluregn:
Yeah, it seems I limited fake selection a little bit too much.

f1ames:
Rebase on the latest `major`.

f1ames:
Extracted two issues #3166 and #3167.

PR #3075, opened 2019-04-25
Add unit type selection for table cell height

### LINKED ISSUE #2084: Cell properties dialog has locked height unit type
## Type of report

New Feature

1. Open http://cdn.ckeditor.com/4.9.2/full-all/samples/index.html
2. Clear editor contents and create a new table.
3. Open context menu and choose cell properties.

### Expected result

Cell height can be changed from pixels to percentage unit.

### Actual result

Cell height unit is text instead of a select.

## Other details

* CKEditor version: 4.9.2
* Installed CKEditor plugins: tabletools
screen shot 2018-06-13 at 16 58 05

### LINKED ISSUE #3098: Unit pickers in cell properties have different width
## Type of report

Bug

1. Go to https://ckeditor.com/docs/ckeditor4/latest/examples/table.html
2. Right-click on any cell in table and choose `Cell -> Cell properties`

### Expected result

`Width` and `Height` boxes should be the same width

### Actual result
Boxes has different width, which leads to different unit pickers width (to be introduced in #3075 - https://github.com/ckeditor/ckeditor-dev/pull/3075#pullrequestreview-236110294):

Screenshot 2019-05-13 at 11 43 44

Screenshot 2019-05-13 at 11 43 54

## Other details

Both elements have exactly the same styling. It may seem that the problem is caused by longer label for `Height` than `Width`, but changing it doesn't resolve the issue (image from #3075 mentioned above):
Screenshot 2019-05-13 at 10 44 02
 

* Browser: every
* OS: macOS
* CKEditor version: 4.11.4, at least as old as 4.7.0
* Installed CKEditor plugins: `tabletools`

### LINKED ISSUE #3164: Accept floating point values for table cell height and width
## Type of report

Feature request

For now floating point values are parsed to integer. We could change it just by replacing `parseInt()` with `parseFloat()` in a few lines of code. Then users will be able to use popular height/width values like `33.3%`.

Dumluregn:

New feature

### This PR contains

- [x] Unit tests
- [x] Manual tests

## What changes did you make?

I copied the solution used for cell width.

Closes #2084.
Closes #3098.
Closes #3164.

### DIFF
--- a/plugins/tabletools/dialogs/tableCell.js
+++ b/plugins/tabletools/dialogs/tableCell.js
@@ -436,14 +436,14 @@ CKEDITOR.dialog.add( 'cellProperties', function( editor ) {
 				},
 
 				setup: setupCells( function( element ) {
-					var attr = parseInt( element.getAttribute( fieldName ), 10 ),
-						style = parseInt( element.getStyle( fieldName ), 10 );
+					var attr = parseFloat( element.getAttribute( fieldName ), 10 ),
+						style = parseFloat( element.getStyle( fieldName ), 10 );
 
 					return !isNaN( style ) ? style :
 						!isNaN( attr ) ? attr : '';
 				} ),
 				commit: function( element ) {
-					var value = parseInt( this.getValue(), 10 ),
+					var value = parseFloat( this.getValue(), 10 ),
 
 						// There might be no unit type, i.e. when multiple cells are
 						// selected but some of them have size expressed in pixels and some
--- a/plugins/tabletools/dialogs/tableCell.js
+++ b/plugins/tabletools/dialogs/tableCell.js
@@ -468,7 +468,7 @@ CKEDITOR.dialog.add( 'cellProperties', function( editor ) {
 				type: 'select',
 				id: fieldName + 'Type',
 				label: editor.lang.table[ fieldName + 'Unit' ],
-				labelStyle: 'visibility:hidden',
+				labelStyle: 'visibility:hidden;display:block;width:0;',
 				'default': 'px',
 				items: [
 					// 'widthPx' and 'widthPc' are also used for height to avoid additional translations.
--- a/plugins/tabletools/dialogs/tableCell.js
+++ b/plugins/tabletools/dialogs/tableCell.js
@@ -440,11 +440,13 @@ CKEDITOR.dialog.add( 'cellProperties', function( editor ) {
 
 					if ( !isNaN( style ) ) {
 						return style;
-					} else if ( !isNaN( attr ) ) {
-						return attr;
-					} else {
-						return;
 					}
+					if ( !isNaN( attr ) ) {
+						return attr;
+					}
+
+					return;
+
 				} ),
 				commit: function( element ) {
 					var value = parseFloat( this.getValue(), 10 ),
--- a/plugins/tabletools/dialogs/tableCell.js
+++ b/plugins/tabletools/dialogs/tableCell.js
@@ -11,396 +11,282 @@ CKEDITOR.dialog.add( 'cellProperties', function( editor ) {
 		widthPattern = /^(\d+(?:\.\d+)?)(px|%)$/,
 		rtl = editor.lang.dir == 'rtl',
 		colorDialog = editor.plugins.colordialog,
-		items = [ {
-			requiredContent: 'td{width}',
-			type: 'hbox',
-			widths: [ '70%', '30%' ],
-			children: [ {
-				type: 'text',
-				id: 'width',
-				width: '100px',
-				label: langCommon.width,
-				validate: validate.number( langCell.invalidWidth ),
-
-				// Extra labelling of width unit type.
-				onLoad: function() {
-					var widthType = this.getDialog().getContentElement( 'info', 'widthType' ),
-						labelElement = widthType.getElement(),
-						inputElement = this.getInputElement(),
-						ariaLabelledByAttr = inputElement.getAttribute( 'aria-labelledby' );
-
-					inputElement.setAttribute( 'aria-labelledby', [ ariaLabelledByAttr, labelElement.$.id ].join( ' ' ) );
-				},
-
+		items = [
+			getCellSizeFieldDefinition( 'width' ),
+			getCellSizeFieldDefinition( 'height' ),
+			createSpacer( [ 'td{width}', 'td{height}' ] ),
+			{
+				type: 'select',
+				id: 'wordWrap',
+				requiredContent: 'td{white-space}',
+				label: langCell.wordWrap,
+				'default': 'yes',
+				items: [
+					[ langCell.yes, 'yes' ],
+					[ langCell.no, 'no' ]
+				],
 				setup: setupCells( function( element ) {
-					var widthAttr = parseInt( element.getAttribute( 'width' ), 10 ),
-						widthStyle = parseInt( element.getStyle( 'width' ), 10 );
+					var wordWrapAttr = element.getAttribute( 'noWrap' ),
+						wordWrapStyle = element.getStyle( 'white-space' );
 
-					return !isNaN( widthStyle ) ? widthStyle :
-						!isNaN( widthAttr ) ? widthAttr : '';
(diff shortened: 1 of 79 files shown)


tests/plugins/tabletools/cellproperties.js
```diff
@@ -1,5 +1,5 @@
 /* bender-tags: editor */
-/* bender-ckeditor-plugins: entities,dialog,tabletools,toolbar */
+/* bender-ckeditor-plugins: entities,dialog,tabletools,toolbar,tableselection */
```
engineering-this: You are only using existing testing function, any reason to add new plugin?

tests/plugins/tabletools/cellproperties.js
```diff
@@ -139,6 +140,48 @@
 			} );
 		},
 
+		// (#2084)
+		'test load and update field values (#11)': function() {
+			this.doTest( 'table-12', function( dialog ) {
```
engineering-this: Maybe `doTest` function could be reworked to return a function, so we could have few lines less.

It would result in the following:
```
'test load and update field values (#11)': doTest( 'table-12, function( dialog ) {
```

Dumluregn:
Problem with pickers width reported - #3098

plugins/tabletools/dialogs/tableCell.js
```diff
@@ -76,9 +76,8 @@ CKEDITOR.dialog.add( 'cellProperties', function( editor ) {
 			children: [ {
 				type: 'text',
 				id: 'height',
-				label: langCommon.height,
 				width: '100px',
-				'default': '',
+				label: langCommon.height,
```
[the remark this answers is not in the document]
Dumluregn: Properties order was changed to be the same as in `Width` item for readability. However with refactoring you proposed there will be only one function left so it won't matter anyway.

tests/plugins/tabletools/cellproperties.js
Dumluregn: There was - tests were failing in built full-all version and the reason was `tableselection` plugin. Therefore I added it to tests to be sure if it's working correctly. I fixed the issue in the next commit, so right now plugin is not necessary, it can be removed.

plugins/tabletools/dialogs/tableCell.js
```diff
@@ -518,6 +412,76 @@ CKEDITOR.dialog.add( 'cellProperties', function( editor ) {
...
+					inputElement.setAttribute( 'aria-labelledby', [ ariaLabelledByAttr, labelElement.$.id ].join( ' ' ) );
+				},
+
+				setup: setupCells( function( element ) {
+					var attr = parseFloat( element.getAttribute( fieldName ), 10 ),
+						style = parseFloat( element.getStyle( fieldName ), 10 );
+
+					if ( !isNaN( style ) ) {
+						return style;
+					} else if ( !isNaN( attr ) ) {
```
engineering-this: When there is a return statement inside `if`, `else` keyword is redundant.

```suggestion
					}
					if ( !isNaN( attr ) ) {
```

plugins/tabletools/dialogs/tableCell.js
```diff
@@ -518,6 +412,76 @@ CKEDITOR.dialog.add( 'cellProperties', function( editor ) {
...
+						return attr;
+					} else {
```
engineering-this: `return;` can be moved out of `else`. This convention is called [guard clause](https://elliotekj.com/2016/12/02/guard-clauses-in-javascript/). It has nice readability, and it is used when possible in our codebase.

tests/plugins/tabletools/manual/cellheight.md
```diff
@@ -0,0 +1,34 @@
+@bender-tags: 4.12.0, 2084, 3098
+@bender-ui: collapsed
+@bender-ckeditor-plugins: wysiwygarea, toolbar, table, tabletools, undo
+
+1. Open the cell properties dialog.
+
+	#### Expected:
+
+	Cell height unit is changeable (pixels or percents).
```
engineering-this: I suggest using lists when there are multiple expected.
```suggestion
	- Cell height unit is changeable (pixels or percents).
```

Also
```suggestion
	- Cell height unit is changeable (pixels or percentages).
```
I'm giving two suggestions, as I'm not fully sure about the second one. I think that `percent` means a symbol `%`, while `percentages` is an actual value `33%`.

plugins/tabletools/dialogs/tableCell.js
```diff
@@ -103,22 +97,36 @@ CKEDITOR.dialog.add( 'cellProperties', function( editor ) {
 					return !isNaN( heightStyle ) ? heightStyle :
 						!isNaN( heightAttr ) ? heightAttr : '';
 				} ),
+
 				commit: function( element ) {
-					var value = parseInt( this.getValue(), 10 );
+					var value = parseInt( this.getValue(), 10 ),
```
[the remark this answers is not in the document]
Comandeer: It can be a potentially breaking change, as it changes the behaviour of dialog. At least it should be noted as separate feature request.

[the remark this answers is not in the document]
engineering-this: Floats are accepted by editor anyway, now they can also be set from dialog. I don't think this change could do any harm. But I agree on including this change as separate entry in changelog under features.

plugins/tabletools/dialogs/tableCell.js
```diff
@@ -518,6 +412,78 @@ CKEDITOR.dialog.add( 'cellProperties', function( editor ) {
...
+					}
+
+					element.removeAttribute( fieldName );
+				},
+				'default': ''
+			}, {
+				type: 'select',
+				id: fieldName + 'Type',
+				label: editor.lang.table[ fieldName + 'Unit' ],
+				labelStyle: 'visibility:hidden;display:block;width:0;',
```
Comandeer: There is `.cke_screen_reader_only` class inside `copyformatting/copyformatting.css`, it can be reused here. But it's probably a separate task.

plugins/tabletools/dialogs/tableCell.js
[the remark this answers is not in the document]
Dumluregn: Reported as #3164.

plugins/tabletools/dialogs/tableCell.js
Dumluregn: Reported in #3165.

Dumluregn:
Wild scrollbar tamed, it will no longer cause us trouble.I also fixed the changelog.

dev/langtool/meta/ckeditor.core/meta.txt
```diff
@@ -1,5 +1,5 @@
 # Copyright (c) 2003-2019, CKSource - Frederico Knabben. All rights reserved.
-# For licensing, see LICENSE.md or https://ckeditor.com/legal/ckeditor-oss-license
+# For licensing, see LICENSE.md or http://ckeditor.com/license
```
f1ames: Looks like some issue with `langtool` and some outdated headers, cc @Comandeer

Comandeer: It seems to be an issue with the langtool itself, as I don't see such header anywhere in the code 🤔

plugins/tabletools/dialogs/tableCell.js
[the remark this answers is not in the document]
f1ames: I think considering it as breaking change would be kind of overuse. It changes the way how dialog works, but more like extending and not limiting the previous state so it should be considered a new feature (which you did so 👍).
OTOH, changing it the other way around (from `parseFloat` to `parseInt` could be considered as breaking change).
Anyway, good catch @Comandeer🔎

dev/langtool/meta/ckeditor.core/meta.txt
Dumluregn: Yeah, something is wrong. I checked what `langtool` did exactly and it turns out that besides messing up license header it also added trailing spaces after `=` sign and changed the order of entries from alphabetical to some random one (probably there are some kind of rules but I don't see them). So instead of using this tool I just added the necessary entry manually.

Issue #3094, opened 2019-05-10
Emoji Dropdown is enabled before Emoji List is loaded


engineering-this:
## Type of report

Bug

## Provide detailed reproduction steps (if any)

No standard steps on this one.

By design Emoji List is loaded asynchronously, after editor is ready. Without Emoji List loaded Emoji Autocomplete and Emoji Dropdown can't work.

### Expected result

Emoji Dropdown is disabled until the Emoji List is loaded.

### Actual result

Emoji Dropdown is enabled as soon as editor is `ready`.

## Other details

https://github.com/ckeditor/ckeditor-dev/blob/08a87a28d7e18b6dae0fe96a6927c330696efa8b/plugins/emoji/plugin.js#L571-L590

https://github.com/ckeditor/ckeditor-dev/blob/08a87a28d7e18b6dae0fe96a6927c330696efa8b/plugins/emoji/plugin.js#L594-L601

PR #3095, opened 2019-05-10
Fix for: Failing test `tests/plugins/emoji/dropdown`

### LINKED ISSUE #2831: Failing test `tests/plugins/emoji/dropdown`
## Type of report

Failing test

Failing test `tests/plugins/emoji/dropdown`.

The test fails on dev and built version from the dashboard.

image

engineering-this:

Fix failing tests

## What changes did you make?

Original issue reported only one case, however I could see three different reasons for tests to fail on Firefox.

### Emoji List is loaded asynchronously

Sometimes tests were run earlier resulting in type error. This is fixed by adding `waitForEmoji` which is delaying each test case until Emoji List is loaded. Related issue: #3094.

### After fix above some tests started to fail on Firefox, because of wrong assertion.
	- expected: `'<p>⭐</p>'`
	- actual: `'<p>⭐</p><p> </p>'`

	Fixed by adding `result.replace( regex )` to cut empty paragraph.

### Lastly the original issue.
Firefox seems to not respond very well to multiple synchronous changes of `scrollTop` of various elements, and that's what we do in our implementation of `scrollIntoView`. When this TC fails scroll position is incorrect. Fixed by calling native `scrollIntoView` for Firefox.

Additionally take a look at the code:
https://github.com/ckeditor/ckeditor-dev/blob/08ea0051cedd75c0bbb55d88d47fe9e05cddd1f5/tests/plugins/emoji/dropdown.js#L135-L137
	`scrollIntoView` is already causing a `scroll` event, so no need to manually trigger it. Although I didn't find it to interfere with tests, it's redundant so I'm removing it.

### Note
All of the issues are causing tests to fail randomly. After applying the fixes I did run tests multiple times and they seem to not fail, but Reviewer should check few times too.

Closes #2831

### DIFF
--- a/tests/plugins/emoji/dropdown.js
+++ b/tests/plugins/emoji/dropdown.js
@@ -32,169 +32,190 @@
 		},
 		'test emoji dropdown has proper components': function() {
 			var bot = this.editorBot;
-			bot.panel( 'EmojiPanel', function( panel ) {
-				try {
-					var doc = panel._.iframe.getFrameDocument();
-					assert.areNotSame( 0, doc.find( '.cke_emoji-navigation_item' ).count(), 'There is no navigation in Emoji Panel.' );
-					assert.areSame( 1, doc.find( 'input' ).count(), 'There is no search input in Emoji Panel.' );
-					assert.areSame( 1, doc.find( '.cke_emoji-outer_emoji_block' ).count(), 'There is no emoji block in Emoji Panel.' );
-					assert.areSame( 1, doc.find( '.cke_emoji-status_bar' ).count(), 'There is no emoji status bar in Emoji Panel.' );
 
-					assert.isTrue( doc.findOne( 'input' ).hasListeners( 'input' ), 'Searchbox doesn\'t have listener.' );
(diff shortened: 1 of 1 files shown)


f1ames:
Rebased and retargeted onto latest `major`.

engineering-this:
>It looks like the test is still failing 😞

That's the problem with unstable tests, you think you fixed it, because it doesn't fail after xx runs, but it will fail when Reviewer tries it 😞

It looks like both native `scrollIntoView` and our fallback doesn't scroll correctly on Firefox. Some kind of race condition. I've tried one more thing - wrapping just the `scrollIntoView` in `setTimeout`. It seems to fix the issue.

f1ames:

engineering-this:
Screenshot 2019-06-11 at 11 44 03

From dashboard it passes almost every time, but yeah it's still not stable 😡

It feels like, spending more time on trying to fix whatever is causing this test to fail is pointless, as it's caused by some browser quirk which is very hard to reproduce outside of this TC. IMO we should go with one of following:
- Ignore TC on Firefox
- Use mocked event to rework it

WDUT @f1ames ?

engineering-this:
Ignored test on Firefox, and reverted recent changes that were aimed for FF.

PR #3125, opened 2019-05-22
CKEDITOR.tools improvements

### LINKED ISSUE #3120: CKEDITOR.tools.createClass is unable to overwrite members
## Type of report

Bug

1. Open IE8 with CKEditor4 sample page e.g. https://ckeditor.com/latest/samples
2. Open console.
3. Execute script:
```javascript
var CreateClass = CKEDITOR.tools.createClass( {
	$: function() {
		this.message = 'Message from CreateClass!';
	},
	proto: {
			toString: function() {
				return this.message;
			},
		}
} );

function NativeClass () {
	this.message = 'Message from Native class!';
}

NativeClass.prototype.toString = function() {
	return this.message;
}

console.log( 'CreateClass', new CreateClass().toString() );
console.log( 'NativeClass', new NativeClass().toString() );
```

### Expected result

In both console logs, you can see overwritten `toString` methods.

### Actual result

Console log for the class created using `CKEDITOR.tools.createClass` prints `[object Object]` which means that the `toString` method has not been overwritten.

## Other details

It may be related only to the native members.

* CKEditor version: from implementation

### LINKED ISSUE #3123: CKEDITOR.tools.object helper methods improvements
## Type of report

Feature request

https://github.com/ckeditor/ckeditor-dev/issues/3120 issue is caused by the ECMA3 [`DontEnum` attribute](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Data_structures#Properties). The attribute is set on some object properties which cannot be enumerated, so `CKEDITOR.tools.extend` method is unable to correctly copy object prototype properties.

I would like to take advantage of the situation and improve our `CKEDITOR.tools.object` helper methods during bug fixing:

* `CKEDITOR.tools.objectKeys` - should include `DontEnum` properties to fix the issue for IE8
* `CKEDITOR.tools.object.forEach` - fixing the issue with `for in` loop for IE8
* `CKEDITOR.tools.object.reduce` - the same as above
* `CKEDITOR.tools.extend` - fixed by using `CKEDITOR.tools.object.forEach` function instead of `for in` loop

Both `forEach` and `reduce` function will have pre-implemented `DontEnum` bug fix (by using `CKEDITOR.tools.objectKeys` which will reduce the possibility of the issue in the future.

jacekbogdanski:

New feature/Bug fix

### This PR contains

- [x] Unit tests
- [ ] Manual tests

## What changes did you make?

Closes #3120 
Closes #3123


jacekbogdanski:
Updated according to discussion #3123

CHANGES.md
```diff
@@ -15,12 +15,16 @@ Fixed Issues:
 * [#1478](https://github.com/ckeditor/ckeditor-dev/issues/1478): Fixed: Custom colors added to [Color Button](https://ckeditor.com/cke4/addon/colorbutton) via [`config.colorButton_colors`](https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_config.html#cfg-colorButton_colors) in form label/code don't work correctly.
 * [#1469](https://github.com/ckeditor/ckeditor-dev/issues/1469): Fixed: Trying to get data from nested editable inside freshly pasted widget throws an error.
 * [#2923](https://github.com/ckeditor/ckeditor-dev/issues/2923): Fixed: CSS `windowtext` color is not correctly recognized by [`CKEDITOR.tools.style.parse`](https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_tools_style_parse.html) functions.
+* [#3120](https://github.com/ckeditor/ckeditor-dev/issues/3120): Fixed: [`CKEDITOR.tools.extend`](https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_tool.html#method-extend) function doesn't work with [`DontEnum`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Data_structures#Properties) object property attribute.
```
f1ames: Shouldn't it be marked as `IE8` only?

core/tools.js
```diff
@@ -1129,21 +1131,12 @@
...
-		 * Returns an array of passed object's keys.
-		 *
-		 *		console.log( CKEDITOR.tools.objectKeys( { foo: 1, bar: false } );
-		 *		// -> [ 'foo', 'bar' ]
-		 *
+		 * @inheritdoc CKEDITOR.tools.object#keys
 		 * @since 4.1
-		 * @param {Object} obj
-		 * @returns {Array} Object's keys.
+		 * @deprecated 4.12.0 Use {@link CKEDITOR.tools.object#keys} instead.
```
f1ames: Unfortunately, `@deprecated 4.12.0 Use {@link CKEDITOR.tools.object#keys} instead.` doesn't show the message in our docs, but I don't see any way to improve it:

image

And the issue was already reported in Umberto.

core/tools.js
```diff
@@ -2086,6 +2079,97 @@
...
+
+			/**
+			 * List of ECMA3 object properties with obsolete
+			 * [DontEnum](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Data_structures#Properties)
+			 * attribute.
+			 *
+			 * @private
+			 * @since 4.12.0
+			 */
+			DONT_ENUMS: [
```
f1ames: This should be also a `@member` of  `CKEDITOR.tools.object`, without it, it will be in the `tools` section in the docs

image

core/tools.js
jacekbogdanski: We could eventually add this information into method description itself, although it's smelling like duplication when Umberto will be fixed. Maybe let's just left as it is for now as we already have changelog entry for it?

f1ames: Let's leave it as is and wait for Umberto fix.

jacekbogdanski:
Seems like IE8 played with me a bit when testing this PR. Should be fine now.

PR #718, opened 2017-08-02
Do not load plugins after editor was destroyed

Comandeer:
## What is the purpose of this pull request?

Bug fix

## Does your PR contain necessary tests?

All patches which change the editor code must include tests. You can always read more
on [PR testing](http://docs.ckeditor.com/#!/guide/dev_contributing_code-section-tests),
[how to set the testing environment](http://docs.ckeditor.com/#!/guide/dev_tests) and
[how to create tests](http://docs.ckeditor.com/#!/guide/dev_tests-section-creating-your-own-test)
in the official CKEditor documentation.

### This PR contains

- [x] Unit tests
- [ ] Manual tests

## What changes did you make?

Closes https://dev.ckeditor.com/ticket/14613.

PR #3128, opened 2019-05-23
Fix for: Errors thrown when Editor is detached and destroyed during initialization

### LINKED ISSUE #2257: Editor throws an exception when destroyed shortly after it was created
## Type of report

Bug

There's a possible race condition if you delete the editor right after it was instantiated.

1. Open [codepen demo](https://codepen.io/mlewand/pen/wxejOg).
1. Open JavaScript console.
1. Click the "Load/Destory" button.

### Expected

Nothing happens.

## Actual

`Uncaught TypeError: Cannot read property 'allow' of undefined` exception is being thrown.

Note that it could sometimes require like 3 or 5 attempts to repro.

## Other details

Originaly reported in old tracker as [#14613](https://dev.ckeditor.com/ticket/14613).

* Browser: Any
* OS: Any

### LINKED ISSUE #3115: Editor should be removable from dom right after it is destroyed
## Type of report

Bug

## Other details

Because of how components in modern JS frameworks are working we should address some issues with editor.

1. Warning `[CKEDITOR] Error code: editor-incorrect-destroy.`
We should completely get rid of this warning. It is visible when part of destroy logic is executed after editor is removed from DOM. Cleaning logic should work regardless of iframe being in DOM or not.
So we could do following:
	```
	editor.destroy();
	editorWrapper.remove();
	```

1. Errors are thrown when editor is destroyed and removed from DOM before it is fully initialized. Frameworks routing works in a way that `component.destroy()` is fired either right after or right before elements are removed from DOM. We should somehow allow this.

Currently all our integrations are affected by this.
React: https://github.com/ckeditor/ckeditor4-react/issues/44
Vue: Requires user to add router guards.
Angular: `classic` editor is disabled, editor is rendered outside of a component and then appended to it - ugly hacks.

I have some ideas in mind I'll report here what works or doesn't.

### LINKED ISSUE #3124: `element.isDetached` check if element is detached from DOM
## Type of report

Feature request

This is needed for #3115

engineering-this:

A bit of fix a bit of feature

### This PR contains

- [x] Unit tests
- [x] Manual tests

## Proposed changelog entry

```
Fixed Issues:

* [#3115](https://github.com/ckeditor/ckeditor-dev/issues/3115): Fixed: Destroying editor during initialization throws an error.

API Changes:

* [#3124](https://github.com/ckeditor/ckeditor-dev/issues/3124): Added the [`CKEDITOR.dom.element#isDetached`](https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_dom_element.html#method-isDetached) method.
```

### Fix

Within various places in code I've added checks if editor is destroyed or detached then execution of following code is prevented. When possible `editor.container` is used to check, otherwise `editor.element`.

### Edge && IE

MS browsers sometimes throws `Permission Denied`, however browser won't break on such error, so it's nearly impossible to fix it. I've spend some time checking `timeouts` and `onload` listeners with no success.
Note: This PR shouldn't break anything for Edge or IE, because when editor is detached than it will throws errors regardless of having this fix or not.

### Other stuff

- https://github.com/ckeditor/ckeditor-dev/compare/t/3115?expand=1#diff-162a1b36d2efd62081522822a026e853R195
☝️ I think about removing above message, as now editor can be destroyed without any issues.

- this PR introduces new method `CKEDITOR.dom.element.isDetached` (#3124)

Closes #718.
Closes #2257.
Closes #3115.
Closes #3124.


alinaciuysal:
@jacekbogdanski @f1ames 
Do you have any updates on this?

jacekbogdanski:
@alinaciuysal the issue has been added to the 4.12.0 milestone (see #3115) so if everything goes smooth it will be available since the 4.12.0 release. Bear in mind that the time may be delayed if we will struggle with some unexpected problems, but hopefully we would be able to close it with 4.12.0.

Please, subscribe #3115 for updates.

plugins/wysiwygarea/plugin.js
```diff
@@ -515,15 +525,9 @@
 			detach: function() {
 				var editor = this.editor,
 					doc = editor.document,
-					iframe,
+					iframe = editor.container.findOne( 'iframe' ),
```
jacekbogdanski: Is there no edge case where the container may be already detached? I'm also not sure about your changes. I have a feeling like getting `editor.window.getFrame` is safer than searching container (eg in some weird case where editor's container have additional iframe element added by 3rd party plugins like context menu).

Can't we just check iframe by `iframe && iframe.isDetached`?

plugins/wysiwygarea/plugin.js
```diff
@@ -533,15 +537,17 @@
...
 
 					// IE BUG: When destroying editor DOM with the selection remains inside
 					// editing area would break IE7/8's selection system, we have to put the editing
 					// iframe offline first. (https://dev.ckeditor.com/ticket/3812 and https://dev.ckeditor.com/ticket/5441)
-					iframe.remove();
+					if ( iframe.getParent() ) {
+						iframe.remove();
+					}
 				} else {
 					CKEDITOR.warn( 'editor-destroy-iframe' );
```
jacekbogdanski: This warning probably doesn't seem sense anymore after changes.

plugins/wysiwygarea/plugin.js
jacekbogdanski: Also, please, cover these changes by unit tests.

tests/core/creators/inline-textarea.js
```diff
@@ -109,6 +109,27 @@
...
+
+		'test destroy when editor.container is absent': function() {
+			bender.editorBot.create( {
+				creator: 'inline',
+				name: 'editor4'
+			}, function( bot ) {
+				var editor = bot.editor,
+					container = editor.container;
+
+				delete editor.container;
```
jacekbogdanski: I'm missing verification when the container is alive but in detached mode.

engineering-this:
The general problem with unit test is that editor has to be detached at the right moment, so that given change in code is executed, if it's too early than it might be not reached. With that amount of async code it's quite tricky. I'll try to add some more cases, but I can't guarantee to cover them all.

plugins/wysiwygarea/plugin.js
engineering-this: `editor.window.getFrame();` returns `null` when iframe is detached, but this doesn't guarantee that custom data has been cleaned, so we have potential memory leaks.

I can improve selector.
```suggestion
					iframe = editor.container.findOne( 'iframe.cke_wysiwyg_frame' ),
```

core/creators/inline.js
```diff
@@ -93,11 +93,16 @@
 
 		// Handle editor destroying.
 		editor.on( 'destroy', function() {
+			var container = editor.container;
 			// Remove container from DOM if inline-textarea editor.
 			// Show <textarea> back again.
+			// Editor can be destroyed before container is created (#3115).
+			if ( textarea && container ) {
```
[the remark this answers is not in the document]
engineering-this: It can, but we are using `isDetached()` condition to prevent further initialization. Here we are destroying editor, I see no reason to alter this behaviour when editor is detached.

core/creators/themedui.js
```diff
@@ -239,6 +239,9 @@ CKEDITOR.replaceClass = 'ckeditor';
 
 			// Delay to avoid race conditions (setMode inside setMode).
 			setTimeout( function() {
+				if ( editor.status === 'destroyed' || editor.container.isDetached() ) {
```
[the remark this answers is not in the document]
engineering-this: Actually this case is supposed to be covered by
`test editor set mode when editor is detached'`
and
`test editor set mode when editor is destroyed'`

But these tests are false positive, because they use fictional mode :sadfrog:.

plugins/clipboard/plugin.js
```diff
@@ -1203,8 +1203,10 @@
 		}
 
 		function setToolbarStates() {
-			if ( editor.mode != 'wysiwyg' )
+			if ( editor.mode != 'wysiwyg' || editor.status === 'destroyed' || editor.container.isDetached() ) {
```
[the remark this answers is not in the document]
engineering-this: It produced an error for me, but I see after other changes it doesn't occur anymore, so probably now it can be reverted.

tests/core/creators/inline-textarea.js
```diff
@@ -109,6 +109,27 @@
 					assert.isNull( editor.container.getCustomData( 'x' ), 'Custom data purged' );
 				} );
 			} );
+		},
+
+		'test destroy when editor.container is absent': function() {
```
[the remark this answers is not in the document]
engineering-this: ☝️ I have same errors on `major` branch. If you check on [createConcurrentEditorTest](https://github.com/ckeditor/ckeditor-dev/blob/major/tests/core/creators/inline-textarea.js#L115-L127) helper you will see that it is intentional to test this warning.

CHANGES.md
```diff
@@ -12,6 +12,7 @@ Fixed Issues:
 
 * [#2672](https://github.com/ckeditor/ckeditor-dev/issues/2672): Fixed: When resizing [Enhanced Image](https://ckeditor.com/cke4/addon/image2) to minimum size with a resizer the image dialog doesn't show actual values.
 * [#1478](https://github.com/ckeditor/ckeditor-dev/issues/1478): Fixed: Custom colors added to [Color Button](https://ckeditor.com/cke4/addon/colorbutton) via [`config.colorButton_colors`](https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_config.html#cfg-colorButton_colors) in form label/code don't work correctly.
+* [#3115](https://github.com/ckeditor/ckeditor-dev/issues/3115): Fixed: Errors thrown when Editor is detached and destroyed during initialization.
```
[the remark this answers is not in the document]
engineering-this:
```suggestion
* [#3115](https://github.com/ckeditor/ckeditor-dev/issues/3115): Fixed: Destroying editor during initialization throws an error.
```

There was one issue with destroying inline editors based on `textarea` element.

engineering-this:
With fine amount of dirty hack replacing methods and stealing provided arguments I managed to cover all the cases with tests.

tests/core/skin/skin.js
```diff
@@ -55,6 +55,12 @@ bender.test( {
 		var skin = CKEDITOR.skin;
 
 		var editor = new CKEDITOR.editor();
+
+		editor.element = {
```
jacekbogdanski: Seems like you already fixed the issue with missing editor element. If so, do we still require this dirty stub among unit tests?

core/editor.js
```diff
@@ -1576,6 +1579,21 @@
...
+		},
+
+		/**
+		 * Returns boolean whether editor is destroyed or detached.
+		 *
+		 * @since 4.12.0
+		 * @private
+		 * @return {boolean}
+		 */
+		_shouldPreventInit: function() {
```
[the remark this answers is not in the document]
engineering-this: `isDestroyed` would be very misleading. Even when editor is detached it still should be destroyed for proper cleanup, such name would suggest that if editor `isDestroyed` you might not need to call `editor.destroy` anymore.

Sometimes I have this dilemma which name is better:
- what it does
- what's the purpose

In this case second option feels better for me. Eventually condition might be updated, but with such a name it wouldn't be a problem. If we went for first option it'd be named like `isDestroyedOrDetached`, and in that case updating condition might be not possible without changing a name. Unless we want to have outdated name which doesn't tell what it does any more.

core/creators/inline.js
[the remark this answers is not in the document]
engineering-this: > I'm not sure if I understand. Clearing custom data / removing editor from DOM won't cause any issues when it's detached? If so, additional check if a container is detached may be redundant.

Custom data is stored in one `customData` object, it's not accessible from global scope, but it has only one instance. Each element which has custom data has it's unique number stored as html attribute which is also key for entry in `customData. We can read html attribute of detached elements, so no issue here.
Also calling `element.remove()` on detached element doesn't seem to cause any issue.

>Also, should custom data by cleared only for textarea inline editor? It shouldn't be removed obviously, but we probably should remove our event listeners, etc. On the other hand, not sure if we should introduce such changes with this PR 🤔

In case of non-textarea based inline editor 
```editor.container.equals( editor.element ) === true```

So everything is cleaned here:
https://github.com/ckeditor/ckeditor-dev/blob/0d2458b99b811b5f235e0e1620f32aa22eca27b7/core/creators/inline.js#L109

**Comment by jacekbogdanski** [comment]
I see that I incorrectly tested browser against memory leaks, and the leaks are much smaller than I thought:

image

**Comment by engineering-this** [comment]
>i.e. in the manual test you assumed that the tester uses Chrome, but it's available also for Safari and Edge.

When initially working on memory leaks #2969 we decided to test only Chrome. We reduced amount of leaks there, so we can compare if given branch produces any new leaks or not. For other browsers we didn't test or fix leaks, doing so is out of scope of this ticket.

Other browser have memory profiling tools they aren't nearly as good as Chromes. Even if they allow to see how much memory is leaking they aren't really helpful to find out what is causing leaks.

## New changes

- Enabled unit tests for IE and Edge.
- Added test cases for all editor types.
- Reworked manual test, so that it is more precise at destroying editor where it caused issues.
- Fixed permission denied thrown by Edge. It looks like an upstream issue, as it randomly throws when trying to access `editable.$` of detached editor. It is fixed by wrapping some parts of code with `try/catch`, also caching `cke-data-expando` for proper custom data cleanup.

**Comment by jacekbogdanski** [comment]
Rebased into the latest `major`.

File: tests/core/editor/manual/detaching.html [hunk 11]
```diff
@@ -0,0 +1,158 @@
...
+		</ul>
+	</section>
+	<section id="destroy">
+		<h2>Destroy:</h2>
+		<ul>
+			<li><input type="radio" name="destroy" value="0" checked><label>synchronously</label></li>
+			<li><input type="radio" name="destroy" value="1"><label>async with 0 timeout</label></li>
+			<li><input type="radio" name="destroy" value="2"><label>before 'loaded' event</label></li>
+			<li><input type="radio" name="destroy" value="3"><label>before 'scriptLoader.load' callback called</label></li>
+			<li><input type="radio" name="destroy" value="4"><label>before iframe#onload</label></li>
```
jacekbogdanski: It looks like this button is disabled all the time despite using different options. Couldn't spot a way to enable it.

core/editor.js
```diff
@@ -1576,6 +1579,16 @@
...
+		},
+
+		/**
+		 * Returns boolean whether editor is destroyed or detached.
+		 *
+		 * @since 4.12.0
+		 * @return {boolean}
+		 */
+		shouldPreventInitialization: function() {
+			return this.status === 'destroyed' || this.container && this.container.isDetached();
```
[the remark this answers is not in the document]
engineering-this: The purpose is to tell when editor initialization should be prevented, so this name says everything you need to know. Condition under we prevent initialization might change in the future, so binding function name to the implementation doesn't feel right to me.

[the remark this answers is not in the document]
jacekbogdanski: This function has too low-level implementation for such high abstraction name. But let's leave it for now and back later during final refactoring.

core/editable.js
```diff
@@ -1264,6 +1270,17 @@
 		}
 	} );
 
+	function catchEdgePermissionDenied( callback ) {
+		// Edge randomly throws permission denied when trying to access native elements of detached editor (#3115).
+		try {
+			callback();
+		} catch ( error ) {
+			if ( !CKEDITOR.env.edge || error.number !== -2146828218 ) {
```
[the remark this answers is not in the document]
engineering-this: I think that it's possible but unlikely that error code will change. Yet it is safe, worst thing that could happen is that same error with different code will be thrown. However without using this check we might silent different error.

Previously we did it this way:
https://github.com/ckeditor/ckeditor-dev/blob/c4d250ea5fa0da09781a3ee84c99ca03f09c144b/core/editable.js#L100-L104

But IMO it's not optimal, as it might silent other errors. If error code changes tho, we might repeat logic from above.

tests/core/editor/manual/detaching.html
engineering-this: It is enabled only when browser is Firefox or IE (not Edge), and editor type is `classic`. Reason for that is only in this case `iframe#onload` is used.

https://github.com/ckeditor/ckeditor-dev/blob/9cb74b35db7082189088f47326eb05b02c947f8d/plugins/wysiwygarea/plugin.js#L49-L53

engineering-this:
The problem with Edge permission denied is that it's occurring randomly, I couldn't reproduce it outside of CKEditor. I doubt Edges team would confirm issue without easy reproduction steps.

By fixing it with try/catch on our side, we are not just silenting it, but also allowing execution of code that otherwise wouldn't be reached because of error. This allows to finish the cleanup, otherwise some data of removed elements won't be ever released. What's more that error is very visible in our framework integrations.

>Especially, that at some point MS may decide to fix the issue (when moving to chromium :trollface:) and we will end up with dead code.

When Chromium based MS Edge will be released we will have to investigate carefully how whole editor works on it. We have so many conditional code for Edge, that it all should be revised. I don't think that we should worry as much about how browsers might work in the future, instead we should focus on providing Editor that works right now. Even if it means sometimes we need to fix browser issue on our side.

>the only manual test is still written for Chrome (however, not ignored for other browsers this time). The most visible issue reproduction is error thrown in dev console. Can't it determine the test result for other browsers without more advanced dev tools?

Test is written for all of the browsers. It has EXTRA steps for Chrome to make sure no memory leaks are introduced with changes. When initially working on [memory leaks](https://github.com/ckeditor/ckeditor-dev/issues/589) we decided to work only with Chrome. There is no reason to put our effort to test memory leaks for every browser now if we didn't do it in dedicated ticket.

core/editable.js
```diff
@@ -1264,6 +1270,17 @@
 		}
 	} );
 
+	function catchEdgePermissionDenied( callback ) {
```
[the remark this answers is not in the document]
engineering-this: Replied here: https://github.com/ckeditor/ckeditor-dev/pull/3128#issuecomment-514125036

engineering-this:
Most important changes:
- Slightly reworked manual test so it works on IE8.
- Added indicator to manual test, so status of current editor is visible.

f1ames:
I think this will also fix #2257 and #718.

core/editable.js
```diff
@@ -1231,30 +1233,34 @@
...
-					if ( head.getCustomData( 'stylesheet' ) ) {
-						var refs = doc.getCustomData( 'stylesheet_ref' );
-						if ( !( --refs ) ) {
-							doc.removeCustomData( 'stylesheet_ref' );
-							var sheet = head.removeCustomData( 'stylesheet' );
-							sheet.remove();
-						} else {
-							doc.setCustomData( 'stylesheet_ref', refs );
+
+				catchEdgePermissionDenied( CKEDITOR.tools.bind( function() {
```
jacekbogdanski: As you move here line with `removeClass` method, it should be probably more readable to move `try/catch` clause into `this.detach` method but extract `try` code as a part of private `_.detach`. Something like:
 
```javascript
detach: function() {
	this.status = 'detached';

	// Update the editor cached data with current data.
	this.editor.setData( this.editor.getData(), 0, 1 );

	this.clearListeners();

	// Edge randomly throws permission denied when trying to access native elements of detached editor (#3115).
	try {
		this._.detach();
	} catch ( error ) {
		if ( !CKEDITOR.env.edge || error.number !== -2146828218 ) {
			throw( error );
		}
	}

	this.editor.fire( 'contentDomUnload' );

	delete this.editor.document;
	delete this.editor.window;
	delete this.editor;
},
```
and private helper:

````javascript
_: {
	detach: function() {
		// Update the editor cached data with current data.
		this.removeClass( 'cke_editable' );
		this.restoreAttrs();

		// Cleanup our custom classes.
		var classes;
		if ( ( classes = this.removeCustomData( 'classes' ) ) ) {
			while ( classes.length )
				this.removeClass( classes.pop() );
		}

		// Remove contents stylesheet from document if it's the last usage.
		if ( !this.is( 'textarea' ) ) {
			var doc = this.getDocument(),
				head = doc.getHead();
			if ( head.getCustomData( 'stylesheet' ) ) {
				var refs = doc.getCustomData( 'stylesheet_ref' );
				if ( !( --refs ) ) {
					doc.removeCustomData( 'stylesheet_ref' );
					var sheet = head.removeCustomData( 'stylesheet' );
					sheet.remove();
				} else {
					doc.setCustomData( 'stylesheet_ref', refs );
				}
			}
		}
	}
}
```

Also, renaming `_.detach` into something like `_.cleanUpCustomElementData` would improve readability.

core/dom/domobject.js
```diff
@@ -257,7 +264,22 @@ CKEDITOR.dom.domObject.prototype = ( function() {
...
+			expandoNumber = this.$[ 'data-cke-expando' ] || ( this.$[ 'data-cke-expando' ] = CKEDITOR.tools.getNextNumber() );
+
+			// Editable is cached unlike other elements, so we can use it to store expando number.
+			// We need it to properly cleanup custom data in case of permission denied
+			// thrown by Edge when accessing native element of detached editable (#3115).
+			if ( this instanceof CKEDITOR.editable ) {
+				this._.expandoNumber = expandoNumber;
+			}
+		} catch ( e ) {
+			expandoNumber = this._ && this._.expandoNumber;
```
jacekbogdanski: Could you make sure that this logic is actually cleaning up custom data by unit tests?

tests/core/editor/detaching.js
```diff
@@ -0,0 +1,174 @@
...
+		return function() {
+			var html = elementName === 'div' ? fakeComponent.replace( /textarea/g, 'div' ) : fakeComponent,
+				component = CKEDITOR.dom.element.createFromHtml( html ),
+				config = {};
+
+			container.append( component );
+
+			switch ( editorName ) {
+				case 'divarea':
+					config.extraPlugins = 'divarea';
```
jacekbogdanski: Missing `break`.

tests/core/editor/detaching.js
```diff
@@ -0,0 +1,174 @@
...
+				case 'classic':
+					editor = CKEDITOR.replace( 'editor', config );
+					break;
+				case 'inline':
+					editor = CKEDITOR.inline( 'editor' );
```
jacekbogdanski: Missing `break`.

tests/core/editor/manual/detaching.html
jacekbogdanski: Do you see it readable in a current shape? It looks like there is something broken when testing not listed browsers, not like it has been disabled on purpose. I see 2 options:
* enable it despite browser sniffing, so it will prevent against regressions as browser sniffing should be treated as an internal mechanism
* write some note at this radio so it's known that it should only work for FF and IE

I'm rather for unifying test case among browsers (enabling it).

tests/core/editor/manual/detaching.html
```diff
@@ -0,0 +1,186 @@
...
+		</ul>
+	</div>
+</form>
+<button id="createdestroy">Create editor</button>
```
jacekbogdanski: As you already created `form` for controlling radio buttons, shouldn't this button be a part of form also?

core/dom/element.js
```diff
@@ -2113,6 +2113,16 @@ CKEDITOR.dom.element.clearMarkers = function( database, element, removeFromDatab
...
+		},
+
+		/**
+		 * Checks if element is not attached to DOM.
+		 *
+		 * @since 4.13.0
+		 * @returns {Boolean} Whether element is detached from DOM.
+		 */
+		isDetached: function() {
+			return !this.getDocument().getDocumentElement().contains( this );
```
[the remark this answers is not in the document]
engineering-this: Is there a real use case to check if content of destroyed editor is detached?  If this comes from a listener - it will be removed with `editor.destroy()`.

tests/core/editor/detaching.js
engineering-this: `break` is not necessary here. Adding break requires to have duplicate `editor = CKEDITOR.replace( 'editor', config );`

So whole switch would look like below:
```javascript
			switch ( editorName ) {
				case 'divarea':
					config.extraPlugins = 'divarea';
					editor = CKEDITOR.replace( 'editor', config );
					break;
				case 'classic':
					editor = CKEDITOR.replace( 'editor', config );
					break;
				case 'inline':
					editor = CKEDITOR.inline( 'editor' );
			}
```

tests/core/editor/detaching.js
engineering-this: `break` is redundant at the end of `switch`.

core/dom/domobject.js
engineering-this: We can't check if custom data is cleaned as custom data object is private. I'll add tests to `editable.getUniqueId` to check if id is returned regardless of error.

tests/core/editor/manual/detaching.html
engineering-this: Button inside form has default action that refreshes the page. By having it outside of form we don't need to `evt.preventDefault()`.

tests/core/editor/manual/detaching.html
engineering-this: I'll add an message in test. This condition is the same as one deciding if `iframe#onload` should be used. So it doesn't matter if browser sniffing is working or not. This case is enabled only in environment which uses this listener for creating editor. Also this case could break test in other browsers, so I'm for having it disabled when it shouldn't be tested.

https://github.com/ckeditor/ckeditor-dev/blob/334fef5a0da86a853b788a0d1e8201eb18624ec2/plugins/wysiwygarea/plugin.js#L51-L51

engineering-this:
>Manual test still looks unresponsive when clicking the creation button. Any feedback from the test would be appreciated, even simple destroys counter which will show that actually, something happens when clicking a button.

Screenshot 2019-08-20 at 14 51 43
If this counter is not enough then what do you propose?

>Please, make sure that the issues from #3128 (comment) are covered by your changes. I see that some unit tests could be copied here.

Both PRs has same fix and same unit test, so only one test is copied. Also I added condition which ignores tests when init is called synchronously (reason for test to fail in IE8).

>Add tests for custom element data to make sure that even despite Edge error, everything is correctly cleaned up and editor target element is restored to the initial state (so no custom classes etc).

We can't check if custom classes or data are removed when trying to reach `element.$` throws permission denied.

>It would be really helpful if we could spot Edge upstream issue because as for now it looks like CKEditor upstream, not Edge.

I can't reproduce it without CKEditor. Error appears randomly. With the amount of various tricks that we do in our editor for Edge/IE, for instance [fixDomain](https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_tools.html#method-fixDomain), I think it might take a lot of time to find which set of actions leads to Edge issue.

f1ames:
> I can't reproduce it without CKEditor. Error appears randomly. With the amount of various tricks that we do in our editor for Edge/IE, for instance [fixDomain](https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_tools.html#method-fixDomain), I think it might take a lot of time to find which set of actions leads to Edge issue.

I think there is no sense in spending more time trying to cover Edge `Permission denied` error. One - there is already fix for that implemented in this PR (hacky `try/catch`, but still), two - all major browsers (basically all but Edge) are covered without any workarounds, three - when Chromium based Edge will be released this "special" fix will become obsolete (the date is still unknown, but from what I see it will be the next Edge version - and the [preview is already available](https://www.microsoftedgeinsider.com/en-us/)).

Comandeer:
Rebased onto latest `major`.

tests/core/dom/element/isdetached.js
```diff
@@ -0,0 +1,43 @@
+/* bender-tags: editor,dom */
+( function() {
+	var ancestor = new CKEDITOR.dom.element( 'div' ),
+		parent = new CKEDITOR.dom.element( 'div' ),
+		element = new CKEDITOR.dom.element( 'div' ),
+		body = CKEDITOR.document.getBody();
+
+	bender.test( {
```
Comandeer: I don't like the fact that all tests reuses the same set of elements. To be 100% sure that nothing is interfering with test's results, we should create separate DOM structures for every test.

tests/core/editable/domfix.js
```diff
@@ -8,6 +8,7 @@ bender.test( {
 	'async:init': function() {
 		var tc = this;
 		var editor = new CKEDITOR.editor();
+
```
Comandeer: Why such changes were introduced in this PR?

tests/core/editor/detaching.js
```diff
@@ -0,0 +1,176 @@
...
+
+	function testDetach( callback, editorName, elementName ) {
+		return function() {
+			var html = elementName === 'div' ? fakeComponent.replace( /textarea/g, 'div' ) : fakeComponent,
+				component = CKEDITOR.dom.element.createFromHtml( html ),
+				config = {};
+
+			container.append( component );
+
+			switch ( editorName ) {
```
Comandeer: Is this `switch` even needed? It probably can be done in a simpler, more readable way, e.g.

```javascript
var config = editorName === 'divarea' ? { extraPlugins: 'divarea' } : {},
	editor = CKEDITOR[ editorName === 'inline' ? 'inline' : 'replace' ]( 'editor', config );
```

Additionally isn't overriding config with `config` variable also overriding all plugins loaded via `bender-ckeditor-plugins` comment?

core/dom/element.js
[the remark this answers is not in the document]
f1ames: I think this will occur for some Edge cases in IE8 (and it's still better than the previous version) so maybe we can accept that fact? WDYT?

core/editor.js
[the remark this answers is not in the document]
f1ames: I agree, especially that docs says the same:
> Returns boolean whether editor is destroyed or detached.

👍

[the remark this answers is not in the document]
msamsel: I've come with an idea to replace this method with simple `isDetached()` and implement `or` statement in places where we actually checking if editor is detached or destroyed.
This removes this additional abstraction here and remains much cleaner API.

tests/core/editable/domfix.js
msamsel: I've remove this chaneg together with simmilar cases in other files.

msamsel:
* I deeply refactor tests related to detaching to have a more flat structure. I also implement some sort of cleaning mechanism there.
* I removed a manual test, as main purpose of this PR is to not throw an error when editor is destroyed during initialization. Memory optimization is just nice to have feature, but it wasn't the main goal here.
* I also add fixes for other places where reviewer point out.
* I removed completely method `isDetachedOrDestroyed` and remain only `isDetached` on the editor. Checking if the editor is detached or destroyed is made where it is needed.
* I add and improve docs entries.

msamsel:
I don't know who is reviewing this ticket now. @Comandeer you were the last one who review this issue, that's why I ask you about it. However, if this should be someone else, then feel free to reassign the PR.

core/creators/themedui.js
```diff
@@ -239,6 +239,9 @@ CKEDITOR.replaceClass = 'ckeditor';
 
 			// Delay to avoid race conditions (setMode inside setMode).
 			setTimeout( function() {
+				if ( editor.status === 'destroyed' || editor.isDetached() ) {
```
Comandeer: I wonder about introducing `editor.isDestroyed()`. Basically it would be just a nicer way to write `editor.status === 'destroyed'`.

Also I find

```javascript
if ( editor.isDestroyed() || editor.isDetached() )
```

more elegant than the current version.

Yes, I know, I have obsession about readability and elegance of code 😄

core/dom/domobject.js
```diff
@@ -126,15 +126,24 @@ CKEDITOR.dom.domObject.prototype = ( function() {
...
+					} else if ( this.$.removeEventListener ) {
+						this.$.removeEventListener( eventName, listener, false );
+					}
+
+					delete nativeListeners[ eventName ];
+				}
+			// Catch Edge `Permission denied` error which occurs randomly. Since the error is quite
+			// random, catching allows to continue the code execution and cleanup.
+			} catch ( error ) {
+				if ( !CKEDITOR.env.edge || error.number !== -2146828218 ) {
```
Comandeer: Maybe extract issue about this weird Edge error to a new issue? 🤔 We can investigate it further when we find some time.

core/dom/element.js
```diff
@@ -2148,6 +2148,17 @@ CKEDITOR.dom.element.clearMarkers = function( database, element, removeFromDatab
...
+		},
+
+		/**
+		 * Checks if element is not attached to DOM.
+		 *
+		 * @since 4.13.0
+		 * @returns {Boolean} Whether element is detached from DOM.
+		 */
+		isDetached: function() {
+			return !this.getDocument().getDocumentElement().contains( this ) ||
```
Comandeer: This condition looks a little bit unreadable. I'd split it into two parts (variables):

* `isInDocument` → `this.getDocument().getDocumentElement().contains( this )`,
* `isDocumentAttached` → `( ( !CKEDITOR.env.ie || CKEDITOR.env.version > 8 ) && this.getDocument().$.defaultView )`.

I'm also wondering about the second case in IE8. I'm feeling that we can polyfill it by something like `this.getDocument().$.parentWindow.frameElement` and then check if `frameElement` is inside its own document. However I'm not sure if it's really worth the trouble.

core/editable.js
```diff
@@ -534,19 +534,28 @@
...
-				// Cleanup the element.
-				this.removeClass( 'cke_editable' );
-
 				this.status = 'detached';
 
-				// Save the editor reference which will be lost after
-				// calling detach from super class.
-				var editor = this.editor;
+				// Update the editor cached data with current data.
+				this.editor.setData( this.editor.getData(), 0, 1 );
```
Comandeer: Shouldn't `getData` be marked as internal one here?

core/editable.js
```diff
@@ -534,19 +534,28 @@
...
+
+				this.clearListeners();
+
+				// Edge randomly throws permission denied when trying to access native elements of detached editor (#3115).
+				try {
+					this._.cleanCustomData();
+				} catch ( error ) {
+					var isMSBrowser = CKEDITOR.env.edge || CKEDITOR.env.ie;
```
Comandeer: `CKEDITOR.env.ie` is true also in Edge, so this condition is redundant.

tests/core/dom/element/isdetached.js
```diff
@@ -0,0 +1,88 @@
+/* bender-tags: editor,dom */
```
Comandeer: There is no test for detached document.

tests/core/editor/destroyrace.js
```diff
@@ -0,0 +1,47 @@
+/* bender-tags: editor,unit */
+/* bender-ckeditor-plugins: toolbar */
+
+( function() {
+	'use strict';
+
+	bender.test( {
+		// (##718), (#2257)
```
Comandeer:
```suggestion
		// (#718, #2257)
```

core/dom/element.js
msamsel: I feel like mentioned polyfill is not suitable here or I just don't how it supposed to work :(
In original solution what I understand we cover case where the entire document is detached with method like: `document.implementation.createHTMLDocument()`. In such a case, there is no related `window` object and we get null. I couldn't find any possibility of how to create a detached document on IE8. That's why I'm not sure how checking the `frameElement` and its presence in own document might replace it.
Could you provide more details about this polyfill? For now, I just ignore this check on IE8 and remain current functionality.

msamsel: I made early returns to simplify this function.

core/editable.js
msamsel: It sounds reasonable

tests/core/dom/element/isdetached.js
msamsel: I add test for current document, for new detached document, and some element inside detached document.

msamsel:
I correct test which faileed due to change introduced by: https://github.com/ckeditor/ckeditor4/pull/3128#discussion_r324460294
And rebase to current major

f1ames:
Taking over this review from @Comandeer. Rebased onto latest `major`.

Issue #3131, opened 2019-05-24
Allow override of default CKEditor.config.protectedSource rules


capJavert:
## Type of report

Feature request

## Provide description of the new feature

Hello,

we are using Media embed plugin to load youtube/instagram/facebook posts inside our richtext editor. Problem that we encountered is that ckeditor 4.x auto protects `<script>` tags by default and does not give option to remove that protection. Even through `CKEditor.config.protectedSource`. 

I followed the problem to this line of source:
https://github.com/ckeditor/ckeditor-dev/blob/major/core/htmldataprocessor.js#L967

While I understand why this kind of "hardcoded" protection would be implemented I think it breaks some cool features of Media embed plugin.

What I would like to propose is to add opt-in flag to `CKEditor.confg` like:
```
allowDangerousProtectedSoruce : Boolean

If true, disables CKEditor.protectedSource default restrictions for tags like script, noscript etc.

config.allowDangerousProtectedSoruce = true;
```

I am also open to discuss further for alternatives and if needed implement this as a PR.

Thanks!

f1ames:
Hello @capJavert,

thanks for the request and in-depth analysis of this behavior 👍 

> While I understand why this kind of "hardcoded" protection would be implemented I think it breaks some cool features of Media embed plugin.

Yes, it is mostly done from security reasons as allowing script tags would be big security risk. I also agree that many of media embed scripts loads additional JS which adds some cool stuff/features to the embedded content, so without it the experience is not the same.

---

I see two solutions - feature in CKEditor or additional integration layer on the backend (or content post processing).

Adding a config option in editor is one of the approaches (we already have [ACF](https://ckeditor.com/docs/ckeditor4/latest/guide/dev_acf.html) which is kind of similar and when turned off allows most of the content - apart from `script` tags😉) so it's doable and we are open for PRs so feel free to contribute.

The other solution would be content post-processing. After the content is fetched from CKEditor and saved to the storage you modify it be replacing protected `script` tags by regular ones. This way on the frontend layer (where the content is presented) you will have full featured media embeds. This can be done by 3rd-party integration or using CKEditor data pipeline and [`toDataFormat`](https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_editor.html#event-toDataFormat) event.

capJavert:
Ok cool, I am also open to contribute so I will hopefully start to work on it soon. 

I will reference this issue when PR is opened.

capJavert:
[...]

PR #3132, opened 2019-05-25
feature(allow-protected-overrides): added new allowDangerousTags flag

capJavert:

Adding feature for issue #3131 

Yes

### This PR contains

- [x] Unit tests
- [x] Manual tests

## What changes did you make?

* Added new CKEDITOR.config.allowDangerousTags
* implemented override inside htmldataprocessor
* added console warning if flag is enabled during apply of protectSource method
* added docs for config flag
* added unit tests
* added manual test for dev console warning

### DIFF
--- a/core/config.js
+++ b/core/config.js
@@ -357,15 +357,14 @@ CKEDITOR.config = {
 	protectedSource: [],
 
 	/**
-	 * If set to true, **disables** CKEditor.protectedSource default restrictions
-	 * for tags like script, noscript and other. **This can potentially allow execution
-	 * of malicious scripts inside the editor**. Use with caution.
+	 * If set to true, **disables** default protected tags restrictions allowing `script`, `noscript` and `meta` tags.
+     * **This can potentially allow execution of malicious scripts inside the editor**. Use with caution.
 	 *
 	 *		config.allowDangerousProtectedSoruce = true;
 	 *
 	 * @cfg
 	 */
-	allowDangerousProtectedSource: false,
+	allowDangerousTags: false,
 
 	/**
 	 * The editor `tabindex` value.
--- a/core/config.js
+++ b/core/config.js
@@ -357,6 +357,17 @@ CKEDITOR.config = {
 	protectedSource: [],
 
 	/**
+	 * If set to true, **disables** CKEditor.protectedSource default restrictions
+	 * for tags like script, noscript and other. **This can potentially allow execution
+	 * of malicious scripts inside the editor**. Use with caution.
+	 *
+	 *		config.allowDangerousProtectedSoruce = true;
+	 *
+	 * @cfg
+	 */
+	allowDangerousProtectedSource: false,
+
+	/**
 	 * The editor `tabindex` value.
 	 *
 	 * Read more in the {@glink guide/dev_tabindex documentation} and see the
--- a/core/htmldataprocessor.js
+++ b/core/htmldataprocessor.js
@@ -61,7 +61,7 @@
 			// Also if CKEDITOR.config.allowDangerousTags is enabled
 			// display warning inside the console
 			if ( editor.config.allowDangerousTags === true ) {
-				console.warn( 'Flag CKEDITOR.config.allowDangerousTags is set to true. This can potentially allow execution of malicious scripts inside the editor. Use with caution.' );
+				CKEDITOR.warn('dangerous-tags-allowed')
 			}
 			data = protectSource( data, editor );
 
--- a/core/htmldataprocessor.js
+++ b/core/htmldataprocessor.js
@@ -58,10 +58,10 @@
 
 			// The source data is already HTML, but we need to clean
 			// it up and apply the filter.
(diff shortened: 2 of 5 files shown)


core/config.js
```diff
@@ -356,6 +356,17 @@ CKEDITOR.config = {
...
+	/**
+	 * If set to true, **disables** CKEditor.protectedSource default restrictions
+	 * for tags like script, noscript and other. **This can potentially allow execution
+	 * of malicious scripts inside the editor**. Use with caution.
+	 *
+	 *		config.allowDangerousProtectedSoruce = true;
+	 *
+	 * @cfg
+	 */
+	allowDangerousProtectedSource: false,
```
f1ames: WDYT about renaming it to something like `allowDangerousTags`? Since it has effect only on hard coded protected tags and doesn't have any influence on how `editor.config.protectedSource` it is somehow a separate concept.

And then I will modify docs to something like:

```
	/**
	 * If set to true, **disables** default protected tags restrictions allowing `script`, `noscript` and `meta` tags. 
     * **This can potentially allow execution of malicious scripts inside the editor**. Use with caution.
	 *
	 *		config.allowDangerousTags = true;
	 *
	 * @cfg
	 */
```

core/htmldataprocessor.js
```diff
@@ -58,6 +58,11 @@
 
 			// The source data is already HTML, but we need to clean
 			// it up and apply the filter.
+			// Also if CKEDITOR.config.allowDangerousProtectedSource is enabled
+			// display warning inside the console
+			if ( editor.config.allowDangerousProtectedSource === true ) {
+				console.warn( 'Flag CKEDITOR.config.allowDangerousProtectedSource is set to true. This can potentially allow execution of malicious scripts inside the editor. Use with caution.' );
```
f1ames: Good idea, I wonder if it should be described in docs errors section (https://ckeditor.com/docs/ckeditor4/latest/guide/dev_errors.html) 🤔

capJavert:
@f1ames I added you as a contributor to my fork. Dunno about the settings of PR. I don't see that option anywhere.

> If you could add unit and manual test for this change that will make the PR complete.

I was planning to add tests. No problem.

> One more thing, I wonder if adding such config doesn't pose any security issues 🤔 The config can be easily changed by running one-liner in the browser dev console which will allow easier attack potential attack vectors (manually modifying the code in browser dev console is also doable but requires a little more effort, especially when minified).

I will also think about security a little bit more and get back to you.

capJavert:
> One more thing, I wonder if adding such config doesn't pose any security issues 🤔 The config can be easily changed by running one-liner in the browser dev console which will allow easier attack potential attack vectors (manually modifying the code in browser dev console is also doable but requires a little more effort, especially when minified).

If you think about code modification or accessing global editor scope then attacker can do just about anything if his script is loaded on the page. I think there is nothing we can do about it.

What we can do is to make sure users know that feature exists and what it does. 

Few of those things were already mentioned inside this thread.. like:
- console.warning message
- we should add additional explanation to dev docs as @f1ames said
- naming prop with `Dangerous` suffix

core/config.js
capJavert: Updated.

core/htmldataprocessor.js
capJavert: Where should I add docs for this?

capJavert:
@f1ames added tests adjusted docs. Please check if this is ok. 

Also, should i rebuild `ckeditor.js` in root of project after my changes? How is that handled? 

Tnx

core/htmldataprocessor.js
capJavert: Added [pull request](https://github.com/ckeditor/ckeditor-docs/pull/251) to ckeditor-docs. Also replaced `console.warn` with `CKEDITOR.warn`.

capJavert:
@f1ames just the status update. I think that I am done with this. Tests and docs are also added. 
Can I get any info from you about merging this?

We are planning on releasing a version of our product and would appreciate if this can be merged and released soon.

Tnx

Comandeer:
I'm not a big fan of this solution. If I understand correctly, the use case here is to display rich preview of embedded multimedia content? If so, I don't think that switching off one of the security features for the whole editor is a right approach. I'd think about implementing additional sandbox for the media content only, which allows to fire all scripts connected with this content without any influence on the editor (e.g. using `iframe[sandbox]`).

capJavert:
@Comandeer ill look into this options. Is there some example of such use case/implementation inside current ckeditor codebase so I can look into it?

f1ames:
Hello again @capJavert!
AFAIR this PR was mainly motivated by the ability to proper render media embeds inside CKEditor 4 (#3131)? We already started to work on a PR to tackle this problem (in more safety manner - see https://github.com/ckeditor/ckeditor-dev/issues/3264) and so I'm closing this PR. Hopefully, we will be able to deliver #3264 soon, so you may switch to built-in solution🤞

PR #3168, opened 2019-06-14
Enable widget clipboard manipulation

### LINKED ISSUE #3138: Possibility to override clipboard content when widget is copied/dragged
## Type of report

Feature request

There are some cases when adjusting HTML which lands in the clipboard when widget is copied or dragged is necessary. We need to find a way to add possibility to modify/adjust clipboard HTML when widget is being copied or dragged.

There is already an PoC branch available - https://github.com/ckeditor/ckeditor-dev/tree/widget-improvements.

jacekbogdanski:

Feature request

### This PR contains

- [x] Unit tests
- [x] Manual tests

## Proposed changelog entry

```
* [#3138](https://github.com/ckeditor/ckeditor-dev/issues/3138): Use [`widgetDefinition.getClipboardHtml()`](https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_plugins_widget.html#method-getClipboardHtml) method to customize [Widget](https://ckeditor.com/cke4/addon/widget) HTML during copy, cut and drag operations.
```

## What changes did you make?

Added new widget `getClipboardHtml` member allowing to customize HTML copied to the clipboard during copy, cut and drop operations.

Closes #3138.

### DIFF
--- a/plugins/widget/plugin.js
+++ b/plugins/widget/plugin.js
@@ -3419,7 +3419,7 @@
 				widget.edit();
 				// CTRL+C or CTRL+X.
 			} else if ( keyCode == CKEDITOR.CTRL + 67 || keyCode == CKEDITOR.CTRL + 88 ) {
-				// copyWidgets( widget.editor, keyCode == CKEDITOR.CTRL + 88 );
+				copyWidgets( widget.editor, keyCode == CKEDITOR.CTRL + 88 );
 				return; // Do not preventDefault.
 				// Pass chosen keystrokes to other plugins or default fake sel handlers.
 				// Pass all CTRL/ALT keystrokes.
--- a/plugins/widget/plugin.js
+++ b/plugins/widget/plugin.js
@@ -859,6 +859,23 @@
 
 		// Extend this widget with widgetDef-specific methods and properties.
 		CKEDITOR.tools.extend( this, widgetDef, {
+
+			/**
+			 * Returns HTML of the widget. Can be overridden to customize HTML copied to the clipboard
+			 * during copy&cut and drag&drop events.
+			 *
+			 * @since 4.12.0
+			 * @returns {String} Widget HTML.
+			 */
+			getClipboardHtml: function() {
+				var range = this.editor.createRange();
+
+				range.setStartBefore( this.wrapper );
+				range.setEndAfter( this.wrapper );
+
+				return this.editor.editable().getHtmlFromRange( range ).getHtml();
+			},
+
 			/**
 			 * The editor instance.
 			 *
@@ -2592,7 +2609,7 @@
 			delete CKEDITOR.plugins.clipboard.dragStartContainerChildCount;
 			delete CKEDITOR.plugins.clipboard.dragEndContainerChildCount;
 
-			evt.data.dataTransfer.setData( 'text/html', editor.editable().getHtmlFromRange( dragRange ).getHtml() );
+			evt.data.dataTransfer.setData( 'text/html', sourceWidget.getClipboardHtml() );
 			editor.widgets.destroy( sourceWidget, true );
 		} );
 
@@ -3228,13 +3245,9 @@
 
 		copybin.setStyle( editor.config.contentsLangDirection == 'ltr' ? 'left' : 'right', '-5000px' );
 
-		var range = editor.createRange();
-		range.setStartBefore( widget.wrapper );
-		range.setEndAfter( widget.wrapper );
-
 		copybin.setHtml(
 			'<span data-cke-copybin-start="1">\u200b</span>' +
-			editor.editable().getHtmlFromRange( range ).getHtml() +
+			widget.getClipboardHtml() +
 			'<span data-cke-copybin-end="1">\u200b</span>' );
 
 		// Save snapshot with the current state.
@@ -3257,7 +3270,7 @@
 		// Once the clone of the widget is inside of copybin, select
 		// the entire contents. This selection will be copied by the
 		// native browser's clipboard system.
-		range = editor.createRange();
+		var range = editor.createRange();
 		range.selectNodeContents( copybin );
 		range.select();
 
--- a/plugins/widget/plugin.js
+++ b/plugins/widget/plugin.js
@@ -3221,13 +3221,20 @@
 	function copyWidgets( editor, isCut ) {
 		var doc = editor.document,
 			focused = editor.widgets.focused,
-			isEdge16 = CKEDITOR.env.edge && CKEDITOR.env.version >= 16;
+			isEdge16 = CKEDITOR.env.edge && CKEDITOR.env.version >= 16,
+			bookmarks;
 
 		// We're still handling previous copy/cut.
 		// When keystroke is used to copy/cut this will also prevent
 		// conflict with copySingleWidget called again for native copy/cut event.
 		if ( doc.getById( 'cke_copybin' ) ) {
 			return;
+		}
+
+		// When more than one widget is selected, we must save selection to restore it
+		// after destroying copybin (#3138).
+		if ( !focused && !isCut ) {
+			bookmarks = editor.getSelection().createBookmarks( true );
 		}
 
 		// [IE] Use span for copybin and its container to avoid bug with expanding
@@ -3293,6 +3300,10 @@
 			// [IE] Focus widget before removing copybin to avoid scroll jump.
 			if ( !isCut && focused ) {
 				focused.focus();
+			}
+
+			if ( bookmarks ) {
+				editor.getSelection().selectBookmarks( bookmarks );
 			}
 
 			copybinContainer.remove();
--- a/plugins/widget/plugin.js
+++ b/plugins/widget/plugin.js
@@ -4580,6 +4580,9 @@
  *
  * If not set, current widget HTML will be used instead.
  *
+ * Note: this method will overwrite HTML for the whole widget, **including**
+ * any nested widgets.
+ *
  * @method getClipboardHtml
  * @since 4.13.0
  * @returns {String} Widget HTML.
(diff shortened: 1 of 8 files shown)


Comandeer:
Rebased onto latest `major`.

Comandeer:
I've update the implementation to use internally our `toDataFormat` event. It allows us to downcast widgets before copying their content.

I haven't add unit tests, because I want some early review to see if the whole approach is feasible at all.

f1ames:
@jacekbogdanski I remember you talked about some selection blinking when trying to come up with the solution, can you just take a quick glimpse on @Comandeer proposal to see if the solution is similar and same issue may occur?

plugins/widget/plugin.js
```diff
@@ -2910,13 +2931,20 @@
...
 				widget = toBe.widget;
 				widgetElement = toBe.element;
 				retElement = widget._.downcastFn && widget._.downcastFn.call( widget, widgetElement );
 
+				// In case of copying widgets, we replace the widget with clipboard data (#3138).
+				if ( evt.data.widgetsCopy && widget.getClipboardHtml ) {
+					parserFragment = CKEDITOR.htmlParser.fragment.fromHtml( widget.getClipboardHtml() );
+
+					retElement = parserFragment.children[ 0 ];
+				}
```
f1ames: I wonder how it will work with the code below, managing editables output 🤔 Does the fact that custom widget HTML (`getClipboardHtml()`) is provided means any editables will be simply replaced by this custom HTML? If so there is no need to process them in any way. WDYT?

And second thing, I wonder how it will (or should) work with nested widgets? Can you check it and also add such test (we don't have many nested widget cases but still worth checking)?

plugins/widget/plugin.js
```diff
@@ -4253,6 +4313,17 @@
  * @property {String} button
  */
 
+/**
+ * Customizes widget HTML copied to the clipboard
```
f1ames: Isn't it duplicating `getClipboardHtml()` method defined above (https://github.com/ckeditor/ckeditor-dev/pull/3168/files#diff-d932030cd4e22ec8b0e30a362269b7e8R1244)? 🤔

plugins/widget/plugin.js
```diff
@@ -3245,10 +3282,33 @@
...
+
+			if ( editor.widgets.focused ) {
+				return editor.widgets.focused.getClipboardHtml();
+			}
+
+			editor.once( 'toDataFormat', function( evt ) {
+				evt.data.widgetsCopy = true;
+			}, null, null, -1 );
+
+			return editor.dataProcessor.toDataFormat( selectedHtml );
```
jacekbogdanski: This execution operates on the whole selected HTML, which means it also touches non-widget content. It seems for me that such change is a bit out of the scope of the widget plugin. Maybe it should be implemented directly inside `clipboard` plugin?

Although it would require moving whole copybin logic into the clipboard, which maybe not the best idea as it will complicate whole improvement :thinking:

plugins/widget/plugin.js
```diff
@@ -3245,10 +3282,33 @@
...
-				editor.fire( 'saveSnapshot' );
+				handleCut();
 			}
 		}, 100 ); // Use 100ms, so Chrome (@Mac) will be able to grab the content.
+
+		function getClipboardHtml() {
+			var selectedHtml = editor.getSelectedHtml( true );
+
+			if ( editor.widgets.focused ) {
+				return editor.widgets.focused.getClipboardHtml();
```
jacekbogdanski: I'm wondering if there is still need for a single widget copy handling :thinking: Maybe we could simplify it into a single code path by operating only on selection?

plugins/widget/plugin.js
```diff
@@ -2494,6 +2511,7 @@
 	// * FF tends to copy all blocks up to the copybin container.
 	// * IE tends to copy only the copybin, without its container.
 	// * We use spans on IE and blockless editors, but divs in other cases.
+	// We are really sorry for summoning Zalgo…
```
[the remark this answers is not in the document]
jacekbogdanski: :scream:

Comandeer:
> After cut/paste widget it is not possible to cut again if entire content is selected:

Actually the new logic is not even fired when the whole content is selected due to  #3352. I found it during trying to write unit test for the new logic handling multiple widgets.

plugins/widget/plugin.js
Comandeer: Everything seems to work correctly. Editables and nested widgets are replaced by the outermost widget's `getClipboardHtml` result.

plugins/widget/plugin.js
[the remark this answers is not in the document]
Comandeer: I'd keep it separated, as I find having nice, named variables more readable.

Comandeer: The result should be the same as if we would transform only widgets IMO. Moving whole copybin logic to the `clipboard` plugin would be nice, but I fear that it could be too much work.

plugins/widget/plugin.js
Comandeer: One is defined directly on `CKEDITOR.plugins.widget` class and one on widget's definition.

Comandeer:
> Overall it looks nice, especially that widgets are now downcasted, but as already @f1ames mentioned, I'm wondering if we can go with such big change

Actually by default the upcasted HTML is returned, via default `getClipboardHtml` implementation. So without implementing custom `getClipboardHtml` everything should work the same as previously. `toDataFormat` is used only as a simple way to  utilize our HTML parser.

I'm thinking about making `getClipboardHtml` compliant with `downcast`, so it wouldn't operate on HTML string, but rather `htmlParser.node`s. WDYT of such a change?

plugins/widget/plugin.js
Comandeer: Unfortunately something's wrong with selection then 😞

Comandeer:
I've fixed failing tests in IE8 and Safari. Additionally I've refactored copybin to have a nicer API via `CopyBin` class. Enhancing it further should allow us to finally move it to `clipboard` (see #3358).

The only issue left is the one with selecting all content and cutting, but I'm afraid that it's connected with deeper issues with selecting widgets, as can be seen in unit tests, e.g. Safari has incorrect selection, when only two nested widgets are selected and IE8 does not select widgets if whole wrappers are not contained inside the range.

f1ames:
> The only issue left is the one with selecting all content and cutting, but I'm afraid that it's connected with deeper issues with selecting widgets, as can be seen in unit tests, e.g. Safari has incorrect selection, when only two nested widgets are selected and IE8 does not select widgets if whole wrappers are not contained inside the range.

Does it always happen? I recall you mentioned something about `widgetselection` plugin, but can't find it now 🤔 If it happens always (and not only with/without the plugin) that's pretty bad news😭 Anyway, @Comandeer would you be trying to cover this issue in this PR or prefer to extract as a follow-up (still we will need to fix it soon probably).

Comandeer:
It's already extracted as #3352. Probably it would be much more difficult to fix than the proposed refactor of clipboard handling and because of that I don't see much sense in delaying this change.

f1ames:
Rebased onto latest `major`.

plugins/widget/plugin.js
f1ames: Shouldn't we mentioned in the `getClipboardHtml()` docs that for nested widget it will overwrite/replace entire widget HTML (no matter nesting/editables content)? Or is it obvious? 🤔

plugins/widget/plugin.js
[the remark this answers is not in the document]
f1ames: I wonder if we really need this line 😛

tests/plugins/widget/manual/clipboardhtml.md
```diff
@@ -0,0 +1,34 @@
...
+@bender-ui: collapsed
+@bender-ckeditor-plugins: widget, undo, wysiwygarea, toolbar
+
+**NOTE:** It's expected that if you paste the same copied content multiple times, widget copy counter won't change. Counter changes only during `copy`, `cut` and `drop` operations. Please also note that the counter is the same for every widget instance.
+
+1. Manipulate the content by:
+
+	* copying and pasting
+	* cutting and pasting
+	* dragging and droping
```
f1ames: Not sure why this happens:

image

but there is something wrong with formatting :P

Comandeer:
> I don't see you have addressed blinking issue mentioned earlier (and it's still visible), have you checked if anything could be done there?

It seems that selection is updated asynchronously in every browser and therefore our copybin must be asynchronous. However it seems that replacing `setTimeout` with `requestAnimationFrame` in supported browsers fix the issue. The blinking is visible in IEs (which don't support rAF) and sometimes in other browsers, when rAF is not performing in 60 FPS.

> there is something wrong with formatting :P

It seems that our markdown parser can't into nested lists. The formatting is wrong as everything after "dragging and droping" is inserted into the last item of the nested list. I've just replaced the list with three consecutive paragraphs.

> Can you elaborate a bit more on this one?

I can, but on second thought I realised that it was a really bad idea 😅 Let's keep everything the way it is.

I've rebased everything on `major` and during it I allowed myself to remove changelog related commits. Instead I added latest changelog entry version into the PR's description.

Comandeer:
I haven't fix #3423, however I added a simple workaround. The real fix would probably involve changing the core logic of creating ranges, which seems too much for this PR. There is however one more bug, more general, with pasting and `widgetselection`, #3427.

f1ames:
Rebased onto latest `major`.

f1ames:
Failing cut is caused by removing timeout from Chrome on macOS (this line - https://github.com/ckeditor/ckeditor4/pull/3168/files#diff-d932030cd4e22ec8b0e30a362269b7e8L3286) by replacing it with RequestAnimationFrame.

PR #3213, opened 2019-06-25
Selection optimizations

### LINKED ISSUE #3118: Selecting a paragraph 3 times and applying heading 1 on that paragraph applies heading on next paragraph as well.
## Type of report

Bug

1.  Paste this text in ckeditor.com demo page
The purpose of lorem ipsum is to create a natural looking block of text (sentence, paragraph, page, etc.) that doesn't distract from the layout. A practice not without controversy, laying out pages with meaningless filler text can be very useful when the focus is meant to be on design, not content.

The passage experienced a surge in popularity during the 1960s when Letraset used it on their dry-transfer sheets, and again during the 90s as desktop publishers bundled the text with their software. Today it's seen all around the web; on templates, websites, and stock designs. Use our generator to get your own, or read on for the authoritative history of lorem ipsum.

2. Select first paragraph by clicking on it 3 times.
3. Select Heading 1 from drop down.
4. The next paragraph also becomes heading.

### Expected result

Only first paragraph should change to Heading1

### Actual result

Next paragraph also changes to Heading1

## Other details

* Browser: All
* OS: All
* CKEditor version: 4.11.4
* Installed CKEditor plugins: not relevant

### LINKED ISSUE #3161: Double click on one word span produces buggy selection
## Type of report

Feature

1. Once `t/2813` (#2987) is merged open manual test `tests/core/editable/manual/insert`
2. Place collapsed selection in the middle of one spans `Foo ^bar`.
3. Press `insert element`.
4. Double click on created `bar` span.
5. Press `insert element`

### Expected result

Span containing word `bar` is completely removed.

### Actual result

Span is preserved only it's text is removed.

## Other details

This is an upstream browser issue, so it is labeled as a feature to provide a workaround. To fix this we should introduce logic to optimize selection, which would also fix #3118.

* Browser: Firefox
* OS: macOS
* CKEditor version: development of 4.12.0

### LINKED ISSUE #3175: [Umbrella] Selection ending on the beginning of the following block element causes buggy behaviours
## Type of report

Bug

It is possible to create a selection like:

```html
<p>[Foo Bar</p>
<p>]Baz</p>
```

or 

```html
<p>Foo [Bar</p>
<p>]Baz</p>
```

which causes few buggy behaviours in the editor. Such selection can be created with:

* Triple click (Chrome and IE AFAIR)
* Double click (some IEs)
* Mouse selection (all browsers)
* Keyboard selection (all browsers)

Such selection affects for example - applying block styles (which are applied to both block instead of the first one), removing content (when next block is merged into the previous one) and few other cases. These were reported in #3118, #3119, #3161 and https://github.com/ckeditor/ckeditor-dev/issues/3161#issuecomment-502731204

Since it touches many editor mechanisms and there are few ways such selection could be achieved (e.g. triple click) the proper solution seems to be a generic mechanism which will fix such selection upon selection event (probably `selectionCheck`).

## Other details

* Browser: All
* OS: All
* CKEditor version: It's rather and old issue and partially upstream too (triple click), I would assume 4.5+
* Installed CKEditor plugins: All

### LINKED ISSUE #3256: Triple-click and deleting text in last table cell pulls in content below the table
## Type of report

Bug

1. Go to [ckeditor demo](https://ckeditor.com/ckeditor-4/demo/)
2. Insert a new 3x2 table above some existing content
3. Enter some text in the last cell of the table (bottom-right)
4. Triple-click the text in the table to select it
5. Press the backspace or delete key

### Expected result

The text should be removed

### Actual result

The text is removed but the paragraph of content below the table is pulled into the last cell of the table
Capture1

Capture2

## Other details

* Browser: Chrome - (Firefox does not have the issue)
* CKEditor version: 4.11
* Installed CKEditor plugins: Standard Set (with table plugin)

engineering-this:

Bug fix

### This PR contains

- [X] Unit tests
- [X] Manual tests

## What changes did you make?

Selection might cause various buggy behaviours. As a workaround I propose to adjust selection ranges on `selectionCheck` event. `selection.js` has 2500 lines of code so I created new module.

## Notes

This is an **early PR**, there are still some things to be done, but first I'd like some feedback on proposed solution.

Right now manual tests are enabled on all envs, however some of cases are reproducible on specific browsers, so please check corresponding issues for more details.

TODO:
- Should we make the method private? It's used in one place.
- API docs.
- Fix failing tests:
Screenshot 2019-06-25 at 14 00 26
☝️ Some tests need small adjustments - missing method added within this PR in mocked `selection` instance., others rely on selection which we prevent, so probably these cases can be removed. It doesn't look like anything is broken.

Closes #3175 
Closes #3118
Closes #3161
Closes #3256


engineering-this:
This should be retargetted and rebased to `master`, but first `master` needs to be ready for 4.12.1 commits.

core/selectionoptimizer.js
```diff
@@ -0,0 +1,57 @@
...
+		if ( range.startOffset === limit ) {
+			return true;
+		}
+
+		return startsInText ^ range.endContainer.type === CKEDITOR.NODE_TEXT;
+	}
+
+	// Prevent infinite recurrency when browser doesn't allow expected selection.
+	function preventRecurrency( editor, range ) {
+		editor.once( 'selectionCheck', function( evt ) {
```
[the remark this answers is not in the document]
f1ames: The other thought was to add just regular listener (`on`) and control it with just some flag/switch variable, but it doesn't seem to solve the above (potential) problem.

tests/core/selectionoptimizer.js
```diff
@@ -0,0 +1,68 @@
...
+		return function( editor ) {
+			bender.tools.selection.setWithHtml( editor, options.initial );
+
+			var actual = bender.tools.selection.getWithHtml( editor );
+
+			assert.beautified.html( options.expected, normalizeHtml( actual ) );
+		};
+	}
+
+	function normalizeHtml( html ) {
```
f1ames: Probably bender normalization function would be more useful here? There is no need to reinvent the wheel and if you need some custom replacing you can always do this after HTML is passed through our standard normalization if needed.

tests/core/selectionoptimizer.js
```diff
@@ -0,0 +1,68 @@
...
+		} ),
+		'test selection optimization case 5': testSelection( {
+			initial: '<p>fo[o</p><p>bar]</p><p>baz</p>',
+			expected: '<p>fo[o</p><p>bar]</p><p>baz</p>'
+		} ),
+		'test selection optimization case 6': testSelection( {
+			initial: '<p>foo</p><p>[bar</p><p>]baz</p>',
+			expected: '<p>foo</p><p>[bar]</p><p>baz</p>'
+		} ),
+		'test selection optimization case 7': testSelection( {
```
f1ames: Some more complex use cases with more block elements (e.g. tables, lists, blockquote) and some nested inline elements (styling spans, bold, etc.) should be also added.

tests/core/selection/manual/optimizations.md
```diff
@@ -0,0 +1,31 @@
...
+
+1. Start selection word `Foo` by mouse from the left.
+
+1. Release mouse over the bullet of sub list.
+
+1. Press `insert` button.
+
+## Expected
+
+No red boarder is visible.
```
f1ames:
```suggestion
No red border is visible.
```

core/selectionoptimizer.js
```diff
@@ -0,0 +1,57 @@
...
+		}
+
+		var startsInText = range.startContainer.type === CKEDITOR.NODE_TEXT,
+			limit = startsInText ? range.startContainer.getLength() : range.startContainer.getChildCount();
+
+		if ( range.startOffset === limit ) {
+			return true;
+		}
+
+		return startsInText ^ range.endContainer.type === CKEDITOR.NODE_TEXT;
```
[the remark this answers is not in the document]
engineering-this: I'm not a big fan of using bitwise operators, as they are not commonly used in JS usually when seeing them have to remember myself what they do. But alternative conditional isn't very readable, maybe instead we can just add a comment `// XOR operator.` WDUT?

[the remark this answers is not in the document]
f1ames: It's probably more individual and depends who reads the code. For me alternative conditional is more readable and easier to parse, but I understand it isn't for others.

When it comes to comment, something like `// XOR operator.` is a little too laconic - most people looking on the code knows it's XOR but still will have to check how it works exactly. So in case of a comment you will need more explicit one which describes the condition, e.g.:

```
// Check if range starts or ends in text (not both) using XOR operator.
```

So you end up with two lines, one with actual code and the other one describing it. And on the other end you have alternative conditional which is one line and kind of self explanatory. I don't have very strong opinion here so decide which on you prefer.

engineering-this:
>Now I started to wonder if we really should have it in a separate file or just add it to the selection.js - as separate function wrapping everything inside like:

I know that this file doesn't really fit into our codebase, however `selection.js` has over 2500 lines, my IDE with linters occasionally freezes when editing or browsing such big files. This all leads too poor developer experience.

However when scanning through `selection.js` I see that some bigger portions could be extracted to separate file:
- `CKEDITOR.on( 'instanceCreated', ... )`
- `CKEDITOR.on( 'instanceReady', ... )`
- Adding selection related `prototype` methods of editor, range, and more.
- Adding constans under `CKEDITOR` namespace, eg. `CKEDITOR.SELECTION_NONE = 1;`

So that original file would contain only `CKEDITOR.dom.selection` class and helper functions used by it.

☝️ Such refactoring might make merging some of existing PRs a bit problematic, probably we should extract it to separate ticket, if we even want to touch this.

core/selectionoptimizer.js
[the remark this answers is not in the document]
engineering-this: I tried selecting various ranges and every time it was `selectionCheck` event. Although I'll add a flag for extra security.

engineering-this:
I've found one issue introduced by these changes. When changing selection with <kbd>Shift</kbd>+<kbd>Arrow Key</kbd> optimization reverted change on selection from keystroke resulting in not being able to move selection outside of current element.

I fixed it by adjusting optimization method to consider last keystrokes. This however leads to yet another issue - when you start changing selection with keystroke always only one end of current range changes, however if selection is changed (even if it looks identical) in Firefox, IE and Edge browser doesn't know which end of selection should be moved on next keystroke. This makes selecting things with keyboard a bit problematic, so for this case there won't be optimizations.

f1ames:
Rebased onto latest `major`.

tests/core/selection/manual/optimization.md
```diff
@@ -0,0 +1,43 @@
...
+@bender-ui: collapsed
+@bender-ckeditor-plugins: wysiwygarea, toolbar, elementspath, sourcearea, list, undo, div
+
+1. Double click on word `Bar`.
+
+1. Press `insert` button below the editor.
+
+  ## Expected
+
+  No red boarder is visible.
```
f1ames:
```suggestion
  No red border over inserted text is visible.
```

tests/core/selection/manual/optimization.md
```diff
@@ -0,0 +1,43 @@
...
+
+1. Start selection word `Foo` by mouse from the left.
+
+1. Release mouse over the bullet of sub list.
+
+1. Press `insert` button.
+
+  ## Expected
+
+  No red border is visible.
```
f1ames:
```suggestion
  No red border over inserted text is visible.
```

tests/core/selection/manual/optimization.md
```diff
@@ -0,0 +1,43 @@
...
+
+  ## Unexpected
+
+  Red border is visible.
+
+1. Place collapsed selection inside editor.
+
+1. Use keystroke <kbd>Shift</kbd> + <kbd>Arrow Key</kbd> to select few different elements.
```
f1ames:
```suggestion
1. Use keystroke <kbd>Shift</kbd> + <kbd>Arrow Key</kbd> to select multiline content.
```

tests/core/selection/manual/optimization.md
```diff
@@ -0,0 +1,43 @@
...
+
+## Expected
+
+Anything can be selected.
+
+## Unexpected
+
+Range can't move past end of block.
```
f1ames:
```suggestion
Range can't move past beginning/end of block elements.
```

f1ames:
One more thing, few `tests/core/selection/optimization` tests fails for build editor on CI, please check what's going on there.

core/editor.js
```diff
@@ -168,6 +168,26 @@
 		} );
 		this.on( 'mode', updateCommands );
 
+		// (#3175)
```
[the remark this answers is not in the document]
engineering-this:
```suggestion
		// Store last arrow key keycode so selection optimizations can alter its flow (#3175).
```

core/selection/optimization.js
```diff
@@ -0,0 +1,96 @@
...
+	// Prevent infinite recurrency when browser doesn't allow expected selection.
+	function preventRecurrency( editor, range ) {
+		editor.once( 'selectionCheck', function( evt ) {
+			if ( preventListener ) {
+				return;
+			}
+
+			var newRange = evt.data.getRanges()[ 0 ];
+
+			if ( !CKEDITOR.tools.objectCompare( newRange, range ) ) {
```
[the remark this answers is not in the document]
engineering-this: `objectCompare` doesn't do deep check, so it's safer to use `ranges.equal()`, also comparing ranges is something we do from time to time, so exposing this method might be useful.

```javascript
var editor = ...,
	sel = editor.getSelection();

CKEDITOR.tools.objectCompare( editor.getSelection().getRanges()[0], editor.getSelection().getRanges()[0] ); // false
CKEDITOR.tools.objectCompare( sel.getRanges()[0], sel.getRanges()[0] ); // true
```

In commented case `objectCompare` works, recurrency doesn't appear 😉

tests/core/dom/range/clonecontents.js
```diff
@@ -22,6 +22,11 @@
 			var range,
 				clone;
 
+			// Prevent selection optimizations from breaking tests (#3175).
```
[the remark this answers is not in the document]
engineering-this: These test cases are build around selection which might be not possible after selection optimizations. Optimization would apply to initial selection. These tests are to check if other things than optimizations have effect on selection, so turning off optimizations is required to test for regression caused by other methods.

For instance, commented code relates to a [bug](https://dev.ckeditor.com/ticket/11586) when `range.cloneContents` altered selection. In this case turning off optimizations for TC is the way check for regression without rewriting test. Also note that optimization are applied to editor selection, not just a range, so you can create a range without selecting it, and call `range.cloneContents()`.

IMO turning off optimization in these tests is easiest and safest way to keep these tests. Otherwise we should go back to each TC and analyse it, analyse initial issue, and rewrite it in a way that optimizations doesn't affect it.

>The thing I'm concerned about is that if preventing optimizations here will not mask any issues (unit test will pass while in reality it will fail)
This won't be the same case, if the test passes without optimization, but something is wrong with it, then the optimization itself could be source of the new issue, hence it's different case.

My last argument for keeping these tests in current shape is that in future we might add an option to turn off optimizations.

engineering-this:
>## Improved solution
>We could live with that as it's not that common use case, however I have an idea how to improve it further. What if optimization will not be executed if Shift is pressed? With such approach slowly moving with Arrow Key will work and ignoring for specific browser would not be necessary so code will get simpler.

I haven't thought about this earlier, but that's great idea. Delaying optimizations until <kbd>Shift</kbd> allows simplifying code, and it seems to work better! Also plenty tests can be removed.

core/editor.js
```diff
@@ -168,8 +168,23 @@
...
 		// Handle startup focus.
 		this.on( 'instanceReady', function() {
+			this.editable().on( 'keydown', function( evt ) {
+				this._.shiftPressed = evt.data.$.shiftKey;
+			}, this );
+
+			this.editable().on( 'keyup', function( evt ) {
+				this._.shiftPressed = evt.data.$.shiftKey;
+			}, this );
+
```
f1ames: I would be for moving all the changes from this file to `core/selection/optimization.js `, and here just add something like:

```js
// (#3175)
this.on( 'instanceReady', function( evt ) {
    CKEDITOR.dom.selection.setupOptimizations( evt.editor );
} );
```

and `setupOptimizations();` would attach all listeners. WDYT?

core/selection/optimization.js
```diff
@@ -0,0 +1,69 @@
+( function() {
+	var preventListener = true;
+
+	CKEDITOR.dom.selection.prototype.optimizeInElementEnds = function() {
```
f1ames: Since this is a public method it needs documentation.

core/editor.js
```diff
@@ -168,8 +168,23 @@
...
+		// (#3175)
+		this.on( 'selectionCheck', function( evt ) {
+			if ( evt.data ) {
+				evt.data.optimizeInElementEnds();
+			}
+		} );
+
 		// Handle startup focus.
 		this.on( 'instanceReady', function() {
+			this.editable().on( 'keydown', function( evt ) {
```
f1ames: My concluding thought about not optimizing when <kbd>Shift</kbd> is pressed was that it should be combined with <kbd>Arrow key</kbd> presses so we should prevent optimization only when <kbd>Shift</kbd> + <kbd>Arrow key</kbd> was pressed - I see it never appeared in my previous https://github.com/ckeditor/ckeditor-dev/pull/3213#pullrequestreview-276645022 comment 🤔 

So this way all other combinations with <kbd>Shift</kbd> will trigger optimization, but it will still solve the issue with <kbd>Arrow key</kbd>s, WDYT?

It's not a big change, since instead of `keydown` you will need `key` event checking `keyCode` - the same way it was done in https://github.com/ckeditor/ckeditor-dev/pull/3213/commits/b8901c2eaf138e87e4a7f7d6db500862731dff46#diff-162a1b36d2efd62081522822a026e853R173. And resetting `shiftPressed` flag on `keyup` will still work.

engineering-this: After discussion we leave it as it is.

engineering-this:
## Tests in Safari

Two ranges are created
`range`
```html
<p>foo[<span contenteditable="false" id="el1">bar</span>]</p><div contenteditable="false" id="el2">bom</div>
```

`range2`
```html
<p>foo<span contenteditable="false" id="el1">bar</span></p>[<div contenteditable="false" id="el2">bom</div>]
```

When `selection.selectRanges( [ range, range2 ] );` is called editor is selected with range which is different from ranges above:
```html
<p>foo{<span contenteditable="false" id="el1">bar</span>]</p><div contenteditable="false" id="el2">bom</div>
```

So in Safari
```
range.equals( editor.getSelection().getRanges()[ 0 ]; // false
```
While Chrome returns `true`

This selection triggers optimization (one end anchored in text, another in element), and call on `range.shrink( CKEDITOR.SHRINK_TEXT );`.

Originally tested range is modified by the browser so to me it makes sense to ignore this assertion. Alternatively, we can turn off optimizations, but I'm more for ignoring.

## Tests in IE8

I fixed two tests, however `list` test should be ignored.
In IE8 elements with `dir="rtl"` works as if they were separate contenteditable, so selection can't start in one of them and end in another - it's broken.

1. Go to https://cdn.ckeditor.com/4.12.1/full-all/samples/
2. Paste into editor: 
```html
<p dir="rtl">foo</p>
<p dir="rtl">bar</p>
<p dir="rtl">baz</p>
```html
3. Play with selection.

It looks completely broken to me, so I'm for ignoring failing TCs here.

## Firefox cases

It looks like adding `skipBogus = true` to call on [range.shrink](https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_dom_range.html#method-shrink) solved the issue.

engineering-this:
I've reworked `preventRecurrection` so that it also considers range from before any optimizations. This solves failing tests.

f1ames:
Rebased onto latest `major`.

PR #3230, opened 2019-07-01
Allow passing config.ignoreStyle to widgetTools.addTests as function type

### LINKED ISSUE #3173: Improve widgetTools.addTests to evaulate ignoreStyle option during runtime
## Type of report

Task

This configuration option can depend on some plugin or custom editor configuration which is not available during test file execution. The real use case which should be also corrected after changes are described here: https://github.com/ckeditor/ckeditor-dev/pull/2693#pullrequestreview-249965579 (follow discussion for more details).

As an example, you should be able to use `widgetTools.addTests` like:

```javascript
	tools.addTests( tcs, {
		name: 'default',
		widgetName: 'mathjax',
		extraPlugins: 'mathjax',
                // more configuration goes here...
		ignoreStyle: function( editor ) {
                  return !editor.plugins.mathjax.isSupportedEnvironment();
                }
	} );
```
So the function is executed during test runtime with existing editor instance / already loaded CKEditor plugins.

Again, don't forget to update https://github.com/ckeditor/ckeditor-dev/pull/2693#pullrequestreview-249965579 when working on this issue.

engineering-this:

Task

### This PR contains

- [ ] Unit tests
- [ ] Manual tests

none

## What changes did you make?

`config.ignoreStyle` passed to `widgetTools.addTests` can now be function type. Also updated uses in `mathjax-mock` test.

Closes #3173

### DIFF
--- a/tests/plugins/widget/_helpers/tools.js
+++ b/tests/plugins/widget/_helpers/tools.js
@@ -18,7 +18,7 @@ var widgetTestsTools = ( function() {
 	//
 	// @param config.newData
 	// @param config.newWidgetPattern
-	// @param {Boolean/function} [config.ignoreStyle=false]
+	// @param {Boolean/Function} [config.ignoreStyle=false]
 	function addTests( tcs, config ) {
 		var editor,
 			editorBot,
--- a/tests/plugins/widget/_helpers/tools.js
+++ b/tests/plugins/widget/_helpers/tools.js
@@ -33,10 +33,7 @@ var widgetTestsTools = ( function() {
 					loaded: function( evt ) {
 						editor = evt.editor;
 
-						var ignoreStyle = config.ignoreStyle;
-						ignoreStyle = typeof ignoreStyle === 'function' ? ignoreStyle( editor ) : ignoreStyle;
-
-						initialData = fixHtml( editor.getData(), ignoreStyle );
+						initialData = fixHtml( editor.getData(), config.ignoreStyle, editor );
 
 						editor.dataProcessor.writer.sortAttributes = true;
 					},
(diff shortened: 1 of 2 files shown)


tests/plugins/widget/_helpers/tools.js
```diff
@@ -32,7 +33,10 @@ var widgetTestsTools = ( function() {
 					loaded: function( evt ) {
 						editor = evt.editor;
 
-						initialData = fixHtml( editor.getData(), config.ignoreStyle );
+						var ignoreStyle = config.ignoreStyle;
+						ignoreStyle = typeof ignoreStyle === 'function' ? ignoreStyle( this ) : ignoreStyle;
+
+						initialData = fixHtml( editor.getData(), ignoreStyle );
```
jacekbogdanski: You implemented this feature very high in function aggregation. `ignoreStyle` option for `addTests` function is not used very extensively, AFAIK only for `mathjax-mock` tests. However, `fixHtml` function is used in many places where this improvement may be very useful. Why not move the whole logic into `fixHtml` function itself?

Of course, it makes sense to still pass `editor` instance when possible (for `addTests` etc) to avoid some creepy workarounds when writing tests. Unfortunately, our `CKEDITOR.tools.bind` the function doesn't work correctly, as it doesn't allow to bind function arguments and would make implementation much cleaner

 We have already nice PR for this feature https://github.com/ckeditor/ckeditor-dev/pull/2862 and maybe it's a good time to finish it? I will bump this PR a bit.

jacekbogdanski:
Changed upstream into `major` and rebased.

f1ames:
Fixed one commit message.

f1ames:

PR #3241, opened 2019-07-04
Fix for: Table selection moved into the topmost element when clicking balloon button

### LINKED ISSUE #3136: Table selection moved into the topmost element when clicking balloon button
## Type of report

Bug

Reproduction steps are valid for native Firefox visual selection.

1. Open https://codepen.io/jacekbogdanski/pen/MdXQGd
2. Open console.
3. Select some table cells.
4. Click balloon button (empty one).

### Expected result

Range logged by console reflects the visual selection.

### Actual result

The range is moved into the topmost element inside the editable element. You will also see cleaned FF visual selection.

It looks like natively range is not lost during operations on clickable elements outside contenteditable: https://codepen.io/jacekbogdanski/pen/mYKxGv

The issue is also reproducible with an inline editor (`divarea`), although the selection is kept inside the first cell of the table.

## Other details

* Browser: Firefox
* CKEditor version: at least 4.10.0

engineering-this:

Bug fix

### This PR contains

- [X] Unit tests
- [X] Manual tests

```
* [#3136](https://github.com/ckeditor/ckeditor-dev/issues/3136) [Firefox] Fixed: Clicking on [Balloon Toolbar](https://ckeditor.com/cke4/addon/balloontoolbar) items removes selection.
```

## What changes did you make?

When clicking on button command execution causes `editor.focus()`, however the click itself blurs editor. In case of inline editors we restore last selection on `focus`. However in case of native table selection it was wrongly restored, so I've adjusted conditional to reach proper path of restoring sel.

In case of `iframe` editors selection doesn't need to be restored, however due to upstream issue in Firefox native table selection is lost when `editor.focus()` is triggered by click. As a workaround logic for restoring selection in this case logic for locking/unlocking selection from inline editors is used.

The second case is not covered by unit test because I couldn't reproduce it without a real mouse click.

Closes #3136


Comandeer:
Rebased onto latest `major`.

engineering-this:
Fixed false positive test, and added another editor type, so both classic and inline is tested.

https://github.com/ckeditor/ckeditor-dev/pull/3241/commits/b8d63ecd013fbba2b0d5330269308c2f0d8753db#diff-6c88377ea0037896524f85085616f665R73
I wonder how I didn't spot this, as I'm always checking if my tests fail without fix 🤔 All I can think of is a missclick with IDEs autocomplete.

Issue #972, opened 2017-09-28
Copy&Paste on a table cell, creates new table row


roxgirdu:
## Are you reporting a feature request or a bug?
Bug:
When a table cell is copied and pasted into another table cell, a new table row is created and the copied cell is pasted on the next row. 

## Provide detailed reproduction steps (if any)

1. Add a table
2. Copy a table cell.
3. Paste it into another cell.

### Expected result

Another table row is created and the cell is pasted on the next line.

### Actual result

The copied cell should be pasted within the cell where it wanted to be copied.

## Other details

* Browser: Firefox, Chrome, IE 11
* OS: 10.12.6
* CKEditor version: 4.7.3
Demo: https://cl.ly/110w2V3l2G3b

msamsel:
I was able to reproduce this situation.

PR #3244, opened 2019-07-05
Fix for: Element/Image drag to move within table is no longer available

### LINKED ISSUE #547: Element/Image drag to move within table is no longer available
BUG

Element/Image drag to move within table is no longer available

Steps to reproduce
1. Insert multi-cell table
2. Insert image into a cell
3. Drag image to another cell. (This does not work). 
4. Add text to cell.
5. Select text and move to another cell. (This does not work).

I expected the image or text to move to another cell.

Actual result is the cell itself is selected 

I was able to reproduce with the demos on CKEditor.com

* Browser: Chrome,IE
* OS: Windows

Dumluregn:

Bug fix

### This PR contains

- [ ] Unit tests are not reliable as mentioned [here](https://github.com/ckeditor/ckeditor-dev/pull/3244#pullrequestreview-258421144)
- [x] Manual tests

`[#547](https://github.com/ckeditor/ckeditor-dev/issues/547): Fixed: Element/Image drag to move within table is no longer available when using [Table Selection plugin](https://ckeditor.com/cke4/addon/tableselection).`

## What changes did you make?

Added drag&drop handling so that table selection no longer needs to cancel those events.

Closes #547.

### DIFF
--- a/plugins/tableselection/plugin.js
+++ b/plugins/tableselection/plugin.js
@@ -379,22 +379,18 @@
 	}
 
 	function fakeSelectionDragHandler( evt ) {
-		var table = evt.data.getTarget().getAscendant( 'table', true );
+		var table = evt.data.getTarget().getAscendant( 'table', true ),
+			editor = evt.editor || evt.sender.editor;
 
-		// (#2945)
-		if ( table && table.hasAttribute( ignoredTableAttribute ) ) {
+		// (#547)
+		if ( table && evt.name == 'dragstart' ) {
+			table.data( ignoredTableAttribute, '' );
 			return;
+		} else if ( table && evt.name == 'drop' ) {
+			table.data( ignoredTableAttribute, false );
 		}
-
-		var cell = evt.data.getTarget().getAscendant( { td: 1, th: 1 }, true );
-
-		if ( !cell || cell.hasClass( fakeSelectedClass ) ) {
-			return;
-		}
-
-		// We're not supporting dragging in our table selection for the time being.
-		evt.cancel();
-		evt.data.preventDefault();
+		editor.getSelection().reset();
+		clearFakeCellSelection( editor, true );
 	}
 
 	function copyTable( editor, isCut ) {
@@ -1192,6 +1188,7 @@
 				editable.attachListener( mouseHost, 'mouseup', fakeSelectionMouseHandler, null, evtInfo );
 
 				editable.attachListener( editable, 'dragstart', fakeSelectionDragHandler );
+				editable.attachListener( editable, 'drop', fakeSelectionDragHandler );
 				editable.attachListener( editor, 'selectionCheck', fakeSelectionChangeHandler );
 
 				CKEDITOR.plugins.tableselection.keyboardIntegration( editor );
--- a/plugins/tableselection/plugin.js
+++ b/plugins/tableselection/plugin.js
@@ -10,6 +10,7 @@
 		fakeSelectedEditorClass = fakeSelectedClass + '-editor',
 		fakeSelectedTableDataAttribute = 'cke-table-faked-selection-table',
 		ignoredTableAttribute = 'data-cke-tableselection-ignored',
+		draggingActiveAttribute = 'cke-table-dragging-active',
 		fakeSelection = { active: false },
 		tabletools,
 		getSelectedCells,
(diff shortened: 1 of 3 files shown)


config.js
```diff
@@ -5,7 +5,7 @@
 
 CKEDITOR.editorConfig = function( config ) {
 	// Define changes to default configuration here. For example:
-	// config.language = 'fr';
+	config.height = 500;
```
f1ames: The best way to avoid editing samples it's too add manual test right away - a little more work, but then it can be reused for the final PR version.

plugins/clipboard/plugin.js
```diff
@@ -2345,7 +2345,18 @@
...
+				table = ranges.length && ranges[ 0 ]._getTableElement( { table: 1 } ),
+				element = editor.getSelection().getSelectedElement();
+
+			// When element inside cell is dragged with tableselection plugin on, it is sometimes
+			// nested in additional table. We want to filter only the content of a cell.
+			if ( editor.plugins.tableselection && table && element && element.$.draggable ) {
+				this.setData( 'text/html', element.$.outerHTML );
+			} else {
+				this.setData( 'text/html', editor.getSelectedHtml( 1 ) );
+			}
```
f1ames: This is different issue, it also happens on copy/paste. Please see if it is already reported - if not you can report an issue (and maybe link this piece of code as a possible solution). Then we can proceed separately with it.

plugins/tableselection/plugin.js
```diff
@@ -379,22 +379,18 @@
...
 	function fakeSelectionDragHandler( evt ) {
-		var table = evt.data.getTarget().getAscendant( 'table', true );
+		var table = evt.data.getTarget().getAscendant( 'table', true ),
+			editor = evt.editor || evt.sender.editor;
 
-		// (#2945)
-		if ( table && table.hasAttribute( ignoredTableAttribute ) ) {
+		// (#547)
+		if ( table && evt.name == 'dragstart' ) {
+			table.data( ignoredTableAttribute, '' );
```
f1ames: The `element.data()` method adds `data-` prefix so you will end up with `data-data-cke-tableselection-ignored`. This means this code doesn't add proper ignore attribute and still works which means the attribute may be not needed 🤔 

Also the condition for `#2945` was removed so you need special attention here to se if nothing gets broken. Btw. I think it will still work with this condition present.

f1ames: I would consider the other way around approach too (not saying it will be better but worth checking) - instead of just always allowing dragging (this is what happens after this change basically), maybe don't prevent `dragStart` event in more case. The same it is done for `table && table.hasAttribute( ignoredTableAttribute )` or `!cell || cell.hasClass( fakeSelectedClass )` - it should be also possible to be done for image or text elements.

plugins/clipboard/plugin.js
Dumluregn: Reported as #3278. Note that this works fine in FF. It seems to be similar to #972, but that issue is a problem for FF too, so I'm not certain that the cause is the same.

Dumluregn: PR for the fix - #3290, should be merged before this one. I will remove redundant code during rebase onto latest `major`.

Dumluregn: After some digging I think #3278 was just some particular case of #972, so I closed it as duplicate. Still, #3290 should be merged before this as it fixes mentioned clipboard issues (also the one mentioned [in another comment](https://github.com/ckeditor/ckeditor4/pull/3244#pullrequestreview-258421144)).

plugins/tableselection/plugin.js
Dumluregn: Firstly, you're obviously right - the attribute was misused and useless 👎 But later I found out that trying to reuse it in this case isn't trival, as after drop we want to remove the temporal attribute, and I didn't see a way to distinguish whether it was added by user with #2945 in mind, or just added for dragging, so I decided to restore removed conditions and mimic that solution but with a new, dedicated attribute (I came up with `cke-table-dragging-active`), except for the part where I control turning it on and off.
As for the second comment, during tests I didn't encounter and can't really think of any edge-case scenario I would like to cut out, so if you have any of them on mind or find it during review please let me know. For now I would like to give a go to this solution.

Dumluregn:
A little summary:
I rebased PR onto latest `major`. It's ready for review, but will not work properly until #3290 is merged - it fixes the bugs mentioned in this PR like [this](https://github.com/ckeditor/ckeditor4/pull/3244#pullrequestreview-258421144) and [this](https://github.com/ckeditor/ckeditor4/pull/3244#discussion_r300696718). Without it dragged elements will be pasted nested in additional table.

PR #3263, opened 2019-07-12
Pass the same document for detached widget wrapper element as for editor instance

### LINKED ISSUE #3261: Widget created using dialog has incorrect owner document
## Type of report

Bug

When a widget is created using dialog (default widget implementation) `widget.element` has incorrect owner document pointing into the browser root document instead of the editor's one. It causes issues when attaching event listeners directly to the widget document.

1. Open https://codepen.io/jacekbogdanski/pen/MMLezW codepen.
2. Open console.
3. Verify console status.
4. Open placeholder dialog (inserting, not editing).

### Expected result

For both, upcasted widget and created widget when opening placeholder dialog, `editor.document` and `widget.element.getDocument` should be the same.

### Actual result

Upcasted widget during editor initialization has a correct document (the same as editor), although using dialog points into browser root document.

## Other details

* CKEditor version: at least 4.10.0

jacekbogdanski:

Bug fix

### This PR contains

- [x] Unit tests
- [ ] Manual tests

```
* [#3261](https://github.com/ckeditor/ckeditor-dev/issues/3261): Fixed: [Widget](https://ckeditor.com/cke4/addon/widget) initialized using dialog has incorrect owner document.
```

## What changes did you make?

Passing the same document for detached wrapper element as for editor instance.

Closes #3261


jacekbogdanski:
The manual test would be exactly the same as a provided unit test. It doesn't seem to give testing value as it's not something with UI or depending on user interaction. Do you think it's worth the effort writing and clicking during release if the manual test would only differ from the unit by opening dev tools console, instead of bender assertion check? :thinking:

f1ames:
Rebased onto latest `major`.

> The manual test would be exactly the same as a provided unit test.

That's a good point, if the code will be the same that doesn't make sense to duplicate it 👍

Issue #3308, opened 2019-08-01
Drop caching response logic in embedbase plugin in favor of promises


engineering-this:
## Type of report

Task

## Provide description of the task

We have set of methods for sending requests, caching them, and getting cached responses.

https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_plugins_embedBase__jsonp.html#method-sendRequest
https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_plugins_embedBase_baseDefinition.html#method-_getCachedResponse
https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_plugins_embedBase_baseDefinition.html#method-_getCachedResponse

They are all private, so it should be okay for breaking change here. 

Currently if for whatever reason same media is inserted second time before first response is returned, then separate response is send. I propose that sendRequest returns a promise, so that one promise can be used multiple times.

With current implementation only use case is that someone pastes same link with `autoembed` two times quickly - that's not a real case, because there is no reason for user to do that. However after #3264 or at least #2306 (pending PR: #3270) is fixed, then for compatibility reasons we will have to send request for widgets that are created from initial editor data (e.g. fetched from backend). This creates a more realistic use case when same media is inserted two times in larger document. In this case when editor is created with such data, then two requests are send. Although it could be fixed within mentioned tickets, it would make PRs much bigger and complex, so it makes sense to extract it here.

With this change we could easily fix situation when user enters source mode after inserting link with autoembed but before response is returned which break upcasting widget.

## Other details

* Browser: …
* OS: …
* CKEditor version: …
* Installed CKEditor plugins: …

PR #3311, opened 2019-08-02
Fix for: Inconsistent media embeds in editor content and its output

### LINKED ISSUE #2306: check if media embed should try to handle all URLs or only selected that it supports
## Type of report

Bug

1. copy url from the browser,
2. paste into the editor

### Expected result

text as a link

### Actual result

untitled421

## Other details

* Browser: Chrome, Firefox
* OS: OSX
* CKEditor version: demo on the ckeditor website
* Installed CKEditor plugins: ?

### LINKED ISSUE #3264: Inconsistent media embeds in editor content and its output
## Type of report

Bug

Extracted from https://github.com/ckeditor/ckeditor-dev/issues/2306#issuecomment-510122599.

The main provider we are using (`IFramely`), but also regular embeds use scripts to enrich the embedded content. Since scripts are not executed inside CKEditor, the embedded content looks quite different in CKEditor itself and its output (where scripts are usually executed):

| In editor | Output |
|-----------|--------|
| image | image |
| image | image |
| image | image |

One reason for that is not executing scripts as mentioned above and the other is that CKEditor outputs original iframely response instead of what we have displayed in CKEditor. So there are two ways to make embeds consistent:

* Output should use HTML which is displayed inside CKEditor (and no scripts).
* CKEditor should be able to show rich embeds.

## Output should use HTML which is displayed inside CKEditor

That will make content consistent and would be easy to implement, however then we lost most of the rich media content (which is now visible at last on the frontend layer) so that would be a step backward IMHO.

## CKEditor should be able to show rich embeds

Showing rich embeds in CKEditor itself is another alternative. It would make embeds consistent and always looking up to date (if embedding script changes it will be automatically handled).

The idea is to create some kind of a sandbox (using `iframe` or shadow DOM) where entire response HTML can be attached, then script will load and enrich the content and then rich content will be moved/shown inside CKEditor. @engineering-this did some research on this:

> both with `iframe` and shadow DOM sandboxing (both inside and outside of the editor). However, there are few problems we are unable to overcome (problem with detecting final dimension and moment in time when embed script has finished modifying the content).

and this can be rather tricky to implement, however I think it could be doable - using a combination of a different event listeners and some simple heuristic maybe 🤔 

---

Related issues #2264, #3132.

engineering-this:

Bug fix/ new feature

### This PR contains

- [ ] Unit tests
- [ ] Manual tests

```
* [#3264](https://github.com/ckeditor/ckeditor-dev/issues/3264): Fixed: [Media Embed](https://ckeditor.com/cke4/addon/embed) widgets in editor are inconsistent with its output.
```

## What changes did you make?

This is an early PR, PoC. So it still needs lots of polishing, and tests. It is based on branch from #3270, however I think all the changes can be emerged within current PR. I suggest that reviewer should check what's new in #3270 and take a quick look at new code and test, to know what's going on there.

This PR brings new way of dealing with media embed widgets.
Instead of just inserting raw html from response, an iframe is created, to which is inserted html code. The script provided with response is now created as an separate element and inserted, so it executes - when script is element is created from html string instead of `document.createElement` then it won't be executed.

Iframe has one issue in all browsers - when it's detached from DOM it's documentElement `<html>` is removed, so when you attach same iframe again it's contents are lost. This leads to empty widget when recreating iframe. An alternative proposed by @iparamonau from iframely was to use `srcdoc`. That solution is doesn't have issue mentioned above, but have the `mathjax` effect. Each time widget is recreated (undo/redo, cut/paste, sourcemode) it is created from initial html response, and it takes a second for widget to look the same again.

I fixed it by not using srcdoc, but instead elements are manually appended to iframe. iframes `html` element reference is stored, and it's content are appended to iframe each time widget is recreated. It works well unless widget is downcasted before it is fully initialized - this is fixed by manually recreating provided script element inside iframe.

When there are multiple widgets with same url, that we can't share same html elements, so all references are cached. When new widget is inserted with same url, then first detached element will be used, widget is rendered from `response.html` only if no detached elements are available. So for example if user creates lengthy article and puts same image from flickr two times, then each time he goes to source or does undo/redo, both widgets will appear instantly.

When widget is downcasted iframe is replaced with `response.html`. That`s for backward compatibility.

New solution doesn't really work in Edge/IEs, explained here: https://github.com/ckeditor/ckeditor-dev/issues/3264#issuecomment-517317703 and https://github.com/ckeditor/ckeditor-dev/issues/3264#issuecomment-517649610
So instead logic #3270 is used as a fallback.

## Todos

- First of all it should be manually tested if everything really works, if this solution is good enough and no there are no big issues that could be blockers.
- Check if any actions with multiple widgets doesn't produce any unexpected behaviour
- There are some tests, but more cases should be covered.
- Refactoring, which function we want to expose as part of API, which should be just local helpers, API docs, ect. There are multiple plugins `embed` `embedbase` `embedsemantic` `autoembed`, so probably we should rethink which logic should stay in which plugin.
- Consider if #3308 should be included here or not.

Closes #3264
Closes #2306


engineering-this:
Removed IE/Edge fallback, which means removing code inherited from #3270 see https://github.com/ckeditor/ckeditor-dev/issues/3264#issuecomment-522524474
Working on tests.

tests/plugins/embedbase/definition.js
```diff
@@ -300,7 +300,7 @@ bender.test( {
 			that.listeners.push( widget.on( 'handleResponse', function( evt ) {
 				eventsCount += 1;
 
-				evt.data.html = evt.data.response.html + '<p>customized</p>';
```
engineering-this: This use case is wrong. When widget removed and new one is created from same url, then cached response is used. so this event won't be fired second time. If `evt.data.html` is modified then the second widget will be different from the first one. Only way to make it work as suggested in comment below is to modify the response itself, so `_getCachedResponse` will return modified html.

PR #3360, opened 2019-08-27
Add AMD support for CKEDITOR.tools.promise

### LINKED ISSUE #3287: ReferenceError: Can't find variable: ES6Promise
## Type of report

Bug

I'm running integration tests using **PhantomJS** (which doesn't implement ``Promise``) and I'm using **RequireJS** (AMD) to load other modules besides CKEditor.

### Expected result

The tests were passing before upgrading to from CKEditor 4.11.4 to CKEditor 4.12.1.

### Actual result

The tests are failing now with:

    ReferenceError: Can't find variable: ES6Promise

## Other details

* Browser: PhantomJS
* OS: Ubuntu 19.4
* CKEditor version: 4.12.1
* Installed CKEditor plugins: standard set

Upon checking the code I believe the culprit is this https://github.com/ckeditor/ckeditor-dev/commit/438a019cf196255a97cb5bde76bd857e768d391a#r34380433 .

The workaround for me was to run this before my tests:

    window.Promise = window.Promise || function() {};

It works because CKEditor doesn't use the ``Promise`` that much at the moment (and so it doesn't affect my tests), but this may change in the future.

Comandeer:

Bug fix

### This PR contains

- [x] Unit tests
- [x] Manual tests

```
* [#3287](https://github.com/ckeditor/ckeditor-dev/issues/3287) Fixed: [`CKEDITOR.tools.promise`](https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_tools_promise.html) incorrectly initialized if an AMD loader is present.
```

## What changes did you make?

I've added a small workaround that loads `Promise` polyfill using AMD loader.

As tests uses RequireJS, I've added appropriate info to our license.

Closes #3287.


core/promise.js
```diff
@@ -11,16 +11,25 @@
 	if ( window.Promise ) {
 		CKEDITOR.tools.promise = Promise;
 	} else {
-		var polyfillURL = CKEDITOR.getUrl( 'vendor/promise.js' );
+		var polyfillURL = CKEDITOR.getUrl( 'vendor/promise.js' ),
+			isAmdEnv = typeof window.define === 'function' && window.define.amd && typeof window.require === 'function';
+
+		if ( isAmdEnv ) {
+			return window.require( [ polyfillURL ], function( Promise ) {
```
f1ames: From what I can see, `window.require` loads a given script - this means `CKEDITOR.scriptLoader.load` below is not needed for AMD env (and should be probably placed in `else`)?

I wonder about build version, how it will behave there, since the file is bundled (AFAIR) 🤔Could you check it? Will AMD try to load this file separately?

tests/core/tools/manual/amd.md
```diff
@@ -0,0 +1,20 @@
...
+	Everything works correctly.
+
+	## Unexpected
+
+	Some options are disabled or does not work correctly.
+2. Open console.
+
+	## Expected
+
+	There are no errors;
```
f1ames:
```suggestion
	There are no errors.
```

core/promise.js
Comandeer: > and should be probably placed in `else`

`else` is not necessary as execution is halted on `require` due to `return` statement.

In case of built version, promise file seems to be left unbundled (however it's nicely minified) and it's loaded separately.

f1ames: Ok, somehow missed the `return` statement 🤦‍♂

f1ames:
Rebased onto latest `major`.

PR #3365, opened 2019-08-29
Fix for: range._getTableElement which sometimes not return a table element

### LINKED ISSUE #3101: range._getTableElement sometimes returns invalid result
## Type of report

Bug

1. Open https://codepen.io/jacekbogdanski/pen/BeVmxX
1. Open console.

## Expected

Table element for both console logs.

## Actual

Null for both console logs.

---

The issue has been already covered for Safari, although it also occurs on Chrome. It probably makes sense to remove browser sniffing or at least enable the fix for Chrome. The issue is still reproducible on Safari when start container points outside the table.

https://github.com/ckeditor/ckeditor-dev/blob/36ad5d9fddff8173290854e36f25cfcea2b83cfb/core/dom/range.js#L2760-L2762

Also, the condition should be probably reversed for the `endContainer` i.e.
```javascript
if ( endTable && start.equals( this.root ) ) {
	return end.getAscendant( tableElements, true );
}
```

The issue can be reproduced by making partial table selection without `tableselection` plugin. Mostly, you want to create selection where start or end container points into element outside the table - so selection should be placed on the edge of the table.

## Other details

* Browser: Chrome, Safari
* CKEditor version: 4.7.0

msamsel:

Bugfix

### This PR contains

- [x] Unit tests
- [x] Manual tests <- might be not required and removed from this PR if you consider it 
redundant

```
* [#3101](https://github.com/ckeditor/ckeditor-dev/issues/3101): Fixed: [`range._getTableElement()`](https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_dom_range.html#method-_getTableElement) return `null` instead of a table element for edge cases. 
```

## What changes did you make?

Replace `range.startContainer` and `range.endContainer` with `range.getTouchedStartNode()` and `range.getTouchedEndNode()`.
That results with accepting the situation when range starts just before the table for example
```
<p>foo</p>[<table><tr><td>bar</td>]<td>baz</td></tr></table>
```
or ends just after the table
```
<p>foo</p><table><tr><td>bar</td>[<td>baz</td></tr></table>]
```
Previously in such cases start or end containers returned editable (or outer element) and prevent table detection.

Closes #3101.

### DIFF
--- a/tests/core/dom/range/gettableelement.js
+++ b/tests/core/dom/range/gettableelement.js
@@ -42,11 +42,11 @@
 		return range;
 	}
 
-	function assertElements( expected, range, selectors ) {
+	function assertElements( expected, range, selectors, message ) {
 		if ( !expected ) {
 			assert.isNull( range._getTableElement() );
 		} else {
-			assert.isTrue( expected.equals( range._getTableElement( selectors ) ), 'Correct element is returned' );
+			assert.isTrue( expected.equals( range._getTableElement( selectors ) ), message || 'Correct element is returned' );
 		}
 	}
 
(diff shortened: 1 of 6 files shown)


tests/core/dom/range/gettableelement.js
```diff
@@ -42,11 +42,11 @@
 		return range;
 	}
 
-	function assertElements( expected, range, selectors ) {
+	function assertElements( expected, range, selectors, message ) {
```
Comandeer: TBH when I see function with more than three arguments, I always think that it should be replaced simply by `options`:

```javascript
function assertElement( options )
```

msamsel:
I search for some cases where this might be an issue, however, I couldn't find anything. I see that the dialog in the link plugin, the balloonpanel and the tableselection use `isInTable` method. I wasn't able to figure out any case where it might break. If you have some case on your mind which might break please share it.

About breaking change, I agree it's a breaking change, but this is a requirement of this issue. To treat as table, selection which start just before table. As it is defined in the provided sample:
```
  range.setStartBefore( table );
  range.setEndAfter( table.findOne( 'td' ) );
```

tests/core/dom/range/gettableelement.js
msamsel: I've changed it a little bit different. As `expected` and `range` are required parameters and has to be provided in every assertion I remain it without change. Those are not options form my perspective, and that's why should be hidden inside option object. Other parameters are moved to the options. 
```
function assertElement( expected, range, options )
```

I also add little docs describing parameters.

msamsel:
I've made a better description for the manual test, with instruction on how to create such selection. Right now test should have more sense. :)

PR #3371, opened 2019-09-02
Enable mask for widget parts

### LINKED ISSUE #3240: It should be possible to use mask only on a widget part
## Type of report

Feature request

The [widget `mask` property](https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_plugins_widget.html#property-mask) works in a way that it covers entire widget. There are cases where it could be useful to define only a [part](https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_plugins_widget.html#property-parts) of a widget which will be masked.

This is real use case encountered during [Drupal `embed-entity` plugin](https://www.drupal.org/project/entity_embed) development, where there is an captionable media embed. The media itself should have interaction blocked while the caption should be editable in place (and the mask covers entire widget blocking caption editing).

The proposal is to make `mask` property accept string too which will be the name of the widget part which should be masked.

Dumluregn:

New feature

### This PR contains

- [x] Unit tests
- [x] Manual tests

API section:
```* [#3240](https://github.com/ckeditor/ckeditor-dev/issues/3240): Extended [`CKEDITOR.plugins.widget#mask`](https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_plugins_widget.html#property-mask) property to accept string with widget part name.```

## What changes did you make?

`setupMask()` function is now checking whether `widget.mask` property is just `true` (and creates full mask for widget as it always did) or not - then it checks if there is a widget part with the given name and if so, creates a partial mask positioned to cover only requested widget part.

Note that the old, complete mask has it's width and height set just as 100%, so any changes to widget size are applied automatically - which is not the case with partial mask, having it's dimensions calculated exactly when created. For that reason partial mask is recreated each time the widget is changed anyhow - to track this a listener for `saveSnapshot` event was added.

Closes #3240.

### CODE REVIEW

File: plugins/widget/plugin.js [hunk 1]
```diff
@@ -3530,22 +3534,73 @@
 	}
 
 	function setupMask( widget ) {
-		if ( !widget.mask )
+		if ( !widget.mask || !widget.wrapper ) {
```
f1ames: The `setupMask()` function for complete mask (`widget.mask == true`) should be called only once on widget initialization as before. Now, since you merged the logic for handling all mask types into this one method, `setupMask()` for complete mask is called also on editor changes which doesn't make sense. So there should be 2 functions - `setupMask()`/`setupFull/CompleteMask()` which is only called once, on widget init for complete mask and `setupPartialMask()` which is called on widget init and then on widget resize.

f1ames:
@Dumluregn as for failing IE8 test, it looks like https://github.com/ckeditor/ckeditor-dev/issues/3381. Rebase onto latest `major` should be enough to solve this.

Dumluregn:
Failing tests - #3407.

Dumluregn:
Rebased onto latest `major`.

plugins/widget/plugin.js
```diff
@@ -3310,70 +3327,268 @@
...
+				} );
 			}
+			maskBuffer.input();
 
-			var label = this.getLabel ? this.getLabel() : getLabelDefault.call( this );
+			// FF renders image-like widget very late (much after 'instanceReady') so mask creation
+			// has to be postponed. The same is true for switching between WYSIWYG and source mode,
+			// just different event and smaller delay are needed.
+			if ( CKEDITOR.env.gecko ) {
+				widget.editor.on( 'instanceReady', function() {
```
f1ames: I know we have already talked about it, but timeouts are quite a last resort solutions (and could be quite unreliable). We also talked about `img.complete` property, that it doesn't do the job, but have you tried to combine it with `img.load` event like below:

```js
var imgs = widget.element.find( 'img' );
CKEDITOR.tools.array.forEach( imgs.toArray(), function( img ) {
	if ( !img.$.complete ) {
		img.on( 'load', function() {
			maskBuffer.input();
		} );
	}
} );
```

It seems to be working covering this behaviour in FF (I have also tested it with services like https://picsum.photos which loads different photos each time so it's not cached and with throttling in dev console and it works fine). Could you check this one? Of if you already checked it, described a case in which it failed?

tests/plugins/widget/manual/partialmask.html
```diff
@@ -0,0 +1,1582 @@
...
+	var template = '<img alt="" src="" />',
+		templateBlock = new CKEDITOR.template(
+			'<figure class="{captionedClass}">' +
+				template +
+				'<figcaption>{captionPlaceholder}</figcaption>' +
+			'</figure>' ),
+		alignmentsObj = { left: 0, center: 1, right: 2 },
+		regexPercent = /^\s*(\d+\%)\s*$/i;
+
+	CKEDITOR.plugins.add( 'image3', {
```
f1ames: That's a lot of duplication, maybe there is some simple way (even hacky) to overwrite `image2` widget definition to change it's mask property so the partial mask is applied? 🤔  I would leave dialog and context menu untouched (so one can test e.g. changing widget size by dialog).

plugins/widget/plugin.js
```diff
@@ -3310,70 +3327,268 @@
...
+			mask;
+
+		// If requested part is invalid, don't create mask.
+		if ( !part ) {
+			return;
+		}
+
+		mask = this.wrapper.findOne( '.cke_widget_partial_mask' );
+
+		if ( mask ) {
```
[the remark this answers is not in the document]
Dumluregn: It's a nice way to reduce one level of conditionals 👍

plugins/widget/plugin.js
```diff
@@ -3310,70 +3327,268 @@
...
+	}
+
+	function setupPartialMask() {
+		if ( !this.wrapper ) {
+			return;
+		}
+
+		// Original value of 'widget.mask' is substituted with actual mask element, so
+		// 'widget.maskPart' property was added to be able to adjust partial mask e.g. after resizing.
+		this.maskPart = this.maskPart || this.mask;
```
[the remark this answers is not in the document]
Dumluregn: I was thinking about it too and wanted to drop this for both flows, but I noticed that we have a test checking if mask is DOM element - but this may be both a business use case or just a way to determine if mask was created. What was first - 🥚 or 🐔? I have no clue, so decided to let it be as it is. However I'm open-minded here.

plugins/widget/plugin.js
```diff
@@ -3310,70 +3327,268 @@
...
+		}
+
+		for ( editableName in definedEditables ) {
+			editableDef = definedEditables[ editableName ];
+			widget.initEditable( editableName, typeof editableDef == 'string' ? { selector: editableDef } : editableDef );
+		}
+	}
+
+	function setupMask( widget ) {
+		if ( widget.mask === true ) {
```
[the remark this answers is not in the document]
Dumluregn: It was also my conclusion that `setupWidget()` can only be called once for a widget and I tested it on block (`codesnippet`) and inline( `mathjax`) plugins, but if you think it's dangerous I can make this condition safer.

plugins/widget/plugin.js
Dumluregn: I just tested the code you proposed. The problem with it occurs when the images aren't different each time, to be more precisely - when image is already cached. Then it seems to always have `img.complete === true` and doesn't attach listener on `load` event.
So I tried to abandon this condition and just always attach listener (which is obviously less efficient) and it works pretty ok, in fact I still managed to break it a few times but `1.` it was rather rare (about 1 per 20-30 page refreshes) and `2.` I can't guarantee that the reason wasn't connected to abusive browser refreshing. So I think it can be called quite reliable solution.

Dumluregn:
Rebased onto latest major.

plugins/widget/plugin.js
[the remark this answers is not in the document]
f1ames: Ok, I've looked into mentioned issue - https://dev.ckeditor.com/ticket/11281. It looks like in some case the widget HTML is reused - that explains reusing DOM mask element. The whole setup and widget state is for a new widget instance so I think we are safe here 👍

tests/plugins/widget/manual/partialmask.md
```diff
@@ -0,0 +1,26 @@
+@bender-ui: collapsed
+@bender-tags: 4.13.0, feature, 3240
+@bender-ckeditor-plugins: wysiwygarea,toolbar,clipboard,image2,sourcearea,list,undo
+
+**Note:**
+
+**Please test with dev tools open and after each step check mask size and alignment. Each time it should adjust to the changes made.**
```
f1ames:
```suggestion
**Please test with dev tools open and after each step check mask size and position. Each time it should adjust to the changes made.**
```

Dumluregn:
Right, it seems that Chrome is creating mask in some weird moment, when widget's offset position is not final yet, and it places mask in the wrong place. I think that extending the mentioned fix to all browsers would be just fine, because looking for another solution would probably bring us to just listening to different event. WDYT?

f1ames:
> Right, it seems that Chrome is creating mask in some weird moment, when widget's offset position is not final yet, and it places mask in the wrong place. I think that extending the mentioned fix to all browsers would be just fine, because looking for another solution would probably bring us to just listening to different event. WDYT?

I wonder if the same issue occurs for widgets without images. This will mean that we need another event here. If it's only a widget with image issue we could go for enabling this fix for all browsers.

Could you check if this issue occurs for any widgets without plugin (it will probably require adding alignment styles for some imageless widgets (e.g. media embed) like in https://ckeditor.com/docs/ckeditor4/latest/examples/styles.html#widget-styles).

Issue #3306, opened 2019-07-31
Make applyStyle and removeStyle hookable


f1ames:
## Type of report

Feature request

## Provide description of the new feature

Some of the features use [`commands`](https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_command.html) (e.g. bold, italics, alignment, etc) which makes it easy to hook into with [`beforeCommandExec`](https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_editor.html#event-beforeCommandExec) and [`afterCommandExec`](https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_editor.html#event-afterCommandExec).

However, others use [`applyStyle`](https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_editor.html#method-applyStyle) and [`removeStyle`](https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_editor.html#method-removeStyle) (e.g. font, stylescombo, colorbutton, colordialog) which makes it impossible to hook into as there are no events or any other forms of intercepting the flow (apart from hacky methods proxying). This makes it quite difficult to integrate with other features.

So the idea is to make `applyStyle` and `removeStyle` methods hookable. We were thinking about refactoring using commands however it requires quite significant amount of work and may be problematic. So we decided to go with events which will be fired before/after styles applying/removing.

I'm just wondering if one event will be enough - something like `beforeApplyStyle`/`removeApplyStyle` (which will also allow canceling styles applying) or we could also add `afterApplyStyle`/`afterApplyStyle`? cc @jacekbogdanski

jacekbogdanski:
Fortunately, these methods don't return anything, so probably canceling them shouldn't cause side effects. Yeah, I think that additional events instead of dedicated methods will be much easier to implement (single place) with the same profit.

Just don't forget about passing style object into an event, otherwise, it will be useless.

`before` events looks redundant for me as if I'm correct styles method are called synchronously, so you can always attach a listener at the end of the event queue. In that case, maybe let's simplify naming into `applyStyle/removeStyle` event?

Comandeer:
You can look into `copyformatting` plugin. We made there commands and events around applying styles. Maybe similar strategy can be used also for styles core?

f1ames:
[...]

PR #3375, opened 2019-09-03
Make font plugin use command instead of direct style operations

### LINKED ISSUE #1306: Reapplying active font size/family with collapsed selection adds nested spans with same styling

Bug

1. Go to e.g. `/tests/plugins/font/manual/preservestyle` (can be performed in any editor instance with `font` plugin).
2. Add text `foo bar`.
3. Select whole text.
4. Apply some font size (e.g. **24**).
5. Put collapsed selection inside styled text.
6. Apply same font size few times.

### Expected result

Font keeps its size. Additional HTML elements are not added.

### Actual result

Font keep its size. However, additional nested spans are added and if you start to type you end up with something like `<p><span style="font-size:24px">Hel<span style="font-size:24px"><span style="font-size:24px"><span style="font-size:24px"><span style="font-size:24px">foo</span></span></span></span>lo</span> world</p>`.

As we updated the logic of dropdown (_Default_ value), previously reapplying the same style just removed it. So it is regression in a sense that it got broken not that it does not work like before.

## Other details

* Browser: All (checked on Chrome, FF)
* OS: All
* CKEditor version: 4.8.0
* Installed CKEditor plugins: `font`

### LINKED ISSUE #3728: Expose font commands
## Type of report

Feature request

Font plugin should expose font name and size change as dedicated commands as an addition to internal content style manipulation.

Extracted from https://github.com/ckeditor/ckeditor4/issues/3306, read https://github.com/ckeditor/ckeditor4/issues/3306#issuecomment-526525004 for more details about implementation.

msamsel:

New feature

yes

### This PR contains

- [x] Unit tests
- [x] Manual tests

```
* [#3306](https://github.com/ckeditor/ckeditor-dev/issues/3306): Added new commands `font` and `fontSize` which applies the selected font style from the toolbar.
* [#1306](https://github.com/ckeditor/ckeditor-dev/issues/1306): Fixed: Creating nested spans with font's styles after reapplying it. 
```
## What changes did you make?

Mostly refactoring of the `onClick()` method in font's richcombo.
The old mechanism which always removes and applies styles is now replaced with detection if in the given range there are nodes required change. And if so, then the appropriate command is executed over the editor.

1. There are two functions which were just extracted from the main thread for better separation, without logic change:
    * `_prepareStylesAndNames()`
    * `_splitElementOnCollapsedRange()`
2. Function `_matchElementToStyleDefinition()` with all related functions could fit better to `CKEDITOR.style`, however, moving it there will introduce more API changes and will require better testing. I think right now it might be kept here and I can create a ticket to move it to CKEDITOR.style one day.
3. I see there is a bunch of bugs with `label:plugins:font`, which are not fixed. Probably implementing walker during command execution would fix most of them or adding some dedicated CKEDITOR.styles method which might help for such cases.
4. There also case with default styles, which probably could be implemented as CKEDITOR.style and maybe fix another part of the reported issue. However, it wasn't a scope of origin issue as well.
5. I add options object which will be utilized by the `select` method added with this PR: #3392 

Closes #3728
Closes #1306


msamsel:
About `Command State` it is huge change which you propose and from my perspective is practically rewriting entire plugin from the scratch.
1. Currently, the state is kept in UI element, not in the command (which wasn't present there before).
2. As you also noticed command class is not ready to support customized values and will require another huge modification.
3. What's more, if I understand correctly the logic here, the state is somehow shared with `listblock` plugin and flow between those components seems to be another good candidate to refactoring and improvements.

> Still, we need some way to update combo box value,

There wasn't any word about it in the specification. Moving logic to command is completely different task than allowing on update values displayed in the richcombo. Is it required here?
Right now available styles are recognized and add during component render. Providing API to modify it later, sounds like a new feature to me.

You have great ideas about how to simplify it, however, those changes might consume a tremendous amount of time and work, where we could provide something more meaningful for the users. That's why I would prefer to split this work into parts and find those things which are "essential" to the spreadsheet, and all things which are "nice to have" fix if there will be time or there will be another round of refactoring this plugin.

f1ames:
> There wasn't any word about it in the specification. Moving logic to command is completely different task than allowing on update values displayed in the richcombo. Is it required here?
Right now available styles are recognized and add during component render. Providing API to modify it later, sounds like a new feature to me.

I agree with @msamsel here, we haven't discussed it earlier and it's not the part of this task. I suspect, you got this idea @jacekbogdanski after some testing which showed that our integration with the new plugin doesn't entirely work - `font` combo doesn't correctly recognizes selection styles and does not update it's state?

jacekbogdanski:
> combo doesn't correctly recognizes selection styles and does not update it's state?

Exactly, but we already discussed some time ago that this case should be also covered within this PR (and #https://github.com/ckeditor/ckeditor4/issues/3387 followup).

Due to the fact that this PR may highly depend on changes done in #3405 and #3387, I suggest to wait with a review once we finish these PRs.

msamsel:
> > combo doesn't correctly recognizes selection styles and does not update it's state?
> 
> Exactly, but we already discussed some time ago that this case should be also covered within this PR (and ##3387 followup).
> 
> Due to the fact that this PR may highly depend on changes done in #3405 and #3387, I suggest to wait with a review once we finish these PRs.

It's referenced in PR description at the top:
> Close #1306.
Together with pr #3405, it closes #3306.

jacekbogdanski:
Yes, I know, but I already reviewed #3405 and #3387 and requested some changes which can invalidate changes proposed in this PR (like `options` property and synchronization between command and button).

msamsel:
I just rebase it to major. Now it requires adaptation to changes introduced by #3387

msamsel:
@jacekbogdanski I dig out this PR and having changes from #3405 and #3387 on my mind, I wonder how we can solve state synchronization for richCombo.

Rich combo update preview of its value in toolbar when selection is changed, but button state remain OFF. So in comparison to colorbutton, there is a visual feedback to the user that current selection has different style. That's why I wonder, how we should bind together those styles?

Maybe command should have 3 states, based on current selection. But UI will work differently here and switch ON and OFF only when richcombo panel is open? This will cause desynchronization of states between UI and Command, however, disabling Command should also disable UI.
Or maybe somehow trick the css, and apply the proper state, but without triggering css switch?

I also wonder how important is to keep state synchronization for richcombo commands? Maybe we could only synchronize disabling and better improvements extract as follow up?

I appreciate some ideas because now implementing similar command synchronization, as it was done for panelbutton with color command ticket, looks rather like a big task. What's more, maybe we won't event need so big refactor and so tight state synchronization?

plugins/font/plugin.js
```diff
@@ -1,45 +1,75 @@
...
+				}
 			}
-		}
+		} );
 
 		editor.ui.addRichCombo( comboName, {
 			label: lang.label,
 			title: lang.panelTitle,
 			toolbar: 'styles,' + order,
-			defaultValue: 'cke-default',
```
jacekbogdanski: I'm wondering why default value has been set into `cke-default` in the first place and if it's safe to replace it with an empty string. It looks like later it's added to richcombo entries as an index key, so it will result in object like:

```javascript
items = {
'': '(Default)',
'Arial': 'Arial'
...
}
```
It's a valid object key because it's just expected that an object key will be just a string type, however, it may be error-prone because the empty string is also a falsy value. Imagine something like:

```javascript
for ( var key in items ) {
  // This will never execute for empty string key,
  // however, from the code logic point of view, it's correct key with some value.
 if ( key ) {
    console.log( items[ key ] );
}
```

This may work for now, but later it could create hard to find issue. Also, if it's done on purpose, so the default value is skipped for some algorithms, it still seems better to explicitly check if it's the default value for better readability. Could you tell the reason for this change?

plugins/font/plugin.js
```diff
@@ -94,7 +124,7 @@
 				this.startGroup( lang.panelTitle );
 
 				// Add `(Default)` item as a first element on the drop-down list.
-				this.add( this.defaultValue, defaultText, defaultText );
+				this.add( '', defaultText, defaultText );
```
jacekbogdanski: `this.defaultValue` actually tells more than a magic empty string, so maybe it's worth to keep it as named property? Also easier to update value later if needed.

plugins/font/plugin.js
```diff
@@ -1,45 +1,75 @@
...
  * @license Copyright (c) 2003-2019, CKSource - Frederico Knabben. All rights reserved.
  * For licensing, see LICENSE.md or https://ckeditor.com/legal/ckeditor-oss-license
  */
 
 ( function() {
-	function addCombo( editor, comboName, styleType, lang, entries, defaultLabel, styleDefinition, order ) {
+	function addCombo( editor, comboName, styleType, lang, entries, defaultLabel, configStyleDefinition, order ) {
 		var config = editor.config,
-			style = new CKEDITOR.style( styleDefinition );
+			allowedAndRequiredContent = new CKEDITOR.style( configStyleDefinition ),
```
jacekbogdanski: Maybe something like `contentStyle` or `defaultContentStyle` as it's also used in refresh methods?

plugins/font/plugin.js
```diff
@@ -1,45 +1,75 @@
...
+				if ( !allowedAndRequiredContent.checkApplicable( path, editor, editor.activeFilter ) ) {
+					this.setState( CKEDITOR.TRISTATE_DISABLED );
+				} else {
+					this.setState( CKEDITOR.TRISTATE_OFF );
+				}
 			}
-		}
+		} );
 
 		editor.ui.addRichCombo( comboName, {
```
jacekbogdanski: Interestingly, you can still provide `command` for richcombo and it will be correctly added as a command UI element:
https://github.com/ckeditor/ckeditor4/blob/066ab2b99e4c634a61bb1d6f292ab297cfe464f0/core/ui.js#L92
But not sure if it does anything special as rich combo doesn't use it. Still, nice to keep it unified with the rest of the UI items, at least partially.

plugins/font/plugin.js
```diff
@@ -209,10 +163,323 @@
 			},
 
 			refresh: function() {
-				if ( !editor.activeFilter.check( style ) )
+				if ( !editor.activeFilter.check( allowedAndRequiredContent ) )
 					this.setState( CKEDITOR.TRISTATE_DISABLED );
 			}
```
jacekbogdanski: Not sure, but probably once we synchronize UI element state on command refresh, this code could be removed?

plugins/font/plugin.js
```diff
@@ -209,10 +163,323 @@
...
+				continue;
+			}
+
+			if ( !_hasValidAttributes( el, currentStyleDefinition ) ) {
+				continue;
+			}
+
+			if ( !_hasValidStyles( el, currentStyleDefinition ) ) {
+				continue;
+			}
```
jacekbogdanski: **nit:** Maybe marge it into single if statement? Won't be necessary more complicated if separated by a new line:

```javascript
if ( condition1 ||
     condition2 ||
     condition3 ) {
      continue;
}
```

tests/plugins/font/manual/fontcommand.md
```diff
@@ -0,0 +1,30 @@
+@bender-tags: 4.14.0, feature, 3306, 3728
+@bender-ui: collapsed
+@bender-ckeditor-plugins: wysiwygarea, toolbar, font, enterkey, elementspath, notification
```
jacekbogdanski: It would be nice to also add sourcearea plugin, so we can see if modes work correctly.

tests/plugins/font/manual/fontcommand.md
```diff
@@ -0,0 +1,30 @@
...
+
+## Expected:
+Command should be executed only once. Styling the same content with the same style should not trigger command.
+
+## Unexpected:
+Command is not executed or is executed more than once.
+
+# Scenario 2:
+1. Select styled text and remove font/fontSize from it.
+2. Try to remove style from not styled text.
```
jacekbogdanski: I know it's easier to keep multiple related case studies in the same file to remove HTML/JS code duplication, however, more complicated reproduction steps make harder manual testing, especially on smaller devices. Maybe both scenarios could be extracted into separate files, but with shared `__template__.html` file? It would probably require structure like:

* `manual/command`
  * `__template__.html`
  * `applystyle.md`
  * `removestyle.md`

WDYT?

tests/plugins/font/fontcommand.js
```diff
@@ -0,0 +1,853 @@
+/* bender-ckeditor-plugins: font,toolbar,basicstyles,table,list */
```
jacekbogdanski: I'm wondering what to do with missing issue tags (`(#3728)`) here. Adding issue tags for every unit test seems weird as it's a lot of duplication but will pay off with better test discoverability. Maybe do you have some idea how we could handle this case?

tests/plugins/font/manual/collapsedandnestedspans.md
```diff
@@ -0,0 +1,14 @@
+@bender-tags: 4.13.0, feature, 1306, 3306
```
jacekbogdanski: We probably could drop `3306` issue ticket in favor of `3728`

tests/plugins/font/manual/fontcommand.md
```diff
@@ -0,0 +1,30 @@
+@bender-tags: 4.14.0, feature, 3306, 3728
```
jacekbogdanski: `3306` has been treated as general issue, but as we split it into smaller pieces, maybe only `3728` is just fine?

tests/plugins/font/fontcommand.js
msamsel: It definitely looks weird added to every single test :(

Maybe there could be added a global tag `/* bender-tags: 3728 */`. WDYT?

I also wonder how useful this tag could be here for general purpose unit tests? Maybe it's not necessary to add it at all?
The issue doesn't fix any specific edge case situation, which reference might be useful to understand reproduction steps. Do you have some idea how the issue reference here might be useful?

plugins/font/plugin.js
msamsel: It was suggested here: https://github.com/ckeditor/ckeditor4/pull/3375#pullrequestreview-283067208
However, I haven't noticed this scenario.

plugins/font/plugin.js
```diff
@@ -1,45 +1,75 @@
...
  */
 
 ( function() {
-	function addCombo( editor, comboName, styleType, lang, entries, defaultLabel, styleDefinition, order ) {
+	function addCombo( editor, comboName, styleType, lang, entries, defaultLabel, configStyleDefinition, order ) {
 		var config = editor.config,
-			style = new CKEDITOR.style( styleDefinition );
+			allowedAndRequiredContent = new CKEDITOR.style( configStyleDefinition ),
+			commandName = comboName.slice( 0, 1 ).toLowerCase() + comboName.slice( 1 ),
+			preparedStylesAndNames = _prepareStylesAndNames( {
```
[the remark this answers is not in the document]
msamsel: For now, I wrap it in a small internal abstraction. It seems to look better.

plugins/font/plugin.js
```diff
@@ -1,45 +1,75 @@
...
+					isRemove = newStyle === undefined;
 
-		// Create style objects for all fonts.
-		var styles = {};
-		for ( var i = 0; i < names.length; i++ ) {
-			var parts = names[ i ];
+				// Prevent of throw error when situation from this bug appear (#1116).
+				if ( !oldStyle && !newStyle ) {
+					return;
+				}
```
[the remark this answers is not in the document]
msamsel: It's a confusing comment. I removed it. It was only pointing to a specific test-case when I noticed some additional error which could be introduced without this change.

plugins/font/plugin.js
msamsel: Now it checks command state and disables it, when the command is disabled.
Unfortunately richcombo by itself determines what state should have, and only allows on disabling it in refresh method.

https://github.com/ckeditor/ckeditor4/blob/16c60f9c79c0658f37480377f9ba91d0aec5c69a/plugins/richcombo/plugin.js#L184-L187

plugins/font/plugin.js
jacekbogdanski: Oh, shoot, sorry for that. Self-destruction:smile: But that's what happens when continuing review a couple months later :cry:

tests/plugins/font/fontcommand.js
jacekbogdanski: Maybe let's just go with
```
/* bender-tags: 3728 */
```
? 
Even when there are more tests added to this test file, it still applies to the main ticket about this feature. WDYT?

tests/plugins/font/fontcommand.js
```diff
@@ -0,0 +1,883 @@
...
+			element: 'span',
+			styles: { 'font-size': '#(size)' }
+		},
+		ARIAL = new CKEDITOR.style( font_style, { family: 'Arial,Helvetica,sans-serif' } ),
+		COMIC_SANS = new CKEDITOR.style( font_style, { family: 'Comic Sans MS,cursive' } ),
+		COURIER_NEW = new CKEDITOR.style( font_style, { family: 'Courier New,Courier,monospace' } ),
+		SIZE_24PX = new CKEDITOR.style( fontSize_style, { size: '24px' } ),
+		SIZE_48PX = new CKEDITOR.style( fontSize_style, { size: '48px' } );
+
+	var CONTENT_TEMPLATE = '[<p>^test$@</p>' +
```
jacekbogdanski: Maybe let's merge these 2 `var` statements?

tests/plugins/font/manual/command/applystyle.md
```diff
@@ -0,0 +1,18 @@
...
+@bender-ckeditor-plugins: wysiwygarea, toolbar, font, enterkey, elementspath, notification, sourcearea
+
+General Note:
+* Font and font size now use commands instead of direct operation on editor's content.
+* When font or fontSize command is executed then notification should be displayed in the editor and browser's console.
+* Please repeat steps for both font and font size features
+
+1. Select unstyled text and apply new font/fontSize.
+2. Try to apply this same font/fontSize again.
+3. Try to apply different font for already styled text.
```
jacekbogdanski: Looks like the following numbered list have been merged with a bullet list. It probably requires a double new line. Maybe also splitting them with additional heading would help?

tests/plugins/font/manual/command/removestyle.md
```diff
@@ -0,0 +1,18 @@
...
+@bender-ui: collapsed
+@bender-ckeditor-plugins: wysiwygarea, toolbar, font, enterkey, elementspath, notification, sourcearea
+
+General Note:
+* Font and font size now use commands instead of direct operation on editor's content.
+* When font or fontSize command is executed then notification should be displayed in the editor and browser's console.
+* Please repeat steps for both font and font size features
+
+1. Select styled text and remove font/fontSize from it.
+2. Try to remove style from not styled text.
```
jacekbogdanski: Also here, bullet list and numbered list are mixed which results in weird-looking bullet list.

plugins/font/plugin.js
```diff
@@ -88,131 +193,345 @@
...
-									this.setValue( value );
-								return;
-							}
-						}
+				command.on( 'state', function() {
+					if ( command.state === CKEDITOR.TRISTATE_DISABLED ) {
+						this.setState( CKEDITOR.TRISTATE_DISABLED );
+					} else {
+						this.setState( CKEDITOR.TRISTATE_OFF );
 					}
```
jacekbogdanski: Do we actually need to check the current command state? :thinking: I see that it won't allow `CKEDITOR.TRISTATE_ON` but if someone really wants to set `ON` state, we should probably allow forcing it. It's the only reason, or there is some additional quirk fixed by checking current command states?

jacekbogdanski:
Rebased to the latest upstream.

plugins/font/plugin.js
```diff
@@ -4,42 +4,147 @@
...
 		var config = editor.config,
-			style = new CKEDITOR.style( styleDefinition );
+			lang = definition.lang,
+			defaultContentStyle = new CKEDITOR.style( definition.configStyleDefinition ),
+			stylesData = new StyleData( {
+				entries: definition.entries,
+				styleVariable: definition.styleType,
+				styleDefinition: definition.configStyleDefinition
+			} ),
+			removeStyleValue = 'cke-remove-style',
```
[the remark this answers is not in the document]
msamsel: I prefer to have defined default value outside of an UI element. Obtaining it from UI might result in a situation that command won't be working without an UI element. I feel that better is to separate command from toolbar/UI.
I will bring back te the `defaultValue` inside font panel combo. However, this value is not documented anywhere, what is a clear signal that shouldn't be used anywhere outside.

plugins/font/plugin.js
```diff
@@ -88,131 +193,345 @@
...
+				this.add( removeStyleValue, defaultText, defaultText );
 
-				for ( var i = 0; i < names.length; i++ ) {
-					name = names[ i ];
-					// Add the tag entry to the panel list.
-					this.add( name, styles[ name ].buildPreview(), name );
-				}
+				stylesData.executeForEach( function( name, styleValue ) {
+					this.add( name, styleValue.buildPreview(), name );
+				}, this );
```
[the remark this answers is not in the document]
msamsel: It looks like a really nice improvement.

plugins/font/plugin.js
```diff
@@ -4,42 +4,147 @@
...
+					entries.splice( i, 1 );
+					i--;
+				}
+			}
+		},
+
+		proto: {
+			getStyle: function( value ) {
+				return this._.data[ value ];
+			},
```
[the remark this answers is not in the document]
msamsel: 🤦‍♂ definitely `name` name is a better name here

plugins/font/plugin.js
```diff
@@ -88,131 +193,345 @@
...
+						value = getMatchingValue( editor, elementPath, stylesData );
 
-							editor.getSelection().selectRanges( [ range ] );
+					if ( value ) {
+						if ( value != currentValue ) {
+							this.setValue( value );
 						}
-					} else {
-						editor.removeStyle( previousStyle );
-					}
```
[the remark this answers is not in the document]
msamsel: The path is refactored. Right now this check together with collapsed selection support is moved to function `hasAlreadyAppliedNewStyle`.

Previously the style was always removed from selection and then applied once again. So subsequent execution of applying the same style result with DOM operation which removed the style and applied the same once again.

To avoid that and not repeating the same command, again and again, it is checked if selection is not styled or contains non-styled fragment. If yes then command is executed, and `removeStyle` operation is the inside of the command.
This generates splits of responsibilities that UI determines if current state of the editor should apply font command or not. And all DOM related operations are made only inside the command.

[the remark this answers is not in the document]
msamsel: The code which reflecting this else is now located here:
https://github.com/ckeditor/ckeditor4/blob/abb9d6cba0c3dc60706e206f630b479b9f16164a/plugins/font/plugin.js#L114-L116

plugins/font/plugin.js
```diff
@@ -88,131 +193,345 @@
...
+			} );
+
+			hasFragmentsWithoutNewStyle = !isRemoveOperation && !hasAlreadyAppliedNewStyle( {
+				range: range,
+				style: newStyle,
+				configStyleDefinition: definition.configStyleDefinition
+			} );
+
+			if ( styleToRemove || hasFragmentsWithoutNewStyle ) {
+				editor.fire( 'saveSnapshot' );
```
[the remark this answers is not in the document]
msamsel: The idea is the same as in the above comment. Firstly it is determined if style should be applied or not. So it is analyzed selection in the editor. This is nonintrusive check and doesn't sound to require `saveSnapshot`. The snapshot is taken before the command is going to be executed, so when is decision to change something in the editor's content.
From my perspective in this flow, it is not required to save a snapshot at the beginning. There might be a situation when the font command won't be executed, so firing snapshot is not necessary in such cases. Do you find some idea when it might be useful?

plugins/font/plugin.js
```diff
@@ -88,131 +193,345 @@
...
+			styleToRemove = isRemoveOperation && hasStyleToRemove( {
+				range: range,
+				configStyleDefinition: definition.configStyleDefinition
+			} );
+
+			hasFragmentsWithoutNewStyle = !isRemoveOperation && !hasAlreadyAppliedNewStyle( {
+				range: range,
+				style: newStyle,
+				configStyleDefinition: definition.configStyleDefinition
+			} );
```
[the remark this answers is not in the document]
msamsel: Yeah, I noticed that I miss such a feature in `CKEDITOR.style`. However, I feel it is beyond the scope of this task and should be done as a separate one. Together with the improvement in a font plugin. Adding new features to `CKEDITOR.style` require more work on tests and analyse of potential edge cases. From that reason, I prefer extracting this case as a separate issue.

plugins/font/plugin.js
```diff
@@ -88,131 +193,345 @@
...
+				value = stylesData.findMatchingStyleName( function( style ) {
+					return style.checkElementMatch( element, true, editor );
+				}, this );
+
+				if ( value ) {
+					return value;
+				}
+			}
+
+			return;
```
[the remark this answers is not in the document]
msamsel: I've changed it to `null`. But I also wonder if this return is necessary at all. Maybe it should be removed completely. WDYT?

plugins/font/plugin.js
msamsel: > Do we actually need to check the current command state? 

yes

> I see that it won't allow CKEDITOR.TRISTATE_ON but if someone really wants to set ON state, we should probably allow forcing it.

The code doesn't prevent setting `ON` state. You can still use `editor.ui.get( 'Font' ).setState( CKEDITOR.TRISTATE_ON )`

> It's the only reason, or there is some additional quirk fixed by checking current command states?

This is a reason why it is here.
UI doesn't synchronize with the command by itself. This would require to introduce a new feature to rich combo, which is out of the scope for this issue. Probably similar as it was done for the PanelButton for the color commands.

For example, when you would call `editor.getCommand( 'fontSize' ).disable()`, the command will change state to `DISABLED`, however, the richcombo UI won't get this info and remain in `OFF` state. To gave synchronization from command to UI, it is required to listen on command state change. I can extract and describe this case together with information that this part of the font command should be removed. WDYT?
I also don't think, it is a good idea to embed such changes directly in this PR and also I would not recommend making this PR dependent on another one.

jacekbogdanski: Sorry I could have said it wrong. I meant we could simplify this code into something like:

```javascript
command.on( 'state', function() {
  this.setState( command.state );
} );
```
so, UI change is totally dependant on command behavior in that case, without additional logic. However, as we already talked F2F we are also setting `CKEDITOR.TRISTATE_ON` within command refresh, so this place should be also simplified:

https://github.com/ckeditor/ckeditor4/blob/b183d83bb103f3bf2e8d6555356ef9b59875205b/plugins/font/plugin.js#L143-L156

Into something like:

```javascript
refresh: function( editor, path ) {
	if ( !defaultContentStyle.checkApplicable( path, editor, editor.activeFilter ) ) {
		this.setState( CKEDITOR.TRISTATE_DISABLED );
	}
}
```

plugins/font/plugin.js
```diff
@@ -88,131 +215,325 @@
...
-					this.setValue( '', defaultLabel );
 				}, this );
 			},
 
 			refresh: function() {
-				if ( !editor.activeFilter.check( style ) )
+				if ( command.state === CKEDITOR.TRISTATE_DISABLED ) {
 					this.setState( CKEDITOR.TRISTATE_DISABLED );
+				}
 			}
```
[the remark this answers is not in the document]
msamsel: yes
During `activeFilterChange`, the command state is updated before `activeFilterChange`. So state of UI will be set according to current mode as it is done inside richcombo and won't be disabled.

You can remove this part, open manual test with widget and check that state is not disabled.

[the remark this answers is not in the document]
jacekbogdanski: Yep, you are right. Thanks for explanation and good manual test :+1:

tests/plugins/font/manual/command/__template__.html
```diff
@@ -0,0 +1,79 @@
...
+			message += 'which removes style: <code>' + styleName + '</code> with value: <code>' + commandData.oldStyle.getDefinition().styles[ styleName ] + '</code>.'
+		} else if ( !commandData.oldStyle ){
+			message += 'which applies style: <code>' + styleName + '</code> with value: <code>' + commandData.newStyle.getDefinition().styles[ styleName ] + '</code>.';
+		} else {
+			message += 'which applies style: <code>' + styleName + '</code> with value: <code>' + commandData.newStyle.getDefinition().styles[ styleName ] +
+				'</code> and removes style: <code>' + styleName + '</code> with value: <code>' + commandData.oldStyle.getDefinition().styles[ styleName ] + '</code>.';
+		}
+
+		messages.push( message );
+		console.log( message );
```
jacekbogdanski: `console` breaks this test on IE8-9, probably simple check against `window.console` should be enough even for test steps as once dev tools are opened, a console is also available.

tests/plugins/font/fontcommand.js
jacekbogdanski: @msamsel could you refer to issue tags?

msamsel:
I've brought back the old path which doesn't `removeStyle` when the old and the new value was the same.

PR #3378, opened 2019-09-04
Fix for: CKEDITOR.plugins.addExternal will not load a plugin if the path does not contain slash at the end

Dumluregn:

Bug fix

### This PR contains

- [x] Unit tests
- [x] Manual tests

```
* [#917](https://github.com/ckeditor/ckeditor-dev/issues/917): Fixed: [`CKEDITOR.plugins.addExternal()`](https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_plugins.html#method-addExternal) method does not load a plugin if the path does not contain slash at the end.
```

## What changes did you make?

If the `path` argument for [`CKEDITOR.plugins.addExternal()`](https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_plugins.html#method-addExternal) method doesn't end with `/` and `fileName` is not an empty string (if it is it means that `path` contains `fileName` so we shouldn't add `/` there), we assume that `/` should be concatenated to `path`.


Dumluregn:
After the fix there was a failing test in [/tests/tickets/10913/1.js](https://github.com/ckeditor/ckeditor-dev/blob/major/tests/tickets/10913/1.js#L5) but it was: 
1. making this issue practically unfixable;
2. validating a case which [according to our documentation](https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_plugins.html#method-addExternal) was incorrect (if there is no `fileName` parameter set, then the `plugin.js` should be used - and instead `addExternal()` was still trying to find a name in path). Therefore with @f1ames we decided to remove this test and make people obey the documentation if anyone has an issue with it.

tests/core/plugins/manual/addexternal.md
```diff
@@ -0,0 +1,11 @@
...
+@bender-ui: collapsed
+@bender-ckeditor-plugins: wysiwygarea, toolbar
+
+1. Check the editor content.
+
+### Expected result:
+Editor is loaded and `image2` plugin is working (image is visible, can be resized etc.)
+
+### Unexpected result:
+Editor or image crushed.
```
Comandeer:
```suggestion
Editor or image crashed.
```

core/resourcemanager.js
```diff
@@ -136,14 +136,21 @@ CKEDITOR.resourceManager.prototype = {
 	 * @param {String} names The resource names, separated by commas.
 	 * @param {String} path The path of the folder containing the resource.
 	 * @param {String} [fileName] The resource file name. If not provided, the
-	 * default name is used. If provided with a empty string, will implicitly indicates that `path` argument
+	 * default name is used. If provided with an empty string, will implicitly indicates that `path` argument
 	 * is already the full path.
 	 */
 	addExternal: function( names, path, fileName ) {
 		names = names.split( ',' );
 		for ( var i = 0; i < names.length; i++ ) {
```
Comandeer: I know that it's out of the scope of this issue, but it seems that the [logic for setting correct path and filename](https://github.com/ckeditor/ckeditor-dev/blob/4033248f9e6fead2b7ffc9d2b973e67ec309be34/core/resourcemanager.js#L147-L154) can be moved outside the loop. The whole function can look something like:

```javascript
addExternal: function( names, path, fileName ) {
	names = names.split( ',' );

	// If there is no '/' at the end of path, but 'fileName' is anything different
	// than an empty string (so either is not set at all or is a valid name), assume
	// that path is a directory and concatenate '/' to it (#917).
	if ( path.charAt( path.length - 1 ) != '/' && fileName !== '' ) {
		path += '/';
	}

	// If "fileName" is not provided, we assume that it may be available
	// in "path". Try to extract it in this case.
	if ( !fileName ) {
		path = path.replace( /[^\/]+$/, function( match ) {
			fileName = match;
			return '';
		} );
	}

	// Use the default file name if there is no "fileName" and it
	// was not found in "path".
	fileName = fileName || ( this.fileName + '.js' );

	CKEDITOR.tools.array.forEach( names, function( name ) {
		this.externals[ name ] = {
			dir: path,
			file: fileName
		};
	} );
}
```

As we look inside this file rarely, let's fix this small issue now and just forget about it :D

core/resourcemanager.js
```diff
@@ -136,14 +136,21 @@ CKEDITOR.resourceManager.prototype = {
...
 			var name = names[ i ];
 
+			// If there is no '/' at the end of path, but 'fileName' is anything different
+			// than an empty string (so either is not set at all or is a valid name), assume
+			// that path is a directory and concatenate '/' to it (#917).
+			if ( path.charAt( path.length - 1 ) != '/' && fileName !== '' ) {
```
Comandeer: Not sure about it, but don't you think that simple `path[ path.length - 1 ]` would be more readable? 🤔 

```suggestion
			if ( path[ path.length - 1 ] !== '/' && fileName !== '' ) {
```

WDYT?

tests/core/plugins/plugins.js
```diff
@@ -2,17 +2,15 @@
...
-				assert.isTrue( CKEDITOR.plugins.get( 'myplugin' ).definition );
-			} );
-		}, this );
+	'test: Loading self defined external plugin file paths': testExternal( { plugin: 'myplugin', path: '%TEST_DIR%_assets/myplugins/sample/', file: 'my_plugin.js' } ),
 
-		this.wait();
-	},
+	// (#917)
+	'loading with slash and without fileName': testExternal( { plugin: 'myplugin1', path: '%TEST_DIR%_assets/myplugins/myplugin1' } ),
+	'loading with slash and empty fileName': testExternal( { plugin: 'myplugin3', path: '%TEST_DIR%_assets/myplugins/myplugin3/', file: '' } ),
```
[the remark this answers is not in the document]
Dumluregn: Four subsequent tests were prepared for the same issue, but I admit that putting test case for `myplugin2` after `myplugin4` was not the best idea, so I renamed tests and added ticket reference to each one.

tests/core/plugins/plugins.js
```diff
@@ -132,3 +130,20 @@ bender.test(
...
+		CKEDITOR.plugins.load( options.plugin, function() {
+			this.resume( function() {
+				if ( options.wrongFile ) {
+					assert.areEqual( options.wrongFile, CKEDITOR.plugins.externals[ options.plugin ].file );
+				}
+				assert.isTrue( CKEDITOR.plugins.get( options.plugin ).definition );
+			} );
+		}, this );
+
+		this.wait();
```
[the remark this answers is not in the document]
Dumluregn: If I just delete `this.wait` and `this.resume` the tests fail with `Test has no asserts.` comment. Code looks like that:
```
function testExternal( options ) {
	return function() {
		CKEDITOR.plugins.addExternal( options.plugin, options.path, options.file );

		CKEDITOR.plugins.load( options.plugin, function() {
			assert.areEqual( options.expectedFileName, CKEDITOR.plugins.externals[ options.plugin ].file );
			assert.isTrue( CKEDITOR.plugins.get( options.plugin ).definition );
		}, this );
	};
}
```
What's wrong?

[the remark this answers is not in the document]
Comandeer: I meant that `this.wait` and `this.resume` are also available as global functions – `wait` and `resume`. There is no need to use them as `this` methods.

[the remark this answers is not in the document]
Dumluregn: Oooook, makes sense 😂

core/resourcemanager.js
Comandeer: We won't do any unnecessary operations several times, inside the loop; just once before it. And the code will become more logical.

Dumluregn: Ok, I meant that we never use functionality for loading multiple files at once, but I understand it's JIC. Thanks for tip.

Dumluregn:
Applied suggestions from review and rebased onto latest `major`.

Issue #3387, opened 2019-09-05
API for changing selected item for rich combo based plugins


jacekbogdanski:
## Type of report

Feature request

## Provide description of the new feature

As we already started working on https://github.com/ckeditor/ckeditor-dev/issues/3306 we will also need API allowing to change combo box selected item. We agreed together with @f1ames and @msamsel that the feature should be possible generic and easy to use.

Current proposition is to extend rich combo API with additional `select` method, accepting filtering function (based mostly on find algorithm) allowing to choose correct style definition. As an example, `font` plugin resulting with rich combo component:

```javascript
// First match will decide about selected item.
editor.ui.get( 'Font' ).select( function( definition ) {
  return definition.style[ 'font-family' ] === 'Arial';
}
```
As some plugins like `font` uses more complex object to resolve correct content style, `select` method callback should accept `CKEDITOR.style.getDefinition()` object, extensively used by rich combo related plugins to set content style.

It will also require providing some optional method to implement required definition mapping to avoid breaking changes with other rich combo plugins.  It may make sense to fully delegate `select` filtering callback into plugin implementing rich combo to avoid rich combo and `CKEDITOR.style` dependency or just by providing simple ID to `CKEDITOR.style` mapping. Not totally sure which method would be better here, I would leave it to developer decision which seems more verbose / easier to implement.

msamsel:
@jacekbogdanski, @f1ames.
I think it could be nice idea to return `true` or `false` value. However, I wonder what information should be indicated by this value. I see 2 possibilities:
1. Information that value was set.
2. Information that value is a new one and was changed.

For example.
If you have already set in richcombo "Tahoma" font, and you perform a select method which is going to select "Tahoma", then what result would you expect on the output?
* `true` - because was sucesfull set the Tahoma font (even though this value was already set up)
* `false` - because there wasn't changed anything in the richcombo.

What can be more useful in this case from your perspective?

jacekbogdanski:
I'm thinking about usage of boolean return from `select` method as it seems for me more like fire and forget. But it's not that easy as the filter method may not return with any truthy value. Overall it seems like implementation error as it should always match something. Maybe it should just return false if nothing has been returned by filter callback otherwise true? We could also go more pythonic version and just throw an error which seems also good for me.

On the other hand, I'm not sure if we should inform successful/failure value change - treating this method as a simple setter may simplify API. Do you see some use cases where such information would be useful?

msamsel:
[...]

PR #3392, opened 2019-09-09
Add "select" method to rich combo

msamsel:

New feature

### This PR contains

- [x] Unit tests
- [x] Manual tests

```
[#3387](https://github.com/ckeditor/ckeditor-dev/issues/3387): Added the [CKEDITOR.ui.richCombo#select](https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_ui_richCombo.html#method-select) method.
```

## What changes did you make?

Add new method in richcombo to select proper value based on a callback function. 

Closes: #3387

### DIFF
--- a/plugins/richcombo/plugin.js
+++ b/plugins/richcombo/plugin.js
@@ -120,7 +120,9 @@ CKEDITOR.plugins.add( 'richcombo', {
 			 * to this button will be appended to.
 			 */
 			render: function( editor, output ) {
-				var env = CKEDITOR.env;
+				var env = CKEDITOR.env,
+					instance,
+					selLocked;
 
 				var id = 'cke_' + this.id;
 				var clickFn = CKEDITOR.tools.addFunction( function( el ) {
@@ -133,7 +135,8 @@ CKEDITOR.plugins.add( 'richcombo', {
 				}, this );
 
 				var combo = this;
-				var instance = {
+
+				instance = {
 					id: id,
 					combo: this,
 					focus: function() {
@@ -215,7 +218,7 @@ CKEDITOR.plugins.add( 'richcombo', {
 					instance.onfocus && instance.onfocus();
 				} );
 
-				var selLocked = 0;
+				selLocked = 0;
 
 				// For clean up
 				instance.keyDownFn = keyDownFn;
--- a/plugins/richcombo/plugin.js
+++ b/plugins/richcombo/plugin.js
@@ -402,21 +402,26 @@ CKEDITOR.plugins.add( 'richcombo', {
 			},
 
 			/**
-			 * Method expects that options object is provided with richcombo definitiont. Options should contain a key-value pairs mapping,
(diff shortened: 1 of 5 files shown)


tests/plugins/richcombo/manual/select.html
```diff
@@ -0,0 +1,53 @@
...
+
+				combo = editor.ui.get( 'custom_richcombo' );
+
+				combo.createPanel( editor );
+			}
+		}
+	} );
+
+	function setComboValue( value ) {
+		combo.select( function( option ) {
```
jacekbogdanski: This variable is defined pretty deep in code execution (`instanceReady`) which is a bit hard to find. Maybe you could use  `editor.ui.get` here and expose `editor` instance (which could be also used in `instanceReady`) when replacing the element with an editor (`CKEDITOR.replace` returns editor instance also)?

tests/plugins/richcombo/manual/select.md
```diff
@@ -0,0 +1,16 @@
+@bender-ui: collapsed
+@bender-tags: 3387, 4.13.0, feature
```
jacekbogdanski: Please, update the tag version to `4.14.0`.

plugins/richcombo/plugin.js
```diff
@@ -399,6 +404,37 @@ CKEDITOR.plugins.add( 'richcombo', {
...
+			 * Filter function takes an object as an argumnet with `value` and `text` fields. Values of those
+			 * fields match to argument passed in `CKEDITOR.ui.richcombo.add` method.
+			 *
+			 * @since 4.14.0
+			 * @param {Function} callback function should return `true` if found matching element
+			 * @param {*} callback.item object containing value and text fields which are compared by this callback
+			 */
+			select: function( callback ) {
+				if ( CKEDITOR.tools.isEmpty( this._.items ) ) {
+					this.createPanel( this._.editor );
```
jacekbogdanski:
```suggestion
					return;
```

I could mislead you with the previous comment about `this.createPanel` method - I didn't mean that it should be a part of `select` method. Creating a panel based on the value of `this._.items` property seems very implementation-dependent, so I'm for removing this part of code. However, early return if ` CKEDITOR.tools.isEmpty( this._.items )` seems like a good approach.

What I mean was that it should be developer responsibility to make sure that panel is already created when using `panel.select` method. It's rather edged case when someone is trying to use this method when a panel has not been created yet and panel creation should be the toolbar responsibility. But we may add some docs information about that. WDYT?

plugins/richcombo/plugin.js
```diff
@@ -399,6 +404,37 @@ CKEDITOR.plugins.add( 'richcombo', {
...
+							value: value,
+							text: this._.items[ value ]
+						} ) ) {
+							this.setValue( value );
+							return;
+						}
+					}
+				};
+
+				this.select( callback );
```
jacekbogdanski: It seems like it could be simplified when you drop panel creation into simple for loop (inner implementation of `this.select` method).

PR #3405, opened 2019-09-11
Implement textColor and bgColor commands

### LINKED ISSUE #3306: Make applyStyle and removeStyle hookable

### LINKED ISSUE #3727: Expose colorbutton commands
## Type of report

Feature request

Color Button should expose back and front text color change as dedicated commands as an addition to internal content style manipulation.

## Specification:
* color button registers 'textColor' and 'bgColor' commands.
* color commands takes as data an object with a `newStyle` key.
  * new style should be empty if we want to remove styles
  * new style should be an instance of [`CKEDITOR.style`](https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_style.html) if we want to apply new style
* [extra] color command should should be [`contextSensitive`](https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_commandDefinition.html#property-contextSensitive) and react on selection changes
* UI state should react to changes of related color command state

Extracted from https://github.com/ckeditor/ckeditor4/issues/3306, read https://github.com/ckeditor/ckeditor4/issues/3306#issuecomment-526525004 for more details about implementation.

### LINKED ISSUE #3748: Refresh colorbutton state during selection change
## Type of report

Feature request

Currently, `colorbutton` doesn't update its state during selection change, which lacks correct feedback for the user. We could improve this behavior by making button active when selection is placed inside color-styled content.

msamsel:

New feature

### This PR contains

- [x] Unit tests
- [x] Manual tests

```
* [#3306](https://github.com/ckeditor/ckeditor-dev/issues/3306): Added new commands `textColor` and `bgColro` which applies the selected color from color panel.
```

## What changes did you make?

* extract logic which applies color to a command.

Closes #3727
Closes #3748

### DIFF
--- a/plugins/colorbutton/plugin.js
+++ b/plugins/colorbutton/plugin.js
@@ -1,4 +1,4 @@
-﻿/**
+/**
  * @license Copyright (c) 2003-2019, CKSource - Frederico Knabben. All rights reserved.
  * For licensing, see LICENSE.md or https://ckeditor.com/legal/ckeditor-oss-license
  */
--- a/plugins/colorbutton/plugin.js
+++ b/plugins/colorbutton/plugin.js
@@ -105,6 +105,7 @@ CKEDITOR.plugins.add( 'colorbutton', {
 				style = new CKEDITOR.style( config[ 'colorButton_' + type + 'Style' ] ),
 				colorBoxId = CKEDITOR.tools.getNextId() + '_colorBox',
 				colorData = { type: type },
+				defaultColorStyle = new CKEDITOR.style( config[ 'colorButton_' + type + 'Style' ], { color: 'inherit' } ),
 				panelBlock;
 
 			editor.ui.add( name, CKEDITOR.UI_PANELBUTTON, {
@@ -231,11 +232,9 @@ CKEDITOR.plugins.add( 'colorbutton', {
 						return;
 					}
 
-					var newStyle = data.newStyle,
-						config = editor.config,
-						colorStyleTemplate = config[ 'colorButton_' + type + 'Style' ];
+					var newStyle = data.newStyle;
 
-					editor.removeStyle( new CKEDITOR.style( colorStyleTemplate, { color: 'inherit' } ) );
+					editor.removeStyle( defaultColorStyle );
 
 					editor.focus();
 
@@ -247,19 +246,13 @@ CKEDITOR.plugins.add( 'colorbutton', {
 				}
 			} );
 
-			editor.once( 'instanceReady', function() {
-				var uiElement = editor.ui.get( name ),
-					command = editor.getCommand( commandName );
-
-				if ( !command || !uiElement ) {
+			editor.attachStyleStateChange( defaultColorStyle, function( state ) {
+				if ( editor.readOnly ) {
 					return;
 				}
 
-				if ( uiElement.state !== command.state ) {
-					command.setState( uiElement.state );
-				}
-
-				_registerStateSynchronization( command, uiElement );
+				editor.getCommand( commandName ).setState( state );
+				editor.ui.get( name ).setState( state );
 			} );
 		}
 
@@ -419,15 +412,6 @@ CKEDITOR.plugins.add( 'colorbutton', {
 		function normalizeColor( color ) {
 			// Replace 3-character hexadecimal notation with a 6-character hexadecimal notation (#1008).
 			return CKEDITOR.tools.normalizeHex( '#' + CKEDITOR.tools.convertRgbToHex( color || '' ) ).replace( /#/g, '' );
-		}
-
-		function _registerStateSynchronization( command, uiElement ) {
-			// The reverse order is not synchronized as we don't do it anywhere in the editor.
-			command.on( 'state', function() {
-				if ( this.state !== uiElement.getState() ) {
-					uiElement.setState( this.state );
-				}
-			} );
 		}
 	}
 } );
--- a/plugins/colorbutton/plugin.js
+++ b/plugins/colorbutton/plugin.js
@@ -1,4 +1,4 @@
-/**
+﻿/**
  * @license Copyright (c) 2003-2019, CKSource - Frederico Knabben. All rights reserved.
  * For licensing, see LICENSE.md or https://ckeditor.com/legal/ckeditor-oss-license
  */
@@ -422,19 +422,12 @@ CKEDITOR.plugins.add( 'colorbutton', {
 		}
 
 		function _registerStateSynchronization( command, uiElement ) {
+			// The reverse order is not synchronized as we don't do it anywhere in the editor.
 			command.on( 'state', function() {
 				if ( this.state !== uiElement.getState() ) {
 					uiElement.setState( this.state );
 				}
 			} );
-
-			uiElement.setState = function( state ) {
-				uiElement.constructor.prototype.setState.call( uiElement, state );
-
-				if ( command.state !== state ) {
-					command.setState( state );
-				}
-			};
 		}
 	}
 } );
--- a/plugins/colorbutton/plugin.js
+++ b/plugins/colorbutton/plugin.js
@@ -20,7 +20,12 @@ CKEDITOR.plugins.add( 'colorbutton', {
 			lang = editor.lang.colorbutton;
 
 		if ( !CKEDITOR.env.hc ) {
-			addButton( 'TextColor', 'fore', lang.textColorTitle, 10, {
+			addButton( {
+				name: 'TextColor',
+				type: 'fore',
+				commandName: 'textColor',
+				title: lang.textColorTitle,
+				order: 10,
 				contentTransformations: [
 					[
 						{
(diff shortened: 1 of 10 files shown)


msamsel:
About failing CI, I don't know what happen. The tests are passing locally and seem to be completely indented from made change. Similar errors might be found on other PR branches.

plugins/colorbutton/plugin.js
```diff
@@ -364,6 +405,22 @@ CKEDITOR.plugins.add( 'colorbutton', {
...
 		}
+
+		function _registerStateSynchronization( command, uiElement ) {
+			command.on( 'state', function() {
+				if ( this.state !== uiElement.getState() ) {
+					uiElement.setState( this.state );
+				}
+			} );
+
+			uiElement.setState = function( state ) {
```
jacekbogdanski: The reverse order of updating the state seems unnecessary. You can test it with other buttons in CKEditor, e.g.:

```javascript
var button = editor.ui.get( 'Bold' );
button.setState( 0 );
var command = editor.getCommand( 'bold' );
command.state; // => 2
```

It may seem like they should synchronize, but because we are not doing it in our codebase, I recommend to stick to adopted logic.

tests/plugins/colorbutton/colorcommand.js
```diff
@@ -0,0 +1,515 @@
...
+		return CKEDITOR.style.getStyleText( style.getDefinition() ).replace( ';', '' );
+	}
+
+	function assertColorCommand( options ) {
+		var expected = options.expected,
+			editor = options.editor,
+			newStyle = options.newStyle,
+			collapsed = options.collapsed || false,
+			commandName = options.commandName;
+
```
jacekbogdanski: Stray line.

tests/plugins/colorbutton/colorcommand.js
```diff
@@ -0,0 +1,515 @@
...
+				}
+			} );
+		},
+
+		'test should synchronize state between text color command and text color panel': function() {
+			var editor = this.editor,
+				colorPanel = editor.ui.get( 'TextColor' ),
+				command = editor.getCommand( 'textColor' );
+
+			assert.areEqual( colorPanel.getState(), command.state, 'should have same initial state' );
```
jacekbogdanski: Start message sentence with upper case, please.

tests/plugins/colorbutton/manual/colorcommand.html
```diff
@@ -0,0 +1,25 @@
...
+
+	<p><span style="background-color:#f39c12">This text has background color.</span></p>
+</div>
+
+<script>
+	var index = 1;
+	CKEDITOR.replace( 'editor', {
+		height: 400,
+		on: {
+			'beforeCommandExec': function( evt ) {
```
jacekbogdanski: It would be nice to also verify the color change, so maybe instead of using console logs and notification, you could add some element (outside editor) like small rectangle changing color to the same as text inside the editor?

plugins/colorbutton/plugin.js
```diff
@@ -87,7 +87,8 @@ CKEDITOR.plugins.add( 'colorbutton', {
 			var style = new CKEDITOR.style( config[ 'colorButton_' + type + 'Style' ] ),
 				colorBoxId = CKEDITOR.tools.getNextId() + '_colorBox',
 				colorData = { type: type },
-				panelBlock;
+				panelBlock,
+				commandName = type === 'back' ? 'bgColor' : 'textColor';
```
jacekbogdanski: It seems like a command name could be also passed as an argument reducing magic done by function :thinking: 
Could you also refactor `addButton` like `renderColors` to accept options object instead of so may arguments?

plugins/colorbutton/plugin.js
```diff
@@ -203,10 +209,51 @@ CKEDITOR.plugins.add( 'colorbutton', {
...
+
+				if ( uiElement.state !== command.state ) {
+					command.setState( uiElement.state );
+				}
+
+				_registerStateSynchronization( command, uiElement );
+			} );
 		}
 
-		function renderColors( panel, type, colorBoxId, colorData ) {
```
jacekbogdanski: :+1: for removing primitive obsession.

tests/plugins/colorbutton/manual/colorcommand.md
```diff
@@ -0,0 +1,13 @@
+@bender-tags: feature, 4.13.0, 3306
```
jacekbogdanski: Please, update the version tag as it should point into `4.14.0` now.

msamsel:
I've used the same way as it is done for basic style synchornization.
I've not eager to modify panel button plugin with this issue, as it requires to preserve full backward compatibility. There might be 3rd party plugins which rely on the current state and changing it should be made with caution. This would also generate writing a bunch of new test checking behaviour of new features of the panel button, which will extend this issue even more.
WDYT about such approach?

msamsel:
Ok it appears to be more complicated as I initially thought. The `editor.attachStyleStateChange` doesn't update properly when `activeFilter` is changed and might appear situation when command together with UI should be disabled.
This might happen inside widget which contains custom rules what is allowed inside their editables.
So right now I made the command [`contextsensitive`](https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_commandDefinition.html#property-contextSensitive). The state of the command is refreshed when selection is changed in the editor. So the logic seems to be the same however is closer to the command where is the source of truth. Together with the command is refreshed the state of the UI button to have both bound together.

msamsel:
Issue related to invalid behaviour with custom config already exists in the editor. I've made a separate issue for that: #3741.

plugins/colorbutton/plugin.js
```diff
@@ -203,10 +227,57 @@ CKEDITOR.plugins.add( 'colorbutton', {
...
+					} else {
+						newState = CKEDITOR.TRISTATE_OFF;
+					}
+
+					this.setState( newState );
+
+					if ( uiElement ) {
+						uiElement.setState( newState );
+					}
+				}
```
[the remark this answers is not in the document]
msamsel: I removed the refresh method from colorbutton panel, as now everything is handled by the command.
The problem with custom configuration is extracted as a separate ticket as it is not a regression. #3741

msamsel:
I move command synchronization to the panelbutton logic and also add 3 unit test checking it.
Please also notice that I add a new manual test, which checks `activeFilter` settings and panelbutton state, by a custom widget with restricted editable settings.

plugins/panelbutton/plugin.js
```diff
@@ -70,54 +71,86 @@ CKEDITOR.plugins.add( 'panelbutton', {
...
+					if ( this.command ) {
+						var me = this;
+
+						editor.getCommand( this.command ).on( 'state', function() {
+							me.setStateFromCommand( this );
+						} );
+					}
+
+					return CKEDITOR.ui.button.prototype.render.call( this, editor, output );
+				},
```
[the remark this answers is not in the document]
msamsel: Yeah it is working. Previously  I haven't notice that `modes` are still available for panel button and the code was not reachable.

plugins/panelbutton/plugin.js
```diff
@@ -70,54 +71,86 @@ CKEDITOR.plugins.add( 'panelbutton', {
...
+						if ( me.className ) {
 							this.element.getFirst().removeClass( me.className + '_panel' );
+						}
 
-						me.setState( me.modes && me.modes[ editor.mode ] ? CKEDITOR.TRISTATE_OFF : CKEDITOR.TRISTATE_DISABLED );
+						if ( command ) {
+							me.setStateFromCommand( command );
+						} else {
+							me.setState( me.modes && me.modes[ editor.mode ] ? CKEDITOR.TRISTATE_OFF : CKEDITOR.TRISTATE_DISABLED );
+						}
```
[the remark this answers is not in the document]
msamsel: I also add 2 new unit test checking cases when `modes` and `command` are defined inside one panel button.

msamsel:
> There is a small problem with synchronization between UI and command when the panel is opened here:

I would remain this case for now. As it seems that it shouldn't generate any issues and is fixed just after panel hide.
I think it would be better to wait if this will be an issue for the user and fix it when specific requirements will appear. Right now it is really hard to say if it is a problem at all.

PR #3421, opened 2019-09-16
Add link protocol field customization

### LINKED ISSUE #2227: Use "https" as default protocol for links
## Type of report

Feature request

Change the default protocol for links to "https" from "http".  https://github.com/ckeditor/ckeditor-dev/blob/master/plugins/link/dialogs/link.js line 268

jacekbogdanski:

New feature

### This PR contains

- [x] Unit tests
- [x] Manual tests

```
[#2227](https://github.com/ckeditor/ckeditor-dev/issues/2227): Added [`config.link_defaultProcotol`](https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_config.html#cfg-link_defaultProtocol) configuration option that allows setting default URL protocol for [Link](https://ckeditor.com/cke4/addon/link) plugin dialog.
```

## What changes did you make?

Closes #2227

### DIFF
--- a/plugins/link/dialogs/link.js
+++ b/plugins/link/dialogs/link.js
@@ -280,12 +280,11 @@
 								[ 'news://\u200E', 'news://' ],
 								[ linkLang.other, '' ]
 							],
+							'default': editor.config.link_defaultProtocol,
 							setup: function( data ) {
-								var protocol = data.url && data.url.protocol;
-
-								protocol = protocol !== undefined ? protocol : editor.config.link_defaultProtocol;
-
-								this.setValue( protocol );
+								if ( data.url ) {
+									this.setValue( data.url.protocol || '' );
+								}
 							},
 							commit: function( data ) {
 								if ( !data.url )
--- a/plugins/link/dialogs/link.js
+++ b/plugins/link/dialogs/link.js
@@ -280,7 +280,7 @@
 								[ 'news://\u200E', 'news://' ],
 								[ linkLang.other, '' ]
 							],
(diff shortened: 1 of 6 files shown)


plugins/link/plugin.js
```diff
@@ -895,7 +895,16 @@
...
+		linkShowTargetTab: true,
+
+		/**
+		 * Default URL protocol used for link dialog.
+		 *
+		 * @cfg {String}
+		 * @member CKEDITOR.config
+		 * @since 4.13.0
+		 */
+		link_defaultProtocol: 'http://',
```
f1ames: Shouldn't we be consistent with rest of the config options and have `linkDefaultProtocol`?

plugins/link/dialogs/link.js
```diff
@@ -272,7 +272,6 @@
 							id: 'protocol',
 							type: 'select',
 							label: commonLang.protocol,
-							'default': 'http://',
```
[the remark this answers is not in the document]
jacekbogdanski: I incorrectly supposed that a single instance of the dialog is created for whole CKEditor (like plugins), so in that case, your solution won't work. But it looks like it's created for every editor separately so :+1:

plugins/link/plugin.js
jacekbogdanski: `link_defaultProtocol` is actually consistent with our current guideline for config values - but as you noticed other link config values have full camel case names, so I agree it's probably better to stick with old naming if it's already available :+1:

PR #3480, opened 2019-09-23
Fix broken paste in iOS 13

### LINKED ISSUE #3354: Pasting, among other things, broken with iOS 13 (beta)
## Type of report

Bug

1. Click into an editor
2. Attempt to paste text from your clipboard

Additional information from reporting users:

> Pasting any text is not working on the cke4 demo but is working on the cke5 demo. 
>  
> There’s a few weird issues I’m noticing on iOS13 too. Sometimes the double space for a full stop doesn’t work. Auto-correct doesn’t work. And I’ll often have large gaps in between paragraphs that never happened on ios12. 

### Expected result

The text from your clipboard displays in the editor

### Actual result

The editor "flickers" and your text does not display

## Other details

* Browser: Safari
* OS: **iOS 13 Beta, up to and including public beta 7**
* CKEditor version: CKE4.12.1

We have an integrated instance of CKEditor 4 (currently 4.12.1) in our software release, and at least 5 users have reported to us that pasting in CKEditor 4, including the default demo on the site, is not working, however pasting does work with the CKEditor 5 demo. While we anticipate moving towards CKEditor 5 in the future, _and_ iOS 13 is currently only in beta, I felt it prudent to pass along these reports.

msamsel:

Bug fix

### This PR contains

- [x] Unit tests
- [x] Manual tests

```
*[#3354](https://github.com/ckeditor/ckeditor4/issues/3354) Fix: Broken paste in iOS 13.

```

## What changes did you make?

Mark current iOS Safari version as supported for custom copy/cut and mark it as trusted browsers. This flags generally change the way how we treat pasted and copied data:
https://github.com/ckeditor/ckeditor4/blob/98cae96bbe874496a63cdedd0dd9a8f869e3a0b3/plugins/clipboard/plugin.js#L2249-L2271
It looks like a bug which is reported here: https://bugs.webkit.org/show_bug.cgi?id=19893#c34 no longer occur in iOS Safari.
The current version of Safari supports custom data types applied in data transfer.

As tests were already there covering those features I just simple activate them. I also re-activate manual tests.

Closes #3354

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -42,6 +42,7 @@ Fixed Issues:
 * [#14613](https://dev.ckeditor.com/ticket/14613): Fixed: Race condition when loading plugins for an already destroyed editor instance throws an error.
 * [#2257](https://github.com/ckeditor/ckeditor-dev/issues/2257): Fixed: The editor throws an exception when destroyed shortly after it was created.
 * [#3115](https://github.com/ckeditor/ckeditor-dev/issues/3115): Fixed: Destroying the editor during the initialization throws an error.
+* [#3354](https://github.com/ckeditor/ckeditor4/issues/3354): [iOS] Fixed: Pasting no longer works on iOS version 13.
 
 API Changes:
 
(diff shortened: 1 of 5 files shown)


f1ames:
Rebased onto latest `major`.

Issue #3498, opened 2019-09-25
Editor throws an error during copy when selection starts in text an ends in widget


msamsel:
## Type of report

Bug

## Provide detailed reproduction steps (if any)

1. Open console
2. Open test: /tests/plugins/widget/manual/block
3. set content as paragraf and image:
```
<p>foo</p>

<figure class="image"><img alt="Lena Söderberg" src="/tests/_assets/lena.jpg" width="150" />
<figcaption>Image of <a href="http://en.
			wikipedia.org/wiki/Lena_S%C3%B6derberg">Lena Söderberg</a> used in<br />
many image processing experiments.</figcaption>
</figure>
```
4. Make selection which starts in paragraph and ends on image (but before caption)
5. Press <kbd>CTRL</kbd>/<kbd>CMD</kbd>+<kbd>C</kbd> 

### Expected result

no error in the console

### Actual result

error in the console

## Other details
![Screenshot 2019-09-25 at 14 29 10](https://user-images.githubusercontent.com/9017322/65600659-de5e6380-dfa0-11e9-9f3c-a8a45cd63382.png)

* Browser: Chrome
* OS: MacOS
* CKEditor version: 4.13.0
* Installed CKEditor plugins: …

msamsel:
Related to: #3551

Issue #3007, opened 2019-04-01
Can't delete/edit content when widget is in selection


Comandeer:
## Type of report

Bug

## Provide detailed reproduction steps (if any)

1. Open https://jsfiddle.net/Comandeer/zf5eu26g/
2. Start selecting from right text under the widget and end on bottom of the widget.
3. Press <kbd>Backspace</kbd>/<kbd>Delete</kbd> or just start typing.

### Expected result

Selected content is deleted/replaced with typed text.

### Actual result

Chrome/Firefox: editor's content is scrolled, but widget and text stays untouched.
Safari: selected text is deleted alongside widget's caption; selection is inside caption.

## Other details

* Browser: Chrome, Firefox, Safari
* CKEditor version: 4.5.0+
* Installed CKEditor plugins: widget, image2

Issue #2517, opened 2018-10-25
Image selection: Uncaught TypeError: Cannot read property 'checkReadOnly' of undefined


ashklianko:
## Type of report

Bug

## Provide detailed reproduction steps (if any)

1. Insert captioned image as a first entry in editor (nothing should be above it, even empty p or div)
2. Type some text after image
3. Start selecting text and release mouse button over image
4. Open image dialog and try inserting other image

### Expected result

New image inserted in place of selected text and image

### Actual result

Nothing happens, error in dev console:

> Uncaught TypeError: Cannot read property 'checkReadOnly' of undefined

## Other details

Reproduced even on https://ckeditor.com/ckeditor-4/

## Other details

* Browser: Chrome, Firefox

mlewand:
On Firefox a different exception is thrown: `TypeError: b is null`, works fine on Edge.

engineering-this:
There is issue with selection.

## Chrome:

```javascript
startContainer: div.cke_widget_wrapper.cke_widget_block.cke_widget_image,
startOffset: 2
```
Focusing editor in this case removes selection ranges.

## Firefox

```javascript
startContainer: span.cke_image_resizer_wrapper,
startOffset: 1
```
In this case widget is not inserted because `range.checkReadOnly` returns true, and then editor tries to focus widget that wasn't inserted which results in error.

engineering-this:
See https://github.com/ckeditor/ckeditor-dev/pull/2866#pullrequestreview-221070086. There are other closely related issues #3007 and #3008 caused by buggy selection. Currently #2866 contains fix for selection on widget `finalizeCreation`. However as there are more related issues and every browser allows different selection it would be better to normalize selection. `widgetselection` plugin is right place for such logic.

If one end of range is in widget, and it's not at the start of it selection should be moved out of a widget.

engineering-this:
There is another steps to reproduce this issue:

## Steps

1. Open https://codepen.io/anon/pen/dLNPGg
1. Open browser console.
1. Press button below editor.
1. Press `Image2` button.
1. Paste or type any src, and press `OK`.

## Expected

Image is inserted into editor.

## Actual

- Image is not inserted.
- Error is thrown. `TypeError: Cannot read property 'checkReadOnly' of undefined`

Note: such selection is created when I try to select whole editor content in IE11 with mouse.

Issue #3551, opened 2019-10-08
Improve support for widget's selection in editor


msamsel:
## Type of report

Feature request

## Provide a description of the new feature

There bunch of bugs in the editor which are mostly the result of partially selected widgets. The behaviour might vary between browsers. For example, in Chrome text inside widget cannot be selected, where for FireFox it's possible if selection starts before widget.
![Oct-08-2019 13-32-00](https://user-images.githubusercontent.com/9017322/66392416-19648c00-e9d0-11e9-93ae-24c4c0a3c433.gif)

This kind of selection cases which contains only partially selected widget triggers bugs in various places, which are fixed individually. Maybe there could be some common mechanisms, which will prevent the situation of the partially selected widget. It could also allow to truly treat widget as a single entity during selection, which cannot be partially selected.

This issue might be treated as a hub for smaller bugs detected during the editor usage. The most problematic cases are usually operations which rely on the selection or ranges. Such as inserting/removing content, clipboard operation, undo, etc.

msamsel:
Another case of how it interferes with image2 under FF:
![Oct-08-2019 14-11-07](https://user-images.githubusercontent.com/9017322/66394507-96decb00-e9d5-11e9-9557-c970fb7bded1.gif)

msamsel:
Might be somehow related: #3698, #3699

f1ames:
There is already related PR, however it concerns `widgetselection` so the cases when widget is on the beginning/end of the content only - https://github.com/ckeditor/ckeditor4/pull/2866.

PR #3553, opened 2019-10-08
Prevent throwing error when partially selected widget is copied

msamsel:

Bug fix

### This PR contains

- [x] Unit tests
- [x] Manual tests

```
Fixed Issues:

* [#3498](https://github.com/ckeditor/ckeditor4/issues/3498): Fixed: The editor threw an error during copy operation when a widget was partially selected.
* [#2517](https://github.com/ckeditor/ckeditor4/issues/2517): Fixed: Selection over [Enhanced Image](https://ckeditor.com/cke4/addon/image2) can result in an error. 
* [#3007](https://github.com/ckeditor/ckeditor4/issues/3007): Fixed: Cannot modify editor's content once a selection is released over a widget.
* [#3698](https://github.com/ckeditor/ckeditor4/issues/3698): Fixed: Cutting selected text and widget merges paragraphs.

API changes:

* [#3842](https://github.com/ckeditor/ckeditor4/issues/3842): Added [`editor.getSelectedRanges`](https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_editor.html#method-getSelectedRanges) alias.
```

## What changes did you make?

* replace `editor.getSelectedHtml()` with custom method which fix situation of partially selected widget on start/end.

Closes: #3498
Closes: #2517.
Closes: #3007.
Closes: #3698.
Closes: #3842


f1ames:
Rebased onto latest `master`.

tests/plugins/widget/copycutintegration.js
```diff
@@ -0,0 +1,70 @@
...
+
+		init: function( editor ) {
+			editor.widgets.add( 'customwidget', {
+				button: 'Add widget',
+				template: '<div>Widget</div>'
+			} );
+		}
+	} );
+
+	bender.editor = {
```
f1ames: Could you add divarea and inline editors here?

tests/plugins/widget/copycutintegration.js
```diff
@@ -0,0 +1,70 @@
...
+
+			range.setStart( editable.findOne( 'p' ).getFirst(), 4 );
+			range.setEnd( editable.findOne( 'div' ), 0 );
+			range.select();
+
+			editable.fire( 'copy', new CKEDITOR.dom.event( {} ) );
+			assert.pass();
+		},
+
+		'test cutting partially selected widget remain collapsed selection in editor': function() {
```
f1ames: I would be for adding one more test, similar to this one but testing more complex widgets (e.g. with nested editable in which selection starts/ends).

f1ames: Shouldn't this test also check what was copied/cut (or maybe you need separate test for that)?

tests/plugins/widget/manual/copycutintegration.html
```diff
@@ -0,0 +1,35 @@
+<div id="editor1">
+	<p>Adipisicing corporis rem repellendus vel mollitia vero? Consectetur dolores voluptatibus illo ipsam eveniet? Lorem ipsum dolor sit amet.</p>
+	<div data-widget="customwidget">Widget</div>
```
f1ames: This widget is not the best fit for this test, since it's hard to select part of it (e.g on Chrome you can select non or entire widget). Could you make it more complex (e.g. with few elements ore editable part) to make sure that partial selection is possible?

plugins/widget/plugin.js
```diff
@@ -3399,7 +3399,7 @@
 		copyBin.handle( getClipboardHtml() );
 
 		function getClipboardHtml() {
-			var selectedHtml = editor.getSelectedHtml( true );
+			var selectedHtml = getSelectedHtmlWithFullWidgets( editor );
```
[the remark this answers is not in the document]
f1ames: Now I started to wonder, that it returns selected content with widgets when they are fully selected only - maybe it should be named `getSelectedHtmlWithCompleteWidgetsOnly`? WDYT?

plugins/widget/plugin.js
```diff
@@ -3423,6 +3423,41 @@
...
+	// Fix might be obsolete after (#3551).
+	function getSelectedHtmlWithFullWidgets( editor ) {
+		var selection = editor.getSelection( true ),
+			ranges = selection.getRanges(),
+			editable = editor.editable(),
+			range,
+			wrapper,
+			outputHtml = '';
+
+		for ( var i = 0; i < ranges.length; i++ ) {
```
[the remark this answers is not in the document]
msamsel: I was inspired by this method when I created this solution. https://github.com/ckeditor/ckeditor4/blob/t/3498/core/editor.js#L705

I think it's not required and in most cases, it will be a single range each time. I'll remove it, as you consider this for loop as increased complexity.

plugins/widget/plugin.js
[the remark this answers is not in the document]
msamsel: I change it to `getHtmlWithFullySelectedWidgets`, WDYT? Or better stick to your proposition?

msamsel:
> but not sure if you already marked this PR as review ready 🤔

Nope I have not :D However, thx for checking it. Most of the issue mentioned in your comment already exists in the editor and those are not a result of my change.

msamsel:
1. Extracted here: #3698

msamsel:
Point 2. is now extracted here: #3699

msamsel:
> Due to pt 3. I'm wondering if it's not better to rathe copy/cut full widget if a selection is inside, instead of preventing copy/cut 🤔 It seems to work this way on IE11. WDYT?

IE11 doesn't look like a good candidate to copy behavior ;P.
I don't think is a good idea, because there is no visual information for the user that the widget is selected. It means that when you start selecting text and release the mouse button over a widget, there is no blue border which would inform, that entire widget is now selected (together with text).
Currently, such cases are poorly supported in the editor. That's why I would stick with removing such widgets from the selection, as those are not marked. Widgets are not always treated as a single entity in the editor and there are multiple hacks which fixing it. I think we should make wider improvements to integrate widgets and selection, however, it is a separate task and it is already reported here: #3551.

msamsel:
I rebased it to current `master`

tests/plugins/widget/copycutintegration.js
msamsel: I tried to achieve that, however, the way how the copybin is working doesn't allow on that.
Single tests work fine, but when there are multiple editors, running one test after another the problematic situation starts to appear. The copybin is destroyed asynchronously and there is no easy way to track it. :(
I feel that still, those tests require improvements, however, I don't see any easy nice way to properly attach to it.

msamsel:
I've come with the idea to fix the problem inside `toDataFormat` event, which runs when data are obtained from the copybin. Previously we don't handle cases when widget could be selected partially. So in the copybin there could be only the widget's wrapper, or there was an element which doesn't have an appropriate widget instance in the editor. Currently, such elements are not processed in the `cut`/`copy`.

I noticed a few edge cases, which seems to be other independent issues.
1. Selecting content from down to widget and from top to widget, works a little bit different. When selection is made from the top, the widget selection observer marks such widget is selected.
When it is done from the bottom, then widget selection observer doesn't recognize such widget is selected.
This result with errors described in point 3 https://github.com/ckeditor/ckeditor4/pull/3553#pullrequestreview-325225473. Selecting from top triggers `copyWidget` logic, selecting from the bottom not.
https://github.com/ckeditor/ckeditor4/blob/135cef730245994c7041ef69103d5d45d44baf88/plugins/widget/plugin.js#L2802-L2804
2. Unit tests are really poor, as there is no easy way to interact with the copybin. Having multiple tests using the same copybin (what happen, for editors with the same `editor.document`), can create a situation where the old copybin from the previous test still exists. What later prevents of creating the new copybin required for a new test case, as there is logic with an early return for such cases:
https://github.com/ckeditor/ckeditor4/blob/135cef730245994c7041ef69103d5d45d44baf88/plugins/widget/plugin.js#L3364-L3366

jacekbogdanski:
As we already have some more time for this issue thus we are no longer targeting it `4.13.1`, I propose to go with more valid fix which would prevent incorrect selection than cleaning up after the bug occurs. I see that selection from the bottom of the widget has not been correctly recognized by copybin because widget repository is unable to tell if widget has been selected, see: https://github.com/ckeditor/ckeditor4/blob/e9c2070c8d3ae68e662159fc83abd965c765b5c1/plugins/widget/plugin.js#L307-L313
range walker here goes only forward, and because the selection is placed at the end of the widget block, it cannot find widget wrapper. Quick prototyping by adding `walker.previous` seems to work, but requires some code to exclude duplicates (if moving both forward and backward). Probably we should be able to reset the walker, so it starts at the beginning of the range instead.

Once widget repository can correctly recognize widgets inside a selection, we should be able to correctly set selection bookmarks here: https://github.com/ckeditor/ckeditor4/blob/e9c2070c8d3ae68e662159fc83abd965c765b5c1/plugins/widget/plugin.js#L3406 which may cause other issues where selection is incorrectly merged on a cut event (not totally sure if it will fix the issue, but at least will match correct selection). 

However, probably whole selection should be fixed before we are copying HTML to the clipboard (so, before bookmarks also) somewhere around `stateUpdater`: https://github.com/ckeditor/ckeditor4/blob/e9c2070c8d3ae68e662159fc83abd965c765b5c1/plugins/widget/plugin.js#L3128-L3166

@msamsel could you take a second look if, without time restriction, we can propose a better solution to prevent incorrect widget selection on chromium?

jacekbogdanski:
I updated the proposed PR with fixing selection before it gets corrupted by copy&paste. The fix is only targeted on non-IE browsers as it seems like IE works correctly.

f1ames:
Rebased onto latest `major`.

f1ames:
Looking at how the fix works (modyfing selection), it seems it may be also fixing other issues caused by invalid selection:

* https://github.com/ckeditor/ckeditor4/issues/2517
* https://github.com/ckeditor/ckeditor4/issues/3007
* https://github.com/ckeditor/ckeditor4/issues/3008
* #3698 and #3699 - both extracted when working on this  PR, I have some doubts here but still worth checking.

I will check the above issue during review.

jacekbogdanski:
To sum up our F2F talk we should mostly focus on fixing Safari if possible. Feature about improved cross-content widget selection has been extracted here https://github.com/ckeditor/ckeditor4/issues/3777. We will skip changes to Edge within this PR mostly due to announced engine change to Chromium which should cover the issue (https://github.com/ckeditor/ckeditor4/issues/3778).

Blinking visible on the https://github.com/ckeditor/ckeditor4/pull/3553#pullrequestreview-341170841 for Firefox is not easily fixable and should be treated as a part of the design.

jacekbogdanski:
As we are not really able to easily find out why the issue on Safari occurs and it takes a lot of debugging time, let's check out how significant Safari regression would be and if it could be safely extracted into a separate issue instead.

f1ames:
@Mgsy could you take a look at this PR and test for regressions? It only introduces changes to Chrome, Firefox and Safari browsers.

The initial issue fixed is #3498, but there are few more covered (see end of the https://github.com/ckeditor/ckeditor4/pull/3553#pullrequestreview-341170841 comment).

I would like to ask you to take a more detailed look on Safari where we introduced minor regression to see if nothing strange happens in other similar cases. The regression is described in https://github.com/ckeditor/ckeditor4/pull/3553#pullrequestreview-341170841 comment (under "Safari" section).

You can ping me or @jacekbogdanski if you have any questions.

FilipTokarski:
Chrome and FF seem to be fine, apart from what's mentioned here: https://github.com/ckeditor/ckeditor4/pull/3553#pullrequestreview-341170841, I can't find any serious bugs.

f1ames:
Rebased onto latest `major`.

f1ames:
Checked #3041 and it's not fixed by this PR :cry:

jacekbogdanski:
Seems like  there is a small error: 
image

#3498 shouldn't be replaced with #3683. I already fixed the issue but please confirm if it's fine now.

jacekbogdanski:
> Since this is not the part of the initial fix, probably we should skip this test and leave the issue open. It looks like it needs more attention. WDYT?

:+1: removed from changelog proposal/closes

> The tests/plugins/image2/manual/crosscontentselection test fails in Safari due to regression we have introduced:

Yes but you can still follow test steps to check if a selection is inside the editor. I updated the test steps to make it more clear.

f1ames:
Rebased onto latest `major`.

Issue #1653, opened 2018-02-12
Balloontolbar in divarea doesn't react on scrollbar movement


msamsel:
## Are you reporting a feature request or a bug?

bug

## Provide detailed reproduction steps (if any)

1. Open test: `/tests/plugins/easyimage/manual/customstyleicons`
2. Click into image to show balloontoolbar
3. Move scrollbar of editable in divarea editor
![screen shot 2018-02-12 at 14 28 55](https://user-images.githubusercontent.com/9017322/36099113-c9812580-1001-11e8-8369-2789fed1f239.png)

### Expected result

ballontoolabr should be repositioned simmilar to classic editor

### Actual result

ballontolbar is not moved

## Other details

* Browser: Safari, Chrome
* OS: MacOS
* CKEditor version: 4.9.0
* Installed CKEditor plugins: full-all

jacekbogdanski:
Closed via #3562

PR #3562, opened 2019-10-09
Fix missing reposition of balloon toolbar after scrolling the editable

msamsel:

Bug fix

### This PR contains

- [x] Unit tests
- [x] Manual tests

```
* [#1653](https://github.com/ckeditor/ckeditor4/issues/1653) Fix: Missing reposition of [balloon toolbar](https://ckeditor.com/cke4/addon/balloontoolbar) when the editable was scrolled in the [divarea](https://ckeditor.com/cke4/addon/divarea) editor.
```

## What changes did you make?

Improve listener which now detects elements with scrollbar holding the editable and attach scroll listener to it.

Closes: #1653

### DIFF
--- a/plugins/balloontoolbar/plugin.js
+++ b/plugins/balloontoolbar/plugin.js
@@ -697,9 +697,8 @@
 				CKEDITOR.ui.balloonPanel.prototype.show.call( this );
 
 				// It's necessary to execute code in a closure, as reposition is defined in a prototype.
-				// Direct assign `this.reposition` to an evenListener might result with an error when multiple editors are on a page.
-				// There might be case where `this._detachListeners`, will be executed from 2 editors. So listener will be detached by first one,
-				// but second editor will try to remove it once again what will throw an error, becasue it would have reference to the same reposition method.
+				// Otherwise hiding balloon toolbar may remove event listeners from different editors,
+				// as removing listeners are done by function delegate.
 				function repositionClosure() {
 					this.reposition();
 				}
--- a/plugins/balloontoolbar/plugin.js
+++ b/plugins/balloontoolbar/plugin.js
@@ -688,13 +688,21 @@
 
 				this._detachListeners();
 
-				this._listeners.push( editor.on( 'change', this.reposition, this ) );
-				this._listeners.push( editor.on( 'resize', this.reposition, this ) );
-				this._listeners.push( win.on( 'resize', this.reposition, this ) );
-				this._listeners.push( win.on( 'scroll', this.reposition, this ) );
-				this._listeners.push( editorScrollableElement.on( 'scroll', this.reposition, this ) );
+				this._listeners.push( editor.on( 'change', repositionClosure, this ) );
+				this._listeners.push( editor.on( 'resize', repositionClosure, this ) );
+				this._listeners.push( win.on( 'resize', repositionClosure, this ) );
+				this._listeners.push( win.on( 'scroll', repositionClosure, this ) );
+				this._listeners.push( editorScrollableElement.on( 'scroll', repositionClosure, this ) );
 
 				CKEDITOR.ui.balloonPanel.prototype.show.call( this );
+
+				// It's necessary to execute code in a closure, as reposition is defined in a prototype.
+				// Direct assign `this.reposition` to an evenListener might result with an error when multiple editors are on a page.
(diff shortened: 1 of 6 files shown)


plugins/balloontoolbar/plugin.js
```diff
@@ -675,19 +675,32 @@
...
+				this._listeners.push( win.on( 'resize', attachListener, this ) );
+				this._listeners.push( win.on( 'scroll', attachListener, this ) );
+
+				this._listeners.push( editorScrollableElement.on( 'scroll', attachListener, this ) );
 
 				CKEDITOR.ui.balloonPanel.prototype.show.call( this );
+
+				function attachListener() {
+					this.reposition();
+				}
```
jacekbogdanski: Could we skip this middleman function and pass `this.reposition` directly? E.g.

```javascript
this._listeners.push( editor.on( 'change', this.reposition, this ) );
```

`this` object should be correctly resolved here, wdyt?

tests/plugins/balloontoolbar/manual/divareascroll.html
```diff
@@ -0,0 +1,142 @@
+<style>
+	body {
+		height: 5000px;
+	}
+	.top-margin {
+		margin-top: 500px;
+	}
+</style>
+<h2 class="top-margin">Divarea:</h2>
+<div contenteditable="true" id="editor1" class="editor">
```
jacekbogdanski: I really like the simplicity of this manual test. I'm only wondering if we could avoid paragraph duplication by using proper CSS for that, something which has been done for autocomplete tests: https://github.com/ckeditor/ckeditor4/blob/master/tests/plugins/autocomplete/manual/viewposition.html

tests/plugins/balloontoolbar/manual/divareascroll.md
```diff
@@ -0,0 +1,15 @@
...
+@bender-tags: 4.13.1, bug, 1653, balloontoolbar
+@bender-ckeditor-plugins: wysiwygarea,toolbar,balloontoolbar,link
+
+## Please notice that both editable and entire document should have a scrollbar.
+
+If editors are not visible in the test please scroll down the page.
+
+1. Put selection inside a link to show balloontolbar.
+2. Scroll editable.
+3. You can use other links and scroll document to test different variation.
```
jacekbogdanski:
```suggestion
3. Use other links and scroll document to test different variations.
```

Let's make it obligatory :smile:

tests/plugins/balloontoolbar/positioning.js
```diff
@@ -229,6 +229,181 @@
 
 				assert.isTrue( markerElement.equals( spy.args[ 0 ][ 0 ] ) );
 			} );
+		},
+
+		// #1653
```
jacekbogdanski: Please, update with `(#1653)` to our adopted code style.

tests/plugins/balloontoolbar/positioning.js
```diff
@@ -229,6 +229,181 @@
...
+							sinon.assert.called( spy );
+							assert.pass();
+						} );
+					} );
+				}, null, null, 100000 );
+
+				editableScrollElement.fire( 'scroll' );
+
+				wait();
+			} );
```
jacekbogdanski: Tests work well :+1: . Although, there is a lot of code duplication which hurts maintainability. Could you extract common code to helper functions?

msamsel:
> when you are scrolling the browser with divarea editor, where balloon toolbar sticks to the browser window instead of scrolling with a document.

I extracted the case to separate issue: #3632. It seems that `balloonpanel` is a place where fix should be applied.

tests/plugins/balloontoolbar/positioning.js
msamsel: I feel like we don't have defined code style for it, as there is plenty of examples of both approaches with and without parentheses. However, I don't mind to change it.

msamsel:
Rebased on current master.

plugins/balloontoolbar/plugin.js
msamsel: I had to revert this change. `this.reposition` is defined in a prototype. So when there are 2 editors on a page, there might appear a situation that both of them will call `this.detachListener`. This will result with attempt to remove twice the same listener. To avoid it, it's necessary to preserve the context for a given listener, to avoid the situation where a different instance of balloon toolbar removes such listener.

PR #3643, opened 2019-11-12
Fix failing tests /tests/core/editable/misc

### LINKED ISSUE #3633: Failing test - tests/core/editable/misc (since FF 70)
## Type of report

Failing test

1. Run test `tests/core/editable/misc`.

### Expected result

Test passes.

### Actual result

3rd and 4th test cases fail:

Screenshot 2019-11-07 at 16 04 30

## Other details

The issue was introduced in FF 70.

* Browser: Firefox 70
* OS: macOS
* CKEditor version: `4.13.0`

Dumluregn:

Bug fix

### This PR contains

- [x] Unit tests
- [ ] Manual tests

None

## What changes did you make?

If FF 70 or later is detected, test will trim the result before comparing with expected to get rid of trailing `\n`.

Closes #3633.

### DIFF
--- a/tests/core/editable/misc.js
+++ b/tests/core/editable/misc.js
@@ -112,11 +112,7 @@ bender.test( {
 			} ) );
 
 			// From Firefox 70 it returns trailing '\n' (#3633).
-			if ( CKEDITOR.env.gecko && CKEDITOR.env.version >= 700000 ) {
-				assert.areSame( 'foo', editor.getSelection().getSelectedText().trim(), 'Selection has not been changed' );
-			} else {
-				assert.areSame( 'foo', editor.getSelection().getSelectedText(), 'Selection has not been changed' );
-			}
+			assert.areSame( 'foo', editor.getSelection().getSelectedText(), 'Selection has not been changed' );
 		} );
 	},
 
--- a/tests/core/editable/misc.js
+++ b/tests/core/editable/misc.js
@@ -112,7 +112,7 @@ bender.test( {
 			} ) );
 
 			// From Firefox 70 it returns trailing '\n' (#3633).
-			assert.areSame( 'foo', editor.getSelection().getSelectedText(), 'Selection has not been changed' );
+			assert.areSame( 'foo', editor.getSelection().getSelectedText().trim(), 'Selection has not been changed' );
 		} );
 	},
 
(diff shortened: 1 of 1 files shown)


tests/core/editable/misc.js
```diff
@@ -92,7 +92,12 @@ bender.test( {
 				target: img.$
 			} ) );
 
-			assert.areSame( 'foo', editor.getSelection().getSelectedText(), 'Selection has not been changed' );
+			// From Firefox 70 it returns trailing '\n' (#3633).
+			if ( CKEDITOR.env.gecko && CKEDITOR.env.version >= 700000 ) {
```
jacekbogdanski: I'm wondering if we actually need a browser sniffing here? Won't this test still be valid if trimming will be done by default simplifying code into a single execution path?

Also, as we declare browser support for the **latest** stable Firefox (see https://ckeditor.com/docs/ckeditor4/latest/guide/dev_browsers.html#officially-supported-browsers), probably there is no need for version check  - although I would rather remove browser sniffing entirely here if possible.

Dumluregn: Well, yes - modified version will pass on each browser. On one hand I totally agree that having one consistent test form would be prettier and more readable. But on the other one generalising test when it's not really necessary moves the it away from the original thing that test was about to check. And it may also cause us miss some changes (maybe more important than that one 😝) in other browsers in the future. But... I see that the comment is kind of doubling the information in code.

So, to sum up - if you are not convinced by the arguments for sniffing, I'll be glad with the look of the code without it too as I see pros and cons of both.

jacekbogdanski: Generally speaking, I agree that making a test less strict may result in some positive-failing tests. But take a look at what actually this single test does - it checks whatever image or placeholder (`foo`) text is selected - it doesn't actually care about text format - the only important case here is if an image has not been selected.

So, this test fails, if `editor.getSelection().getSelectedText()` would return an empty string instead of `foo` (because selecting image results with an empty string, you can check it yourself), so trimming value could only improve the test.

You are making a good point about generalizing test cases, but in the best scenario unit test checks only a single case and this one seems to do it fine.

Also, note that you actually trimming value for Firefox, so you are already generalizing this test case for this one browser :smile:

Dumluregn:
Ok, so I removed browser sniffing and rebased onto latest master.

PR #3663, opened 2019-11-18
Fix focus inside widget for textarea and input elements

### LINKED ISSUE #3587: Widget gets focus when typing with form input elements
## Type of report

Bug

The issue is caused by the https://github.com/ckeditor/ckeditor4/blob/ad381679b4436f489a65f6f397680da338024b0f/core/selection.js#L942-L954f code which is supposed to prevent editing non-editable elements on Edge and IE. However, it also makes impossible editing correct form elements like textarea, input, etc.

You may have some issue reproducing test steps on IE as it doesn't fire `keydown` event when typing inside form elements, however, this code path is wrong for both Edge and IEs.

1. Open https://codepen.io/jacekbogdanski/pen/vYYXamd
2. Add test widget (no-icon button)
3. Focus the widget and start typing.

### Expected result

Typed text is registered by widgets textarea.

### Actual result

Widget gets focus on every keystroke which makes typing impossible.

## Other details

* Browser: Edge, IE
* OS: Windows
* CKEditor version:  at least 4.11.0

msamsel:

Bugfix

### This PR contains

- [x] Unit tests
- [x] Manual tests

```
* [#3587](https://github.com/ckeditor/ckeditor4/issues/3587): [Edge, IE] Fix: textarea and input elements inside widget are not possible to focus with a mousedown.
```

## What changes did you make?

Prevent to run logic responsible for the entire widget selection on IE and Edge, when the target element is a textarea or an input inside a widget.

Closes #3587.

### DIFF
--- a/core/selection.js
+++ b/core/selection.js
@@ -941,30 +941,7 @@
 
 			if ( CKEDITOR.env.ie ) {
 				// https://dev.ckeditor.com/ticket/14407 - Don't even let anything happen if the selection is in a non-editable element.
-				editable.attachListener( editable, 'keydown', function( evt ) {
-					var sel = this.getSelection( 1 ),
-						ascendant = getNonEditableAscendant( sel ),
-						activeElement,
-						isValidActiveElement;
-
-					if ( !ascendant ) {
-						return;
-					}
-
-					activeElement = this.document.getActive(),
-					isValidActiveElement = activeElement && ( activeElement.getName() === 'input' || activeElement.getName() === 'textarea' );
-
-					// Allow on typing inside editable elements of widgets if those are currently focused (#3587)
-					if ( isValidActiveElement && ascendant.contains( activeElement ) ) {
-						return;
-					}
-
-					// Prevent changing selection when an ascendant is an entire editable (#1632).
-					if ( !ascendant.equals( editable ) ) {
-						sel.selectElement( ascendant );
-						evt.data.preventDefault();
-					}
-				}, editor );
+				editable.attachListener( editable, 'keydown', disableSelectionChangeForNonEditables, editor );
 			}
 
 			// Always fire the selection change on focus gain.
@@ -1051,6 +1028,35 @@
 				// The parentElement may be null for read only mode in IE10 and below (https://dev.ckeditor.com/ticket/9780).
 				if ( sel.type != 'None' && range.parentElement() && range.parentElement().ownerDocument == doc.$ )
 					range.select();
+			}
+
+			function disableSelectionChangeForNonEditables( evt ) {
+				var sel,
+					ascendant;
+
+				// Allow on typing inside editable elements of widgets if those are currently focused (#3587)
+				if ( isTypingElement( this.document.getActive() ) ) {
+					return;
+				}
+
+				sel = this.getSelection( 1 );
+				ascendant = getNonEditableAscendant( sel );
+
+				// Prevent changing selection when an ascendant is an entire editable (#1632).
+				if ( ascendant && !ascendant.equals( editable ) ) {
+					sel.selectElement( ascendant );
+					evt.data.preventDefault();
+				}
+			}
+
+			function isTypingElement( activeElement ) {
+				if ( !activeElement ) {
+					return false;
+				}
+
+				return activeElement.getName() === 'input' ||
+					activeElement.getName() === 'textarea' ||
+					activeElement.getAttribute( 'contenteditable' ) === 'true';
 			}
 
 			function getNonEditableAscendant( sel ) {
--- a/core/selection.js
+++ b/core/selection.js
@@ -943,10 +943,24 @@
 				// https://dev.ckeditor.com/ticket/14407 - Don't even let anything happen if the selection is in a non-editable element.
 				editable.attachListener( editable, 'keydown', function( evt ) {
 					var sel = this.getSelection( 1 ),
-						ascendant = getNonEditableAscendant( sel );
+						ascendant = getNonEditableAscendant( sel ),
+						activeElement,
+						isValidActiveElement;
+
+					if ( !ascendant ) {
+						return;
+					}
+
+					activeElement = this.document.getActive(),
+					isValidActiveElement = activeElement && ( activeElement.getName() === 'input' || activeElement.getName() === 'textarea' );
+
+					// Allow on typing inside editable elements of widgets if those are currently focused (#3587)
+					if ( isValidActiveElement && ascendant.contains( activeElement ) ) {
+						return;
+					}
 
 					// Prevent changing selection when an ascendant is an entire editable (#1632).
-					if ( ascendant && !ascendant.equals( editable ) ) {
+					if ( !ascendant.equals( editable ) ) {
 						sel.selectElement( ascendant );
 						evt.data.preventDefault();
 					}
--- a/core/selection.js
+++ b/core/selection.js
@@ -1055,8 +1055,7 @@
 				}
 
 				return activeElement.getName() === 'input' ||
-					activeElement.getName() === 'textarea' ||
-					activeElement.getAttribute( 'contenteditable' ) === 'true';
+					activeElement.getName() === 'textarea';
 			}
 
 			function getNonEditableAscendant( sel ) {
--- a/tests/plugins/widget/integration/selection/inputselection.js
+++ b/tests/plugins/widget/integration/selection/inputselection.js
@@ -52,7 +52,7 @@
 
 	bender.test( {
 		'test textarea inside widget should allow on typing': function() {
-			if ( !CKEDITOR.env.ie || CKEDITOR.env.version < 11 ) {
+			if ( !CKEDITOR.env.ie || CKEDITOR.env.version < 9 ) {
 				assert.ignore();
 			}
 
@@ -87,7 +87,7 @@
 		},
 
 		'test input inside widget should allow on typing': function() {
-			if ( !CKEDITOR.env.ie || CKEDITOR.env.version < 11 ) {
+			if ( !CKEDITOR.env.ie || CKEDITOR.env.version < 9 ) {
 				assert.ignore();
 			}
 
--- a/tests/plugins/widget/integration/selection/inputselection.js
+++ b/tests/plugins/widget/integration/selection/inputselection.js
@@ -1,8 +1,6 @@
 /* bender-tags: widgetcore */
 /* bender-ckeditor-plugins: widget,undo */
 /* bender-ckeditor-remove-plugins: tableselection */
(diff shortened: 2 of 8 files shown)


tests/plugins/widget/integration/selection/manual/ieinputwidget.html
```diff
@@ -0,0 +1,43 @@
...
+					init: function() {
+						this.on( 'focus', function() {
+							var input = this.element.findOne( 'input' ),
+								range = this.editor.createRange();
+
+							input.focus( 1 );
+
+							range.selectNodeContents( input );
+							range.collapse();
+							range.select();
```
jacekbogdanski: That's crazy we need so many selection manipulations for simple input focus. But I suppose it's different issue where focus doesn't always mean selection change on IE?

tests/plugins/widget/integration/selection/manual/ieinputwidget.html
```diff
@@ -0,0 +1,43 @@
+<div id="editor">
+		<p>Add widget below.</p>
+	</div>
+
+	<script>
+		if ( !CKEDITOR.env.ie || CKEDITOR.env.version < 11) {
```
jacekbogdanski: The changed code path touches all IE browsers (including Edge). Is there a reason why this test is enabled only for IE11 and Edge?

tests/plugins/widget/integration/selection/inputselection.js
```diff
@@ -0,0 +1,123 @@
...
+				}, null, null, 100000 );
+
+				editable.fire( 'keydown', new CKEDITOR.dom.event( {
+					keyCode: 66 // B
+				} ) );
+			} );
+		},
+
+		'test input inside widget should allow on typing': function() {
+			if ( !CKEDITOR.env.ie || CKEDITOR.env.version < 11 ) {
```
jacekbogdanski: The same about IE support.

tests/plugins/widget/integration/selection/inputselection.js
```diff
@@ -0,0 +1,123 @@
...
+						}
+					} );
+				}
+			}
+		}
+	};
+
+	bender.test( {
+		'test textarea inside widget should allow on typing': function() {
+			if ( !CKEDITOR.env.ie || CKEDITOR.env.version < 11 ) {
```
jacekbogdanski: Looks like the test should be enabled for at least all IE browsers https://github.com/ckeditor/ckeditor4/blob/0c34fecc92e8183d2a15c17237e0fde6730321b3/core/selection.js#L942

on the other hand, AFAIK we don't have any tests confirming that this case will work on other browsers (like Chrome, FF). Maybe it makes sense to ignore browser sniffing to also confirm input elements everywhere?

tests/plugins/widget/integration/selection/inputselection.js
```diff
@@ -0,0 +1,123 @@
...
+					}
+				}, null, null, 100000 );
+
+				editable.fire( 'keydown', new CKEDITOR.dom.event( {
+					keyCode: 66 // B
+				} ) );
+			} );
+		},
+
+		'test input inside widget should allow on typing': function() {
```
jacekbogdanski: Both tests share very common logic. Could you extract it into a helper method to avoid duplication?

tests/plugins/widget/integration/selection/inputselection.js
```diff
@@ -0,0 +1,123 @@
...
+							return element.name === 'div' && element.hasClass( 'testwidget' );
+						},
+						init: function() {
+							this.on( 'focus', function() {
+								this.element.findOne( 'textarea' ).focus();
+							} );
+						}
+					} );
+
+					editor.widgets.add( 'testwidget2', {
```
jacekbogdanski: As the widgets are very similar, maybe it makes sense to extract them into the common builder function?

core/selection.js
```diff
@@ -943,10 +943,24 @@
...
+						return;
+					}
+
+					activeElement = this.document.getActive(),
+					isValidActiveElement = activeElement && ( activeElement.getName() === 'input' || activeElement.getName() === 'textarea' );
+
+					// Allow on typing inside editable elements of widgets if those are currently focused (#3587)
+					if ( isValidActiveElement && ascendant.contains( activeElement ) ) {
+						return;
+					}
```
jacekbogdanski: Could we improve code readability here by keeping a single abstraction level? I suggest extracting this code into helper function, something like:  `isTypingElement` - function checking if element is of `input`, `textarea` or `contenteditable=true` type.

Also, it seems like in that case, checking for `nonEditableAscendant` could be moved lower, as if an active element is directly editable, there is no need to traverse through parents.

tests/plugins/widget/integration/selection/inputselection.js
```diff
@@ -0,0 +1,123 @@
...
+				editable.once( 'keydown', function() {
+					var selection = editor.getSelection();
+					// Edge expands selection to inner div. IE keeps selction inside textArea.
+					// Important is to not have focused widget's wrapper.
+					if ( CKEDITOR.env.edge ) {
+						assert.isTrue( selection.getStartElement().equals( widget.element ) );
+					} else {
+						assert.isTrue( selection.getStartElement().equals( textArea ) );
+					}
+				}, null, null, 100000 );
```
jacekbogdanski: Is so high priority important here? `keydown` listener inside `selection` module doesn't manipulate priority, so it should work fine without setting priority explicitly.

core/selection.js
```diff
@@ -943,10 +943,24 @@
 				// https://dev.ckeditor.com/ticket/14407 - Don't even let anything happen if the selection is in a non-editable element.
 				editable.attachListener( editable, 'keydown', function( evt ) {
```
jacekbogdanski: Extracting the whole function into named private function (like it has been done for above listeners) should improve readability.

tests/plugins/widget/integration/selection/inputselection.js
jacekbogdanski: Also, this test passes despite undoing `selection` module changes on IE11.

tests/plugins/widget/integration/selection/manual/ieinputwidget.html
```diff
@@ -0,0 +1,43 @@
...
+		};
+
+		CKEDITOR.plugins.add( 'testplugin', {
+			requires: 'widget',
+			init: function( editor ) {
+				editor.widgets.add( 'testwidget', {
+					requiredContent: 'div(testwidget)',
+					allowedContent: 'div(testwidget)',
+					template: '<div class="testwidget">' +
+						'<input type="text">' +
```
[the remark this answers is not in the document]
msamsel: `textarea` elements have a seaprate test file: https://github.com/ckeditor/ckeditor4/pull/3663/files#diff-f985a7ca0e14601902fcf80a889e7f88

tests/plugins/widget/integration/selection/manual/ieinputwidget.html
msamsel: Those are 2 main branches of browser which we actively support, so I haven't test it on lower IE versions.
However, I've checked and everything seems to work up to IE9 with manual and unit tests. IE8 has some troubles with this fix. It doesn't sound reasonable to spend too much time on preparing dedicated for IE8 anyway. I think we can remain it as it is, and if there appear requests for supporting such cases under IE8, then fix it.

tests/plugins/widget/integration/selection/inputselection.js
msamsel: https://github.com/ckeditor/ckeditor4/pull/3663#discussion_r358143995

tests/plugins/widget/integration/selection/inputselection.js
msamsel: > AFAIK we don't have any tests confirming that this case will work on other browsers (like Chrome, FF). Maybe it makes sense to ignore browser sniffing to also confirm input elements everywhere?

Unfortunately, these tests fail on FF or Chrome. Also, the entire block responsible for the bug was present in the section dedicated to IE:
https://github.com/ckeditor/ckeditor4/blob/cc1939111a2cddfec9019b9105dbde2a84372a1e/core/selection.js#L942-L954

From that reason is really unlikely that bug could appear on non-ie browsers. That's why I feel like having manual tests for other browsers would be sufficient in these cases.
Please also remember that we should immediately know that something is wrong on Chrome or FireFox, because the spreadsheet plugin would fail on those browsers.

tests/plugins/widget/integration/selection/inputselection.js
msamsel: > Also, this test passes despite undoing selection module changes on IE11.

Because there is no issue with input element under IE11, only textarea has an issue.

tests/plugins/widget/integration/selection/inputselection.js
msamsel: I can lower it, however, I want to guarantee that the listener will be fired after all the events related to 'keydown' will be executed inside the editor. That's why it has so high value, to be executed as a last one.

jacekbogdanski:
I accidentally have taken over wrong PR :disappointed: sorry for that, the previous review is still valid.

tests/plugins/widget/integration/selection/manual/ieinputwidget.html
msamsel: to have somehow working input and text area elements inside IE and Edge, we need to make such magic. :(

tests/plugins/widget/integration/selection/manual/ieinputwidget.html
```diff
@@ -0,0 +1,43 @@
...
+							range.selectNodeContents( input );
+							range.collapse();
+							range.select();
+						} );
+					}
+				} );
+			}
+		} );
+
+		window.editor = CKEDITOR.replace( 'editor', {
```
jacekbogdanski: What's the reason for exposing the editor globally?

tests/plugins/widget/integration/selection/manual/ietextareawidget.html
```diff
@@ -0,0 +1,43 @@
...
+						range.selectNodeContents( txtArea );
+						range.collapse();
+						range.select();
+					} );
+				}
+			} );
+		}
+	} );
+
+	window.editor = CKEDITOR.replace( 'editor', {
```
jacekbogdanski: What's the reason for exposing the editor globally?

tests/plugins/widget/integration/selection/manual/ieinputwidget.html
msamsel: Probably remain when I updated the tests.

PR #3672, opened 2019-11-20
Making print plugin reflecting content rendered by preview plugin

### LINKED ISSUE #3661: Printed content is different than the content in print preview
## Type of report

Feature request

1. Go to https://ckeditor.com/docs/ckeditor4/latest/examples/fullpreset.html.
2. Insert page break in the content.
3. Click editors print button.

### Expected result

Page break creates new page / page separation in printable document.

### Actual result

Page break creates new page / page separation in printable document. Page break element is visible in the document on the preceding page.

image

## Other details

* Browser: Tested on Chrome.
* OS: Linux
* CKEditor version: 4.13.0 (4.10.0)
* Installed CKEditor plugins: `pagebreak,print`

Comandeer:

New feature

### This PR contains

- [x] Unit tests
- [x] Manual tests

```
*[#3661](https://github.com/ckeditor/ckeditor4/issues/3661): [Print](https://ckeditor.com/cke4/addon/print) plugin respects styling rendered by [Preview](https://ckeditor.com/cke4/addon/preview) plugin.
```

## What changes did you make?

I've make `preview` plugin a dependency of `print` plugin. The whole logic of creating the preview was extracted to `CKEDITOR.plugins.preview#createPreview()` and is reused inside `print`. Printing is now based on creating preview window and immediately printing it and closing. I've also fixed some issues with preview & printing in different browsers:

* in Safari I fixed the issue with not rendering the preview if external styles aren't inside the cache;
* in Firefox I fixed the issue with creating incorrect paths to images.

Basing the new implementation on `preview` plugin allowed also to enable `print` inside inline editors (as the printed document is in its own window now).

I've also added additional styles to show pagebreak element in the preview, but hide it in the printed content.

I've also make a big refactor of the `preview` plugin. Code should be much more readable now.

Closes #3661.


plugins/preview/plugin.js
```diff
@@ -8,128 +8,262 @@
...
+				// For IE we should use window.location rather than setting url in window.open (https://dev.ckeditor.com/ticket/11146).
+				previewLocation = getPreviewLocation(),
+				previewUrl = getPreviewUrl(),
+				previewWindow,
+				doc;
 
 			// (https://dev.ckeditor.com/ticket/9907) Allow data manipulation before preview is displayed.
 			// Also don't open the preview window when event cancelled.
-			if ( editor.fire( 'contentPreview', eventData = { dataValue: sHTML } ) === false )
+			if ( editor.fire( 'contentPreview', eventData ) === false ) {
```
jacekbogdanski: I'm wondering how this case can be reflected in the print plugin. Previously, the print plugin didn't depend on the preview plugin and has been printing current editor HTML. So, if someone highly customizes `preview` with `contentPreview` event, he will get totally different HTML after these changes. Let's say, someone has been adding custom headers to the preview HTML but assumes that print plugin is a separate feature that shouldn't depend on this event. I'm wondering if we could exclude additional, `contentPrint` event for that case.

On the other hand, `contentPreview` is recommended to use to prevent XSS (https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR.html#event-contentPreview), so if we stop executing this function, it may lead to security issues with the print command which is even the worse scenario.

Maybe, at least, we could pass some information if it's a preview or print feature by passing some flag in `contentPreview` event, so someone can decide what to do? Could be done by listening on `contentPreview` with high priority in print plugin to avoid dome dependency issues, and adding some additional flag or enum. WDYT?

plugins/preview/plugin.js
```diff
@@ -8,128 +8,262 @@
...
+
+	/**
+	 * Namespace providing a set of helper functions for working with print preview, exposed by the
+	 * [Preview](https://ckeditor.com/cke4/addon/tableselection) plugin.
+	 *
+	 * @since 4.14.0
+	 * @singleton
+	 * @class CKEDITOR.plugins.preview
+	 */
+	CKEDITOR.plugins.preview = {
```
jacekbogdanski: Refactoring nicely simplified this module. I'm only wondering if we could go a bit further, and use here our good old `CKEDITOR.tools.createClass` for even better readability (and start using OOP finally in our plugins). I already did something like that during working on quick table plugin and it seems to improve a bit implementation:
https://github.com/ckeditor/ckeditor4/blob/c23ca76d0cf074e713094a074274f2211cb931bb/plugins/table/plugin.js#L166
https://github.com/ckeditor/ckeditor4/blob/c23ca76d0cf074e713094a074274f2211cb931bb/plugins/quicktable/plugin.js#L26

It would be nice to start using a single adopted method for modules. WDYT?

plugins/preview/plugin.js
```diff
@@ -8,128 +8,262 @@
...
+			};
+		} catch ( e ) {
+			return {
+				width: 640,
+				// 800 * 0.8,
+				height: 420,
+				// 600 * 0.7,
+				left: 80 // (800 - 0.8 * 800) /2 = 800 * 0.1
+			};
+		}
```
jacekbogdanski: What's the reason for try..catch here? I suppose some `Permission Denied` errors on IE?

plugins/preview/plugin.js
```diff
@@ -8,128 +8,262 @@
...
 				doc.close();
 			}
 
-			return true;
+			return new CKEDITOR.dom.window( previewWindow );
 		}
 	};
 
-	var pluginName = 'preview';
+	previewCmd = {
```
jacekbogdanski: Defining the command definition separately doesn't seem important. Could it be inlined when adding command in `plugin#init` method?

tests/plugins/print/print.js
```diff
@@ -0,0 +1,68 @@
...
+			resume( function() {
+				createPreviewStub.restore();
+
+				assert.pass();
+			} );
+		} );
+
+		editor.execCommand( 'print' );
+		wait();
+	},
```
jacekbogdanski: Now I see why this test passes, as implementation checks against `false.$` which returns `undefined`. Although it's a bit hacky as it's rather expected that such code should throw an error as boolean is a primitive type. It seems a bit like leveraging language bad design which makes code less readable. See https://github.com/ckeditor/ckeditor4/pull/3672#discussion_r354867438

plugins/print/plugin.js
```diff
@@ -6,41 +6,54 @@
...
+				previewWindow.print();
+			} else {
+				previewWindow.document.execCommand( 'Print' );
+			}
+
+			previewWindow.close();
+		},
+		canUndo: false,
+		readOnly: 1,
+		modes: { wysiwyg: 1 }
```
jacekbogdanski: Mode for `print` is more specific, as `preview` seems to work also in `sourcearea` - which actually seems a bit weird as it shows up editor data, no source code. I'm for using the same modes for both plugins, but it seems like `preview` plugin mode should be decreased to `wysiwyg` only, so it doesn't close us for `sourcearea` preview/print feature. WDYT?

plugins/preview/plugin.js
```diff
@@ -8,128 +8,262 @@
...
+			'resizable=yes',
+			'width=' + dimensions.width,
+			'height=' + dimensions.height,
+			'left=' + dimensions.left
+		].join( ',' );
+	}
+
+	function getPreviewLocation() {
+		if ( !CKEDITOR.env.ie ) {
+			return;
```
jacekbogdanski: As the function returns some value, it would be nice to return `null` here instead of implicit `undefied` for readability.

tests/plugins/print/print.js
Comandeer: TBH it was totally unintended and I didn't plan to use `false.$`.

At the same time this unit test doesn't seem to check anything sensible 🤔 it probably should check if print native command was invoked.

plugins/preview/plugin.js
Comandeer: I don't understand what problem does the new event fix? It would make the whole implementation much more convoluted and lead to producing two versions of content – one for the preview and one for printing. So we would be basically back to the starting point.

plugins/preview/plugin.js
Comandeer: Nearly every plugin uses namespaces to just store some helper functions, often unrelated with each other, e.g. https://github.com/ckeditor/ckeditor4/blob/37e9cb17403336a06f1065d83be376142fd5cb9d/plugins/tableselection/plugin.js#L941-L958. Making it a class, which members would basically share only `editor` property, doesn't sound like OOP to me.

OTOH these namespaces exist only to allow better testing or ability to override some parts of plugins from other plugins (as it was done in Table Selection, which reuses and overrides some parts of Table Tools). They do not serve any other purpose.

plugins/preview/plugin.js
Comandeer: It was in such form from the beginning, I didn't refactor it. I'll do some research to check if it works without the `try`/`catch`.

plugins/print/plugin.js
```diff
@@ -6,41 +6,54 @@
...
-		// Register the command.
-		editor.addCommand( pluginName, CKEDITOR.plugins.print );
+	/**
+	 * Allows to customise implementation of printing the editor's content,
+	 * provided by [Print](https://ckeditor.com/cke4/addon/print) plugin.
+	 * The provided value should be of {@link CKEDITOR.commandDefinition} type.
+	 *
+	 * @singleton
+	 * @class CKEDITOR.plugins.print
+	 */
```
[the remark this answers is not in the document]
Comandeer: Yes, I tried inheriting, but it produced incorrect result.

Comandeer:
> I'm getting some docs builder errors

It's an error in our docs. I'll prepare PR later.

plugins/preview/plugin.js
jacekbogdanski: > Maybe, at least, we could pass some information if it's a preview or print feature by passing some flag in contentPreview event, so someone can decide what to do? Could be done by listening on contentPreview with high priority in print plugin to avoid dependency issues, and adding some additional flag or enum. WDYT?

Someone could add some fancy HTML to the preview page (like a logo) which shouldn't be visible on the print page. Additional information in the event object could slightly minimalize breaking changes in that case, so it's still possible to customize preview content but not print (and vice versa).

plugins/preview/plugin.js
jacekbogdanski: > Making it a class, which members would basically share only editor property, doesn't sound like OOP to me.

OOP is not only about sharing some common properties or state, but also encapsulate domain logic in unit boundaries. Although:

>OTOH these namespaces exist only to allow better testing or ability to override some parts of plugins from other plugins (as it was done in Table Selection, which reuses and overrides some parts of Table Tools). They do not serve any other purpose.

Let's keep it simple then:+1: and worry if it ever grows.

plugins/print/plugin.js
```diff
@@ -6,41 +6,79 @@
...
+			}
+
+			nativePreviewWindow = previewWindow.$;
+
+			// In several browsers (e.g. Safari or Chrome on Linux) print command
+			// seems to be blocking loading of the preview page. Because of that
+			// print must be performed after the document is complete.
+			if ( nativePreviewWindow.document.readyState === 'complete' ) {
+				return print();
+			}
```
jacekbogdanski: This code seems to produce the issue with missing images on Chrome. Removing it executes print command once images have been loaded. That's weird because according to docs `readyState === 'complate'` should be called once images are loaded, so it's seems like a bug of Linux Chrome. Nevertheless, it's also set **before** `load` event (see https://developer.mozilla.org/en-US/docs/Web/API/Document/readyState) which invalidates fix for images.

If I remember correctly, `load` event is incorrectly propagated on Safari? Maybe we just need to print preview immediately on Safari? It seems to work correctly without waiting for load event.

tests/plugins/print/manual/print.md
```diff
@@ -0,0 +1,11 @@
...
+@bender-tags: 3661, 4.14.0, feature
+@bender-ckeditor-plugins: wysiwygarea, print, font, colorbutton, format, clipboard, pagebreak, toolbar, floatingspace, link, image2
+
+Things to check:
+
+* There are both "Preview" and "Print" buttons in the toolbar.
+* Both "Preview" and "Print" commands are disabled in source mode.
+* "Print" button prints the same document as in preview. **Note**: preview shows pagebreak element, in all browsers except IE8, as it is not possible to render new page in HTML. However this element should not be visible on printed document.
+* Preview is closed after confirming/dismissing native print dialog.
+* Check if page is rendered correctly inside native print dialog's preview and in printed content.
```
jacekbogdanski: It's a bit hard to follow. Maybe change it into proper test steps, so they are simpler to reproduce? Currently seems more like a generic test.

jacekbogdanski:

plugins/print/plugin.js
Comandeer: > If I remember correctly, `load` event is incorrectly propagated on Safari? Maybe we just need to print preview immediately on Safari?

It's not that simple… probably `load` propagates correctly, but in some cases, window can be opened _after_ `load` event. At the same moment not waiting for `load` event can cause invoking `print` command on unstyled content without images. In such case, checking `readyState` ensures that `print` is invoked after loading the document.

I'm not sure if we shouldn't just report bug in Chrome and leave it as is. Alternatively maybe adding `setTimeout` inside `readyState` check will do the trick 🤔

jacekbogdanski: So, maybe just use `readyState === 'complete'` only for Safari, for the rest `load` event?

jacekbogdanski: Ok, so it looks like disabling browser addons significantly reduce the issue and as we already checked this issue with another dev on Linux machine - I'm for keeping this implementation as it's clean and should work correctly. Any related issue is rather an edge case with system distribution than the implementation problem.

jacekbogdanski:
> 1. I'm still able to reproduce the issue with missing image for the print plugin when preview has been

TL; dr; we checked it doesn't seem related to implementation, see also https://github.com/ckeditor/ckeditor4/pull/3672#discussion_r372922618

So, the current approach with `load` listener and `readyState` is correct :+1:

Comandeer:
I've reworded manual tests and fixed failing CI test.

plugins/preview/plugin.js
jacekbogdanski: Ok, so as we already talked F2F it's rather an edge case as `contentPreview` event should be mostly used to sanitize HTML which seems also important when using a print plugin. Let's see if someone directly requests such improvement instead of pre-optimizing.

PR #3723, opened 2019-12-06
image2 maxSize enhancement

Jugen:
This PR adds '**scaled**' as an option for image2's maxSize width or height setting.

### This PR contains

- [x] Unit tests
- [ ] Manual tests

```
* Added 'scaled' as an option for image2's maxSize width or height setting.
```

## What changes did you make?

Changed `getMaxSize()` in image2's `resizer.on( 'mousedown',` function to check for and process new '**scaled**' keyword.

Added test to image2's resizehandler.js test suite.

None

### DIFF
--- a/plugins/image2/plugin.js
+++ b/plugins/image2/plugin.js
@@ -1362,8 +1362,13 @@
 				maxSize = CKEDITOR.tools.copy( maxSize );
 				natural = CKEDITOR.plugins.image2.getNatural( image );
 
+				if ( maxSize.width === 'scaled' ) {
+					maxSize.width = natural.width * (isNaN( maxSize.height ) ? 1.0 : maxSize.height / natural.height);
+				}
 				maxSize.width = Math.max( maxSize.width === 'natural' ? natural.width : maxSize.width, min.width );
-				maxSize.height = Math.max( maxSize.height === 'natural' ? natural.height : maxSize.height, min.width );
+
+				if ( maxSize.height === 'scaled' ) maxSize.height = natural.height * maxSize.width / natural.width;
+				maxSize.height = Math.max( maxSize.height === 'natural' ? natural.height : maxSize.height, min.height );
 
 				return maxSize;
 			}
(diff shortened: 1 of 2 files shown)


Jugen:
Hi Jacek, sure and thanks for the review. Two questions though, when you say:

1. "report a ticket" Do you mean, open an issue ticket describing the problem this PR is addressing ?
2. "config change" Do you mean adding the same info in 1 to the "proposed changelog entry" paragraph in this PR description or is it the "What changes did you make" paragraph ?

jacekbogdanski:
> 1. "report a ticket" Do you mean, open an issue ticket describing the problem this PR is addressing ?

Yes, we are referencing such issue inside `CHANGELOG.md` file, so anyone can visit this file and verify what changes have been introduced by a specific entry.

> 2. "config change" Do you mean adding the same info in 1 to the "proposed changelog entry" paragraph in this PR description or is it the "What changes did you make" paragraph ?

I have been referencing to docs description update related to this configuration option change. You wrote:

> Or a number representing a value and the string 'scaled' that limits the maximum size in pixel units:

It's slightly enigmatic as it doesn't describe how this `scaled` option works. E.g. take a look at docs description for `natural` option:

> A string representing the natural image size, so each image resize operation is limited to its own natural height or width

it gives information on how actually this option works. I'm not totally sure what your changes introduce, so both ticket (GH issue) with description and configuration option docs should help.

Jugen:
Hi Jacek, I thought I'd just run this by you first.

My 'problem' initially was that I didn't want to explicitly specify both dimensions because:

1.   I was only interested in constraining the width, and
2.  I didn't know what the max height should be.

So then I found that if you explicitly specify the max of one dimension and use 'natural' for the other AND an image is **smaller** than the max of the explicit dimension then you can never stretch the image to that max as the resize is constrained by the 'natural' dimension limit.

Now in some cases this may be the desired behavior but in mine I wanted to allow the user to stretch the image beyond its natural size but without having to explicitly specify both max dimensions and so this PR was born.

In retrospect it's probably silly as I could just have used 99999 or something for the other dimension and achieved the same result.

So I think you can just close this as it's not really necessary.

jacekbogdanski:
Thanks for the info @Jugen!

> In retrospect it's probably silly as I could just have used 99999 or something for the other dimension and achieved the same result.

Indeed it may be some workaround for the issue. As your case seems specific, it may make sense to just go with it as it's still a valid approach.

In that case, I'm closing this PR, but if you find out that it's still a worthy improvement, please tell us so we can reopen it.

PR #3725, opened 2019-12-10
Fix for: Safari incorrectly removes blocks after selecting all

### LINKED ISSUE #3705: Safari incorrectly removes blocks after selecting all
## Type of report

Bug

1. Open https://jsfiddle.net/Comandeer/jnf5at6z/
2. Open console.
3. Click "Extract HTML" button.

### Expected result

`<p><br></p>` is logged.

### Actual result

`<div><br></div><p><br></p>` is logged.

## Other details

It seems that Safari, instead of just deleting both blocks, removes their content and preserves blocks.

* Browser: Safar
* OS: macOS
* CKEditor version: 4.13.0

Dumluregn:

Bug fix

### This PR contains

- [x] Unit tests
- [x] Manual tests - is an extension of existing test to avoid code duplication.

```
*[#3705](https://github.com/ckeditor/ckeditor4/issues/3705): Fix for: Safari incorrectly removes blocks after selecting all.
```

## What changes did you make?

The actual reason for this issue was selection optimization introduced in #3213 (with breaking commit [06b94a6](https://github.com/ckeditor/ckeditor4/commit/06b94a64713e34ae8cb41196f2ce00618468a76a)). When it is fired on Safari, it trims selection so that in the [example](https://jsfiddle.net/Comandeer/jnf5at6z/ ) actually just the word `Whatever` is selected and then extracted.

As that change was introduced to prevent [this bug](https://user-images.githubusercontent.com/1061942/63528470-4b468f80-c503-11e9-9e95-af86e7622ad8.gif) occurring in FF, IE and Edge, the easiest way to deal with it is using browser sniffing and skip skipping bogus (yeah!), which in a one-liner fix.

The problem is that IE 8, 9 and 10 fail unit test for this case with following results:
* IE 8:
Screenshot 2019-12-10 at 11 35 40

* IE 9 and 10:
Screenshot 2019-12-10 at 12 26 43

but it doesn't look very important and I propose to extract this case as a separate issue.

Closes #3705.

### DIFF
--- a/tests/core/selection/manual/tripleclick.md
+++ b/tests/core/selection/manual/tripleclick.md
@@ -1,15 +1,28 @@
-@bender-tags: selection, 4.13.0, bug, 3118, 3175
+@bender-tags: selection, 4.13.0, 4.13.2, bug, 3118, 3175, 3705
 @bender-ui: collapsed
 @bender-ckeditor-plugins: wysiwygarea, toolbar, list, format, sourcearea, undo
 
 1. Select list item by triple clicking on it.
-1. Pick `heading 1` from format combo.
+1. Pick `Heading 1` from format combo.
 
-## Expected
+	### Expected:
 
-Heading is applied only to list item.
+	Heading is applied only to list item.
 
-## Unexpected
+	### Unexpected:
 
-Paragraph after the list is turned into `h1`
+	Paragraph after the list is turned into `h1`.
 
+1. Open console.
+
+1. Start selection at the end of document and end it in the first line without selecting list item (<a href="https://user-images.githubusercontent.com/1061942/63528470-4b468f80-c503-11e9-9e95-af86e7622ad8.gif" target="_blank">gif</a>).
+
+1. Change format to `Heading 3`.
+
+	### Expected:
+
+	List item is still `h1`, the second paragraph is `h3`. There is no error in the console.
+
+	### Unexpected:
+
+	Style of list item changed or there is an error in the console.
--- a/tests/core/selection/manual/tripleclick.md
+++ b/tests/core/selection/manual/tripleclick.md
@@ -2,7 +2,10 @@
 @bender-ui: collapsed
 @bender-ckeditor-plugins: wysiwygarea, toolbar, list, format, sourcearea, undo
 
+1. Open console.
+
 1. Select list item by triple clicking on it.
+
 1. Pick `Heading 1` from format combo.
 
 	### Expected:
@@ -13,9 +16,7 @@
 
 	Paragraph after the list is turned into `h1`.
 
-1. Open console.
-
-1. Start selection at the end of document and end it in the first line without selecting list item (<a href="https://user-images.githubusercontent.com/1061942/63528470-4b468f80-c503-11e9-9e95-af86e7622ad8.gif" target="_blank">gif</a>).
+1. Start selection at the end of document and end it in the first line without selecting list item (<a href="https://user-images.githubusercontent.com/1061942/63528470-4b468f80-c503-11e9-9e95-af86e7622ad8.gif" target="_blank">see gif</a>).
 
 1. Change format to `Heading 3`.
 
--- a/tests/core/selection/manual/tripleclick.md
+++ b/tests/core/selection/manual/tripleclick.md
@@ -1,4 +1,4 @@
-@bender-tags: selection, 4.13.0, 4.13.2, bug, 3118, 3175, 3705
+@bender-tags: selection, 4.13.0, 4.14.0, bug, 3118, 3175, 3705
 @bender-ui: collapsed
 @bender-ckeditor-plugins: wysiwygarea, toolbar, list, format, sourcearea, undo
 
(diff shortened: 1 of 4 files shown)


Dumluregn:
Retargeted and rebased to `major`.

f1ames:
Rebased onto latest `major`.

tests/core/selection/optimization.js
```diff
@@ -124,6 +124,19 @@
...
+		},
+
+		// (#3705)
+		'test deleting blocks': function( editor, bot ) {
+			bot.setData( '<p>&nbsp;</p><p>Whatever</p>', function() {
+				var range = editor.createRange();
+
+				range.selectNodeContents( editor.editable() );
+				range.select();
+				editor.extractSelectedHtml();
```
f1ames: Can you also assert what `editor.extractSelectedHtml();` is returning (you can pass `true` as a first argument to get a string)?

tests/core/selection/manual/tripleclick.md
```diff
@@ -1,15 +1,28 @@
-@bender-tags: selection, 4.13.0, bug, 3118, 3175
+@bender-tags: selection, 4.13.0, 4.13.2, bug, 3118, 3175, 3705
```
f1ames:
```suggestion
@bender-tags: selection, 4.13.0, 4.14.0, bug, 3118, 3175, 3705
```

Since we retargeted to `major`.

tests/core/selection/optimization.js
Dumluregn: Safari is the only one that doesn't extract `<p>` element (just unwrapped `<br data-cke-eol="1">`. I tried to track down why this happens and it seems that `range.select()` method sets the `endContainer` to be text node instead of body like in other browsers. For that reason unfortunately I had to distinguish this case in unit test.

Dumluregn:
Using `isInnerHtmlMatching()` helped to unify tests for IE9 and 10. Still, in IE8 cutting works differently (it doesn't extract the extra node, just content - yet it worked that way also before) so there is some nasty browser sniffing for that (and also for Safari mentioned [here](https://github.com/ckeditor/ckeditor4/pull/3725#discussion_r358699101)). I'm not super happy with it and wonder if I couldn't do any better. But for Safari it seems more like a browser engine than editor issue and trying to fix it deeper in the code may be even uglier :disappointed:

tests/core/selection/optimization.js
f1ames: AFAIR Safari has some native selection optimization mechanism which always places selection boundaries in text nodes (if it can and even if you explicitly place it in block element) so we can't do nothing more reasonable here.

tests/core/selection/optimization.js
```diff
@@ -124,6 +124,33 @@
...
+				// and empty paragraph is converted to <br/> element.
+				if ( CKEDITOR.env.safari ) {
+					assert.isInnerHtmlMatching( '<br data-cke-eol="1" /><p>Whatever</p>', extractedHtml, 'Extracted HTML is incorrect.' );
+				} else {
+					assert.isInnerHtmlMatching( '<p>@</p><p>Whatever</p>', extractedHtml, 'Extracted HTML is incorrect.' );
+				}
+
+				// And in IE8 only node content is removed.
+				if ( CKEDITOR.env.ie && CKEDITOR.env.version == 8 ) {
+					assert.isInnerHtmlMatching( '<p>@@</p><p></p>', editor.editable().getHtml(), 'Editor content after extraction is incorrect.' );
```
f1ames: Have you checked how it worked before? From what I see (manual tests with jsfiddle provided in initial report) for both `4.12.0` and `4.13.0` `editor.editable().getHtml()` returns `<p>&nbsp;</p>` on IE8 :thinking:

But I also see that providing both `true` or `false` to `skipBogus` doesn't change anything here so it's not caused by this change, maybe it is something with focus behaviour in unit vs manual testing :thinking: Let's leave it like that then :ok_hand:

f1ames: And yes it was `true` before and now with `!CKEDITOR.env.webkit` is still `true` for IE8 for sure it's not this change. Still kind of confusing :thinking:

f1ames:
Rebased also onto latest `major`.

tests/core/selection/optimization.js
```diff
@@ -124,6 +124,33 @@
...
+				range.select();
+				extractedHtml = editor.extractSelectedHtml( true );
+
+				// In Safari the same selection has different range offsets and endContainer. Because of that it is automatically fixed
+				// and empty paragraph is converted to <br/> element.
+				if ( CKEDITOR.env.safari ) {
+					assert.isInnerHtmlMatching( '<br data-cke-eol="1" /><p>Whatever</p>', extractedHtml, 'Extracted HTML is incorrect.' );
+				} else {
+					assert.isInnerHtmlMatching( '<p>@</p><p>Whatever</p>', extractedHtml, 'Extracted HTML is incorrect.' );
+				}
```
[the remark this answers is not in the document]
Dumluregn: Sure, strings are pretty short and keeping them in one line still looks good, so I'll apply this style in both places.

PR #3764, opened 2020-01-09
Move to ESLint

### LINKED ISSUE #3279: ESLint support
## Type of report

Task

Currently, we are using JSHint and JSCS as our linting tools. It would be nice to get ESLint support also, which seems to have much better TE/IDE support and have great [extension mechanism](https://eslint.org/docs/developer-guide/working-with-plugins) for custom rules.

Additionally, JSCS has a terrible performance which slows down any Text Editor including the ones with the best performance (like Sublime, VIM, etc) when opening bigger JS files - almost freezes on `plugins/widget/plugin.js` when using live linting.

Not really sure if we should totally replace current liners, although it makes sense for me to use a single tool, especially that some rules may not be available for JSHint/JSCS when moving to ESLint.

There is already some nice starting point for ESLint config file prepared by @mlewand https://github.com/mlewand/eslint-config-ckeditor4

Please, give your feedback if you also would like to see linting tool switch (or you are against / think that current tooling is fine).

Comandeer:

Task

### This PR contains

- [ ] Unit tests
- [ ] Manual tests

No changelog entry is needed.

## What changes did you make?

I've used rules from https://github.com/mlewand/eslint-config-ckeditor4 to create ESLint config. Additionally I created precommit hook to lint staged files (via `lint-staged` + `husky`) and removed jshint + jscs.

However ESLint rules may need some polishing.

Closes #3279.


msamsel:
It looks like globals defined for tests are not recognized by eslint:
https://github.com/ckeditor/ckeditor4/blob/d1710c198d8f5007e8d04fed9870bfd4b182ae54/.jshintrc#L24
At least I get all tests red, when I used this config and turn on eslint verification inside VSCode.

Comandeer:
I've rebased onto latest `major`.

jacekbogdanski:

Comandeer:
> Could you take a look at which rules should be disabled / enabled to reduce linter errors and warnings?

I've already looked into it… and we will need to disable even rules about always adding curly brackets ¯\_(ツ)_/¯ The style of older code is very inconsistent.

PR #3770, opened 2020-01-10
Fix emoji dropdown tests

### LINKED ISSUE #3744: Failing tests on CI tests/plugins/emoji/dropdown
## Type of report

Failing test

## Provide description

Tests from `tests/plugins/emoji/dropdown` are failing regularly. They seem to pass when run outside Travis CI. See also https://github.com/ckeditor/ckeditor4/issues/3407 for some hints.

image
image

https://travis-ci.org/ckeditor/ckeditor4/jobs/625681191?utm_medium=notification&utm_source=github_status

## Other details

* Browser: Chrome
* OS: Travis CI
* CKEditor version: 4.14.0 built

msamsel:

Test case fix for CI

### This PR contains

- [ ] Unit tests
- [ ] Manual tests

**No entry**

## What changes did you make?

Added transparent div with higher z-index than emoji panel, which prevents mouse interaction with the panel.
It seems that cursor on CI is located in the middle of the screen and since Chrome v79 it has started to interact with the emoji panel.

Closes #3744.

### DIFF
--- a/tests/plugins/emoji/dropdown.js
+++ b/tests/plugins/emoji/dropdown.js
@@ -107,8 +107,7 @@
 
 			waitForEmoji( bot.editor, function() {
 				// Overlay is required to not trigger mouse event what happens on CI (#3744).
-				var overlay = createOverlayElement();
-				CKEDITOR.document.getBody().append( overlay );
+				var overlay = addOverlayCover();
 
 				bot.panel( 'EmojiPanel', function( panel ) {
 					try {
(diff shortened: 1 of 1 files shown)


f1ames:
> Added transparent div with higher z-index than emoji panel, which prevents mouse interaction with the panel.
It seems that cursor on CI is located in the middle of the screen and since Chrome v79 it has started to interact with the emoji panel.

That's interesting issue TBH 🕵 I can get the same result when slightly moving mouse during unit test run so this seems to be confirming this issue 🤔 

From what I understand the issue is caused by the fact that the mouse is placed over the dropdown panel and when it appears the `mouseover` event is fired breaking the test. Now on Travis CI we use XVFB to display browser window.

**The first question is why the test fails for built version and not dev version?** When tests are run the cursor is always positioned in the same viewport position. The difference between dev and built tests is the fact that built version has all plugins (`full-all` presets) which means the emoji dropdown will be opened in different position than in dev tests (under the cursor I assume).

**The second question is why the test started to fail?** As @msamsel noticed it started to happen after Chrome update to 79 version. It might be that some positioning slightly changed or default browser window size or the fact that `mouseover` started to work slightly differently and now event is fired when panel is opened underneath. Still I was not able to reproduce this situation on Chrome 79 on macOS. Checked also on Ubuntu 16.04 (but without XVFB) - still the results are a little different, static cursor doesn't cause the issue, but when it is moved after opening browser window and before running tests (to be over emoji dropdown) the test fails the same as on CI.

PR #3784, opened 2020-01-15
Add preserving custom colors feature

### LINKED ISSUE #1795: Preserving custom colors feature

Feature request

1. Open [Sample](https://sdk.ckeditor.com/samples/colorbutton.html).
1. Click `Text Color` button.
1. When dialog show up click  `More colors`.
1. Select custom color and click `Ok`. 
1. Click `Text Color` button again.

### Expected result

New color should be added to color palette inside `colorbutton` dialog.

screen shot 2018-03-14 at 12 20 54

### Actual result

User is not able to get back to previous selected color from `colorbutton` dialog.

screen shot 2018-03-14 at 12 22 15

## Other details

We had some brain storm with @mlewand how to implement this feature and we find out with some idea:

1. Scan editor content for elements containing color information on dialog open. It could be defined be default as `span[color]` or configured by user e.g. [colorButton_backStyle](https://docs.ckeditor.com/ckeditor4/latest/api/CKEDITOR_config.html#cfg-colorButton_backStyle), [colorButton_foreStyle](https://docs.ckeditor.com/ckeditor4/latest/api/CKEDITOR_config.html#cfg-colorButton_foreStyle). We should find out the most performant way to scan editor content by keeping CSS selectors precise.
1. Filter out default colors. We don't want to duplicate already  #defined default colors.
1. Sort colors by occurrence so we will be able to choose the most frequent ones.
1. Limit custom colors to some value, 12 should be fine (two rows). With this limitation we will keep dialog relatively small so it won't grow too big if a user used many different colors inside editor.
1. Append custom colors to dialog. A user will be able to reapply previously selected color without the need to memorise or use [Copy Formatting](https://sdk.ckeditor.com/samples/copyformatting.html) feature.

### LINKED ISSUE #3783: Track colors used in the editor contents
## Type of report

Feature request

Extracted as a separate issue from specification for #1795 described [here](https://github.com/ckeditor/ckeditor4/issues/1795#issuecomment-559809627). It will probably be closed in one PR, but both deserve separate changelog entries as they can be used independently.

Dumluregn:

New feature

### This PR contains

- [x] Unit tests
- [x] Manual tests

```
* [#1795](https://github.com/ckeditor/ckeditor4/issues/1795): Colors picked from [Color Dialog](https://ckeditor.com/cke4/addon/colordialog) are now stored in [Color Button](https://ckeditor.com/cke4/addon/colorbutton) palette and can be easily reused.
* [#3783](https://github.com/ckeditor/ckeditor4/issues/3783): Colors used in the document are now displayed as a part of the [Color Button](https://ckeditor.com/cke4/addon/colorbutton) palette.
```

## What changes did you make?

* ## Features
Implementation is mainly based on conclusions listed [here](https://github.com/ckeditor/ckeditor4/issues/1795#issuecomment-559809627), but during development I made one significant change - in palette there are two separate sections for `Content Colors` (colors found in the editor content - checked on every `colorbutton` panel open) and `Custom Colors` (colors picked from color dialog). Without this division the palette was changing unintuitive when e.g. some parts of coloured text were deleted. 
Thanks to that it was also possible to implement both features independently, so theoretically this PR could be divided into two if it's easier to work with for reviewer.

* ## Unit tests
1. As we are in the middle of discussion about place of issue references in unit tests (using `bender-tags` or not), for now I just added references for the whole files as it was quicker way. If we determine that adding tags for each test is better, of course I'll change it.
2. Features are implemented for both `Text Color` and `Background Color` buttons and the only difference between them is that they should track colors independently, so only tests checking setup are doubled - the core features are tested only for `Text Color`. If this is insufficient, of course I'll change it.
3. Code for preserving color style is entirely in colorbutton plugin so I also put tests there, but in fact without colordialog it doesn't do a thing, so I wonder if it would be better to place them in the colordialog folder.

* ## Manual tests
1. I'm guess despite manual tests are split into 2 files they may still be to complex, if reviewer will share this point of view then of course I'll change it.
2. The same as point 3. for unit tests - maybe tests for custom colors should be in colordialog folder?

* ## PS
1. There will be probably two new phrases to translate - right now they are "Content Colors" and "Custom Color". Yet I didn't update lang files, as they can still change - let's create new entries just before merge.
2. In our test base I saw e.g. such code snippet:
```javascript
<h2 style="color:FF0000">Hello</h2>
```
and in this format the color will not be displayed in palette (which makes sense with because default ACF settings don't allow this construction).
As for now this feature only finds colors expressed in hex code attached to spans through `style` attribute, like here:
```javascript
<p><span style="color:334455">Hello</span></p>
```
But I just realised that the below snippet is also fine for ACF:
```javascript
<p><span style="color:red">instance</span></p>
```
so probably it should work too, and now it doesn't... So I'll fix that after the first review :)

Closes #1795.
Closes #3783.


plugins/colorbutton/plugin.js
```diff
@@ -331,31 +359,47 @@ CKEDITOR.plugins.add( 'colorbutton', {
...
+				// Render the "Custom Colors" section.
+				output.push( '</tr>' +
+					'<tr>' +
+						'<td colspan=' + colorsPerRow + '" align="center">' +
+							'<span class="cke_colorcustom_label" style="display:none">Custom Colors</span>' +
+						'</td>' +
+					'</tr>' +
+					'<tr class="cke_colorcustom_row">' );
+
+				// Render the "More Colors" button.
```
jacekbogdanski: This function is already pretty huge, adding anything on top of it only reduces readability. Could you extract template functions to clean up it a bit? Also, note [correct level of abstraction](http://principles-wiki.net/principles:single_level_of_abstraction) here.

plugins/colorbutton/plugin.js
```diff
@@ -370,6 +414,152 @@ CKEDITOR.plugins.add( 'colorbutton', {
 			}
 		}
 
+		function renderContentColors( options ) {
+			// This function is called on each dialog opening.
+			var contentColorsRow = options.contentColorsRow,
+				contentColorsLabel = options.contentColorsLabel,
+				cssAttribute = options.cssAttribute,
+				clickFn = options.clickFn,
+				type = options.type,
```
jacekbogdanski: Overall destructuring properties to avoid "talking to strangers" is good practice, however, there is nothing wrong with accesing children members by a single dot. Destructuring `options` here makes this function less readable because later in function code you have to find out from where these variables come from. It would be a bit easier to recognize them as `options` children which also gives some context about its purpose. Also, see https://en.wikipedia.org/wiki/Law_of_Demeter for some tips about the object's loose coupling.

plugins/colorbutton/plugin.js
```diff
@@ -370,6 +414,152 @@ CKEDITOR.plugins.add( 'colorbutton', {
...
+				colors = [],
+				colorOccurrences = {},
+				sortedColors = [],
+				colorsPerRow = config.colorButton_colorsPerRow || 6,
+				tilePosition = 1;
+
+			// It's more reliable and easier to delete exisiting color tiles and recreate them than to manipulate the old ones.
+			CKEDITOR.tools.array.forEach( contentColorsRow.find( 'td' ).toArray(), function( node ) {
+				node.remove();
+			}, 0, 1 );
```
jacekbogdanski: It could be extracted as a separate function.

plugins/colorbutton/plugin.js
```diff
@@ -370,6 +414,152 @@ CKEDITOR.plugins.add( 'colorbutton', {
 			}
 		}
 
+		function renderContentColors( options ) {
```
jacekbogdanski: Maybe we could give it some ubiquitous name that can be used everywhere when talking about this feature? Some proposals:

* Quick Access Colors
* Colors History
* Colors Register

This function could be simplified by extracting code chunks into helper functions. It will make it more readable and self-documenting. Functions also act as living documentation which won't go out of date without notice like code comments, so it's preferred way to document and clean code if no more complicated explanation required. I will give you some hints which parts of this function could be extracted, but of course, use your own instinct during refactoring.

plugins/colorbutton/plugin.js
```diff
@@ -370,6 +414,152 @@ CKEDITOR.plugins.add( 'colorbutton', {
...
+			}
+
+			CKEDITOR.tools.array.forEach( colorSpans, function( span ) {
+				// colorSpans may contain spans with background color when they are not necessary, so here they are being filtered out.
+				var spanColor = span.getStyle( cssAttribute );
+
+				if ( spanColor ) {
+					colors.push( normalizeColor( spanColor ).toUpperCase() );
+				}
+			} );
```
jacekbogdanski: #separateFunction2

plugins/colorbutton/plugin.js
```diff
@@ -370,6 +414,152 @@ CKEDITOR.plugins.add( 'colorbutton', {
...
+				return;
+			}
+
+			CKEDITOR.tools.array.forEach( colors, function( color ) {
+				if ( color in colorOccurrences ) {
+					colorOccurrences[ color ]++;
+				} else {
+					colorOccurrences[ color ] = 1;
+				}
+			} );
```
jacekbogdanski: #separateFunction3

plugins/colorbutton/plugin.js
```diff
@@ -370,6 +414,152 @@ CKEDITOR.plugins.add( 'colorbutton', {
...
+
+			if ( !colors.length ) {
+				contentColorsLabel.hide();
+				return;
+			}
+
+			CKEDITOR.tools.array.forEach( colors, function( color ) {
+				if ( color in colorOccurrences ) {
+					colorOccurrences[ color ]++;
```
jacekbogdanski: It's a bit easier to read something like `colorOccurences[ color ] += 1` or even `colorOccurences[ color ] = colorOccurences[ color ] + 1` than incrementation. Overall, incrementation works fine mostly for step variables like `i++` in for loop, otherwise, it's a bit unreadable and sometimes ends up in unexpected behavior, like legendary
```
y = 0
x = y++
x == 1 // Oh snap ;( we should use ++y instead
```

plugins/colorbutton/plugin.js
```diff
@@ -370,6 +414,152 @@ CKEDITOR.plugins.add( 'colorbutton', {
...
+					colorOccurrences[ color ] = 1;
+				}
+			} );
+
+			for ( var color in colorOccurrences ) {
+				sortedColors.push( [ color, colorOccurrences[ color ] ] );
+			}
+			sortedColors.sort( function( a, b ) {
+				return b[ 1 ] - a[ 1 ];
+			} );
```
jacekbogdanski: #separateFunction4

plugins/colorbutton/plugin.js
```diff
@@ -370,6 +414,152 @@ CKEDITOR.plugins.add( 'colorbutton', {
...
+
+			// Trim the color array to the row size.
+			if ( sortedColors.length > colorsPerRow ) {
+				sortedColors.splice( colorsPerRow - 1, sortedColors.length - colorsPerRow );
+			}
+
+			CKEDITOR.tools.array.forEach( sortedColors, function( color ) {
+				color[ 2 ] = editor.lang.colorbutton.colors[ color[ 0 ] ] || color[ 0 ];
+			} );
```
jacekbogdanski: #separateFunction5

plugins/colorbutton/plugin.js
```diff
@@ -370,6 +414,152 @@ CKEDITOR.plugins.add( 'colorbutton', {
...
+					colorLabel: color[ 2 ],
+					clickFn: clickFn,
+					colorCode: color[ 0 ],
+					type: type,
+					position: tilePosition++,
+					setSize: sortedColors.length
+				} ) );
+
+				contentColorsRow.append( colorTile );
+			} );
```
jacekbogdanski: #separateFunction6

plugins/colorbutton/plugin.js
```diff
@@ -370,6 +414,152 @@ CKEDITOR.plugins.add( 'colorbutton', {
...
+
+			contentColorsLabel.show();
+		}
+
+		function addCustomColorToPanel( options ) {
```
jacekbogdanski: I won't go with #separateFunctionX here, but you already know what we could improve here.

plugins/colorbutton/plugin.js
```diff
@@ -370,6 +414,152 @@ CKEDITOR.plugins.add( 'colorbutton', {
...
+			// This function is called when a color from colordialog is chosen.
+			var colorHexCode = options.colorHexCode,
+				colorLabel = editor.lang.colorbutton.colors[ colorHexCode ] || colorHexCode,
+				clickFn = options.clickFn,
+				type = options.type,
+				colorsPerRow = options.colorsPerRow,
+				customColorsRow = options.customColorsRow,
+				customColorsLabel = options.customColorsLabel,
```
jacekbogdanski: No need to destructure options here also.

plugins/colorbutton/plugin.js
```diff
@@ -370,6 +414,152 @@ CKEDITOR.plugins.add( 'colorbutton', {
...
+				existingTiles;
+
+			if ( chosenColorTile ) {
+				// If the same color is chosen again, find the old tile and move it to the beginning
+				// instead of creating a new one.
+				customColorsRow.append( chosenColorTile.getParent(), true );
+			} else {
+				var colorTile = new CKEDITOR.dom.element( 'td' );
+
+				tilesNumber++;
```
jacekbogdanski: It looks like there is some loop over this variable. But if I see correctly, it's just incremented to match correct count? If so, maybe it could be done right at the definition to avoid misunderstanding?

tests/plugins/colorbutton/_helpers/tools.js
```diff
@@ -0,0 +1,49 @@
+/* exported findInPanel, chooseColorFromDialog, asyncChooseColorFromDialog */
```
jacekbogdanski: You are still exposing all global functions in this file, this comment only serves for linting. It's better to just wrap whole code in iffy module and expose needed API as a part of it. Eg see https://github.com/ckeditor/ckeditor4/blob/f1496f45653f1ee1880da493dd066ad3230232b5/tests/plugins/easyimage/_helpers/tools.js

f1ames:
I have some additional feedback regarding @jacekbogdanski https://github.com/ckeditor/ckeditor4/pull/3784#pullrequestreview-344626190.

The initial specs was about having one custom colors row, you have decide to split it to two - which I'm fine with as long as it works intuitively.

## Content colors row

Content colors - seems to be showing colors used in the content. The problem here is that, both custom colors from the content appears there (so pasted from Word or added manually via source mode) and colors picked from color panel/dialog too :thinking:  So these are just all colors used but sorted in order based on their appearance in the content.

So from my perspective, it should only show colors which were not picked manually from color panel or dialog - only these which appeared there other ways (paste, etc).

And as mentioned in #1795 as we treat them as content colors these should be sorted by number of appearances:

> Colors found in content should be ordered by number of occurrences, counted in <span> elements.

and if the number is the same I would be for sorting them in order of occurrence in the content (we didn't covered this in #1795).

## Custom colors row

I know those are only colors picked from dialog, however it seems unintuitive for me as picking color from dialog or panel has the same effect - the dialog is just an extension of the panel with more colors so these should be treated the same way.

Also since colors added from dialog also appear in the content, they are also shown in the **Content colors row** which might be confusing :thinking: 

---

Getting back to @jacekbogdanski https://github.com/ckeditor/ckeditor4/pull/3784#pullrequestreview-344626190 it seems to be in tact with what we drafted as sepcs some time ago here - https://github.com/ckeditor/ckeditor4/issues/1795#issuecomment-559809627 and so:

> 1. Finds out all available content colors **on the first color button opening** to fill empty quick access panel as at this point I didn't choose anything manually. Sort them in some smart order e.g. from the start of the content.
> ```
> Content colors:
> red | blue | yellow | green
> ```

Agree. The only thing I'm not sure and is different in the initial specs is:

> Document colors should be updated on each panel show. This way we are able to catch custom colors in content pasted from Word etc.

The idea was here that you can open color panel and apply some color (or even just open/close), then paste some colorful content and then open color panel again - shouldn't we fill empty color slots with the colors that appeared in the content?

> 2. If I have chosen color manually it's moved **at the beginning** of the quick access color row. The last color is removed if there is no space left. It doesn't matter if color is chosen from a color dialog or just predefined panel color - at this point, a quick access row is treated as the color history. I know specs tell that we don't want duplicate colors, but if we don't, it breaks feature flow and makes it less user-friendly. You shouldn't really find out how things works, it should just 'click' on the first peek.
> ```
> Chosen black:
> black | red | blue | yellow
> ```

Color should be moved to the beginning when picked (from panel, dialog or even or custom colors row) as mentioned.

What about duplicating colors? Can't we just move picked color to the beginning and left the empty slot on the end (or fill with white or next content color available)?

> 3. If I choose the same color twice, it's moved **at the beginning**, but the rest of the colors stays at their position (-1 because the color from the end has been moved, but nothing removed).
> ```
> Chosen yellow
> yellow | black | red | blue
> ```

Not sure what's the difference between this point and previous one :thinking:

@jacekbogdanski Consider your scenario wit a little different set of colors:

1. Initial colors:

```
Content colors:
red | blue | yellow | green
```

2. Picked yellow from color panel (moving to beginning with duplication)

```
Content colors:
yellow | red | blue | yellow
```

3. Picked yellow again (so picked twice - moving to the beginning):

```
Content colors:
yellow | red | blue | ? / content color
```

wouldn't it be simpler with just two steps:

1. Initial colors:

```
Set 1
red | blue | yellow | green

Set 2
red | blue | yellow | empty

Set 3
red | blue | black | green
```

2. Picked yellow from color panel (moving to beginning, no duplication)

```
Set 1
yellow | red | blue | green

Set 2
yellow | red | blue | empty

Set 3
yellow | red | blue | black
```

f1ames:
@Dumluregn can you add a shot summary what we agreed on during F2F talk?

Dumluregn:
Sure:

1.  We prefer to keep both content colors and custom colors unified in one row, which will be used as a color history.
2.  Only at the first panel open (so onBlock) colors from content will be read and rendered in the history row. When a new color is applied (no matter if it comes from default palette or from color dialog), it is **prepended** to the history row (excessive colors from the end are removed). If this color was present in the row before, it is just relocated to the beginning (it doesn't matter if it's color from palette or from content).
3.  Colors from the content should be sorted according to their frequency of occurrence.
4.  Colors from the content that appeared in the editor **after** the first panel open (e.g. pasted from external source) are ignored for now. Probably handling them will be treated as a follow-up issue and suspended for now.
5.  If a color from default palette is applied, it will be duplicated and appear in both places.
6.  We would like to expose API for finding colors in the editor contents.
7.  For now we will not include any config options. Later on we may think about adding multiple rows, different color finding algorithm (e.g. to search not only by `span` elements) or separating content and custom colors, but as follow-up issues.

f1ames:
> 7. For now we will not include any config options. Later on we may think about adding multiple rows, different color finding algorithm (e.g. to search not only by span elements) or separating content and custom colors, but as follow-up issues.

I think we agreed on config option for number of custom colors rows? @jacekbogdanski?

The rest of the summary looks good 👍

Dumluregn:
Hm you can be right. I edited the summary.

plugins/colorbutton/plugin.js
```diff
@@ -285,25 +301,37 @@ CKEDITOR.plugins.add( 'colorbutton', {
...
-					' role="option" aria-posinset="1" aria-setsize="', total, '">' +
+					' onclick="CKEDITOR.tools.callFunction(' + clickFn + ',null,\'' + type + '\');return false;"' +
+					' href="javascript:void(\'' + lang.auto + '\')"' +
+					' role="option" aria-posinset="1" aria-setsize="' + total + '">' +
 						'<table role="presentation" cellspacing=0 cellpadding=0 width="100%">' +
 							'<tr>' +
-								'<td colspan="' + colorsPerRow + '" align="center"><span class="cke_colorbox" id="', colorBoxId, '"></span>', lang.auto, '</td>' +
+								'<td colspan="' + colorsPerRow + '" align="center">' +
+									'<span class="cke_colorbox" id="' + colorBoxId + '"></span>' + lang.auto +
+								'</td>' +
```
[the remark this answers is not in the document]
Dumluregn: It was working alright, I just don't like inconsistency and fix it when I see it (as you already know), sorry - this time I really tried to control myself :p

jacekbogdanski:
> I think we agreed on config option for number of custom colors rows?

Yes, could be useful to preserve more colors from content if needed.

Dumluregn:
One more thing I think we agreed to do was to ask some non-tech or UI people to test the solution.

jacekbogdanski:
Yeah, but let's create the solution first :smile:

plugins/colorbutton/plugin.js
```diff
@@ -331,31 +359,47 @@ CKEDITOR.plugins.add( 'colorbutton', {
...
+					setSize: total
+				} ) + '</td>' );
 			}
 
-			// Render the "More Colors" button.
+			// Render the "Content Colors" section.
+			output.push( '</tr>' +
+				'<tr>' +
+					'<td colspan=' + colorsPerRow + '" align="center">' +
+						'<span class="cke_colorcontent_label" style="display:none">Content Colors</span>' +
```
msamsel: You should already start to using lang strings here instead of direct name `Content Coloros`. It is really easy to forget later about such cases.
To do so you need to make few things:
1. Add required fields to file `plugins/colorbutton/lang/en.js`
2. Run our langtools: `dev/langtool/langtool.sh`
3. This should modify js files in other languages for colorbutton with `// MISSING` comment
4. Run `dev/langtool/update_meta_files.sh` what should update meta file `dev/langtool/meta/ckeditor.plugin-colorbutton/meta.txt`
5. In this txt file should appear empty entries, where is required to describe your new strings for the translators.

And the same for the `Custom Colors`.

plugins/colorbutton/plugin.js
```diff
@@ -370,6 +414,152 @@ CKEDITOR.plugins.add( 'colorbutton', {
...
+				contentColorsLabel.hide();
+				return;
+			}
+
+			CKEDITOR.tools.array.forEach( colorSpans, function( span ) {
+				// colorSpans may contain spans with background color when they are not necessary, so here they are being filtered out.
+				var spanColor = span.getStyle( cssAttribute );
+
+				if ( spanColor ) {
+					colors.push( normalizeColor( spanColor ).toUpperCase() );
```
msamsel: The next `forEach` loop which calculates occurrences could be integrated somewhere here.
It would prevent of one unnecessary of iteration over the same colors twice. Once for spans and then once for read values.

plugins/colorbutton/plugin.js
```diff
@@ -331,31 +359,47 @@ CKEDITOR.plugins.add( 'colorbutton', {
...
-						' role="option" aria-posinset="', ( i + 2 ), '" aria-setsize="', total, '">' +
-						'<span class="cke_colorbox" style="background-color:#', colorCode, '"></span>' +
-					'</a>' +
-					'</td>' );
+				output.push( '<td>' + generateTileHtml( {
+					colorLabel: colorLabel,
+					clickFn: clickFn,
+					colorCode: colorCode,
+					type: type,
+					position: i + 2,
```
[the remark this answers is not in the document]
Dumluregn: Well, I had no idea how to call a variable, so I wrote a comment - if you have any idea for name I'll welcome it.

plugins/colorbutton/plugin.js
Dumluregn: It would have to be calculated twice then, as it should be incremented when new color is picked, while for existing color it's value won't change. But according to your comment \[here\](https://github.com/ckeditor/ckeditor4/pull/3784#discussion\_r367987985) I changed `++` to `+= 1`.

plugins/colorbutton/plugin.js
```diff
@@ -331,31 +359,47 @@ CKEDITOR.plugins.add( 'colorbutton', {
...
-						' draggable="false"' +
-						' ondragstart="return false;"' + // Draggable attribute is buggy on Firefox.
-						' onclick="CKEDITOR.tools.callFunction(', clickFn, ',\'', colorCode, '\',\'', type, '\'); return false;"' +
-						' href="javascript:void(\'', colorCode, '\')"' +
-						' data-value="' + colorCode + '"' +
-						' role="option" aria-posinset="', ( i + 2 ), '" aria-setsize="', total, '">' +
-						'<span class="cke_colorbox" style="background-color:#', colorCode, '"></span>' +
-					'</a>' +
-					'</td>' );
+				output.push( '<td>' + generateTileHtml( {
```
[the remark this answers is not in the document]
Dumluregn: As I wrote in comment, `CKEDITOR.dom.element.createFromHtml()` can't be used to create `<td>` element. So if I encapsulate also `<td>` element, I will not be able to use this function to create color boxes (I think that's a good name, and it's already present in codebase) in other places 😢. If we could fix `CKEDITOR.dom.element.createFromHtml()` before, it would be obvious to do what you proposed.

plugins/colorbutton/plugin.js
Dumluregn: Actually to save space in panel (and labels are no longer needed) I replaced names with `<hr>`, so maybe there will be no need to add any translations.

Dumluregn: Actually to save space in panel (and labels are no longer needed) I replaced names with `<hr>`, so maybe there will be no need to add any translations.

Dumluregn:
Failing CI can be fixed by rebasing after #3744.

jacekbogdanski:

plugins/colorbutton/plugin.js
```diff
@@ -331,45 +382,277 @@ CKEDITOR.plugins.add( 'colorbutton', {
...
+			return color in list ? list[ color ].substr( 1 ) : normalizeColor( span.getComputedStyle( cssProperty ) ).toUpperCase();
+		}
+
+		function countOccurrences( property, object ) {
+			if ( property in object ) {
+				object[ property ] += 1;
+			} else {
+				object[ property ] = 1;
+			}
+		}
```
jacekbogdanski: AFAIK we talked about this function :thinking: nevertheless, it could be simplified by using `reduce` function instead:
```javascript
function extractColorInfo( spans, cssProperty, colorList ) {
	return CKEDITOR.tools.array.reduce( spans, function( counterObject, span ) {
		var hexCode = getHexCode( span, colorList, cssProperty );

		counterObject[ hexCode ] = counterObject[ hexCode ] || 0;
		counterObject[ hexCode ] += 1;
		return counterObject;

	}, {} );
}

```

plugins/colorbutton/plugin.js
```diff
@@ -408,6 +702,27 @@ CKEDITOR.plugins.add( 'colorbutton', {
...
+		 * @since 4.14.0
+		 * @param {String} property CSS property which will be used in query.
+		 * @returns {Array} Returns an array of {@link CKEDITOR.dom.element}s.
+		 * @member CKEDITOR.editor
+		 */
+		editor.getStyledSpans = function( property ) {
+			var spans = editor.editable().find( 'span[style*=' + property + ']' ).toArray();
+
+			return filterExcessiveSpans( spans, property );
+		};
```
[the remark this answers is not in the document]
Dumluregn: As we agreed I moved this function to `CKEDITOR.tools` namespace to be able to use it's versatility elsewhere.

plugins/colorbutton/plugin.js
```diff
@@ -285,29 +295,34 @@ CKEDITOR.plugins.add( 'colorbutton', {
...
 					}, null, colorData );
 				} else {
 					setColor( color && '#' + color );
+					saveColor( {
+						colorHistoryRows: panel.element.find( '.cke_colorhistory_row' ).toArray(),
+						colorHistorySeparator: panel.element.findOne( '.cke_colorhistory_separator' ),
+						colorHexCode: color.toUpperCase(),
+						clickFn: clickFn,
+						colorsPerRow: colorsPerRow
+					} );
```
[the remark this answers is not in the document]
Dumluregn: This is no longer an issue after the refactor (only color code is now passed as an argument here).

plugins/colorbutton/plugin.js
```diff
@@ -331,45 +346,324 @@ CKEDITOR.plugins.add( 'colorbutton', {
...
+		// This function is called on the first panel opening.
+		function fillColorHistory( options ) {
+			if ( config.colorButton_historyRowLimit === 0 ) {
+				return;
+			}
+
+			var colorSpans = editor.getStyledSpans( options.cssProperty ),
+				htmlColorsList = CKEDITOR.tools.style.parse._colors,
+				colorsPerRow = config.colorButton_colorsPerRow || 6,
+				rowLimit = config.colorButton_historyRowLimit || 1,
```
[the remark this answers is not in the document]
Dumluregn: Moved as suggested.

plugins/colorbutton/plugin.js
Dumluregn: Refactored as suggested.

Dumluregn:
@jacekbogdanski do you want me to rebase this PR onto latest `major`?

jacekbogdanski:
There are some merge conflicts, so yes please @Dumluregn

Dumluregn:
Rebased onto latest `major`.

Dumluregn:
CI failed because some code was lost during rebase, so please pull this branch again.

tests/plugins/colorbutton/manual/colorhistory/4-colororder.md
```diff
@@ -0,0 +1,58 @@
...
+
+1. Click on background color button.
+
+  **Expected:**
+
+  * Color history row doesn't appear.
+
+  **Unexpected:**
+
+  * Color history is the same as for text color.
```
jacekbogdanski: Background color history row is visible as there are some background colors preloaded with the editor.

tests/plugins/colorbutton/manual/colorhistory/0-zerorows.md
```diff
@@ -0,0 +1,33 @@
...
+
+1. Click on color button.
+
+  **Expected:**  No new UI elements appeared.
+
+  **Unexpected:**  Color history row or console errors appeared.
+
+1. Repeat all previous steps for background color button.
+
+1. Repeat all previous steps for each editor type.
```
jacekbogdanski: Could we remove stray lines between test steps? It creates unnecessary test description spacing which is hard to read. It touches all manual tests.

tests/plugins/colorbutton/manual/colorhistory/3-toomanycolors.md
```diff
@@ -0,0 +1,21 @@
+@bender-tags: 4.15.0, feature, 1795
+@bender-ui: collapsed
+@bender-ckeditor-plugins: wysiwygarea, toolbar, colorbutton, colordialog, sourcearea, removeformat, undo
+
+1. Click on color button.
+
+1. Pick any color that isn't in color history row yet (e.g. <span style="color:#F39C12">orange</span>).
```
jacekbogdanski: Maybe we could be more specific here? E.g. the first color in the panel.

tests/plugins/colorbutton/manual/colorhistory/4-colororder.md
```diff
@@ -0,0 +1,58 @@
...
+
+  **Expected:**
+
+  1. Below default palette there is horizontal rule and color history row.
+  1. There are 4 color boxes in color history row.
+  1. Color boxes are ordered by number of occurences and further by order in document:
+      * Strong Blue (occures 2 times in document)
+	  * Red (occures 1 time and is 1st color in doc)
+	  * Grey (occures 1 time and is 2nd color in doc)
+	  * Yellow (occures 1 time and is 3rd color in doc)
```
jacekbogdanski: Maybe we could be less specific about the color position and its occurrences, so basically removing everything inside brackets as it's not really needed to finish the test?

tests/plugins/colorbutton/manual/colorhistory/5-tworows.md
```diff
@@ -0,0 +1,23 @@
...
+
+1. Click on color button.
+
+1. Pick any color that isn't in color history row yet (e.g. <span style="color:#F39C12">orange</span>).
+
+1. Click on color button.
+
+  **Expected:**
+
+  * Second color history row appeared.
```
jacekbogdanski: Maybe we could also extend this test checking if the second row is correctly limited i.e. we are unable to create the third row?

tests/plugins/colorbutton/manual/colorhistory/4-colororder.md
Dumluregn: At first I was going to write that you probably omitted the test step where you should clear editor, but in fact there is a way to produce a bug this way if where an empty span with background color set is left in the editor 🤔 I fixed this behaviour by filtering out empty spans.

tests/plugins/colorbutton/manual/colorhistory/5-tworows.md
Dumluregn: Don't you think this case is already covered in `toomanycolors` test?

plugins/colorbutton/plugin.js
```diff
@@ -95,6 +95,303 @@ CKEDITOR.plugins.add( 'colorbutton', {
...
+							' draggable="false"' +
+							' ondragstart="return false;"' + // Draggable attribute is buggy on Firefox.
+							' onclick="CKEDITOR.tools.callFunction(' + this.clickFn + ',\'' + this.color + '\');' +
+							' return false;"' +
+							' href="javascript:void(\'' + this.color + '\')"' +
+							' data-value="' + this.color + '"' +
+							' role="option">' +
+							'<span class="cke_colorbox" style="background-color:#' + this.color + '"></span>' +
+						'</a>' );
+				},
```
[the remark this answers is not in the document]
Dumluregn: It required a bit of hacky logic to maintain the feature to use own custom palette, but we saved a lot of lines of code thanks to this 👍

plugins/colorbutton/plugin.js
```diff
@@ -95,6 +95,303 @@ CKEDITOR.plugins.add( 'colorbutton', {
...
+
+				getCapacity: function() {
+					return ColorHistory.rowLimit * ColorHistoryRow.maxLength;
+				},
+
+				colorList: CKEDITOR.tools.style.parse._colors
+			},
+
+			_: {
+				gatherColorInfo: function() {
```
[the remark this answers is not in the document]
Dumluregn: `countColors` sounds legit 👍

plugins/colorbutton/plugin.js
```diff
@@ -95,6 +95,303 @@ CKEDITOR.plugins.add( 'colorbutton', {
...
+						return;
+					}
+
+					var existing;
+
+					CKEDITOR.tools.array.forEach( this.rows, function( row ) {
+						existing = existing || row.extractColorBox( colorCode );
+					} );
+
+					return existing;
```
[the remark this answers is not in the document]
Dumluregn: Exactly, thx 👍

plugins/colorbutton/plugin.js
```diff
@@ -95,6 +95,303 @@ CKEDITOR.plugins.add( 'colorbutton', {
...
+				},
+
+				createAtBeginning: function( colorCode ) {
+					var colorBox = new ColorBox( colorCode, this.clickFn );
+
+					if ( !this.rows ) {
+						this._.addNewRow();
+					}
+
+					this.rows[ 0 ].addNewColor( colorBox );
```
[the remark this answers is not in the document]
Dumluregn: Almost - `this._.moveToBeginning()`, but thx 👍

plugins/colorbutton/plugin.js
```diff
@@ -184,8 +483,18 @@ CKEDITOR.plugins.add( 'colorbutton', {
...
 						colorBoxId: colorBoxId,
 						colorData: colorData,
-						commandName: commandName
+						commandName: commandName,
+						panel: block.element,
+						colorHistory: history
 					} ) );
+
+					history.setContainer( block.element.findOne( '.cke_colorhistory' ) );
+					history.setClickFn( block.element.data( 'clickFn' ) );
```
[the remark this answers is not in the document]
Dumluregn: Tried to do this on earlier stage and it was very painful, now was pretty easy 👍

plugins/colorbutton/plugin.js
```diff
@@ -352,38 +653,62 @@ CKEDITOR.plugins.add( 'colorbutton', {
...
+					position: i + 2, // First color box should have position 2 as position 1 is for Automatic Button.
+					setSize: total
+				} ) );
+			}
+
+			if ( ColorHistory.rowLimit ) {
+				output.push( '</tbody>' +
+						ColorHistory.renderContainer( 'tbody' ) +
+					'<tbody>' );
 			}
```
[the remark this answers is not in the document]
Dumluregn: I was being naive and wanted to provide flexible API to provide easy possibility to show color history outside of panel (e.g. in dialog) using `<table>` as a parameter instead 😂 Voiceover doesn't have any issue with it, and could you explain which aria attributes do you have in mind? `<tbody>` on the main palette doesn't seem to have any (in fact I added this element in one of the commits - earlier there were just `<tr>`s inside `<table>`).

Dumluregn:
## Manual test numbering

I added numbers for tests as sometimes during testing other big features (like Spreadsheet 😜) it happened that somewhere at the beginning there was a "Play with a feature" test and later more specific ones with the partial functionalities and sometimes I had an impression that I'm testing something twice (as I already checked that in the previous test) or I didn't check something that would be useful (cause I didn't know it works that way). And sometimes two really similar tests were divided by 10 others because their name differed at the beginning, and I wondered how is this test different than something I already did. So I wanted to introduce an order which will show how the feature works step by step - I know it isn't really big and probably it wasn't essential here, but I didn't think it may bother someone or require us to sustain this convention in other PRs. So if you really think it's more harmful than useful - I can remove numbers, otherwise let's keep them.

Dumluregn:
## UI

For me the empty space between default palette and colour history would look like a rendering bug 🤔 And when it comes to hiding HR - have mixed feelings here. When it was hidden by default, I had an impression that after adding a colour to history the interface changes too much. Maybe we keep it and treat as an indicator that the feature is on (so no HR if custom rows limit is set to 0)?

Dumluregn:
## Behaviour

I fixed the aria-selected issue - now the first color box found in palette is selected, which should solve the problem.

tests/plugins/colorbutton/manual/colorhistory/4-colororder.md
jacekbogdanski: Indeed I omitted these test steps. Why colors are removed once the cleaning editor's content? If we consider this feature as a history register, shouldn't colors be preserved? AFAIK we discussed that it should work that way (I mean cleaning content shouldn't have an impact on color history).

jacekbogdanski: Indeed I omitted these test steps. Why colors are removed once the cleaning editor's content? If we consider this feature as a history register, shouldn't colors be preserved? AFAIK we discussed that it should work that way (I mean cleaning content shouldn't have an impact on color history).

Dumluregn: If a color appears in the history, it will not disappear. The content (and colors) are removed before first background color panel opening, and as color history is created on panel opening, there are no colors to be added to it - it's consistent with what we agreed on.

jacekbogdanski: > it's consistent with what we agreed on.

Not really. We agreed that color history should be filled with colors on editor initialization. So basically, you can reproduce something like:

* 1st user is defining some custom colors and saving it into a database.
* 2nd user is loading editor data and gets previously defined colors.

Do you see some use cases for the current implementation?

Dumluregn: Wow, really? I was pretty sure colors were about to load on panel initialization 🤔 As written [here](https://github.com/ckeditor/ckeditor4/pull/3784#issuecomment-576669938):
> Only at the first panel open (so onBlock) colors from content will be read and rendered in the history row.

I'm not quite sure what do you understand by saving the colors into a database and loading them by another user - should editor somehow keep those colors as a metadata? If user wants to suggest some colors to be used it can be done just by changing default palette, so I don't think we need to duplicate this feature 🤔 

The use case for the current implementation was:
* user opens editor;
* pastes content from external source (e.g. Word);
* opens color panel and has content colors in history to be easily picked up.

jacekbogdanski: > should editor somehow keep those colors as a metadata?

It already does - colors are preserved by content, right? So, opening color history on a styled content will load already defined colors by e.g. another user.

> The use case for the current implementation was:

I'm not sure if we should consider this use case, as it only works for the first-panel opening.

Dumluregn: Ok, as I misunderstood you - in a slack conversation we agreed that it's just about changing the moment when the colors from content is read - should be done on editor load, currently it is on panel load. I'll change it after current review round.

core/tools.js
```diff
@@ -1182,6 +1182,22 @@
...
+		 * @param {String} property CSS property which will be used in query.
+		 * @returns {Array} Returns an array of {@link CKEDITOR.dom.element}s.
+		 */
+		getStyledSpans: function( property, source ) {
+			var spans = source.find( 'span[style*=' + property + ']' ).toArray();
+
+			// This is to filter out spans e.g. with background color when we want text color.
+			return CKEDITOR.tools.array.filter( spans, function( span ) {
+				return ( span.getStyle( property ) && span.getText() );
+			} );
```
jacekbogdanski: Seems like there is no test coverage for this function.

jacekbogdanski: I'm a bit unsure if this function will give expected results as a general function. What if it's expected that the span is empty? Maybe it would be better to just find all spans with `style` attribute and filter the correct property?

plugins/colorbutton/plugin.js
```diff
@@ -95,6 +95,291 @@ CKEDITOR.plugins.add( 'colorbutton', {
...
+					this.getElement().append( colorBox.getElement(), true );
+				},
+
+				extractColorBox: function( colorCode ) {
+					var index = CKEDITOR.tools.getIndex( this.boxes, function( box ) {
+						return box.color === colorCode;
+					} );
+
+					if ( index < 0 ) {
+						return;
```
jacekbogdanski: It still returns void, see https://github.com/ckeditor/ckeditor4/pull/3784#discussion_r395652055 (should be changed to `null`).

plugins/colorbutton/plugin.js
```diff
@@ -95,6 +95,303 @@ CKEDITOR.plugins.add( 'colorbutton', {
...
+						this._.addNewRow();
+					}
+
+					this.rows[ 0 ].addNewColor( colorBox );
+				},
+
+				addNewRow: function() {
+					if ( !this.rows ) {
+						this.rows = [];
+					}
```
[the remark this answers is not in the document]
jacekbogdanski: @Dumluregn could you refer to this review comment?

plugins/colorbutton/plugin.js
```diff
@@ -95,6 +95,291 @@ CKEDITOR.plugins.add( 'colorbutton', {
...
+
+		var ColorHistoryRow = CKEDITOR.tools.createClass( {
+			$: function() {
+				this.$ = new CKEDITOR.dom.element( 'tr' );
+				this.$.addClass( 'cke_colorhistory_row' );
+				this.boxes = [];
+			},
+
+			statics: {
+				maxLength: editor.config.colorButton_colorsPerRow
```
jacekbogdanski: Keeping the same naming as config options improves API discoverability and creates ubiquitous naming.

plugins/colorbutton/plugin.js
```diff
@@ -176,50 +461,94 @@ CKEDITOR.plugins.add( 'colorbutton', {
...
+
+					colorStyleTemplate.childRule = type == 'back' ?
+						function( element ) {
+							// It's better to apply background color as the innermost style. (https://dev.ckeditor.com/ticket/3599)
+							// Except for "unstylable elements". (https://dev.ckeditor.com/ticket/6103)
+							return isUnstylable( element );
+						} : function( element ) {
+							// Fore color style must be applied inside links instead of around it. (https://dev.ckeditor.com/ticket/4772,https://dev.ckeditor.com/ticket/6908)
+							return !( element.is( 'a' ) || element.getElementsByTag( 'a' ).count() ) || isUnstylable( element );
+						};
```
jacekbogdanski: Could you extract it as a separate function e.g. `createColorStyleTemplate`?

plugins/colorbutton/plugin.js
```diff
@@ -176,50 +461,94 @@ CKEDITOR.plugins.add( 'colorbutton', {
...
+						if ( color == '?' ) {
+							editor.getColorFromDialog( function( color ) {
+								if ( color ) {
+									setColor( color, history );
+								}
+							}, null, colorData );
+						} else {
+							setColor( color && '#' + color, history );
+						}
+					} );
```
jacekbogdanski: Could you extract it as a separate function e.g. `createClickFunction`?

plugins/colorbutton/plugin.js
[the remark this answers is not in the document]
Dumluregn: Yeah, I didn't notice it :) Fixed it now.

Dumluregn:
1. I removed manual tests numbering as you requested.
2. > Unfortunately, it only reversed the issue

I had a bit of an issue with the following use case:
* User changes text color picking a color from default palette;
* User selects different part of document and changes color using just created color history box;
* He clicks away to some other place in document;
* He selects one of the text fragments colored earlier;
* We don't have any indication which color box was clicked to change color in this particular place (was it box from default palette or color history).

So what I did was adding custom attribute to the clicked color box, which is removed as soon as selection is changed and if it's set, this box will be selected on panel open. It handles the case you described. If custom attribute isn't found, then custom palette has a priority. If you have any idea how to make it better (or I misunderstood sth again 😅) please let me know.

3. Which screen reader did you use? It seems their interpretation differs - MacOS's VoiceOver treats all panel contents the same (so e.g. if there are 24 colors in default palette and 4 in history, it says there are 30 elements - including `automatic` button and `more colors` button (I'm unsure how I missed that earlier), so here the option would be renumbering of all panel elements each time a new color is picked (yuck), but as I understand on the screen reader you used it works different way, so it may be necessary to change history's structure (resigning from separate `tbody` elements probably). I'll add a test for that as soon as we come to a conclusion.

plugins/colorbutton/plugin.js
```diff
@@ -271,123 +551,146 @@ CKEDITOR.plugins.add( 'colorbutton', {
...
+						setColor( color && '#' + color, history );
+					}
+
+					if ( !colorbox ) {
+						return;
+					}
+					colorbox.setAttribute( 'cke_colorlast', true );
+					editor.once( 'selectionChange', function() {
+						colorbox.removeAttribute( 'cke_colorlast' );
+					} );
```
jacekbogdanski, line 605: It would be nice to add some comment for this change.

f1ames:
One side note - we have recently closed one PR related to `colorbutton` plugin #3977, which may make rebasing this PR PITA (I hope not). The changes are very cosmetic there so you may take a look there in case of any doubts during rebase.

plugins/colorbutton/plugin.js
```diff
@@ -19,6 +19,281 @@ CKEDITOR.plugins.add( 'colorbutton', {
 		var config = editor.config,
 			lang = editor.lang.colorbutton;
 
+		var ColorBox = CKEDITOR.tools.createClass( {
```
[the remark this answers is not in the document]
Dumluregn: I did it and the flow of the file is a bit cleaner now, but this approach means we no longer have `editor` in the scope, so I had to add it in some places (e.g. `ColorBox` class signature rose up to 4 parameters 😒 ). You are welcome to propose further enhancements 😛

Dumluregn:
> A screen reader could read color history buttons a bit differently including information about "history" color

I did it by adding `Color History` annotation to the tile title (try to speak it out loud fast 😂 ), so it's both visible on hover and read by screen reader. I experimented with `aria-label` to only change the SR part, but the problem was that JAWS reads values of both attributes if they are present, and `title` is necessary to keep the tooltip on box hover, so there are some issues either way. But if you think one of them is better than the current one - no problem, let's change it.

Dumluregn:
> It would be nice to have a manual test confirming the accessibility of the extended panel
See `keyboarda11y` manual test.

Dumluregn:
> The above navigation seems invalid as the first set is interrupted by color history one. Probably the easiest way would be unifying all sets in one, so color history row would be a part of the main panel set including all panel buttons.

I did as you proposed - all the panel elements are now correctly renumbered every time a new color is picked (also fixed a few numbering bugs with custom configurations 🙈 ).

Dumluregn:
I added one more commit to fix CI.

Dumluregn:
> I think we could just skip this case and extract as a followup, especially that color dialog is obligatory to use color button with friendly named colors.

Extracted #4137 and reverted the changes so the labels for default palette and color history are the same. I fixed the test (actually plugin 😬 ) and HR is now hidden if there are no colors in color history.

f1ames:
> Sorry for not pointing it out earlier, but it seems like unit tests are failing on IE8

Should we even support IE8 with this feature? Or it doesn't support it but tests are not ignored? TBH I don't really remember initial assumptions about supported browsers :thinking:

jacekbogdanski:
The color button is supported on IE8 and after some clicking, I see that the feature works nicely also. If we can support it without some significant work, it's recommended to do it.

Of course, we can decide that the feature shouldn't be supported on IE8, but in that case, it should be disabled on this browser. Maybe let's wait with fixing these units for QA and if IE8 works good, decide if we want to fix them/ignore. WDYT?

Dumluregn:
Let me spend 10 minutes to just check if there is something wrong with tests or with IE8 and if there is no 5-min fix, I'd go with @jacekbogdanski proposition.

Dumluregn:
From what I see there is a problem with rendering content colors 😬 All unit tests that fail have this in common. Also I think it's a problem in manual tests (see `saveinhistory` test). As it would take a while to debug and fix it, I'd be for dropping support for IE8 👋

PR #3856, opened 2020-02-12
Expose colorbutton#select method

### LINKED ISSUE #3743: Expose color selection method
## Type of report

Feature request

Color Button plugin has a private function allowing to select a chosen color from the panel block. With the addition of https://github.com/ckeditor/ckeditor4/issues/3727 feature, it makes sense to expose this function, so it's possible to intercept color command and provide custom color choosing scenarios for more advanced features like Spreadsheets. Preferably, exposed API should be straightforward, similar to https://github.com/ckeditor/ckeditor4/issues/3387

```javascript
var ui = editor.ui.get( 'TextColor' ); // For background color use BGColor instead.
ui.select( function( item ) {
  return item.value === '#fafafa';
} );
```

Note that this feature should only handle colors from the color button panel as colors from the color dialog are not preserved in UI yet.

jacekbogdanski:

New feature

### This PR contains

- [x] Unit tests
- [x] Manual tests

Not required as it's an internal change.

## What changes did you make?

Introduced new color button selection method. Due to fact that it's unlikely this method will be used by community, let's keep it internal for now, as public API would require more complicated refactoring of the colorbutton plugin.

Closes #3743.

### DIFF
--- a/plugins/colorbutton/plugin.js
+++ b/plugins/colorbutton/plugin.js
@@ -159,10 +159,10 @@ CKEDITOR.plugins.add( 'colorbutton', {
 						colorToSelect;
 
 					for ( var i = 0; i < colors.length; i++ ) {
-						var color = CKEDITOR.tools.normalizeColor( colors[ i ] );
+						var color = colors[ i ];
 
 						if ( callback( color ) ) {
-							colorToSelect = color;
+							colorToSelect = normalizeColor( color );
 							break;
 						}
 					}
@@ -242,7 +242,7 @@ CKEDITOR.plugins.add( 'colorbutton', {
 								element = element.getParent();
 							}
 
-							currentColor = CKEDITOR.tools.normalizeColor( element.getComputedStyle( type == 'back' ? 'background-color' : 'color' ) );
+							currentColor = normalizeColor( element.getComputedStyle( type == 'back' ? 'background-color' : 'color' ) );
 							finalColor = finalColor || currentColor;
 
 							if ( finalColor !== currentColor ) {
@@ -257,7 +257,7 @@ CKEDITOR.plugins.add( 'colorbutton', {
 							finalColor = '';
 						}
 						if ( type == 'fore' ) {
-							colorData.automaticTextColor = '#' + CKEDITOR.tools.normalizeColor( automaticColor );
+							colorData.automaticTextColor = '#' + normalizeColor( automaticColor );
 						}
 						colorData.selectionColor = finalColor ? '#' + finalColor : '';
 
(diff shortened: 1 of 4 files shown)


plugins/colorbutton/plugin.js
```diff
@@ -154,6 +154,23 @@ CKEDITOR.plugins.add( 'colorbutton', {
 					attributes: { role: 'listbox', 'aria-label': lang.panelTitle }
 				},
 
+				select: function( callback ) {
+					var colors = config.colorButton_colors.split( ',' ),
+						colorToSelect;
+
+					for ( var i = 0; i < colors.length; i++ ) {
```
Comandeer: We have `CKEDITOR.tools.array#find()` method – it should increase the readability of this code.

In fact it would allow to pass `callback` directly to `colors`:

```javascript
var colorToSelect = normalizeColor( CKEDITOR.tools.array.find( colors, callback ) );
```

plugins/colorbutton/plugin.js
```diff
@@ -154,6 +154,23 @@ CKEDITOR.plugins.add( 'colorbutton', {
 					attributes: { role: 'listbox', 'aria-label': lang.panelTitle }
 				},
 
+				select: function( callback ) {
```
Comandeer: API docs for the new method are missing.

tests/plugins/colorbutton/manual/select.html
```diff
@@ -0,0 +1,74 @@
...
+<div id="editor"></div>
+<select id="choose-color"></select>
+<div id="selected-color"></span>
+
+<script>
+	var editor = CKEDITOR.replace( 'editor' );
+
+	editor.once( 'instanceReady', overwriteUIOpen( 'TextColor' ) );
+	editor.once( 'instanceReady', overwriteUIOpen( 'BGColor' ) );
+	editor.once( 'instanceReady', initializeColorSelector );
```
Comandeer: Can't it be joined into one listener?

plugins/colorbutton/plugin.js
jacekbogdanski: There was something wrong with this code but I couldn't catch up what (had some brain fart yesterday). Obviously, a find method, thanks :smile:

tests/plugins/colorbutton/manual/select.html
jacekbogdanski: It can, I just like to pass callbacks by reference, so it's more readable. But I'm fine for merging it into a single listener.

plugins/colorbutton/plugin.js
jacekbogdanski: I'm not sure about namespace of the docs :thinking: Probably `CKEDITOR.ui.colorButton` would be best but I'm not a fan of introducing empty namespace for this single private method. WDYT about just private docs i.e. commented out with `//` (so commenting private functions according to https://github.com/ckeditor/ckeditor4/blob/major/dev/docs/codestyle.md#210-comments)?

jacekbogdanski:
> Bender's toolbar hides color buttons by default. Maybe it's way to switch it off?
:+1: 
> I wonder why correct behavior of the select method depends on editor focus. If I don't focus the editor first, color button will be always pointing to the "Automatic":

Removed, this part didn't make sense as nothing bad happens if editor is not focused except that auto color won't be calculated.

plugins/colorbutton/plugin.js
Comandeer: Sounds good 👍

PR #3857, opened 2020-02-12
Expose public API for refreshing widget mask and parts

### LINKED ISSUE #3775: It should be possible to refresh size/position of widget partial mask via public API
## Type of report

Feature request

This is a follow up of #3240, where we have introduce a possibility to mask only part of the widget (see also https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_plugins_widget.html#property-mask). However, as reported in https://www.drupal.org/project/drupal/issues/3064572#comment-13283354 by @wimleers, there is an issue with mask refreshing due to specific Drupal Entity Embed CKEditor 4 plugin.

The Drupal Entity Embed plugin works in a way that it asynchronously fetches the entity HTML content (which may be slightly different depending on the entity type) and so there are two issues:

* It is not possible to properly define [`widget.parts`](https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_plugins_widget.html#property-parts) before entity HTML is fetched. Since fetching happens after widget initialisation and parts are processed on widget initialisation it becomes problematic. So the exact issue is that parts selector can be defined but entity HTML is not there on widget initialisation and would not be find.
* Since HTML is dynamic (and can be refetched) during widget lifecycle, it should be possible to trigger mask refreshing by the API call (so it can be done by Entity Embed plugin after fetching and inserting the entity HTML).

I know the issue with parts was somehow workaround by manually updating `widget.parts` but it is not a perfect solution.

To sum up, we need to enable handling cases where widget content is loaded asynchronously (after initialisation) and part of the loaded content is masked correctly.

Extracting method for refreshing mask should be just aliasing the private function we have now. Not sure about refreshing parts, since we need to check if refreshing parts after the widget is initialised doesn't break anything.

Dumluregn:

New feature

### This PR contains

- [x] Unit tests
- [x] Manual tests

```
*[#3775](https://github.com/ckeditor/ckeditor4/issues/3775): Widget [mask](https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_plugins_widget.html#property-mask) and [parts](https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_plugins_widget.html#property-parts) can now be refreshed dynamically via API calls.
```

## What changes did you make?

I created two public methods: `refreshMask()` and `refreshParts()` aliasing accordingly `setupMask()` and `setupParts()` private methods. They needed to be slightly adjusted to handle the case in which one widget part was created at widget initialisation and the other one didn't (e.g. because it wasn't rendered yet). For that reason I also updated existing unit test for setting up parts.

Also in unit test there was an issue with IE8 that I tried to resolve for a few hours but still didn't find out why the first call of `elementFromPoint()` after refreshing parts and mask returns `null` and just added the second call.

Closes #3775.


plugins/widget/plugin.js
```diff
@@ -3605,8 +3620,13 @@
 				partName;
 
 			for ( partName in widget.parts ) {
+				if ( typeof widget.parts[ partName ] !== "string" ) {
```
f1ames:
```suggestion
				if ( typeof widget.parts[ partName ] !== 'string' ) {
```

Use single quotes.

plugins/widget/plugin.js
```diff
@@ -3605,8 +3620,13 @@
...
+					parts[ partName ] = widget.parts[ partName ];
+					continue;
+				}
+
 				el = widget.wrapper.findOne( widget.parts[ partName ] );
-				parts[ partName ] = el;
+				parts[ partName ] = el ? el : widget.parts[ partName ];
```
f1ames: The CI is red (surprisingly a lot of tests started to fail) and it seems this line is problematic one. My assumption is that other widget functions and different plugins checks parts name for `null`. If it's not `null` they treat it as an `element` and try to perform typical elements operations on it.

It looks like a breaking change TBH and we can't change it easily. The only reasonable solution I'm able to think of as for now is to keep copy of initial `widget.parts` and use it for refreshing parts which weren't present during initialization :thinking: 

The only issue is that `widget.parts` is a part of a public API (see https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_plugins_widget_definition.html#property-parts) so probably one may try to change it directly via `widgetInstance.parts = ...` but I don't think we will have such case. We should just remember to mention in the `refreshParts` method docs that it works on initial parts and wouldn't cover a case when `parts` were updated manually after widget initialization.

plugins/widget/plugin.js
```diff
@@ -1420,6 +1420,21 @@
...
+		/**
+		 * Refreshes widget's mask. Can be used together with {@link CKEDITOR.plugins.widget#refreshParts} method to reinitialize mask
+		 * for dynamically created widgets.
+		 */
+		refreshMask: function() {
+			setupMask( this );
+		},
+
+		/**
+		 * Reinitializes widget's {@link CKEDITOR.plugins.widget#parts parts}.
```
f1ames: Some broader explanation could be useful, e.g. that it covers the case when widget HTML is async or modified during widget lifecycle and this method can be used to link new DOM elements to `parts`. Also that it covers only parts which were not initialized before (so with absent element on widget initialization).

plugins/widget/plugin.js
Dumluregn: As you suggested I created a new object called `partSelectors` which keeps original CSS selectors for parts and is used to reinitialize parts if they are refreshed.
Also we agreed to create a flag for user to decide if already existing parts should be reinitialized, but TBH I don't remember what should be default option 😄 I set it to `false` (so existing parts should be ignored), but I can changed it if we agreed otherwise.

f1ames: I think we agreed that by default it should refresh all parts and the flag will be a safe switch for some specific use cases when one needs to only refresh not yet initialized parts.

Dumluregn: Thanks, so I'll change it in a moment 👍

tests/plugins/widget/refreshmaskapi.js
```diff
@@ -0,0 +1,81 @@
...
+					evt.editor.dataProcessor.writer.sortAttributes = 1;
+				}
+			}
+		}
+	};
+
+	bender.test( {
+		// This test needs to always be the first in the suite because it uses 'elementFromPoint()' method
+		// which on IE breaks if there is any message about earlier tests displayed.
+		'test partial mask refresh': function() {
```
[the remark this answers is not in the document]
Dumluregn: I updated tests.
Two tests for `refreshParts()` method (the first is generic one and second tests flag - it worked 😛) are in `definition.js` file. Under the name `refreshmaskapi.js` we now have a new test for just refreshing mask, and the previous one (combining parts and mask refreshing) was moved under the name `refreshpartsandmask.js`.

f1ames:
Rebased onto latest `major`.

PR #3898, opened 2020-02-27
Introduce placeholdertext plugin

### LINKED ISSUE #3793: Introduce placeholdertext plugin
## Type of report

Feature request

Already covered by @Comandeer, we need to move it to `ckeditor4` repo and test if everything works fine. See https://github.com/Comandeer/ckeditor-plugin-placeholdertext.

Comandeer:

Feature

### This PR contains

- [x] Unit tests
- [x] Manual tests

```
*[#3793](https://github.com/ckeditor/ckeditor4/issues/<ISSUE_NUMBER>): Introduce [Placeholder Text](https://ckeditor.com/cke4/addon/placeholdertext) plugin.
```

## What changes did you make?

I've added `placeholdertext` plugin with basic tests.

Closes #3793.

### DIFF
--- a/plugins/placeholdertext/plugin.js
+++ b/plugins/placeholdertext/plugin.js
@@ -1,0 +1,114 @@
+/**
+ * @license Copyright (c) 2003-2020, CKSource - Frederico Knabben. All rights reserved.
+ * For licensing, see LICENSE.md or https://ckeditor.com/legal/ckeditor-oss-license
+ */
+
+/**
+ * @fileOverview Simple CKEditor 4 plugin that adds placeholder text to the editor.
+ */
+( function() {
+	/**
+	 * Namespace providing configuration for placeholdertext plugin..
+	 *
+	 * @singleton
+	 * @class CKEDITOR.plugins.placeholdertext
+	 */
+	var pluginNamespace = {
+		/**
+		 * Template, which will be used to create placeholder element.
+		 *
+		 * `{PLACEHOLDER}` is replaced by the value of CKEDITOR.config.placeholdertext.
+		 *
+		 * @property {String}
+		 */
+		template: '<p data-cke-placeholdertext>{PLACEHOLDER}</p>',
+
+		/**
+		 * Styles that would be applied to the placeholder element.
+		 *
+		 * @property {String}
+		 */
+		styles: '[data-cke-placeholdertext] {' +
+				'opacity: .8;' +
+				'color: #aaa;' +
+			'}'
+	};
+
+	function isEditorEmpty( editor ) {
+		return editor.getData().length === 0;
+	}
+
+	function applyPlaceholder( evt ) {
+		var editor = evt.listenerData.editor,
+			placeholder = pluginNamespace.template.replace( '{PLACEHOLDER}', editor.config.placeholdertext );
+
+		if ( isEditorEmpty( editor ) ) {
+			editor.editable().setHtml( placeholder );
+		}
+	}
+
+	function removePlaceholder( evt ) {
+		var editor = evt.listenerData.editor,
+			editable = editor.editable(),
+			range = editor.createRange(),
+			placeholder = editable.findOne( '[data-cke-placeholdertext]' );
+
+		if ( !placeholder) {
+			return;
+		}
+
+		editor.fire( 'lockSnapshot' );
+		placeholder.remove();
+		range.moveToElementEditStart( editable );
+		range.select();
+		editor.fire( 'unlockSnapshot' );
+	}
+
+	function fixGetData( evt ) {
+		var fragment = evt.data.dataValue,
+			children = fragment.children;
+
+
+		if ( children.length !== 1 ||
+			typeof children[ 0 ].attributes[ 'data-cke-placeholdertext' ] === 'undefined' ) {
+			return;
+		}
+
+		children[ 0 ].remove();
+	}
+
+	CKEDITOR.plugins.add( 'placeholdertext', {
+		onLoad: function() {
+			CKEDITOR.addCss( pluginNamespace.styles );
+		},
+
+		init: function( editor ) {
+			if ( !editor.config.placeholdertext ) {
+				return;
+			}
+
+			editor.on( 'contentDom', applyPlaceholder, null, { editor: editor } );
+			editor.on( 'focus', removePlaceholder, null, { editor: editor } );
+			editor.on( 'blur', applyPlaceholder, null, { editor: editor } );
+
+			editor.on( 'toDataFormat', fixGetData, null, null, 9 );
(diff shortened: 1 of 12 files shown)


plugins/placeholdertext/plugin.js
```diff
@@ -0,0 +1,125 @@
...
+	 */
+	var pluginNamespace = {
+		/**
+		 * Template, which will be used to create placeholder element.
+		 *
+		 * `{PLACEHOLDER}` is replaced by the value of CKEDITOR.config.placeholdertext.
+		 *
+		 * @property {String}
+		 */
+		template: '<p data-cke-placeholdertext>{PLACEHOLDER}</p>',
```
jacekbogdanski: We already have API for templates https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_template.html

plugins/placeholdertext/plugin.js
```diff
@@ -0,0 +1,125 @@
...
+				return;
+			}
+
+			editor.on( 'contentDom', applyPlaceholder, null, { editor: editor } );
+			editor.on( 'focus', removePlaceholder, null, { editor: editor } );
+			editor.on( 'blur', applyPlaceholder, null, { editor: editor } );
+
+			editor.on( 'toDataFormat', fixGetData, null, null, 9 );
+		}
+	} );
```
jacekbogdanski: Could you refactor this code according to https://github.com/ckeditor/ckeditor4/blob/master/dev/docs/codestyle.md#2231-public-api ?
Private functions should be moved below this code to promote discoverability.

plugins/placeholdertext/plugin.js
Comandeer: I've removed the template altogether.

Comandeer:
I've rewritten the whole code to the `::before` pseudoelement approach and added placeholder toggling to the `change` event (unfortunately there is no sensible way of testing it, as the editor always gain focus after invoking `insert*` methods). Only styles for `divarea` editors are still not perfect, yet I'm not sure how to deal with it. The discrepancy between the placeholder and standard text inside a paragraph is due to the browser's default paragraph styles – which can be different between browsers.

> Did you consider using `:before` pseudo-class instead?

The plugin was written without it as it had to work in IE8. As we now relaxed this requirement, I've switched to `::before`.

f1ames:
> The plugin was written without it as it had to work in IE8. As we now relaxed this requirement, I've switched to `::before`.

Remember to correctly prevent plugin code/initialization execution for not supported browsers (if it hasn't been done already, ofc).

plugins/placeholdertext/plugin.js
```diff
@@ -0,0 +1,97 @@
...
+		},
+
+		onLoad: function() {
+			CKEDITOR.addCss( CKEDITOR.plugins.placeholdertext.styles );
+		},
+
+		init: function( editor ) {
+			if ( !editor.config.placeholdertext ) {
+				return;
+			}
```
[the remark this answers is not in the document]
Comandeer: Plugin mechanism would take care, but it would also bind all listeners that are not needed for such a case.

Comandeer:
Ok, I've added additional styles for `divarea` editors.

> I'm also wondering if we should provide some default placeholder value if `config.placeholdertext` has not been set? In that case, if the user didn't set it, still will be able to see some placeholder which means that the feature is not broken, just not properly configured. If a placeholder would be set explicitly to an empty string (by a user) it should be probably respected, so we end up checking against `undefined`, not falsy value. WDYT?

IMO it's not necessary. First of all, we don't know the context in which the editor will be used. Secondly, we will have to translate the default placeholder to all language versions. Sounds like a lot of work for something that will be probably overwrite by the user.

I've also discovered that we are able to support IE8 if we use `:before` syntax instead of `::before` one, however IE8 has additional issues with correctly handling placeholder disappearing inside `iframe`\-based editor. I'm for not spending time on it and just assume that IE8 is not supported.

plugins/placeholdertext/plugin.js
```diff
@@ -0,0 +1,101 @@
...
+		init: function( editor ) {
+			if ( !this.isSupportedEnvironment() || !editor.config.placeholdertext ) {
+				return;
+			}
+
+			bindPlaceholderEvents( editor, [
+				'contentDom',
+				'focus',
+				'blur',
+				'change'
```
jacekbogdanski, line 28: `change` event is triggered pretty often. I'm not totally sure how big performance impact has `editor.getData` method used by the toggling function and overall the rest of the toggling implementation, but maybe it makes sense to throttle this function a bit? We could use one of the [throttle](https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_tools_buffers_throttle.html) implementations for that. WDYT?

tests/plugins/placeholdertext/manual/placeholdertext.md
```diff
@@ -0,0 +1,13 @@
...
+
+Check if:
+
+* placeholder text is present inside the empty editor after its initialization;
+* placeholder text is not present inside the non-empty editor after its initialization;
+* placeholder text is hidden when editor is focused;
+* after deleting whole content from the editor and blurring it (both via keyboard and mouse), placeholder text appears;
+* removing/adding placeholder does not create undo step;
+* placeholder text is semi-transparent;
+* placeholder text is not returned by call to `editor.getData()`.
```
jacekbogdanski, line 13: It's a bit misleading request as it indicates you should open console for `editor.getData()` call, but there is a button available for that. Maybe we could reword this phrase adding information about the button, or rename button itself into something like `Call editor.getData()`?

plugins/placeholdertext/plugin.js
Comandeer: TBH I'd skip adding throttle unless there are real performance issues with the current solution.

Comandeer:
> There is some weird issue on Safari \[…\]

It's a [well-known bug in Safari](https://bugs.webkit.org/show_bug.cgi?id=112854), AFAIR we had even some kind of workaround for it, but it got removed after being fixed in Chrome.

Comandeer:
Ok, I've

*   added tests for `change` event,
*   fixed jumping of placeholder (`position: absolute` worked; however not in IE9, the jumping is still visible there),
*   reworded manual test,
*   fix placeholder for full-page editors.

jacekbogdanski:
> > There is some weird issue on Safari […]
> 
> It's a [well-known bug in Safari](https://bugs.webkit.org/show_bug.cgi?id=112854), AFAIR we had even some kind of workaround for it, but it got removed after being fixed in Chrome.

:+1: 

> however not in IE9, the jumping is still visible there

We can back to IE9 if someone reports the issue :+1:

PR #3907, opened 2020-03-01
Make getclipboard utility configurable

### LINKED ISSUE #3329: Make getClipboard utility configurable
## Type of report

Task

Currently `dev/pastefromword/getclipboard.html` is very inflexible: it has one, hardcoded configuration. However now we're extending our paste handling with `pastetools` and `pastefromgdocs`. It would be nice to make this utility more flexible:

* choose which paste plugin to use;
* choose which plugins to load;
* provide additional config (e.g. `allowedContent`/`extraAllowedContent`).

## Other details

* Installed CKEditor plugins: pastefromword, pastetools, pastefromgdocs

Comandeer:

Task

### This PR contains

- [ ] Unit tests
- [ ] Manual tests

None, it's internal change.

## What changes did you make?

I've added checkboxes that allow to change the feature set of the editor. Additionally I've provided way to configure ACF.

Maybe it's not pretty, but it works.

Closes #3329.


dev/pastetools/getclipboard.html
```diff
@@ -0,0 +1,309 @@
...
+
+			}, null, null, -1 );
+
+			editor.on( 'paste', function( evt ) {
+				setTimeout( function() {
+					document.getElementById( 'output' ).value = editor.getData();
+				}, 0 );
+			}, null, null, 999 );
+		}
+	}
```
jacekbogdanski: This code is unreadable - I know it's mostly copy&paste from the previous example, but maybe you could extract some common variables (like `document.getElementById( 'emHtml' )`) and overall refactor it a bit, so we are the last one who needs to understand this complicated code?

dev/pastetools/getclipboard.html
```diff
@@ -0,0 +1,309 @@
...
+			overflow: auto;
+			margin-bottom: 20px;;
+		}
+		label {
+			display: block;
+			line-height: 2;
+		}
+		span {
+			background-color: #88FFAA;
+		}
```
jacekbogdanski: Using styles in which such small specificity may also change editor styling. It's e.g. visible when opening emoji completion where items have a green background.

dev/pastetools/getclipboard.html
```diff
@@ -0,0 +1,309 @@
...
+		</div>
+		<div id="editable"></div>
+		<div id="paste"></div>
+		<div id="basic"></div>
+		<div id="image"></div>
+		<div id="other"></div>
+	</div>
+
+	<script>
+	var editablePlugins = [
```
jacekbogdanski: I'm wondering if there is a better method to list plugins. They are nicely labeled which is a +, but we will need to update this file every time when a new plugin has been added. Theoretically, we could just list plugins inside `plugins/*` folder, because this file uses repository editor instance. Not sure if it's worth the effort though, and we will lose plugins sections. WDYT?

dev/pastetools/getclipboard.html
```diff
@@ -0,0 +1,309 @@
...
+
+			return oldInstance.destroy( true );
+		}
+
+		createEditor();
+
+		function createEditor() {
+			var textarea = new CKEDITOR.dom.element( 'textarea' ),
+				allowedContent = CKEDITOR.document.getById( 'allowedContent' ),
+				extraAllowedContent = CKEDITOR.document.getById( 'extraAllowedContent' ),
```
jacekbogdanski: Maybe it makes also sense to add `disallowedContent` option, so we can prefilter HTML a bit for cleaner comparison (where some attributes or styles add noise we would like to skip)?

dev/pastetools/getclipboard.html
Comandeer: That was my first approach. I prepared simple Node.js script that just iterated through all plugins and generate HTML code from it… But there was too much noise, e.g. `button`, `panel` etc. Unfortunately we have too many abstract plugins to use such approach. As I had to filter plugins, I've decided to just do it _my way_ ;)

Comandeer:
Ok, I've restyled the whole thing and refactored code, enhancing also support for IE8.

dev/pastetools/getclipboard.html
jacekbogdanski: Make sense to me :+1:

PR #3930, opened 2020-03-15
Fix setting urlRegex and emailRegex properties in config

### LINKED ISSUE #3156: It is not possible to set autolink_urlRegex and autolink_emailRegex properties of autolink plugin
## Type of report

Bug

1. Download CKEditor with autolink plugin.
2. Use one of below code versions to configure CKEditor:
```
CKEDITOR.config.autolink_urlRegex = /^(https?|ftp|file):\/\/(-\.)?([^\s\/?\.#]+\.?)+(\/[^\s]*)?[^\s\.,]$/i;
var editor = CKEDITOR.replace( 'editor1', {extraPlugins : 'autolink'});
```
```
var editor = CKEDITOR.replace( 'editor1', {
    extraPlugins : 'autolink',
    autolink_urlRegex : /^(https?|ftp|file):\/\/(-\.)?([^\s\/?\.#]+\.?)+(\/[^\s]*)?[^\s\.,]$/i
});
```
2. Load CKEditor and type e.g.  `file://C:/test/`  and press space
(I know the URL is invalid and the regex should be changed becaue it match `file:///C:/test/` but this is just for simplicity of testing)

### Expected result

Text should get linked.

### Actual result

Text doesn't get linked

## Other details

This is happening because plugin code uses e.g. `CKEDITOR.config.autolink_emailRegex` instead of `editor.config.autolink_emailRegex`:
https://github.com/ckeditor/ckeditor-dev/blob/master/plugins/autolink/plugin.js#L128
https://github.com/ckeditor/ckeditor-dev/blob/master/plugins/autolink/plugin.js#L103
https://github.com/ckeditor/ckeditor-dev/blob/master/plugins/autolink/plugin.js#L127
what makes impossible to overwrite values for these properties.

* Browser: Any
* OS: Any
* CKEditor version: 4.11.0+
* Installed CKEditor plugins: autolink

serggoodwill:

Bug fix

### This PR contains

- [x] Unit tests
 - [x] Manual tests

```
*[#3156](https://github.com/ckeditor/ckeditor4/issues/3156): It is not possible to set autolink_urlRegex and autolink_emailRegex properties of autolink plugin
```

## What changes did you make?

Changed variable usage `autolink_urlRegex` and `autolink_emailRegex` to allow use `editor.config`

Closes #3156.

### DIFF
--- a/plugins/autolink/plugin.js
+++ b/plugins/autolink/plugin.js
@@ -38,9 +38,9 @@
 				return;
 			}
 
-			var commitKeystrokes = editor.config.autolink_commitKeystrokes || CKEDITOR.config.autolink_commitKeystrokes;
+			// (#3156)
 			editor.on( 'key', function( evt ) {
-				if ( editor.mode !== 'wysiwyg' || CKEDITOR.tools.indexOf( commitKeystrokes, evt.data.keyCode ) == -1 ) {
+				if ( editor.mode !== 'wysiwyg' || CKEDITOR.tools.indexOf( editor.config.autolink_commitKeystrokes, evt.data.keyCode ) == -1 ) {
 					return;
 				}
 
@@ -100,7 +100,7 @@
 						text: text,
 						link: text.replace( doubleQuoteRegex, '%22' )
 					},
-					template = opts.link.match( CKEDITOR.config.autolink_urlRegex ) ?
+					template = opts.link.match( editor.config.autolink_urlRegex ) ?
 						urlTemplate.output( opts )
 						: emailTemplate.output( opts );
 
@@ -124,8 +124,8 @@
 			}
 
 			function matchLink( query ) {
-				return query.match( CKEDITOR.config.autolink_urlRegex ) ||
-					query.match( CKEDITOR.config.autolink_emailRegex );
+				return query.match( editor.config.autolink_urlRegex ) ||
+					query.match( editor.config.autolink_emailRegex );
 			}
 		}
 	} );
--- a/tests/plugins/autolink/autolink.js
+++ b/tests/plugins/autolink/autolink.js
@@ -14,6 +14,15 @@
 				removePlugins: 'link'
 			}
 		},
+		optionalParameters: {
+			config: {
+				allowedContent: true,
+				pasteFilter: null,
+				removePlugins: 'link',
+				autolink_urlRegex: /^https:\/\/foobar.com$/,
+				autolink_emailRegex: /^foo@foobar\.com$/
+			}
+		},
 		encodedDefault: {
 			config: {
 				allowedContent: true,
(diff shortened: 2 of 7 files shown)


plugins/autolink/plugin.js
```diff
@@ -38,7 +38,11 @@
 				return;
 			}
 
-			var commitKeystrokes = editor.config.autolink_commitKeystrokes || CKEDITOR.config.autolink_commitKeystrokes;
+			// (#3156)
+			var commitKeystrokes = editor.config.autolink_commitKeystrokes || CKEDITOR.config.autolink_commitKeystrokes,
+				emailRegex = editor.config.autolink_emailRegex || CKEDITOR.config.autolink_emailRegex,
+				urlRegex = editor.config.autolink_urlRegex || CKEDITOR.config.autolink_urlRegex;
```
jacekbogdanski, line 44: Indeed, using editor config fixes the issue. Actually, `editor.config` will be populated automatically by editor itself with `CKEDITOR.config` if no option has been directly set on `editor.config.`. That means we could safely remove "or" (`||`) condition for regex aliases (and commit keystrokes also).

tests/plugins/autolink/autolink.js
```diff
@@ -14,6 +14,15 @@
 				removePlugins: 'link'
 			}
 		},
+		optionalParameters: {
+			config: {
+				allowedContent: true,
+				pasteFilter: null,
+				removePlugins: 'link',
+				autolink_urlRegex: /^(https?|ftp):\/\/(-\.)?([^\s\/?\.#]+\.?){5}(\/\S*)?[^\s\.,]$/i,
+				autolink_emailRegex: /^[a-zA-Z0-9.!#$%&'*+\/=?^_`{|}~-]+@[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?(?:\.[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?)*$/
```
jacekbogdanski, line 23: These regexes are pretty complicated and as we only need to check if custom regex has been used (instead of the default one). Maybe it could be simplified to something like `/^http:\/\/foobar\.com$/`, so the expected result is crystal clear? The same for the email address.

tests/plugins/autolink/autolink.js
```diff
@@ -134,6 +143,29 @@
 			}
 		},
 
+		// (#3156)
+		'test various valid URL links with optional regex': function() {
```
jacekbogdanski, line 147: Unfortunately, all unit tests (not only this one) seem to be falsely positive due to similarity with default regex. It can be easily verified by checking out `autolink` plugin file into the upstream branch. E.g. `git checkout major -- plugins/autolink/plugin.js` and running bender tests. Tests should pass only on PR branch. Simplifying regexes should fix the issue.

PR #3933, opened 2020-03-16
https: Update LICENSE.md

dolibit-ut:
update license URLs to https://


jacekbogdanski:
Hello @dolibit-ut,

Since we already updated the license file for the `master` branch, there is no need to update `major` as it will be automatically updated when syncing both branches.

PR #3939, opened 2020-03-18
Typo in Swedish

sergeyvdovareize:
My Swedish colleagues say it should be `Matematik `instead of `Mattematik`


f1ames:
Thank you very much for contribution! We're using [Transifex](https://www.transifex.com/ckeditor/ckeditor/) for managing language strings. Please send your contribution there, so that we can pull it into upcoming release.

Thanks!

PR #3957, opened 2020-03-27
Fix issue with loosing keybindings for autocomplete panel with sourcemode

### LINKED ISSUE #3938: Autocomplete loses navigation keybindings after switching source mode
## Type of report

Bug

1. Open any sample with autocomplete and sourcearea plugins.
2. Switch between source editing and content editing.
3. Trigger autocompletion.
4. Navigate using arrow keys.

### Expected result

You can navigate through the completion panel using a keyboard.

### Actual result

Keybindings are removed.

## Other details

Probably standard issue with incorrect event listener on editable where we should use `attachListener` instead:
https://github.com/ckeditor/ckeditor4/blob/5ace43e789f84c74f74fb1b163715aa2ba04e826/plugins/autocomplete/plugin.js#L307-L309

jacekbogdanski:

Bug fix

### This PR contains

- [x] Unit tests
- [x] Manual tests

```
*[#3938](https://github.com/ckeditor/ckeditor4/issues/3938): Fixed: Autocomplete loses navigation keybindings after switching source mode.
```

## What changes did you make?

Corrected attaching events to editable.

Closes #3938

### DIFF
--- a/plugins/autocomplete/plugin.js
+++ b/plugins/autocomplete/plugin.js
@@ -288,26 +288,28 @@
 				this.viewRepositionListener();
 			}, this ) );
 
-			this._listeners.push( editor.on( 'contentDom', onContentDom, this ) );
-
 			// Don't let browser to focus dropdown element (#2107).
 			this._listeners.push( this.view.element.on( 'mousedown', function( e ) {
 				e.data.preventDefault();
 			}, null, null, 9999 ) );
 
-			// Attach if editor is already initialized.
+			// Register keybindings if editor is already initialized.
 			if ( editable ) {
-				onContentDom.call( this );
+				this.registerPanelNavigation();
 			}
 
-			function onContentDom() {
-				// Priority 5 to get before the enterkey.
-				// Note: CKEditor's event system has a limitation that one function (in this case this.onKeyDown)
-				// cannot be used as listener for the same event more than once. Hence, wrapper function.
-				this._listeners.push( editable.on( 'keydown', function( evt ) {
-					this.onKeyDown( evt );
-				}, this, null, 5 ) );
-			}
+			editor.on( 'contentDom', this.registerPanelNavigation, this );
+		},
+
+		registerPanelNavigation: function() {
+			var editable = this.editor.editable();
+
+			// Priority 5 to get before the enterkey.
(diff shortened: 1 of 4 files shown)


tests/plugins/autocomplete/autocomplete.js
```diff
@@ -157,6 +162,32 @@
 			wait();
 		},
 
+		// (#3938)
+		'test navigation keybindings are registered after changing mode': function() {
```
f1ames: This test fails on IE9:

image

Since manual test passes it seems the fix is working correctly there. If this test is not super easy to fix there, we may just ignore it.

plugins/autocomplete/plugin.js
```diff
@@ -288,26 +288,28 @@
...
+			editor.on( 'contentDom', this.registerPanelNavigation, this );
+		},
+
+		registerPanelNavigation: function() {
+			var editable = this.editor.editable();
+
+			// Priority 5 to get before the enterkey.
+			// Note: CKEditor's event system has a limitation that one function (in this case this.onKeyDown)
+			// cannot be used as listener for the same event more than once. Hence, wrapper function.
+			this._listeners.push( editable.attachListener( editable, 'keydown', function( evt ) {
```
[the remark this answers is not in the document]
jacekbogdanski: Yes, keeping old editable instance in a closure was the main issue here, but as you wrote, `attachListener` suits better here.

tests/plugins/autocomplete/autocomplete.js
jacekbogdanski: It looks like IE9 has some weird problem with removing editable listeners once editable has been detached. I was able to reproduce the issue only using bender (it works fine using samples). It seems like dropping destroying autocomplete removes this issue. Destroying autocomplete is not really important in this test as it uses its own editor instance, so it's probably better than ignoring this test on IE9, WDYT?

f1ames: > Destroying autocomplete is not really important in this test as it uses its own editor instance, so it's probably better than ignoring this test on IE9, WDYT?

Makes sense for me. It's always better to have a working test :+1:

f1ames:
Rebased onto latest `master`.

PR #4010, opened 2020-04-30
Allow users to provide absolute css units to editor.resize function

### LINKED ISSUE #1883: editor.resize() doesn't work with CSS units

Bug

1. Open sample CKEditor.
2. Open console, and type `CKEDITOR.instances[ 'your-instance-name' ].resize( '500em', '500em' )`

### Expected result

Editor is resized.

### Actual result

Nothing happens

## Other details

According to [docs](https://docs.ckeditor.com/ckeditor4/latest/api/CKEDITOR_editor.html#method-resize) you should be able to pass
> a CSS size value with unit as an argument.

Note: None of other CSS are working.

* Browser: Any
* OS: Any
* CKEditor version: 4.9.2
* Installed CKEditor plugins: resize

hub33k:

Bug fix

### This PR contains

- [x] Unit tests
- [x] Manual tests

```
*[#1883](https://github.com/ckeditor/ckeditor4/issues/1883): Allow users to pass absolute css units to `editor.resize` function.
```

## What changes did you make?

Allow users to pass absolute css units to `editor.resize` function.

Closes #1883.


core/dom/element.js
```diff
@@ -2243,15 +2243,22 @@ CKEDITOR.dom.element.clearMarkers = function( database, element, removeFromDatab
...
 		if ( typeof size == 'number' ) {
 			if ( isBorderBox && !( CKEDITOR.env.ie && CKEDITOR.env.quirks ) )
 				size -= marginAndPaddingSize.call( this, type );
 
-			this.setStyle( type, size + 'px' );
+			this.setStyle( type, CKEDITOR.tools.cssLength( size ) );
+		} else if ( typeof size == 'string' ) {
+			size = CKEDITOR.tools.convertToPx( size );
+
+			if ( typeof size != 'string' && isBorderBox && !( CKEDITOR.env.ie && CKEDITOR.env.quirks ) )
```
jacekbogdanski, line 2258: Is `typeof size != 'string' && ` condition required at this point?

core/dom/element.js
```diff
@@ -2243,15 +2243,22 @@ CKEDITOR.dom.element.clearMarkers = function( database, element, removeFromDatab
...
+				size -= marginAndPaddingSize.call( this, type );
+
+			this.setStyle( type, CKEDITOR.tools.cssLength( size ) );
```
jacekbogdanski, line 2261: Note that currently, the whole logic is duplicated with small difference between unit type. If we would convert the unit into pixels at the beginning like it has been done for `themedui`, the rest of the code could be left untouched.

tests/core/creators/manual/resizewithcssunits.md
```diff
@@ -0,0 +1,15 @@
+@bender-tags: bug, 4.14.1, 1883
+@bender-ui: collapsed
+@bender-ckeditor-plugins: toolbar, wysiwygarea, table
+
+1. Open browser developer tools.
+
+1. In console paste: `CKEDITOR.instances[ 'editor' ].resize( '20em', '10em' )`
```
jacekbogdanski, line 7: Using developer tools to test this issue is not the most optimal approach as we cannot test it on mobiles. It also requires additional test steps. Maybe we could go with more complex test e.g. by adding CSS style inputs?
image

core/creators/themedui.js
```diff
@@ -289,14 +289,20 @@ CKEDITOR.replaceClass = 'ckeditor';
...
+
 		// Get the height delta between the outer table and the content area.
 		var contentsOuterDelta = ( outer.$.offsetHeight || 0 ) - ( contents.$.clientHeight || 0 ),
 
 		// If we're setting the content area's height, then we don't need the delta.
 			resultContentsHeight = Math.max( height - ( isContentHeight ? 0 : contentsOuterDelta ), 0 ),
 			resultOuterHeight = ( isContentHeight ? height + contentsOuterDelta : height );
 
-		contents.setStyle( 'height', resultContentsHeight + 'px' );
+		if ( !isNaN( resultContentsHeight ) ) {
```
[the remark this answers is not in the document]
hub33k: > At what point `resultContentsHeight` could not be a number?

If user pass height as percent value `convertToPx` function will leave this value as is so these calculation will return `NaN` value.

I think we shouldn't support percent value for height here, WDYT?

core/dom/element.js
hub33k: Yes, if user passes percent value as width then `size` is still type of string. I can simplify it.

core/dom/element.js
hub33k: Yup, I can simplify it :+1:

jacekbogdanski:
https://github.com/ckeditor/ckeditor4/pull/4010#discussion_r419413939

You are right, sorry I point you to wrong direction, didn't now that `convertToPx` works soo picky and seems to handle only absolute units. I'm wondering if we would be able to improve `convertToPx`, so it also correctly handles relative units like percentage and also viewport, initial etc.

It may be possible that it will be easier to fix the issue directly inside `editor#resize` method than fight with `convertToPx`, because converting units like % or viewport size could require iterating to parent. It seems like `convertToPx` may require some rewriting or docs rewording though, as it incorrectly implies that handles `CSS length value`, which is not 100% true.

Could you check if we can operate on CSS lengths instead inside `editor#resize` method or successfully fix `convertToPx` method? The latter may be more complicated, but more useful in the long run.

tests/core/creators/manual/resizewithcssunits.md
```diff
@@ -0,0 +1,39 @@
...
+
+1. Press `Resize!` button.
+
+## Scenario 2
+
+1. In Width input type `20cm`.
+
+1. In Height input type `10cm`.
+
+1. Tick `is content height` checkbox.
```
jacekbogdanski, line 19: It doesn't explain how this option affects editor size. E.g. that the new size should be applied only to the editor's content. That means for each scenario we should have different expected results.

tests/core/creators/manual/resizewithcssunits.md
```diff
@@ -0,0 +1,39 @@
...
+
+1. Press `Resize!` button.
+
+## Scenario 3
+
+1. In Width input type `20cm`.
+
+1. In Height input type `10cm`.
+
+1. Tick `resize inner` checkbox.
```
jacekbogdanski, line 29: The same, it doesn't explain how this option affects editor size.

tests/core/creators/manual/resizewithcssunits.md
```diff
@@ -0,0 +1,39 @@
+@bender-tags: bug, 4.14.1, 1883
+@bender-ui: collapsed
+@bender-ckeditor-plugins: toolbar, wysiwygarea, table
+
+## Scenario 1
+
+1. In Width input type `20cm`.
```
jacekbogdanski, line 7: It would be nice to add additional units to test - the most important check out if this feature still works with number units. You can get a list of absolute units from https://developer.mozilla.org/en-US/docs/Learn/CSS/Building_blocks/Values_and_units

A simple list with these units and information to use them for calculation (by repeating test steps) should be enough.

tests/core/creators/manual/resizewithcssunits.html
```diff
@@ -0,0 +1,65 @@
...
+	[type="checkbox"] + label {
+		display: inline-block;
+	}
+
+	button {
+		margin-bottom: 15px;
+	}
+</style>
+
+<form class="input-container">
```
jacekbogdanski, line 19: There is no need to create a form for this simple manual test and fight with whole page refreshing on IE. We are not submitting anything here, so it may be better to just use JS.

tests/core/creators/manual/resizewithcssunits.html
```diff
@@ -0,0 +1,65 @@
...
+<script>
+	CKEDITOR.replace( 'editor' );
+
+	var resizeButton = document.getElementById( 'resize-button' ),
+		editorWidth = document.getElementById( 'width' ),
+		editorHeight = document.getElementById( 'height' ),
+		isContentHeight = document.getElementById( 'isContentHeight' ),
+		resizeInner = document.getElementById( 'resizeInner' );
+
+	resizeButton.addEventListener( 'click', function ( ev ) {
```
jacekbogdanski, line 56: This test won't work on IE8, because [`addEventListener`](https://caniuse.com/#search=addeventlistener) is not supported. As a rule of thumb, it's best to use CKEditor API when possible as it includes all necessary polyfills to work with older browsers.

Here, elements could be fetched using `CKEDITOR.document.getById` method and listener attached by `element.on`.

core/creators/themedui.js
```diff
@@ -253,12 +253,12 @@ CKEDITOR.replaceClass = 'ckeditor';
 	 *
 	 *		editor.resize( 900, 300 );
 	 *
-	 *		editor.resize( '100%', 450, true );
+	 *		editor.resize( '5in', 450, true );
```
jacekbogdanski, line 256: Using tabs for code listings is a deprecated method. Could you wrap it with markdown code listings i.e. backticks and programming language? E.g. see https://github.com/ckeditor/ckeditor4/blob/8d5d12dae4ee1fbd02c68db4c2a916957c204d87/plugins/autocomplete/plugin.js#L49-L101

core/creators/themedui.js
```diff
@@ -253,12 +253,12 @@ CKEDITOR.replaceClass = 'ckeditor';
...
 	 *
 	 * @param {Number/String} width The new width. It can be an integer denoting a value
-	 * in pixels or a CSS size value with unit.
+	 * in pixels or a CSS size value with absolute unit.
 	 * @param {Number/String} height The new height. It can be an integer denoting a value
-	 * in pixels or a CSS size value with unit.
+	 * in pixels or a CSS size value with absolute unit.
```
jacekbogdanski, line 261: I'm wondering about being more explicit here, as previously these docs have been incorrect. Maybe instead of adding information about absolute units for each param, we could add more generic information right below `Resizes the editor interface`?

```
Resizes the editor interface.

**Note:** Since 4.14.1 this method accepts only numeric or absolute CSS length units.
```

core/creators/themedui.js
```diff
@@ -283,20 +283,28 @@ CKEDITOR.replaceClass = 'ckeditor';
...
+
 		// Set as border box width. (https://dev.ckeditor.com/ticket/5353)
 		outer.setSize( 'width', width, true );
 
 		// WebKit needs to refresh the iframe size to avoid rendering issues. (1/2) (https://dev.ckeditor.com/ticket/8348)
 		contentsFrame && ( contentsFrame.style.width = '1%' );
 
+		if ( typeof height == 'string' ) {
+			height = CKEDITOR.tools.convertToPx( height );
+		}
```
jacekbogdanski, line 298: Due to the issue where `convertToPx` works only with CSS lengths, we could convert every unit into CSS length at the beginning instead. I can suggest extracting separate function (for converting units) for readability.

tests/core/creators/themedui.js
```diff
@@ -47,6 +47,37 @@
...
+				assert.areSame( CKEDITOR.tools.convertToPx( height ), lastResizeData.outerHeight );
+				assert.areSame( CKEDITOR.tools.convertToPx( width ), lastResizeData.outerWidth );
+				assert.areSame( getEditorContentHeight( editor ), lastResizeData.contentsHeight );
+				assert.areSame( CKEDITOR.tools.convertToPx( height ), getEditorOuterHeight( editor ) );
+
+				editor.resize( width, height, true );
+				assert.areSame( getEditorOuterHeight( editor ), lastResizeData.outerHeight );
+				assert.areSame( CKEDITOR.tools.convertToPx( width ), lastResizeData.outerWidth );
+				assert.areSame( getEditorContentHeight( editor ), lastResizeData.contentsHeight );
+				assert.areSame( CKEDITOR.tools.convertToPx( height ), getEditorContentHeight( editor ) );
```
jacekbogdanski, line 79: It would be nice to extract it as a separate test. Could you also add a similar test for `resizeInner` parameter?

Issue #2451, opened 2018-10-03
Remove Format changes selection when applied to unsupported element


jswiderski:
## Type of report

Bug

## Provide detailed reproduction steps (if any)

1. Open nightly sample - http://nightly.ckeditor.com/18-10-02-06-04/full/samples/.
2. Clear editor contents and type any text.
3. Select the text and set Formatting to "Heading1" and then apply the Numbered list.
4. Click on Remove Format button (Notice that Format dropdown no longer shows "Heading 1" at this point)
5. Change the formatting in the Format dropdown to "Heading2"

### Expected result

Format dropdown should display "Heading 2".

### Actual result

Format dropdown displays "Format".
Please note however that new Format gets applied properly. Only the selection is being moved from `H1` element to its parent `ol`. If you change the selection and click inside the `H1`, the Format dropdown will show it correctly again.

## Other details

* Browser: Any
* OS: Any
* CKEditor version: 4.0+
* Installed CKEditor plugins: Format Dropdown.

When Remove Format is applied, it looks like selection is being moved from the actual element to its parent. This behavior is in my opinion correct for inline elements because when format is removed, the inline element (like span or strong) is being removed as well. When it comes to block-level elements, things get more complicated. By default [removeFormatTags](https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_config.html#cfg-removeFormatTags) is only be applied to inline elements and documentation say that only inline elements can be used on that list. There is however nothing stopping from setting this configuration option to `config.removeFormatTags = 'h1,div,blockquote'`. What is more, such setup will work properly (provided that start of element is selected) when removing formatting form H1 for example (tag will get removed). Since the plugin works ok for both types of tags and the only thing which says otherwise is one sentence in documentation, what should we do?
1. Apply Remove Format to tag only if it is on the list? That way we could avoid selection changes for not supported tags and leave current behavior untouched. Best solution IMHO.
2. Block the block-level tags permanently? Disallow them on the list and ignore them when Remove Format is applied?
3. Ignore it and leave it as it is? Kind of makes sense because this plugin should not be used for not supported tag in the first place but on the other hand how could the user know that.

Screen cast: [remove format.zip](https://github.com/ckeditor/ckeditor-dev/files/2441540/remove.format.zip)

Comandeer:
The fix was merged into 4.11.2.

PR #4049, opened 2020-05-12
Remove format button should work with collapsed selection

### LINKED ISSUE #4008: Remove Format button no longer works when no text is selected
## Type of report

Bug

1. Load the editor version 4.14.0: https://jsfiddle.net/xda4omnr/1/
2. Select "bold" and "underlined"
3. Type: `Hello, `
4. Click "Remove Format" button
5. Type: `world!`

### Expected result

The text `world!` has normal style.

### Actual result

The text `world!` is bold and underlined.

## Other details

This feature used to work before - up to and including version 4.11.1. See https://jsfiddle.net/xda4omnr/2/ where the feature works.

The culprit which introduced the regression is most likely this commit: https://github.com/ckeditor/ckeditor4/pull/2625/commits/4a2929658c62f15d0c721e0e5c3027b7be40cf09

Note that you can still remove formatting after selecting some text, but some users use this button just to reset their styling from a particular moment on without selecting any text.

* Browser: Chrome
* OS: Mac OS
* CKEditor version: 4.14.0
* Installed CKEditor plugins: full/all

jacekbogdanski:

Bug fix

### This PR contains

- [x] Unit tests
- [x] Manual tests

```
*[#4008](https://github.com/ckeditor/ckeditor4/issues/4008): Fixed: [Remove format](https://ckeditor.com/cke4/addon/removeformat) doesn't work with collapsed selection.
```

## What changes did you make?

I reverted changes introduced in https://github.com/ckeditor/ckeditor4/pull/2625 as they seem not necessary and adds this regression. Remove format operates only on inline elements, which is nicely documented here: https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_config.html#cfg-removeFormatTags
Also, it seems like remove format indeed only breaks inline elements:https://github.com/ckeditor/ckeditor4/blob/846b07e0d7780ff1826685ed7a9f69c169c0c6d1/plugins/removeformat/plugin.js#L63-L64

so, enlarging range only for inline elements may be a good approach. I didn't note any special issues with skipping block elements and the issue https://github.com/ckeditor/ckeditor4/issues/2451 is still covered.

Closes #4008

### DIFF
--- a/tests/plugins/removeformat/plugin.js
+++ b/tests/plugins/removeformat/plugin.js
@@ -173,5 +173,15 @@ bender.test(
 		editor.execCommand( 'removeFormat' );
 
 		assert.beautified.html( html, bender.tools.selection.getWithHtml( editor ), { customFilters: [ filter ] } );
+	},
+
+	// (#4008)
+	'test remove format with collapsed selection': function() {
+		this.editorBot.setHtmlWithSelection( '<p><strong>Hello,^</strong></p>' );
+
+		this.editor.execCommand( 'removeFormat' );
+		this.editor.insertText( 'World!' );
+
+		assert.areEqual( '<p><strong>Hello,</strong>World!</p>', this.editor.getData() );
 	}
 } );
--- a/tests/plugins/removeformat/plugin.js
+++ b/tests/plugins/removeformat/plugin.js
@@ -177,11 +177,13 @@ bender.test(
 
 	// (#4008)
 	'test remove format with collapsed selection': function() {
+		var editor = this.editor;
+
 		this.editorBot.setHtmlWithSelection( '<p><strong>Hello,^</strong></p>' );
 
-		this.editor.execCommand( 'removeFormat' );
-		this.editor.insertText( 'World!' );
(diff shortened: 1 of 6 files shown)


tests/plugins/removeformat/plugin.js
```diff
@@ -173,5 +173,15 @@ bender.test(
 		editor.execCommand( 'removeFormat' );
 
 		assert.beautified.html( html, bender.tools.selection.getWithHtml( editor ), { customFilters: [ filter ] } );
+	},
+
+	// (#4008)
+	'test remove format with collapsed selection': function() {
+		this.editorBot.setHtmlWithSelection( '<p><strong>Hello,^</strong></p>' );
+
+		this.editor.execCommand( 'removeFormat' );
```
Comandeer, line 182: `this.editor` is used several times in the test – maybe extract it to a variable?

jacekbogdanski:
Weird, it seems to work correctly for me (on IE11):

removeformat

Nevertheless, changing this test to always request full input should cover some edge cases :+1:

Issue #3713, opened 2019-12-05
Failing tests /tests/plugins/pastetools/filter/parsestyles


jacekbogdanski:
## Type of report

Failing test

The whole test suite is failing `/tests/plugins/pastetools/filter/parsestyles`.

Also, see https://github.com/ckeditor/ckeditor4/issues/3424

* Browser: Edge
* CKEditor version: 4.13.1

Dumluregn:
This issue is no longer reproducible in the latest Edge browser version.

Issue #3794, opened 2020-01-17
Failing tests on major


Comandeer:
## Type of report

Failing tests

### IE 8

<img width="967" alt="" src="https://user-images.githubusercontent.com/1078728/72612895-5d035600-392e-11ea-9a2f-46da557a6ae1.png">

### IE 11

<img width="570" alt="" src="https://user-images.githubusercontent.com/1078728/72612930-72788000-392e-11ea-887c-2d30185ae18b.png">

### Edge

<img width="1258" alt="" src="https://user-images.githubusercontent.com/1078728/72612954-83c18c80-392e-11ea-8e0e-e60e8cf2857d.png">

### Safari

<img width="598" alt="" src="https://user-images.githubusercontent.com/1078728/72613019-c1beb080-392e-11ea-97ef-00aa5918b984.png">

Dumluregn:
Updated the post after the `4.14.1` testing phase.

sculpt0r:
Due to the fact that all listed issues are closed - I'm also closing this one.

Issue #1291, opened 2017-11-30
Html data processor failing test


Mgsy:
## Are you reporting a feature request or a bug?

Failing test

## Provide detailed reproduction steps (if any)

1. Go to http://localhost:1030/tests/core/htmldataprocessor.

### Expected result

All tests should pass.

### Actual result

One test is failing - `tests/core/htmldataprocessor test xss - iframe with src=javascript 3` - with alert about unknown protocol `jjavascript:`. It's about this line - https://github.com/ckeditor/ckeditor-dev/blob/216581c42d4dfeb98abac1fd4d79312604dae411/tests/core/htmldataprocessor.js#L1339

## Other details

* Browser: Opera
* OS: Windows 10, MacOS X
* CKEditor version: 4.8.0

Issue #1070, opened 2017-10-19
Failing test: /tests/core/htmldataprocessor on Chrome 62+


msamsel:
## Are you reporting a feature request or a bug?

Bug

## [Check if the issue is already reported](http://docs.ckeditor.com/#!/guide/dev_issues_readme-section-avoid-duplicates)

*Put all reference links here…*

## Provide detailed reproduction steps (if any)

1. run test `/tests/core/htmldataprocessor` -> `test xss - iframe with src=javascript 3`

### Expected result
Test pass

### Actual result
Test fails

## Other details

* Browser: Chrome 62+
* OS: Win7 / MacOs
* CKEditor version: 4.7.3
* Installed CKEditor plugins: full-all

PR #4114, opened 2020-06-17
Ignore failing tests

### LINKED ISSUE #3483: failing test - tests/core/htmldataprocessor/htmldataprocessor
## Type of report

Bug

1. run `tests/core/htmldataprocessor/htmldataprocessor` on iOS13 with Safari

### Expected result

passing test

### Actual result

failing test

## Other details

IMG_0005

* Browser: Safari
* OS: iOS 13 (iPhone 11)
* CKEditor version: adcefdd71 (4.13.0)

### LINKED ISSUE #4111: Ignore failing tests
## Type of report

Task

Tests that are reported to be failing or unstable but would take too much time to fix should be ignored to reduce distraction during future testing.

Dumluregn:

Task

### This PR contains

- ignorancy

No changelog entry.

## What changes did you make?

Ignored failing tests according to the list from #3794. In general I was trying to only ignore particular cases, but sometimes it wasn't possible as the next tests started to fail as they depended on the outcome of the previous ones. Also sometimes it wasn't clearly indicated in the tickets which test cases are unstable, so it is possible we will need to ignore more than I did.

Closes #4111.
Closes #3483.

### DIFF
--- a/tests/core/editor/detaching.js
+++ b/tests/core/editor/detaching.js
@@ -305,6 +305,11 @@
 	// @returns {Function}
 	function getAfterIframeLoadTestCase( options ) {
 		return function() {
+			// (#3714)
+			if ( CKEDITOR.env.edge && CKEDITOR.env.version < 80 ) {
+				assert.ignore();
+			}
+
 			options.wrapper = this.wrapper;
 
 			var data = initializeEditor( options );
--- a/tests/core/editor/detaching.js
+++ b/tests/core/editor/detaching.js
@@ -305,11 +305,6 @@
 	// @returns {Function}
 	function getAfterIframeLoadTestCase( options ) {
 		return function() {
-			// (#3714)
-			if ( CKEDITOR.env.edge && CKEDITOR.env.version < 80 ) {
-				assert.ignore();
-			}
-
 			options.wrapper = this.wrapper;
 
 			var data = initializeEditor( options );
--- a/tests/core/htmldataprocessor/htmldataprocessor.js
+++ b/tests/core/htmldataprocessor/htmldataprocessor.js
@@ -30,11 +30,6 @@
 			output = output ? output.replace( /%xss%/g, 'testXss()' ) : input;
 
 		tcs[ name ] = function() {
-			// (#3483)
-			if ( name == 'test xss - iframe with src=javascript 3' && CKEDITOR.env.safari ) {
-				assert.ignore();
-			}
-
 			var editor = this.editors.editor2,
 				xssed = false;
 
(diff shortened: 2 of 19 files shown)


tests/plugins/pastetools/filter/parsestyles.js
```diff
@@ -6,6 +6,11 @@
 ( function() {
 	'use strict';
 
+	// (#3713)
+	if ( CKEDITOR.env.edge && CKEDITOR.env.version < 80 ) {
+		bender.ignore();
+	}
+
```
f1ames, line 13: Not sure if we would ever need to run this tests on older Edge versions (\`edgeHTML\`) :thinking: Maybe we should just leave it and also close #3713 as Edge migrated to Chromium.

tests/core/editor/detaching.js
```diff
@@ -305,6 +305,11 @@
 	// @returns {Function}
 	function getAfterIframeLoadTestCase( options ) {
 		return function() {
+			// (#3714)
+			if ( CKEDITOR.env.edge && CKEDITOR.env.version < 80 ) {
+				assert.ignore();
+			}
+
```
f1ames, line 312: See my other comment regarding edgeHtml.

tests/plugins/easyimage/widget.js
```diff
@@ -84,10 +84,15 @@
...
-				bender.tools.ignoreUnsupportedEnvironment( 'easyimage' );
+			bender.tools.ignoreUnsupportedEnvironment( 'easyimage' );
 		},
 
 		'test upcasting image widget (figure)': function( editor, bot ) {
+			// (#3715)
+			if ( CKEDITOR.env.edge && CKEDITOR.env.version < 80 ) {
+				assert.ignore();
+			}
+
```
f1ames, line 95: See my other comment regarding edgeHtml.

tests/core/htmldataprocessor/htmldataprocessor.js
```diff
@@ -30,6 +30,11 @@
 			output = output ? output.replace( /%xss%/g, 'testXss()' ) : input;
 
 		tcs[ name ] = function() {
+			// (#3483)
```
f1ames, line 33: I wonder if we shouldn't just try to fix this test (#3483) as it seems like the matter of porper HTML normalization before assertion?

Dumluregn: It's caused by changes in Safari engine - browsers were resigning from removing preceding spaces from attribute value (see #1070 and #1291 for reference), Safari was the only one left on that field, but it also adapted. I changed the test accordingly (actually removing browser sniffing instead of doing it twice, so it's a good change :joy: ).

f1ames:
Rebased onto latest `master`.

PR #4144, opened 2020-07-08
Fix BSA 181 - Fix error on paste with ckeditor relative to upgrade to new version of that library

habemuscode:
**Related to this PR**

https://github.com/oat-sa/tao-core-shared-libs-fe/pull/5

**Related to this issue**

https://oat-sa.atlassian.net/browse/BSA-188

**Description**

When the candidate answers an open-ended question, it is not possible to copy using the toolbar button. An error message appears inside the text block. The same behavior is identified during the work with the same item in the item authoring mode.

**Preconditions**

- Created test, which includes the item with Extended text common interaction with set XHTML Format in the properties;

**Steps**

- Start the delivery;
- Open the item with Extended text interaction and enter the text;
- Cope the text by clicking the Copy button in the tools bar;
- Try to paste the copied text by the Paste button from the tools bar;

**Actual behavior**

The text is not pasted; there is a warning message, but it appeared inside the text block. not in the pop-up or modal window. 

**Expected behavior**

The system should enable the test taker to copy and paste a word or a phrase in the content area of the open question.

image


jacekbogdanski:
@habemuscode it seems like you opened this PR on the wrong repository. I'm closing it for now. If you can confirm that there is no mistake, please tell us so we can reopen it.

PR #4173, opened 2020-07-24
Fix customConfig race condition

### LINKED ISSUE #3361: customConfig conflicts with other CKEDitor instance customConfig
## Type of report

Bug

I'm using [Pimcore](https://pimcore.com/de) as Content Management System, which implements CKEditor as their WYSIWYG editor for rich text fields. It allows specifying a `customConfig` file ([Reference](https://pimcore.com/docs/5.x/Development_Documentation/Documents/Editables/WYSIWYG.html#page_Configuration)) for each field, which will be passed to CKEditor. This is a follow-up from https://github.com/pimcore/pimcore/issues/1124 since Pimcore mentioned that the following bug is related to CKEditor itself.

In the backend interface, where users can edit fields using CKEditor, I have several different fields on one page, all with an individual CKEditor instance and sometimes with different configuration files (`customConfig`).

Imagine the following situation:

One instance of CKEditor for field A with the following custom config file contents (acutally doesn't matter, it's just differently to the other field below):

```javascript
CKEDITOR.editorConfig = function(config) {
  config.enterMode = CKEDITOR.ENTER_BR;
  config.shiftEnterMode = CKEDITOR.ENTER_BR;
  config.autoParagraph = false;
  config.removeButtons = 'Form,Checkbox,Radio,TextField,Textarea,Select,Button,HiddenField,Anchor,Image,Flash,Table,HorizontalRule,Smiley,SpecialChar,PageBreak,Iframe,Format,Font,FontSize,Styles,About,NumberedList,BulletedList,Outdent,Indent,CreateDiv,JustifyLeft,JustifyRight,JustifyBlock,BidiLtr,BidiRtl,JustifyCenter,Language,Link,Unlink,Anchor,Image,Flash,Table,HorizontalRule,Smiley,Iframe,PageBreak,Blockquote';
  config.disallowedContent = 'a';
};
```

Then another instance of CKEditor for field B with the following custom config file contents (acutally doesn't matter, it's just differently to the other field above):

```javascript
CKEDITOR.editorConfig = function(config) {
  config.removeButtons = 'Image';
  config.disallowedContent = 'img';
};
```

And finally, another CKEditor instance for field C with the exact same custom config file like field A.

Now, in this case field C should have the exact same toolbar like field A, because they are referencing to the same `customConfig` file. But because field B - with a separate CKEditor instance - is rendered between those two fields, with a different configuration, field C and any subsequent CKEditor instance gets the configuration of field B - __the wrong__ configuration.

So, just to visualize the issue, this is the toolbar for field A:

In this screenshot you can also see field B ("Content"), which has a different `customConfig`.

And this is the config for field C:

In this screenshot you can see that field C get's a the wrong toolbar, it should look like field A.

### Expected result

A CKEditor instance should always use the provided `customConfig`.

### Actual result

When mixing several CKEditor instances with different `customConfig` files, CKEditor uses the wrong configuration.

## Other details

* Browser: Firefox (latest, 70)
* OS: Windows 10
* CKEditor version: 4.11.2
* Installed CKEditor plugins: None

jacekbogdanski:

Bug fix

### This PR contains

- [x] Unit tests
- [x] Manual tests

```
* [#3361](https://github.com/ckeditor/ckeditor4/issues/3361): Fixed: Loading multiple [custom editor configurations](https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_config.html#cfg-customConfig) is prone to race condition between them.
```

## What changes did you make?

Honestly, I was really wondering about rewriting the whole `customConfig` loading as it could be nicely simplified and the current implementation is pretty bloated. Although, it's not a very good idea to do with a core code 10 years old. I propose simple fix instead, making sure that we won't overwrite cached config functor by a closure race condition. 

Closes #3361.

### DIFF
--- a/core/editor.js
+++ b/core/editor.js
@@ -260,8 +260,9 @@
 		var customConfig = editor.config.customConfig;
 
 		// Check if there is a custom config to load.
-		if ( !customConfig )
+		if ( !customConfig ) {
 			return false;
+		}
 
 		customConfig = CKEDITOR.getUrl( customConfig );
 
@@ -275,19 +276,19 @@
 
 			// If there is no other customConfig in the chain, fire the
 			// "configLoaded" event.
-			if ( CKEDITOR.getUrl( editor.config.customConfig ) == customConfig || !loadConfig( editor ) )
+			if ( CKEDITOR.getUrl( editor.config.customConfig ) == customConfig || !loadConfig( editor ) ) {
 				editor.fireOnce( 'customConfigLoaded' );
+			}
+
 		} else {
 			// Load the custom configuration file.
 			// To resolve customConfig race conflicts, use scriptLoader#queue
 			// instead of scriptLoader#load (https://dev.ckeditor.com/ticket/6504).
 			CKEDITOR.scriptLoader.queue( customConfig, function() {
-				// If the CKEDITOR.editorConfig function has been properly
-				// defined in the custom configuration file, cache it.
-				if ( CKEDITOR.editorConfig )
-					loadedConfig.fn = CKEDITOR.editorConfig;
-				else
-					loadedConfig.fn = function() {};
+				// Cache config if it has been properly set using `editorConfig`,
+				// but make sure to not overwrite existing cache if the same config has
+				// been loaded multiple times by different editors (#3361).
+				loadedConfig.fn = loadedConfig.fn || CKEDITOR.editorConfig || function() {};
 
 				// Call the load config again. This time the custom
 				// config is already cached and so it will get loaded.
--- a/tests/core/editor/custom_config_race.js
+++ b/tests/core/editor/custom_config_race.js
@@ -1,41 +1,66 @@
 /* bender-tags: editor */
 
-var assetsPath = '%TEST_DIR%_assets/',
-	instances = {},
-	initEditors = ( function() {
-		var count = 0;
+var assetsPath = '%TEST_DIR%_assets/';
 
-		return function( editors, callback ) {
-			var total = CKEDITOR.tools.object.keys( editors ).length;
+function initEditors( editors ) {
+	var count = 0;
+	var instances = {};
 
-			for ( var e in editors ) {
-				instances[ e ] = CKEDITOR.replace( e,
-					CKEDITOR.tools.extend( editors[ e ], {
-						plugins: 'wysiwygarea',
-						on: {
-							instanceReady: function() {
-								if ( ++count == total )
-									callback();
+	return function( callback ) {
+		var total = CKEDITOR.tools.object.keys( editors ).length;
+
+		for ( var e in editors ) {
+			instances[ e ] = CKEDITOR.replace( e,
+				CKEDITOR.tools.extend( editors[ e ], {
+					plugins: 'wysiwygarea',
+					on: {
+						instanceReady: function() {
+							if ( ++count == total ) {
+								resume( function() {
+									callback( instances );
+								} );
 							}
 						}
-					} )
-				);
-			}
-		};
-	} )();
+					}
+				} )
+			);
+		}
+
+		wait();
+	};
+}
 
 bender.test( {
-	'async:init': function() {
+	tearDown: function() {
+		for ( var editorName in CKEDITOR.instances ) {
+			CKEDITOR.instances[ editorName ].destroy();
+		}
+	},
+
+	// (#3361)
+	'test race of customConfigs with reused config': function() {
+		initEditors( {
+			editor1: { customConfig: assetsPath + 'raceconfig1.js' },
+			editor2: { customConfig: assetsPath + 'raceconfig2.js' },
+			editor3: { customConfig: assetsPath + 'raceconfig1.js' }
+		} )
+		( function( instances ) {
+			assert.areSame( '200px', instances.editor1.config.width, 'Instance uses own customConfig.' );
+			assert.areSame( '300px', instances.editor2.config.width, 'Instance uses own customConfig.' );
+			assert.areSame( '200px', instances.editor3.config.width, 'Instance uses own customConfig.' );
+		} );
+	},
+
+	'test race of customConfigs': function() {
 		initEditors( {
 			editor1: { customConfig: assetsPath + 'raceconfig1.js' },
 			editor2: { customConfig: assetsPath + 'raceconfig2.js' },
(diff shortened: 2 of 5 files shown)


tests/core/editor/custom_config_race.js
```diff
@@ -1,41 +1,66 @@
...
-						plugins: 'wysiwygarea',
-						on: {
-							instanceReady: function() {
-								if ( ++count == total )
-									callback();
+var assetsPath = '%TEST_DIR%_assets/';
+
+function initEditors( editors ) {
+	var count = 0;
+	var instances = {};
```
Dumluregn, line 7: No need for duplicated `var` keyword I guess?

tests/core/editor/custom_config_race.js
```diff
@@ -1,41 +1,66 @@
...
+	return function( callback ) {
+		var total = CKEDITOR.tools.object.keys( editors ).length;
+
+		for ( var e in editors ) {
+			instances[ e ] = CKEDITOR.replace( e,
+				CKEDITOR.tools.extend( editors[ e ], {
+					plugins: 'wysiwygarea',
+					on: {
+						instanceReady: function() {
+							if ( ++count == total ) {
```
Dumluregn, line 18: ~Someone once told me that using increment operator is good only in `for` loop declaration as it decreases the readability everywhere else, WDYT? 😄~ Now I see you just outdented this part of code so my joke was inappropriate 🤷‍♂️ Still, maybe let's switch it to a simple
```js
count++;

if ( count == total ) {
```
for the sake of future CKE4 maintainers.

tests/core/editor/custom_config_race.js
```diff
@@ -1,41 +1,66 @@
...
 			editor1: { customConfig: assetsPath + 'raceconfig1.js' },
 			editor2: { customConfig: assetsPath + 'raceconfig2.js' },
-			editor3: { customConfig: assetsPath + 'raceconfig3.js' }
-		}, this.callback );
+			editor3: { customConfig: assetsPath + 'raceconfig1.js' }
+		} )
+		( function( instances ) {
+			assert.areSame( '200px', instances.editor1.config.width, 'Instance uses own customConfig.' );
+			assert.areSame( '300px', instances.editor2.config.width, 'Instance uses own customConfig.' );
+			assert.areSame( '200px', instances.editor3.config.width, 'Instance uses own customConfig.' );
```
Dumluregn, line 50: AFAIK the last parameter here is an error message, so something like `Instance should use own customConfig.` or `Instance didn't use own customConfig.` could be a little more precise communicate.

tests/core/editor/custom_config_race.js
```diff
@@ -1,41 +1,66 @@
...
-		assert.areSame( '400px', instances.editor3.config.width, 'Instance uses own customConfig.' );
+		initEditors( {
+			editor1: { customConfig: assetsPath + 'raceconfig1.js' },
+			editor2: { customConfig: assetsPath + 'raceconfig2.js' },
+			editor3: { customConfig: assetsPath + 'raceconfig3.js' }
+		} )
+		( function( instances ) {
+			assert.areSame( '200px', instances.editor1.config.width, 'Instance uses own customConfig.' );
+			assert.areSame( '300px', instances.editor2.config.width, 'Instance uses own customConfig.' );
+			assert.areSame( '400px', instances.editor3.config.width, 'Instance uses own customConfig.' );
```
Dumluregn, line 63: The comment about error message applies here too.

tests/core/editor/manual/customconfigrace.html
```diff
@@ -0,0 +1,24 @@
...
+<div id="editor1"></div>
+<h2>Editor 2</h2>
+<div id="editor2"></div>
+<h2>Editor 3</h2>
+<div id="editor3"></div>
+
+<script>
+	var assetsPath = '%BASE_PATH%core/editor/_assets/',
+		config1 = assetsPath + 'raceconfig1.js',
+		config2 = assetsPath + 'raceconfig3.js';
```
Dumluregn, line 11: I guess you chose `raceconfig3.js` instead of `raceconfig2.js` here to make the difference in editors width as clear as possible, but at first glance it looks like a typo. I checked how does it look with `raceconfig2.js` and I think it will still be easy to verify 👍

tests/core/editor/manual/customconfigrace.md
```diff
@@ -0,0 +1,14 @@
+@bender-tags: bug, 4.14.2, 3361
+@bender-ui: collapsed
+@bender-ckeditor-plugins: toolbar, wysiwygarea, basicstyles
+
+Verify editors width.
+
+## Expected
+
+* Editor 1 and Editor 3 have the same width (`200px`).
+* Editor 2 have different width (`400px`).
```
Dumluregn, line 10: If you agree with me on the change of `configfile` used in the test, this should be changed to `300px`.

tests/core/editor/manual/customconfigrace.md
```diff
@@ -0,0 +1,14 @@
...
+
+## Unexpected
+
+Editor 2 and Editor 3 have the same width. (`400px`).
```
Dumluregn, line 14: And also here - `400px` -> `300px`.

core/editor.js
```diff
@@ -275,19 +276,19 @@
 
 			// If there is no other customConfig in the chain, fire the
 			// "configLoaded" event.
-			if ( CKEDITOR.getUrl( editor.config.customConfig ) == customConfig || !loadConfig( editor ) )
+			if ( CKEDITOR.getUrl( editor.config.customConfig ) == customConfig || !loadConfig( editor ) ) {
 				editor.fireOnce( 'customConfigLoaded' );
+			}
+
```
Dumluregn, line 282: I think we don't usually leave an empty line before the closing brackets, so there is no need for adding one here.

tests/core/editor/custom_config_race.js
jacekbogdanski: It's really nice to see how sides turned out! Indeed I didn't pay enough attention to leave this code better then I found it.

core/editor.js
jacekbogdanski: I've added this line because at the first glance outer `else` looks like it belongs to this inner `if`.
image
**vs**
image

Dumluregn: Ok, we can leave it this way :+1:

PR #4181, opened 2020-07-29
Remove block styles when changing format if a new block format doesn't contain them

### LINKED ISSUE #3207: Add `Default` or `Normal` option to `Styles` dropdown
## Type of report

Feature request

If user changes block style using combo:
Screenshot 2019-06-24 at 12 40 36
and wants to return to default one, he/she needs to click the style again (like with format buttons - `B`, `I` etc.). I think more intuitive would be to add the option `Default` or `Normal` removing formatting on block level. Such an approach is used for Font dropdown:
Screenshot 2019-06-24 at 12 47 02

### LINKED ISSUE #3649: Can't remove block style after changing format
## Type of report

Bug

1. Clear editor contents and type a few characters.
2. Select what you have typed and apply "Italic Title" from Styles Dropdown.
3. Use the Format dropdown to change the formatting to normal Paragraph.

Old issue: https://dev.ckeditor.com/ticket/12526

### Expected result

Plain paragraph of text should be produced - `<p>test</p>`

One style/format should reset other style.

### Actual result

Paragraph with italic title is produced - `<p style="font-style:italic">test</p>`

The same thing is happening when you have a style without any styles defined (which maybe isn't super correct but it is possible in the editor) e.g. 
```
{name: 'Heading1', element: 'h1',: {"styles": {"font-family": "Arial", "color": "red"}}},
{name: 'Normal Para', element: 'p'}
```
If you first apply 'Heading 1' and then 'Normal Para', you will end up with Red Paragraph having Arial font family assigned. 

## Other details

* Browser: Any
* OS: Any
* CKEditor version: 4.0+
* Installed CKEditor plugins: Styles, Format

Dumluregn:

Bug fix

### This PR contains

- [x] Unit tests
- [x] Manual tests

```
* [#3649](https://github.com/ckeditor/ckeditor4/issues/3649): Fixed: Applying a [block format](https://ckeditor.com/docs/ckeditor4/latest/features/format.html) should remove existing block styles.
```

## What changes did you make?

The fix works in the way described in https://github.com/ckeditor/ckeditor4/issues/3649#issuecomment-633466405, so if a new block format is applied, the old block styles are removed (even if no new ones are added). The inline styles remain intact.

Please note that this fix seems to be suspiciously simple, so even though all the current tests are passing it should be treated with a big attention.

Closes #3649.
In a way it also closes #3207 - not exactly what I meant while creating that issue, but this PR gives an intuitive way of clearing the formatting, so I'd be satisfied with it.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,6 +6,11 @@ CKEditor 4 Changelog
 Fixed Issues:
 
 * [#3961](https://github.com/ckeditor/ckeditor4/issues/3961): Fixed: [Table Resize](https://ckeditor.com/cke4/addon/tableresize) prevents editing of merged cells.
+* [#3649](https://github.com/ckeditor/ckeditor4/issues/3649): Fixed: Applying a [block format](https://ckeditor.com/docs/ckeditor4/latest/features/format.html) should remove existing block styles.
+
+API Changes:
+
+* [#3649](https://github.com/ckeditor/ckeditor4/issues/3649): Added new [stylesRemove](https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_editor.html#event-stylesRemove) editor event.
 
 Other Changes:
 
--- a/core/style.js
+++ b/core/style.js
@@ -1659,23 +1659,26 @@ CKEDITOR.STYLE_OBJECT = 3;
 			elementName = style.element;
 
 		// The "*" element name will always be a span for this function.
-		if ( elementName == '*' )
+		if ( elementName == '*' ) {
 			elementName = 'span';
+		}
 
 		// Create the element.
 		el = new CKEDITOR.dom.element( elementName, targetDocument );
 
 		// https://dev.ckeditor.com/ticket/6226: attributes should be copied before the new ones are applied
-		if ( element )
+		if ( element ) {
 			element.copyAttributes( el );
+		}
 
 		el = setupElement( el, style );
 
 		// Avoid ID duplication.
-		if ( targetDocument.getCustomData( 'doc_processing_style' ) && el.hasAttribute( 'id' ) )
+		if ( targetDocument.getCustomData( 'doc_processing_style' ) && el.hasAttribute( 'id' ) ) {
 			el.removeAttribute( 'id' );
-		else
+		} else {
 			targetDocument.setCustomData( 'doc_processing_style', 1 );
+		}
 
 		return el;
 	}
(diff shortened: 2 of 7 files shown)


Dumluregn:
@jswiderski could you take a look at this and check if the behaviour proposed in this PR is consistent with what you wanted to achieve in #3649?

jacekbogdanski:
@Dumluregn I took a look at the issue and in my opinion, we can't just change the current behavior of `core/style.js`, as it may cause a lot of regressions. Note that we are willing to change the behavior of `stylescombo` plugin but as a result, we will change the whole core style handling.

Instead of fixing the code which could be considered "not a bug but a feature", maybe we could implement additional option for such cases? It could be e.g. a flag named `override` resulting in overriding the whole element attributes, not the one which has been specified. We would just need to change our block style definitions with new property: https://github.com/ckeditor/ckeditor4/blob/4f621d0ead1d2cc4ba2ff618f723f80159b6ae22/styles.js#L36-L46

e.g. 

```js
        // ...
	{ name: 'Italic Title', element: 'h2', styles: { 'font-style': 'italic' }, override: true }
       // ...
```

I was thinking about just removing styles from the whole selection for `format` plugin only: https://github.com/ckeditor/ckeditor4/blob/4f621d0ead1d2cc4ba2ff618f723f80159b6ae22/plugins/format/plugin.js#L63-L79

But in that case, we won't be able to fix the issue for `stylescombo` block styles. I'm not really sure if we should, because it kind of makes sense that `stylescombo` plugin behaves differently that `format` and keeps styling despite changing the block element. Something to discuss (RFC?). There are some cons and pros for additional option:

Pros:
+ more control over the feature, someone may don't agree with these changes and just disable overriding for block styles by defining them again
+ additional feature which could be useful in other places

Cons:
- requires changing format and style definitions, so won't be applied for custom implementations
- pollutes API a bit

Despite the chosen approach,  simply removing attributes is not enough. Note, that some styles will produce semantic HTML instead of additional attributes on the target element. As an example, `Big` style wraps content with `big` element, so we should implement a feature that would cover such cases. For example, we already have `removeformat` plugin. We could move logic responsible for removing styles to `core/style` namespace and reuse it in both places. I didn't really dig into if we already have some API for that, so it's worth checking how e.g. `CKEDITOR.editor.removeStyle` method will behave with some dummy, default style.

jacekbogdanski:
Since we've finally know what should be covered in this PR, I proposed a simpler approach to the issue in https://github.com/ckeditor/ckeditor4/pull/4181/commits/291d7f819615254e4b27b74d57f2ae8b9a21b937 commit. 

@Dumluregn please take a look at what do you think about it. I didn't really test it thoroughly, so it may require some polishing.

jacekbogdanski (approved):
Looks good!

PR #4210, opened 2020-08-18
Determine CKEDITOR_BASEPATH when assets are hashed

### LINKED ISSUE #4209: Support inferring CKEDITOR_BASEPATH from hashed ckeditor file
## Type of report

Task

Currently, CKEditor will infer the CKeditor basepath by finding a script that matches a particular regex (looking for `<basepath>/ckeditor.js`), and extracting `<basepath>` for further use. In many build processes, we hash files for long-term-caching benefits across build versions. Due to this hashing, CKEDITOR_BASEPATH could not be inferred and had to be set manually. With these proposed changes, we can extend the ability for inference to cases where ckeditor has a trailing hash.

## Other details
n/a

markjm:

Bugfix/ extensibility feature

### This PR contains

- [x] Unit tests

```
* [#4209](https://github.com/ckeditor/ckeditor4/issues/4209): Support CKEDITOR_BASEPATH inference from hashed assets.
```

## What changes did you make?

Currently, CKEditor will infer the CKeditor basepath by finding a script that matches a particular regex (looking for `<basepath>/ckeditor.js`), and extracting `<basepath>` for further use. In many build processes, we hash files for long-term-caching benefits across build versions. Before, due to this hashing, CKEDITOR_BASEPATH could not be inferred and had to be set manually. With these changes, we can extend the ability for inference to cases where ckeditor has a trailing hash.

Closes #4209.


markjm:
@jacekbogdanski , these changes were helpful for my organization, so I figured I'd try and merge them back to major branch. Feel free to reject, if you prefer.

jacekbogdanski (changes requested):
@markjm thanks for the contribution!

I have some small doubts about this code change. It seems very implementation-specific, as the proposed regex targets some fixed format. E.g, it will match `http://foobar/ckeditor-F123.js` but not `http://foobar/ckeditor-G123.js`.

Also, that means we are actually changing the name of `ckeditor.js` file, which may result in some regression if someone is already using a file which would match updated regex, like some custom `ckeditor-f.js` file which adds some specific improvements to the editor. Making regex less strict due to the previous issue will make it worse.

Since this change may introduce regression I have to reject it, but please feel free to propose code improvements which would make this code safer.

jacekbogdanski:
See https://github.com/ckeditor/ckeditor4/issues/4209#issuecomment-688176911

PR #4215, opened 2020-08-20
Update generic tests

### LINKED ISSUE #2849: Improve embed based manual tests with correct test URLs
## Type of report

Task

Some of the embed/autoembed tests contain incorrect test URLs. It would be nice to replace them among tests and check if the issue is not caused by any bug. Invalid URLs:

```
https://www.google.dk/maps/place/Zygmunta+S%C5%82omi%C5%84skiego+15,+Warszawa
http://tellto.tumblr.com/post/112160646328
https://open.spotify.com/track/2XsTC2dPbmWREu5RQ1mQC0
```

manual tests requiring improvement:

```
tests/plugins/embed/manual/embed
tests/plugins/embedsemantic/manual/embedsemantic
tests/plugins/autoembed/manual/autoembed
```

### LINKED ISSUE #4214: Generic manual tests revamp
Some generic tests are outdated or features covered by them are not likely to change. Let's remove them and add generics which makes more sense from the perspective of the current editor version.

jacekbogdanski:

Task

### This PR contains

- [ ] Unit tests
- [ ] Manual tests

None.

## What changes did you make?

Removed old generics and added a new one which makes more sense for the current editor version.

Note that despite Easy Image is already a pretty dated plugin and shouldn't introduce any regressions, it's also a paid plugin so it probably makes sense to test it during every release.

Closes #4214
Closes #2849


Dumluregn:
@jacekbogdanski maybe it would be a good moment to introduce a `generic` tag instead of our counterintuitive choosing tests without any tags, WDYT?

jacekbogdanski:
That sounds like a very good idea because the generic tests can still be a bug/feature in the same time. I will check how it works and update this PR.

Dumluregn:
Cool, if we do this we should also update our [dev docs](https://github.com/cksource/ckeditor4-dev-docs/blob/master/posts/2016-07-15-testing.md#steps-for-chrome-firefox-safari-edge-ie11-chromeandriod-safariios).

f1ames:
One side note from me - some generic tests (like PFW one) should be ignored for older IE browsers. OTOH I see we have static list of generic tests for IE8-10 (see https://github.com/cksource/ckeditor4-dev-docs/blob/master/posts/2016-07-15-testing.md#steps-for-ie8-10) so maybe it's not necessary? :thinking:

Dumluregn:
Another great URL: `http://tests.ckeditor.test:1030/#tests/is:manual,name:language,name:buttons,name:image,name:block,name:inline` 😄 maybe other tag, like `iegeneric`, would do? But I wonder if IE is worth our effort.

tests/plugins/autoembed/manual/autoembed.md
```diff
@@ -1,3 +1,4 @@
+@bender-tags: feature, generic
```
Dumluregn, line 1: Please remove broken links:
`https://www.google.dk/maps/place/Zygmunta+S%C5%82omi%C5%84skiego+15,+Warszawa` and
`http://tellto.tumblr.com/post/112160646328`
from the list.

tests/plugins/embed/manual/embed.md
```diff
@@ -1,3 +1,4 @@
+@bender-tags: feature, generic
```
Dumluregn, line 1: Please remove broken links:
`https://www.google.dk/maps/place/Zygmunta+S%C5%82omi%C5%84skiego+15,+Warszawa` and
`http://tellto.tumblr.com/post/112160646328`
from the list.

tests/plugins/embedsemantic/manual/embedsemantic.md
```diff
@@ -1,3 +1,4 @@
+@bender-tags: feature, generic
```
Dumluregn, line 1: Please remove broken links:
`https://www.google.dk/maps/place/Zygmunta+S%C5%82omi%C5%84skiego+15,+Warszawa` and
`http://tellto.tumblr.com/post/112160646328`
from the list.

tests/plugins/pastefromgdocs/manual/pastefromgdocs.md
```diff
@@ -1,4 +1,4 @@
-@bender-tags: feature, 4.13.0, 835, pastetools
```
Dumluregn, line 1: Let's add `target="_blank"` attribute to the links in this test to avoid accidental redirection.

tests/plugins/pastefromlibreoffice/manual/main.md
```diff
@@ -1,4 +1,4 @@
-@bender-tags: feature, 4.14.0, libreoffice
```
Dumluregn, line 1: It would be easier to verify if content was pasted correctly if the editor window was bigger. Let's set editor's height to `500` maybe?

tests/plugins/pastefromword/manual/pastefromword.html
```diff
@@ -0,0 +1,14 @@
+<h2>Classic editor</h2>
```
Dumluregn, line 1: It would be easier to verify if content was pasted correctly if the editor window was bigger. Let's set editor's height to `500` maybe?

tests/plugins/pastefromword/manual/pastefromword.md
```diff
@@ -0,0 +1,11 @@
+@bender-tags: 4.15.0, word, feature, generic
```
Dumluregn, line 1: It would be easier to verify if content was pasted correctly if the editor window was bigger. Let's set editor's height to `500` maybe?

Dumluregn (changes requested):
1. I'd like to better understand decisions you made here, so I'm asking you to give a **really** short explanation why did you decide to add/remove/leave each test. Below you'll find a table with the test names - keywords like `core plugin`, `old feature` or `checked in different test` will be just enough.

| Test path  | Added/Removed/Left | Explanation |
| --- | --- | --- |
| `tests/core/language/manual/language` | R | |
| `tests/plugins/autoembed/manual/autoembed` | L | |
| `tests/plugins/clipboard/manual/buttons` | L | |
| `tests/plugins/dialog/manual/startupfocustab` | R | |
| `tests/plugins/easyimage/manual/easyimage` | A | |
| `tests/plugins/embed/manual/embed` | L | |
| `tests/plugins/embedsemantic/manual/embedsemantic` | L | |
| `tests/plugins/mentions/manual/mentions` | R | |
| `tests/plugins/notification/manual/classic` | R | |
| `tests/plugins/notification/manual/inline` | R | |
| `tests/plugins/notificationaggregator/manual/classic` | R | |
| `tests/plugins/pastefromgdocs/manual/pastefromgdocs` | A | |
| `tests/plugins/pastefromlibreoffice/manual/main` | A | |
| `tests/plugins/pastefromword/manual/pastefromword` | A | |
| `tests/plugins/tableselection/manual/tableselection` | R | |
| `tests/plugins/uploadwidget/manual/image` | R | |
| `tests/plugins/widget/manual/block` | L | |
| `tests/plugins/widget/manual/inline` | L | |

2. As we are going to reuse those tests a few times, it makes sense to polish them a little to make testing painless. See the comments for particular tests - sorry but I couldn't reference the exact lines of code, so I've commented the first lines in those tests. They are mostly repetitive (3 actual changes).

tests/plugins/autoembed/manual/autoembed.md
jacekbogdanski: `http://tellto.tumblr.com/post/112160646328` is no brainer here as the URL no longer exists. However, in the case of Google maps, I'm not sure. It looks like we are just not authorized to use this API but it correctly pings to the provider. I suppose that running tests from authorized domains would work. That means we are actually getting expected behavior.

Also, removing invalid tests should close: https://github.com/ckeditor/ckeditor4/issues/2849 I will add it to the closing footnote. Funny that Spotify started working.

tests/plugins/pastefromgdocs/manual/pastefromgdocs.md
jacekbogdanski: I'm not super happy mixing HTML with markdown, but indeed your suggestion makes sense.

tests/plugins/pastefromlibreoffice/manual/main.md
jacekbogdanski: We are keeping the default height for every pasting manual test. I'm for keeping it for consistency since you can just use a resize handler.

jacekbogdanski:
> I'd like to better understand decisions you've made here, so I'm asking you to give a really short explanation why did you decide to add/remove/leave each test. Below you'll find a table with the test names - keywords like core plugin, old feature or checked in different test will be just enough.

I actually hoped that you will give your opinion if you agree with this changes to make sure that they were correct, but fine let's go with explanation:

| Test path                                             | Added/Removed/Left | Explanation                                                                                            |
|-------------------------------------------------------|--------------------|--------------------------------------------------------------------------------------------------------|
| `tests/core/language/manual/language`                 | R                  | unlikely to introduce regression as we are testing it every day                                        |
| `tests/plugins/autoembed/manual/autoembed`            | L                  | using 3rd party provider service which may break implementation                               |
| `tests/plugins/clipboard/manual/buttons`              | L                  | our clipboard handling is seriously complex and small upstream change may break it                     |
| `tests/plugins/dialog/manual/startupfocustab`         | R                  | unlikely to break (probably never did)                                                                 |
| `tests/plugins/easyimage/manual/easyimage`            | A                  | commercial plugin, we should make sure it always work                                                  |
| `tests/plugins/embed/manual/embed`                    | L                  | as autoembed                                                                                           |
| `tests/plugins/embedsemantic/manual/embedsemantic`    | L                  | as autoembed                                                                                           |
| `tests/plugins/mentions/manual/mentions`              | R                  | unlikely to break, we already had a couple of testing phases for this feature to make sure it's stable |
| `tests/plugins/notification/manual/classic`           | R                  | so old and didn't break                                                                                |
| `tests/plugins/notification/manual/inline`            | R                  | ditto                                                                                                  |
| `tests/plugins/notificationaggregator/manual/classic` | R                  | ditto                                                                                                  |
| `tests/plugins/pastefromgdocs/manual/pastefromgdocs`  | A                  | new addition to our codebase, based on 3rd party software                                              |
| `tests/plugins/pastefromlibreoffice/manual/main`      | A                  | ditto                                                                                                  |
| `tests/plugins/pastefromword/manual/pastefromword`    | A                  | based on 3rd party software, we also did huge refactoring here recently                                |
| `tests/plugins/tableselection/manual/tableselection`  | R                  | old, unlikely to break                                                                                 |
| `tests/plugins/uploadwidget/manual/image`             | R                  | uses fake HTTP adapter, survived so many testing phases it's unlikely to break                         |
| `tests/plugins/widget/manual/block`                   | L                  | widgets are complex and are foundation for many plugins, we often change something there               |
| `tests/plugins/widget/manual/inline`                  | L                  | ditto                                                                                                  |

tests/plugins/pastefromlibreoffice/manual/main.md
Dumluregn: Ok, using resizer makes sense. We just need to enable it in `tests/plugins/pastefromgdocs/manual/pastefromgdocs` then.

Dumluregn (changes requested):
Thanks for the table 👍 It all makes sense to me, especially the inclusion of pasting plugins and Easy Image. The only thing I wonder about is if we shouldn't leave the `tableselection` test as generic because IIRC it works differently on FF, so if anything changes there it would be good to know - WDYT?

As for my proposals I think it's worth to always test `cloudservices` plugin too as it is necessary in Easy Image plugin and may be used in the future (#4226). Note that although there is an Easy Image test on the new list, it doesn't really test the cloud services as it works the same when you are offline - there are just errors in the console which you won't see if you don't open it.

I also want to ask for your opinion about testing `print` plugin (than was under heavy refactor lately, but OTOH it didn't work for a long time and nobody complained 🙈 ) and `emoji` plugin - this one is not in any preset so we rarely use it, but it isn't commercial and I guess there is not much to break.

There are other features worth testing but they're mostly covered in the tests we already have on the list (e.g. `image2` plugin is tested in `widget` tests).

jacekbogdanski:
> As for my proposals I think it's worth to always test cloudservices plugin too as it is necessary in Easy Image plugin and may be used in the future (#4226). Note that although there is an Easy Image test on the new list, it doesn't really test the cloud services as it works the same when you are offline - there are just errors in the console which you won't see if you don't open it.

Not sure if I understand, easy image test is using real endpoint service:
https://github.com/ckeditor/ckeditor4/blob/d963d6058c89d0a3b55b78c03a21a6da6ae36c58/tests/plugins/easyimage/manual/easyimage.html#L36-L45
could you tell more about which test would you like to enable?

> I also want to ask for your opinion about testing print plugin

It's not really a complex plugin, there is not much to break. As it may be a subjective opinion, overall it seems best to just test low-level plugins that are used by other plugins as a base, e.g. widget or plugins which may break due to 3rd party regressions.

> and emoji plugin - this one is not in any preset so we rarely use it, but it isn't commercial and I guess there is not much to break.

Since this plugin incorporates a couple of features like autocomplete, panel etc it makes sense to make it generic.

Dumluregn (approved):
> Not sure if I understand, easy image test is using real endpoint service

What I meant is that for the Easy Image test it doesn't matter if the token was fetched or not - the image appears anyway, so unless you take a look at console (and see errors there) or at the uploaded image `href` in source mode (and see it begins with `localhost` instead of the CS url) you don't know if the `cloudservices` token works well or not. But I've just realised there isn't really any (important) feature of this plugin that is not covered by unit tests, so let's leave it as is now.

The rest LGTM, good job 👍  Generic testing should be a bit more interesting and serious from now on (for a while 😛 ).

PR #4228, opened 2020-08-25
Docs: add note about registered event listener (on event method).

### LINKED ISSUE #4123: Add info that a named function can't be listening for the same event more than once
## Type of report

Docs

## Provide a description of requested docs changes

In `editor.on()` method if named listener function is already registered, it can't be added for the second time. However it isn't mentioned in docs, so surprises occur like in #4107. As we are definitely not going to modify the event engine, we could add a mention about this behaviour to docs together with a workaround - using a wrapper function, like in https://github.com/ckeditor/ckeditor4/blob/4b9240f59b4effd86ce152ee07a5fbcb89e0b8ae/plugins/autocomplete/plugin.js#L282-L286.

hub33k:

Docs (api)

### This PR contains

- [ ] Unit tests
- [ ] Manual tests

```
* [#4123](https://github.com/ckeditor/ckeditor4/issues/4123): Add info that a named function can't be listening for the same event more than once.
```

## What changes did you make?

Added mention that a named function can't be listening for the same event more than once.

Closes #4123.

### DIFF
--- a/core/event.js
+++ b/core/event.js
@@ -96,6 +96,15 @@ if ( !CKEDITOR.event ) {
 			/**
 			 * Registers a listener to a specific event in the current object.
 			 *
+			 * **Note**: CKEditor's event system has a limitation that one function cannot be used as listener for the same event more than once.
+			 * Hence, wrapper functions:
+			 *
+			 * ```javascript
+			 * this._listeners.push( win.on( 'scroll', function() {
+			 * 	this.viewRepositionListener();
+			 * }, this ) );
+			 * ```
+			 *
 			 *		someObject.on( 'someEvent', function() {
 			 *			alert( this == someObject );		// true
 			 *		} );
--- a/core/event.js
+++ b/core/event.js
@@ -96,15 +96,6 @@ if ( !CKEDITOR.event ) {
 			/**
 			 * Registers a listener to a specific event in the current object.
 			 *
-			 * **Note**: CKEditor's event system has a limitation that one function cannot be used as listener for the same event more than once.
-			 * Hence, wrapper functions:
-			 *
-			 * ```javascript
-			 * this._listeners.push( win.on( 'scroll', function() {
-			 * 	this.viewRepositionListener();
-			 * }, this ) );
-			 * ```
-			 *
 			 *		someObject.on( 'someEvent', function() {
 			 *			alert( this == someObject );		// true
 			 *		} );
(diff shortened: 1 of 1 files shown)


core/event.js
```diff
@@ -96,6 +96,15 @@ if ( !CKEDITOR.event ) {
 			/**
 			 * Registers a listener to a specific event in the current object.
 			 *
+			 * **Note**: CKEditor's event system has a limitation that one function cannot be used as listener for the same event more than once.
```
f1ames, line 99: I would move it to the end of the description, because now it separates method description from main code sample. It would be more readable to add it at the end as this is a note which one should read after getting grasp of the main concept here (so how attaching listeners works).

f1ames, line 99:
```suggestion
			 * **Note**: CKEditor's event system has a limitation that one function cannot be used as a listener for the same event more than once.
```

core/event.js
```diff
@@ -96,6 +96,15 @@ if ( !CKEDITOR.event ) {
...
+			 * Hence, wrapper functions:
```
f1ames, line 100:
```suggestion
			 * Hence, to reuse it with multiple listeners, it should be wrapped into additional wrapper function:
```

core/event.js
```diff
@@ -96,6 +96,15 @@ if ( !CKEDITOR.event ) {
...
+			 *
+			 * ```javascript
+			 * this._listeners.push( win.on( 'scroll', function() {
```
f1ames, line 103: I would rework this example to be simpler and to show attaching to multiple events. It should use the same convention (`someObject.on...`) as rest of code samples here so something like:

```javascript
function listener( evt ) { ... };

someObject.on( 'someEvent', function() {
    listener();
} );

someObject( 'anotherEvent', function( evt ) {
    listener( evt );
} );
```

f1ames (changes requested):

core/event.js
```diff
@@ -96,21 +96,38 @@ if ( !CKEDITOR.event ) {
...
+			 * function listener( evt ) { ... };
+			 *
+			 * someObject.on( 'someEvent', function() {
+			 * 	listener();
+			 * } );
 			 *
-			 *		someObject.on( 'someEvent', function() { ... } );						// 2nd called
-			 *		someObject.on( 'someEvent', function() { ... }, null, null, 100 );		// 3rd called
-			 *		someObject.on( 'someEvent', function() { ... }, null, null, 1 );		// 1st called
+			 * someObject( 'anotherEvent', function( evt ) {
```
f1ames, line 127:
```suggestion
			 * someObject( 'someEvent', function( evt ) {
```

> CKEditor's event system has a limitation that one function cannot be used as a listener for the same event more than once

Well, @hub33k it seems you should always verify my suggestions (and from any other reviewer too) since I suggested incorrect example (with different and not the same event :see\_no\_evil:) :smile:

f1ames (approved):
LGTM :+1:

f1ames:
Rebased onto latest `master`.

f1ames:
I'm also skipping changelog entry since this is very minor docs correction.

PR #4239, opened 2020-08-31
Fix issue #848, arabic text not being "bound" correctly when pasting

### LINKED ISSUE #848: Arabic text pasted incorrectly

Bug

_none found_

1. Paste the following text twice to CKEditor in Chrome:

    > عام

### Expected result

Characters are "bound" to:

> عامعام

### Actual result

Characters are not bound and showing as:

> عام عام

(space inserted just because the GitHub editor does fine here)

## Other details

* Browser: Chrome (it works for Firefox and IE)
* OS: Windows 10
* CKEditor version: 4.7.2 (tested at https://ckeditor.com/)
* Works for Chrome for example here:
    * GitHub's Issue Text Editor
    * FroalaEditor: https://www.froala.com/wysiwyg-editor
* Assumption is, that RTL does not work in CKEditor in Chrome as also the cursor navigation direction is wrong comparing Chrome to IE, Firefox, Edge

devoidfury:

Bug fix.

### This PR contains

- [x] Unit tests
- [x] Manual tests

```
* [#848](https://github.com/ckeditor/ckeditor4/issues/848): Fixes arabic text not being "bound" correctly when pasting.
```

## What changes did you make?

Edit by @f1ames:

> This change uses the native Element.prototype.normalize method to merge sibling TextNodes automatically, if the method is available.
> Should cause Chrome to trigger ligatures/"bounding" correctly.

I have rework the fix to only normalize text node in Chrome since it is Chrome only issue. Since this is a core change, the changes should be precise to not touch browsers where it works fine.

> I'm not entirely clear on how to add a test here, since it purely addresses a rendering bug in chrome.

I have added manual tests (one for #848 and one IME generic) and also unit tests.

As mentioned, this fix touches core mechanics. I haven't found any issues and other IMEs seems to also work fine. Still it should be checked and tested thoroughly. 

Closes #848.

### DIFF
--- a/core/editable.js
+++ b/core/editable.js
@@ -2030,7 +2030,7 @@
 			while ( ( node = that.mergeCandidates.pop() ) )
 				node.mergeSiblings();
 
-			range.startPath().block.$.normalize()
+			range.startPath().block.$.normalize();
 			range.moveToBookmark( bm );
 
 			// Rule 3.
--- a/core/editable.js
+++ b/core/editable.js
@@ -1219,6 +1219,7 @@
 
 						if ( range.collapsed ) {
 							if ( !mergeBlocksCollapsedSelection( editor, range, backspace, startPath ) )
+								startPath.block.$.normalize();
 								return;
 						} else {
 							if ( !mergeBlocksNonCollapsedSelection( editor, range, startPath ) )
@@ -2029,7 +2030,7 @@
 			while ( ( node = that.mergeCandidates.pop() ) )
 				node.mergeSiblings();
 
-			range.root.$.normalize && range.root.$.normalize();
+			range.startPath().block.$.normalize()
 			range.moveToBookmark( bm );
 
 			// Rule 3.
--- a/core/editable.js
+++ b/core/editable.js
@@ -2029,6 +2029,7 @@
 			while ( ( node = that.mergeCandidates.pop() ) )
 				node.mergeSiblings();
 
+			range.root.$.normalize && range.root.$.normalize();
 			range.moveToBookmark( bm );
 
 			// Rule 3.
--- a/core/editable.js
+++ b/core/editable.js
@@ -1218,9 +1218,12 @@
 							startPath = range.startPath();
 
 						if ( range.collapsed ) {
-							if ( !mergeBlocksCollapsedSelection( editor, range, backspace, startPath ) )
-								startPath.block.$.normalize();
+							if ( !mergeBlocksCollapsedSelection( editor, range, backspace, startPath ) ) {
+								if ( startPath.block ) {
+									startPath.block.$.normalize();
+								}
 								return;
+							}
 						} else {
 							if ( !mergeBlocksNonCollapsedSelection( editor, range, startPath ) )
 								return;
@@ -2030,7 +2033,9 @@
 			while ( ( node = that.mergeCandidates.pop() ) )
 				node.mergeSiblings();
 
-			range.startPath().block.$.normalize();
+			if ( range.startPath().block ) {
+				range.startPath().block.$.normalize();
+			}
 			range.moveToBookmark( bm );
 
 			// Rule 3.
--- a/core/editable.js
+++ b/core/editable.js
@@ -1219,9 +1219,6 @@
 
 						if ( range.collapsed ) {
 							if ( !mergeBlocksCollapsedSelection( editor, range, backspace, startPath ) ) {
-								if ( startPath.block ) {
-									startPath.block.$.normalize();
-								}
 								return;
 							}
 						} else {
@@ -2030,8 +2027,9 @@
 			}
 
 			// Eventually merge identical inline elements.
-			while ( ( node = that.mergeCandidates.pop() ) )
+			while ( ( node = that.mergeCandidates.pop() ) ) {
 				node.mergeSiblings();
+			}
 
 			if ( range.startPath().block ) {
 				range.startPath().block.$.normalize();
--- a/core/editable.js
+++ b/core/editable.js
@@ -2032,31 +2032,21 @@
 				node.mergeSiblings();
 			}
 
-			range.moveToBookmark( bm );
-
 			// Normalize text nodes (#848).
 			if ( CKEDITOR.env.webkit && range.startPath() ) {
-				var path = range.startPath(),
-					container = null;
+				var path = range.startPath();
 
-				// Get parent block element.
 				if ( path.block ) {
-					container = path.block;
+					path.block.$.normalize();
 				} else if ( path.blockLimit ) {
 					// Handle ENTER_BR mode when text is direct root/body child.
 					// This will call native `normalize` on entire editor content in this case
 					// normalizing text nodes in entire editor content.
-					container = path.blockLimit;
-				}
-
-				if ( container ) {
-					var bookmark = range.createBookmark2( true );
-
-					container.$.normalize();
-
-					range.moveToBookmark( bookmark );
+					path.blockLimit.$.normalize();
 				}
 			}
+
+			range.moveToBookmark( bm );
 
 			// Rule 3.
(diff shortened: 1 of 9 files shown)


core/editable.js
```diff
@@ -2029,6 +2029,7 @@
 			while ( ( node = that.mergeCandidates.pop() ) )
 				node.mergeSiblings();
 
+			range.root.$.normalize && range.root.$.normalize();
```
f1ames, line 2032: It would be safer to just normalize paragraph / block limit element which was modified instead of range root (which might be entire document), but not sure if we will be able to get such element correctly in all cases :thinking:

f1ames (dismissed):
Thanks for the contribution @devoidfury! :+1: 

> I'm not entirely clear on how to add a test here, since it purely addresses a rendering bug in chrome.

It might be tricky to add unit tests which checks the way how the text was rendered - as for that it will be enough to add manual test with simple steps (similar to ones described in #848).

As for unit test, it could check if two separate text nodes get merged into one after modified code is run. However, it requires some CKEditor 4 internals and API knowledge - if you would like to try to write we will be more than happy to review it. If it turns out too complex we can do it on our side :+1: 

You may refer to  [Contributing Code](http://docs.ckeditor.com/#!/guide/dev_contributing_code) guide for more details.

core/editable.js
jidrone-mbm: I create a PR against the devoidfury fork replacing that line with:
`range.startPath().block.$.normalize()`

Also, I fixed the deleting issue in that PR https://github.com/devoidfury/ckeditor4/pull/1

jidrone-mbm: I created a PR against the devoidfury fork, replacing that line with
`range.startPath().block.$.normalize();`

Also, I added a possible fix to the deleting issue https://github.com/devoidfury/ckeditor4/pull/1.

jidrone-mbm:
I was able to replicate the issue also in Safari, so I think it is related with webkit browsers.

f1ames:
Thanks @devoidfury, @jidrone-mbm for pushing this PR forward :clap:

Do you plan to work on tests as I asked in https://github.com/ckeditor/ckeditor4/pull/4239#pullrequestreview-479834125 or we should just take it over from there?

jidrone-mbm:
Hi @f1ames, as @devoidfury I don't have enough time, would be great if you can do the tests.

f1ames:
Thanks @devoidfury and @jidrone-mbm for the update, we will try to push this PR forward in CKEditor 4 team :+1:

Comandeer (changes requested):
The fix works, however there are some issues:  
  
1\. unit tests fail in Safari,

2\. fix does not work in `CKEDITOR.ENTER_BR` mode (I've prepared manual test for it).

f1ames:
Rebased and retargeted onto latest `master`.

f1ames:
@Comandeer I fixed `ENTER_BR` mode by normalizing `blockLimit` container if there is no `block` one. It will result in normalizing entire editor content when text is direct root child but I haven't come up with anything better. It's still less intrusive than trying to wrap text nodes into some temporary block element or merging text nodes manually I suppose.

As for tests in Safari it turned out that `Text: ^` after inserting e.g. `foo` becomes `Text:foo ` (the space is moved to an end) so I replaced regular space with `&nbsp;`. I tried to reproduce it outside of unit tests, but space at the end always becomes `&nbsp;`. So it seems to be some race condition issue as the insertion is executed first before space becomes `&nbsp;` (or maybe our test `setHtmlWithSelection()` method simply doesn't trigger space to `&nbsp;` transformation :thinking:).

Comandeer (changes requested):
I have found one more case, in which the fix does not work:

1.  Paste `عام` several times.
2.  Move cursor to the middle of the pasted text.
3.  Paste.

Expected:

*   The newly pasted `عام` is correctly joined with the rest of the text.

Actual:

*   The newly pasted fragment is not joined with the rest.

The case works in the native `textarea`.

f1ames:
@Comandeer well... regarding https://github.com/ckeditor/ckeditor4/pull/4239#pullrequestreview-498505505 it looks like some kind of a race condition TBH :disappointed:

If second normalization is added and deffered (with `setTimeout( ..., 0 )`) it will work fine, however it may break a lot of things...

The same way, duplicating normalization after `range.moveToBookmark( bm );` fixes paste issue too. However, moving it (and not duplicating) breaks initial fix... which mens we need to normalize two times... also second normalization moves selection to block element end... You can see what I mean on `t/848-2` branch (and eccf57f).

I'm afraid this place is really sensitive to changes and fiddling too much with text nodes and selection may break something :thinking: 

So we either:

1.  Extract this case as a follow-up and go with current solution.
2.  Try to find some reasonable way to cover both in this PR.
3.  ?

@Comandeer WDYT?

f1ames:
Regarding my previous https://github.com/ckeditor/ckeditor4/pull/4239#issuecomment-701317028, I had an "aha moment" :bulb:The fact that normalization doesn't work when text is pasted in the middle is caused by bookmark `span` element:



After `range.moveToBookmark( bm );` call, `span` is removed and there are only two separate text nodes which can be normalized correctly (but changing selection as a side-effect...).

OTOH, after `range.moveToBookmark( bm );` call, bookmark `span`s are not there, but we have multiple filling char sequences there which prevents correct normalization:



which is the reason it works fine only if native normalization is called two times - before and after `range.moveToBookmark( bm );` call...

So the current solution does look really like a partial solution from this perspective. And we are basically back to my previous https://github.com/ckeditor/ckeditor4/pull/4239#issuecomment-701317028 wondering which approach to take... :thinking:

Comandeer:
TBH I'm for approach 2 – trying to fit it in this PR. It would require more effort, but we will cover the whole issue. Because now, as you mentioned, the fix seems like a partial one.

f1ames:
Rebased onto latest `master`.

f1ames:
> TBH I'm for approach 2 – trying to fit it in this PR. It would require more effort, but we will cover the whole issue. Because now, as you mentioned, the fix seems like a partial one.

Ok, let's then try with:

> 3. Another idea is to try with our own normalization (gluing text nodes manually), but not sure if it will change anything.

Since it looks like it might be valid way to solve this issue.

f1ames:
Ok, so I tried both with [manual](https://github.com/ckeditor/ckeditor4/pull/4239/commits/ea01df422bf0f26aea9a820e1c13898d2e95a986) and [native](https://github.com/ckeditor/ckeditor4/pull/4239/commits/7b379fb4dbd09d06512721dbd009cd1684f8ba4b) normalization and results are basically the same. It works (when run after bookmark is restored), but the selection is moved to the text end. If I try to preserve the selection with `bookmark2` (non intrusive one to not break normalization), the `startOffset` seems to be correct but when selection is restored via `moveToBookmark` it goes to text node beginning (again for both manual and native normalization).

The `moveToBookmark` method is pretty simple, so it's eaither some hidden issue here or browser upstream... It needs some further investigation :confused:

f1ames:
Having in mind that introducing a fix covering all the edge cases requires more work that we initially anticipated, let's go with the solution which covers initial issue and extract https://github.com/ckeditor/ckeditor4/pull/4239#pullrequestreview-498505505 as follow-up.

f1ames:
Extracted case mentioned in https://github.com/ckeditor/ckeditor4/pull/4239#pullrequestreview-498505505 to https://github.com/ckeditor/ckeditor4/issues/4328.

Comandeer:
Rebased onto latest `master`.

Comandeer:
It looks good to me, but I'd like to ask [@ckeditor/qa-team](https://github.com/orgs/ckeditor/teams/qa-team) to take a look and test this PR too to check if everything works fine without any regressions.

Mgsy:
We've tested it with @FilipTokarski and it seems to be working fine :+1:

Comandeer (approved):
LGTM!

PR #4242, opened 2020-09-01
Force language to ensure assertions

jacekbogdanski:
Some tests are failing due to localization differences.


Dumluregn (approved):
:+1: for caching it, 👎 for the guys who missed this during development 😜

PR #4251, opened 2020-09-02
Fix for crashing full-page editor with editorplaceholder

### LINKED ISSUE #4249: Crash when typing in editor placeholder test
## Type of report

Bug

1. Create `full-all` build in `ckeditor4-presets`
2. Open test `/tests/plugins/editorplaceholder/manual/fullpage`
3. Type something

### Expected result

No errors in console

### Actual result

```js
Uncaught TypeError: Cannot read property '1' of null
    at a.f (ckeditor.js:849)
    at a.e (ckeditor.js:10)
    at a.<anonymous> (ckeditor.js:11)
    at a.CKEDITOR.editor.CKEDITOR.editor.fire (ckeditor.js:13)
    at CKEDITOR.plugins.undo.UndoManager.type (ckeditor.js:1263)
    at CKEDITOR.plugins.undo.NativeEditingHandler.onKeyup (ckeditor.js:1274)
    at $.e (ckeditor.js:10)
    at $.fire (ckeditor.js:11)
    at HTMLBodyElement.<anonymous> (ckeditor.js:69)
```

## Other details

* latest `major` branch

Comandeer:

Bug fix

### This PR contains

- [x] Unit tests
- [x] Manual tests

Not needed, as it's a new feature.

## What changes did you make?

Thanks to `htmlwriter`, the data returned by the editor is nicely formatted. Unfortunately regex used in `editorplaceholder` plugin to detect full-page editor didn't handle content with newlines correctly. This fix includes newlines into the regex.

Closes #4249.


f1ames (changes requested):
Generally, it looks good :+1: 

Since we are reworking this regexp, I tried to find anything which still may break it :detective: Well... `docprops` plugin produces HTML like:

```html
<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.1//EN" "http://www.w3.org/TR/xhtml11/DTD/xhtml11.dtd">
<?xml version="1.0" encoding="iso-2022-jp"?>
<html xmlns="http://www.w3.org/1999/xhtml">
<head>
	<title>Title</title>
	<meta content="text/html; charset=iso-2022-jp" http-equiv="content-type" />
</head>
<body dir="rtl">
<p>a</p>
</body>
</html>
```

and body `dir` attribute cause regex not to match anything and so the same error as in initial report is thrown. Very rare edge case but I would say it will be good to take care of that too.

Comandeer:
Updated.

f1ames (approved):
LGTM :+1:

---

I have found one more case which produces the same error - `fullpage` + `bbcode`... There is no HTML structure and `body` in such case so the regex fails.

Also for some reason for empty editor, `getData()` returns one empty space (so length 1 instead of 0 which doesn't help here). And yes, it is an empty space, so there is now way to distinguish if editor is empty or someone just typed space because `charCode` is the same.

That being said I propose to extract it as a follow-up since it's a very rare edge case (why one would use `fullpage` with `bbcode` while the `bbcode` plugin basically removes entire HTML structure added by `fullpage`).

PR #4266, opened 2020-09-08
Detect if mention panel goes beyond window borders

### LINKED ISSUE #3582: Mention panel should consider window borders for positioning
## Type of report

I use CKEditor in a chat panel in the lower right corner of the website:
grafik
The mention panel is always shown below the typed word and so never shows up completely.

1. Goto https://ckeditor.com/docs/ckeditor4/latest/examples/mentions.html
2. Type @ in CKEditor

### Expected result

Mentions panel should be shown above the "@"

### Actual result

Mentions panel is shown below the "@" and so not much of it is visible:
grafik

## Other details

* Browser: Chrome
* OS: Windows 10
* CKEditor version: 4.13.0
* Installed CKEditor plugins: mentions

hub33k:

New feature

### This PR contains

- [x] Unit tests
- [x] Manual tests

```
* [#3582](https://github.com/ckeditor/ckeditor4/issues/3582): Introduced smart positioning of [Autocomplete](https://ckeditor.com/cke4/addon/autocomplete) panel used by [Mentions](https://ckeditor.com/cke4/addon/mentions) and [Emoji](https://ckeditor.com/cke4/addon/emoji) plugins.
```

## What changes did you make?

When mention panel is about to goes beyond window borders it will open in proper position.

Closes #3582.


hub33k:
Rebased onto latest `major`.

hub33k:
Rebased onto latest `major`.

hub33k:
I'm not entirely sure about the correctness of the unit test. I think it can be improved :thinking:

f1ames (changes requested):
Good start, but there are few more cases to consider (see review comments).

hub33k:
Rebased onto latest `major`.

hub33k:
I have doubts about unit tests here. Do you have any suggestions what should be proper way to handle them? :thinking:

f1ames:
Rebased onto latest `major`.

tests/plugins/autocomplete/viewposition.js
```diff
@@ -152,6 +152,26 @@
...
+
+		// (#3582)
+		'test view position beyond window borders': function( editor ) {
+			var orginalWindowHeight = editor.element.getWindow().getViewPaneSize().height;
+
+			// Simulate small window height.
+			editor.element.getWindow().$.innerHeight = 200;
+
+			var view = createPositionedView( editor, {
+				caretRect: { top: 100, bottom: 110, left: 50 },
```
f1ames, line 165: You are missing `right` here (which you added in this PR too). This causes some `NaN`s in the positioning code:

image

plugins/autocomplete/plugin.js
```diff
@@ -941,6 +941,13 @@
 				top = Math.max( rect.bottom, editorViewportRect.top );
 			}
 
+			var windowHeight = this.element.getWindow().$.innerHeight;
```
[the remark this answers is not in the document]
f1ames: I see you are still using `this.element.getWindow().$.innerHeight` and `this.element.getWindow().$.innerWidth` in your code instead of `getViewPaneSize()`. Is there any reason for that? Any edge cases which behaves differently? I see there are some slight differences (https://stackoverflow.com/a/7205786/646871) but we use it most cases so it should do the job :thinking:

f1ames:
Strange, it seems adding a comment to a previous review comment, posts pending review comments :thinking: I'm still reviewing this PR, will post update when ready.

f1ames (changes requested):
## The code

Generally looks good :+1:

I see that we are not able to cover and test RTL left placement case due to lack of autocomplete RTL support so let's leave it for #4315. But let's leave manual tests in the current state too so we will be able to reuse them when working on it.

It seems that top placement case is not covered too. It is kind of an edge case but can happen with longer editor content:

Peek 2020-10-06 16-09

## Manual tests

Very nice idea with manual tests and buttons which changes editors placement :+1: Only some wording suggestions here and there.

Thanks to these tests I have discovered that autocomplete/mentions doesn't support RTL in fact... as already mentioned above, extracted to https://github.com/ckeditor/ckeditor4/issues/4315.

## Unit tests

Strangely, some unit tests fails when run from dashboard (http://localhost:1030/#tests/is:unit,path:/tests/plugins/autocomplete) but passes when run separately. Probably something wrong with positioning/sizing stubbing :thinking: Tested on Chrome:







Also you have introduced `rect.right` which is passed to `getViewPositon()` so now all unit tests stubbing the rect should use it too.

As for approach to unit tests it looks fine. We need to stub things here since there is no other reasonable solution TBH. If you decide to switch to `getViewPaneSize()` it should be even easier to stub using sinon.

hub33k:
Rebased onto latest `major`.

hub33k:
TODO:
- [ ] handle top placement case
- [ ] finish unit tests
- [ ] fix failing tests

Dumluregn:
I took this PR over.

Dumluregn:
Rebased onto latest `major`.

Dumluregn:
I revoked adding the `rect.right` to the returns of [getViewPosition()](https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_plugins_autocomplete_view.html#method-getViewPosition) as without RTL support it is unnecessary here.

I would like to clarify two things:
1. As I was adding the integration manual tests I noticed, that with (very) custom bender styling the fix for the right window border doesn't work - but if you run a normal page like dev sample, it's fine. 

2. There is also an edge case with scrolling:
    1. You need a long (high) editor (like `1500px`).
    1. Scroll down until the editor toolbar isn't visible but editor contents is.
    1. Open the autocomplete panel close to the top.
    1. Scroll down keeping the panel open.

Expected: the panel sticks to the top border and is completely visible.
Actual: a small part of panel hides before the panel sticks.

I'm not sure if those two things are worth digging deeper - waiting for you opinion on this.

Dumluregn:
CI failed because of #4323.

f1ames (changes requested):
Looks good :+1: Few things to polish:

## The fix

Aren't we missing handling for cases when panel is hidden under top browser border? This is an edge case where the is not enough space below the caret, enough space above the caret but this space is not in browser viewport:

image

Btw. it's not possible to check with current manual tests so we will need a dedicated one for this case (or adjust existing ones but it seems for me it will make them too complicated).

## Manual tests

I would add resize plugin so one can experiment a little more with editor sizing and placement. However, I see it breaks manual tests layout (due to flex, absolut position and inability to scroll) so you will have to check if it makes sense.

## Unit tests

Looks good, I'm only missing a test for edge case mentioned in **The fix** section above.

f1ames:
As for https://github.com/ckeditor/ckeditor4/pull/4266#issuecomment-724570957:

> 1.  As I was adding the integration manual tests I noticed, that with (very) custom bender styling the fix for the right window border doesn't work - but if you run a normal page like dev sample, it's fine.

I think we can ignore this, since bender manual tests styling is very specific.

> 2\. There is also an edge case with scrolling:
> 
> 1. You need a long (high) editor (like `1500px`).
> 2. Scroll down until the editor toolbar isn't visible but editor contents is.
> 3. Open the autocomplete panel close to the top.
> 4. Scroll down keeping the panel open.
> 
> Expected: the panel sticks to the top border and is completely visible.  
> Actual: a small part of panel hides before the panel sticks.

Not sure if I understand correctly what's going on :thinking: Do yo mean something like that (the moment when mentions panel cover the toolbar):

Peek 2020-11-17 08-21

Or the fact that editor is partially hidden under browser borders on window scroll:
Peek 2020-11-17 08-23

Dumluregn:
> Aren't we missing handling for cases when panel is hidden under top browser border? 

I see what you mean. In such case it would probably be best to show a smaller panel, but it seems like a separate feature - so for now I suppose we should just display panel in the default way (i.e. below the caret), right?

> Not sure if I understand correctly what's going on

It can be explained better than I did previously 😄 What I meant is that when scrolling, the panel sticks to the editor toolbar instead of the viewport top - I've recorded gif showing it. I wonder if changing it may cause any unintuitive behaviours 🤔 
mentions

f1ames:
> I see what you mean. In such case it would probably be best to show a smaller panel, but it seems like a separate feature - so for now I suppose we should just display panel in the default way (i.e. below the caret), right?

Yes, below the caret should work fine I suppose. I mean the cas is there is not enough space above but there is enough below (in terms of browser viewport size). The edge edge case is that there is no enough space above and below and that could be a candidate for smaller panel but we don't need it at the moment.

> It can be explained better than I did previously What I meant is that when scrolling, the panel sticks to the editor toolbar instead of the viewport top - I've recorded gif showing it. I wonder if changing it may cause any unintuitive behaviours

Ok, so that's the same as second `.gif` I've added in https://github.com/ckeditor/ckeditor4/pull/4266#issuecomment-728740061. Does the fix for the above ("handling for cases when panel is hidden under top browser border") will also change this behavior? Or do you mean it shouldn't/won't be included in this PR?

I guess it now works the same way (sticking to window borders) for bottom positioning on scroll?

tests/plugins/autocomplete/integration/emoji/emoji.html
```diff
@@ -0,0 +1,107 @@
...
+
+	.inline {
+		width: 300px;
+		height: 194px;
+	}
+</style>
+
+<div>
+	<p>Select mode</p>
+	<button type="button" id="left-button">Left</button>
```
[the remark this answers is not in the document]
Dumluregn: I'm aware that this increases the tests overhead, but I think it's worth to ensure nothing breaks in the meantime, so I'd like to keep it as is. The RTL support tests will be able to reuse this code with just changing language in config and some description.

tests/plugins/autocomplete/integration/emoji/emoji.md
```diff
@@ -0,0 +1,19 @@
+@bender-tags: 4.16.0, feature, 3582
+@bender-ckeditor-plugins: wysiwygarea, toolbar, elementspath, sourcearea, emoji, clipboard, undo
+@bender-ui: collapsed
+
+*Repeat steps below for all 3 editors and modes (Left, Right and Bottom - but skip the last one on IE 10 and older).*
```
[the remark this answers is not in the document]
Dumluregn: Sorry, I copied the whole template from my other tests 🙈

Dumluregn:
Rebased onto latest `major` (with fixed CI).

As for the resizer - I've added it in the additional manual test for an edge case you mentioned in https://github.com/ckeditor/ckeditor4/pull/4266#pullrequestreview-532054417. In the original test it was indeed breaking the editors flow.

> I guess it now works the same way (sticking to window borders) for bottom positioning on scroll?

Yes, when I mentioned it for the first time the behaviour seemed strange, but after I understood what is going on it seems to be fine and I think I'd leave it as is.

f1ames (changes requested):
Almost there :sweat_drops: :+1: 

## The fix

:+1: 

## Unit tests

* Chrome :heavy_check_mark: 
* Firefox :heavy_check_mark: 
* Safari :heavy_check_mark: 
* Safari@iOS :heavy_check_mark: 
* Chrome@Android :heavy_check_mark: 
* Edge :heavy_check_mark: 
* IE11 :heavy_check_mark: 
* IE8 :heavy_check_mark: 
* IE9 :heavy_check_mark: 
* IE10 :heavy_check_mark: 

## Manual tests

* Chrome :heavy_check_mark: 
* Firefox :heavy_check_mark: 
* Safari :heavy_check_mark:
* Safari@iOS :thinking: 
* Chrome@Android :thinking: 
* Edge :heavy_check_mark: 
* IE11 :heavy_check_mark: 
* IE8 :heavy_check_mark: 
* IE9 :heavy_check_mark:
* IE10 :heavy_check_mark:

Those tests are quite painful on mobiles (not mobile friendly bender UI + custom absolute positioning) so I we should probably skip them there, WDYT?

Dumluregn:
We agreed with @f1ames to conduct an additional refactor in [`setPosition()`](https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_plugins_autocomplete_view.html#method-setPosition) method.

plugins/autocomplete/plugin.js
```diff
@@ -857,94 +857,165 @@
...
 		 */
 		setPosition: function( rect ) {
-			var editor = this.editor,
-				viewHeight = this.element.getSize( 'height' ),
-				editable = editor.editable(),
-				// Bounding rect where the view should fit (visible editor viewport).
-				editorViewportRect;
+			var documentWindow = this.element.getWindow(),
+				windowRect = documentWindow.getViewPaneSize(),
+				top = setVerticalPosition( {
```
f1ames, line 862: Well, it doesn't really set the position it just calculates it returning `top` placement. The position is set in [line 875 with `setStyle`](https://github.com/ckeditor/ckeditor4/pull/4266/files#diff-679aa38255fe9479e45442e51e5b5daca8bc8df2e3b06102929c33fd2e7b1853R875).

plugins/autocomplete/plugin.js
```diff
@@ -857,94 +857,165 @@
...
+					return editable.isInline() ? editable.getClientRect( true ) : editor.window.getFrame().getClientRect( true );
+				}
 			}
 
-			this.element.setStyles( {
-				left: rect.left + 'px',
-				top: top + 'px'
-			} );
+			function setHorizontalPosition( options ) {
+				var leftPosition = options.leftPosition,
```
f1ames, line 994:
```suggestion
              var caretLeftPosition = options.leftPosition,
```

?

plugins/autocomplete/plugin.js
```diff
@@ -857,94 +857,165 @@
...
-			// |     |     view     |                        |
-			// |     |              |                        |
-			// |     +--------------+                        |
-			// |																						 |
-			// |       editor viewport                       |
-			// +---------------------------------------------+
-			if ( editorViewportRect.top > rect.top ) {
-				top = Math.max( rect.bottom, editorViewportRect.top );
+			// Bounding rect where the view should fit (visible editor viewport).
+			function getEditorViewportRect( editor ) {
```
f1ames, line 982: I would be for moving this function below `setHorizontalPosition`.

plugins/autocomplete/plugin.js
```diff
@@ -857,94 +857,165 @@
...
-				editable = editor.editable(),
-				// Bounding rect where the view should fit (visible editor viewport).
-				editorViewportRect;
+			var documentWindow = this.element.getWindow(),
+				windowRect = documentWindow.getViewPaneSize(),
+				top = setVerticalPosition( {
+					editorViewportRect: getEditorViewportRect( this.editor ),
+					caretRect: rect,
+					viewHeight: this.element.getSize( 'height' ),
+					verticalScroll: documentWindow.getScrollPosition().y,
```
f1ames, line 866: Maybe `verticalScrollPosition` or `scrollPositionY`?

tests/plugins/autocomplete/integration/mentions/manual/mentions.html
```diff
@@ -0,0 +1,166 @@
...
+	};
+
+	// We need one editor reference to pass it to the bender.tools.ignoreUnsupportedEnvironment() method below.
+	var editor = CKEDITOR.replace( 'classic', config );
+	CKEDITOR.replace( 'divarea', CKEDITOR.tools.extend( { extraPlugins: 'divarea' }, config ) );
+	CKEDITOR.inline( 'inline', config );
+
+	bender.tools.ignoreUnsupportedEnvironment( 'mentions', editor );
+
+	// This could be earlier, but let's keep ignore statements in one place.
```
f1ames, line 152: I guess this comment is not necessary :thinking:

f1ames (changes requested):
Splitting calculating vertical and horizontal position into separate function really makes it more readable and easier to understand :+1: 

Just some minor polishing here and there and we will be good to go :tada:

plugins/autocomplete/plugin.js
Dumluregn: I didn't add `caret` prefix here because according to docs it doesn't have to be a caret:
```
rect : Object
    Represents the position of a vertical (e.g. a caret) line relative to which the panel should be positioned.
```
but I see I forgot about it in the `setVerticalPosition()` function anyway 🙈 And since it is rather hard to imagine something else, I agree we should go with `caretLeftPosition` here.

Dumluregn:
Rebased onto the latest `major`.

f1ames (approved):
LGTM :+1:

PR #4270, opened 2020-09-11
Fix: Incomplete CSS margin parsing if "auto" or "0" without unit is used

### LINKED ISSUE #1330: Bug: Incomplete CSS margin parsing if "auto" or "0" without unit is used

Bug

Related to #450 

Open HTML with inline CSS like `margin: 10px 0 0;` switch to Source mode, or
open HTML with inline CSS like `margin: 10px auto 0;` switch to Source mode.

### Expected result

`margin-bottom:0; margin-left:0; margin-right:0; margin-top:10px`, or
`margin-bottom:0; margin-left:auto; margin-right:auto; margin-top:10px`

### Actual result

`margin-bottom:10px; margin-left:10px; margin-right:10px; margin-top:10px`

Incorrect margins are applied

hub33k:

Bug fix

### This PR contains

- [x] Unit tests
- [ ] Manual tests

```
* [#1330>](https://github.com/ckeditor/ckeditor4/issues/1330): Fix converting margin style shorthand to separate styles.
```

## What changes did you make?

Changed regex for `splitMarginShorthand` function so it now converts margin shorthands correctly.

Closes #1330.


tests/core/filter/editor.js
```diff
@@ -339,6 +339,35 @@
 				assert.isFalse( 'filter' in editor, 'editor.filter was deleted' );
 				assert.isFalse( 'activeFilter' in editor, 'editor.activeFilter was deleted' );
 			} );
+		},
+
+		// (#1330)
+		'test converting margin shortcut to separate styles': function() {
```
f1ames, line 345: I think you could go with more unit test like approach, checking only modified method. It is available under `CKEDITOR.filter.transformationTools.splitMarginShorthand`.

Or, there is already a similar test (more like integration one) in filter transformation test: https://github.com/ckeditor/ckeditor4/blob/8fccb33c878b0ec2bcc414d3ca2b1b85c0eafd2e/tests/core/filter/transformations.js#L416

So I'm for the second approach here - introducing similar test to what we have `test margin transformations`, checking all possible cases with `0`s - this are mentioned in #1330. You can also [take a look on MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/margin#Syntax).

f1ames (changes requested):
I'm not sure if we need manual test here since it doesn't bring any additional value over unit tests. We just need to check input - output of this method. And since it is more an integration test than unit one (checking entire flow using editor instance) we have (almost) exactly the same execution flow as in manual test.

core/filter.js
```diff
@@ -2242,7 +2242,7 @@
 				return;
 			}
 
-			var widths = element.styles.margin.match( /(\-?[\.\d]+\w+)/g ) || [ '0px' ];
+			var widths = element.styles.margin.match( /(\-?[\.\d]*\w+)/g ) || [ '0px' ];
```
f1ames, line 2245: Changes in this regex made it work with `0`/`auto` values and it works better than before indeed. However, analyzing it deeper I'm not sure if it shouldn't be changed completely :thinking: 

The previous version also accepts invalid values (e.g. two or more digit numbers without units):



And yours while fixing initial issue also accepts invalid values (any digit without units):

image

I'm not sure how far we should polish it TBH because it may get complex. CSS supports percentage here too, which our regex can't handle (well it can but results in extracting number only which is not correct handling)...

But ok, let's be reasonable here - this function job is to simply split margin shorthand and not to validate what's there. The browser will simply not render the margin if values are incorrect. So this means all valid values should be used in the same form in split properties. And the ones we handle incorrectly are percentage values (because e.g. `10%` will be translated to `10` which is incorrect) and so lets try to additionally handle something like (please also add unit test for the `%` case):

```
5%
auto 10%
123px -10.5% 11.5
0% 10em 12.3px 1234%
```

---

I also wonder about `|| [ '0px' ]` part. Since we check if styles are empty in the beginning, the regex should always return some result, unless... and we are back to introduced change in regex. Previous one didn't match text values:
image

while this one does:

image

---

To sum up, I would be for modifying the regex in a way to not accepted many more than required - so `0` and `auto`, which are in fact special cases and the rest are just numbers + units. And so the regex could be extended like:

```
(auto|0|(?:\-?[\.\d]+\w+))
```
It will match `0` (unit doesn't matter here so will be skipped), match `auto` or match number + unit. Or more precisely:

```
(auto|0|(?:\-?[\.\d]+(?:\w+|%)))
``` 
to match `%` too (if you see it can be improved or made shorter feel free to change it).

Still it kind of validates input not allowing single digits without units, but allowing e.g `12` (due to `\w+` which also includes digits), not allowing text only, etc.  But let's keep it like that, fixing the way `0`, `auto` and `%` are handled as mentioned above.

f1ames (changes requested):
Looks good :+1: Unit tests are now where they should be too.

I only had some doubts related to used regex (both versions in fact), please see my comment.

hub33k:
Rebased onto latest `master`.

hub33k:
Red CI is caused by #4285.

f1ames:
Rebased onto latest `master`.

f1ames (changes requested):
I added two unit tests checking backwards compatibility for handling invalid (or partially invalid) values. It would be good to keep this behavior consistent since they may be some editor parts (hopefully not) which relays on this behavior. It turned out that `|| [ '0px' ]` alternative (mentioned somewhere in https://github.com/ckeditor/ckeditor4/pull/4270#discussion_r489507738) is still needed :see\_no\_evil: 

Now it handles invalid values the same way it did before, but some tests fails on IEs. It would be good to check what's going on. Since also invalid values tests fails there it might be good to check how it behaved with the previous regexp version there.

**IE8**:



**IE9/IE10/IE11**:

hub33k:
One test is failing (checked on Chrome)

hub33k:
It seems that IE completely removes any invalid styles that occurs in code :thinking:

f1ames:
> It seems that IE completely removes any invalid styles that occurs in code

Yes, if IE just doesn't accept invalid values we may skip those tests :+1: Btw. have you checked if it works the same with the previous regexp (if it's a browser fault it should).

> `rem` is  invalid in IE8, becouse it's supported from >= IE9 https://caniuse.com/rem

I assume it is also removed in IE8 as invalid value. Is there any unit we can use here instead of `rem` (apart from `px` and `%`) to work also on IE8? I guess `pt` would be sufficient for example?

f1ames (changes requested):
See https://github.com/ckeditor/ckeditor4/pull/4270#issuecomment-701987405.

hub33k:
With previous regexp on IE8 these tests fails:



With previous regexp on IE11 these tests fails:

image

hub33k:
> I guess `pt` would be sufficient for example?

Yes, it works on IE8. I assume we want to change `rem` to `pt`?

hub33k:
@f1ames I've replaced \`rem\` with `pt` and ignored testing invalid values on IE.

f1ames (approved):
LGTM :+1:

PR #4276, opened 2020-09-15
Add stalebot GitHub action

jacekbogdanski:
Small automation for closing stale issues.

### DIFF
--- a/.github/workflows/stalebot.yml
+++ b/.github/workflows/stalebot.yml
@@ -19,8 +19,8 @@ jobs:
         stale-pr-label: 'pr:stale'
         exempt-issue-labels: 'status:confirmed'
         exempt-pr-labels: 'status:confirmed'
-		close-issue-label: 'resolution:expired'
-		close-pr-label: 'pr:frozen'
+        close-issue-label: 'resolution:expired'
+        close-pr-label: 'pr:frozen'
         remove-stale-when-updated: true
         days-before-stale: 7
         days-before-close: 7
--- a/.github/workflows/stalebot.yml
+++ b/.github/workflows/stalebot.yml
@@ -16,7 +16,7 @@ jobs:
         close-issue-message: "There was no activity regarding this issue in the last 14 days. We're closing it for now. Still, feel free to provide us with requested feedback so that we can reopen it."
         close-pr-message: "There was no activity regarding this pull request in the last 14 days. We're closing it for now. Still, feel free to provide us with requested feedback so that we can reopen it."
         stale-issue-label: 'status:stale'
-        stale-pr-label: 'pr:stale'
+        stale-pr-label: 'pr:stale :moyai:'
         exempt-issue-labels: 'status:confirmed'
         exempt-pr-labels: 'status:confirmed'
         close-issue-label: 'resolution:expired'
--- a/.github/workflows/stalebot.yml
+++ b/.github/workflows/stalebot.yml
@@ -13,14 +13,14 @@ jobs:
         repo-token: ${{ secrets.GITHUB_TOKEN }}
         stale-issue-message: "It's been a while since we last heard from you. We are marking this issue as stale due to inactivity. Please provide the requested feedback or the issue will be closed after next 7 days."
         stale-pr-message: "It's been a while since we last heard from you. We are marking this pull request as stale due to inactivity. Please provide the requested feedback or the pull request will be closed after next 7 days."
-        close-issue-message: "There was no activity regarding this issue in the last 14 days. We're closing it for now. Still, feel free to provide us requested feedback so that we can reopen it."
-        close-pr-message: "There was no activity regarding this pull request in the last 14 days. We're closing it for now. Still, feel free to provide us requested feedback so that we can reopen it."
+        close-issue-message: "There was no activity regarding this issue in the last 14 days. We're closing it for now. Still, feel free to provide us with requested feedback so that we can reopen it."
+        close-pr-message: "There was no activity regarding this pull request in the last 14 days. We're closing it for now. Still, feel free to provide us with requested feedback so that we can reopen it."
         stale-issue-label: 'status:stale'
         stale-pr-label: 'pr:stale'
         exempt-issue-labels: 'status:confirmed'
         exempt-pr-labels: 'status:confirmed'
         close-issue-label: 'resolution:expired'
-        close-pr-label: 'pr:frozen'
+        close-pr-label: 'pr:frozen ❄'
         remove-stale-when-updated: true
         days-before-stale: 7
         days-before-close: 7
--- a/.github/workflows/stalebot.yml
+++ b/.github/workflows/stalebot.yml
@@ -11,14 +11,16 @@ jobs:
     - uses: actions/stale@v3
       with:
         repo-token: ${{ secrets.GITHUB_TOKEN }}
-        stale-issue-message: "Hi, It's been a while since we last heard from you. We are marking this issue as stale due to inactivity. Please, provide the requested feedback, or we will have to close the issue in the next 7 days."
-        stale-pr-message: "Hi, It's been a while since we last heard from you. We are marking this pull request as stale due to inactivity. Please, provide the requested feedback, or we will have to close the pull request in the next 7 days."
-        close-issue-message: "This issue has not been active for 14 days already. We're closing it for now. Still, feel free to provide us requested feedback so that we can reopen it."
-        close-pr-message: "This pull request has not been active for 14 days already. We're closing it for now. Still, feel free to provide us requested feedback so that we can reopen it."
+        stale-issue-message: "It's been a while since we last heard from you. We are marking this issue as stale due to inactivity. Please provide the requested feedback or the issue will be closed after next 7 days."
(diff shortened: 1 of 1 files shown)


.github/workflows/stalebot.yml
```diff
@@ -0,0 +1,24 @@
...
+  schedule:
+  - cron: "0 9 * * *"
+
+jobs:
+  stale:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/stale@v3
+      with:
+        repo-token: ${{ secrets.GITHUB_TOKEN }}
```
f1ames, line 13: Do we need to generate some token or it happens automatically? :thinking:

.github/workflows/stalebot.yml
```diff
@@ -0,0 +1,24 @@
...
+        stale-issue-message: "Hi, It's been a while since we last heard from you. We are marking this issue as stale due to inactivity. Please, provide the requested feedback, or we will have to close the issue in the next 7 days."
+        stale-pr-message: "Hi, It's been a while since we last heard from you. We are marking this pull request as stale due to inactivity. Please, provide the requested feedback, or we will have to close the pull request in the next 7 days."
+        close-issue-message: "This issue has not been active for 14 days already. We're closing it for now. Still, feel free to provide us requested feedback so that we can reopen it."
+        close-pr-message: "This pull request has not been active for 14 days already. We're closing it for now. Still, feel free to provide us requested feedback so that we can reopen it."
+        stale-issue-label: 'status:stale'
+        stale-pr-label: 'status:stale'
```
f1ames, line 19: I wonder if we should use the same labels for PRs. Generally makes sense, OTOH we never labeled PRs so it might be a little odd on the beginning. Maybe different labels would make it easier to distinguish what is the purpose here and there? Something simple like `pr:stale`?

Also for closed PRs we can add our `pr:frozen` label (with `close-pr-label:`) so it will make sense to have `pr:*` labels for better consistency. And for closed issues, we should add `resoltuion:expired` probably?

.github/workflows/stalebot.yml
```diff
@@ -0,0 +1,24 @@
...
+  - cron: "0 9 * * *"
+
+jobs:
+  stale:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/stale@v3
+      with:
+        repo-token: ${{ secrets.GITHUB_TOKEN }}
+        stale-issue-message: "Hi, It's been a while since we last heard from you. We are marking this issue as stale due to inactivity. Please, provide the requested feedback, or we will have to close the issue in the next 7 days."
```
f1ames, line 14:
```suggestion
        stale-issue-message: "Hi, it's been a while since we last heard from you. We are marking this issue as stale due to inactivity. Please provide the requested feedback or the issue will be closed after next 7 days."
```

I know we have "Hi, It's..." in our templates but I think it's not correct that way? Unless, there are line breaks and it's like e.g. email when you write:

```
Hello Jacek,

It's been...
```

:thinking:

f1ames: Not 100% sure about the grammar but it seems there were too many `,` there :thinking: Can you also update pr message accordingly?

.github/workflows/stalebot.yml
```diff
@@ -0,0 +1,24 @@
...
+        stale-pr-message: "Hi, It's been a while since we last heard from you. We are marking this pull request as stale due to inactivity. Please, provide the requested feedback, or we will have to close the pull request in the next 7 days."
+        close-issue-message: "This issue has not been active for 14 days already. We're closing it for now. Still, feel free to provide us requested feedback so that we can reopen it."
```
f1ames, line 16:
```suggestion
        close-issue-message: "There was no activity regarding this issue in the last 14 days. We're closing it for now. Still, feel free to provide us requested feedback so that we can reopen it."
```

f1ames (changes requested):
Generally looks good :+1: Some minor polishing here and there.

.github/workflows/stalebot.yml
jacekbogdanski: No, GH will take care of it.

.github/workflows/stalebot.yml
jacekbogdanski: Using the same label seems just simpler. Not sure why we would need to distinguish both, because they are still from different views (https://github.com/ckeditor/ckeditor4/issues vs https://github.com/ckeditor/ckeditor4/pulls). But since we already are doing such distinction, I have no problem changing it.

> And for closed issues, we should add resoltuion:expired probably?

I see there is an option for that, I will update PR accordingly.

.github/workflows/stalebot.yml
jacekbogdanski: I would just remove "Hi" since it should be a continuation of the discussion, so a user should be already greeted by us.

.github/workflows/stalebot.yml
```diff
@@ -0,0 +1,26 @@
...
+      with:
+        repo-token: ${{ secrets.GITHUB_TOKEN }}
+        stale-issue-message: "It's been a while since we last heard from you. We are marking this issue as stale due to inactivity. Please provide the requested feedback or the issue will be closed after next 7 days."
+        stale-pr-message: "It's been a while since we last heard from you. We are marking this pull request as stale due to inactivity. Please provide the requested feedback or the pull request will be closed after next 7 days."
+        close-issue-message: "There was no activity regarding this issue in the last 14 days. We're closing it for now. Still, feel free to provide us requested feedback so that we can reopen it."
+        close-pr-message: "There was no activity regarding this pull request in the last 14 days. We're closing it for now. Still, feel free to provide us requested feedback so that we can reopen it."
+        stale-issue-label: 'status:stale'
+        stale-pr-label: 'pr:stale'
+        exempt-issue-labels: 'status:confirmed'
+        exempt-pr-labels: 'status:confirmed'
```
f1ames, line 21: I wonder if we really need this? We never label PRs, but OTOH it's kind of a workaround if we would like to keep some PRs opened longer... Still I recall we discussed that we shouldn't keep such PRs open and if there is some work going on it will be updated.

I'm for removing it TBH. WDYT?

.github/workflows/stalebot.yml
```diff
@@ -0,0 +1,26 @@
...
+        stale-issue-message: "It's been a while since we last heard from you. We are marking this issue as stale due to inactivity. Please provide the requested feedback or the issue will be closed after next 7 days."
+        stale-pr-message: "It's been a while since we last heard from you. We are marking this pull request as stale due to inactivity. Please provide the requested feedback or the pull request will be closed after next 7 days."
+        close-issue-message: "There was no activity regarding this issue in the last 14 days. We're closing it for now. Still, feel free to provide us with requested feedback so that we can reopen it."
+        close-pr-message: "There was no activity regarding this pull request in the last 14 days. We're closing it for now. Still, feel free to provide us with requested feedback so that we can reopen it."
+        stale-issue-label: 'status:stale'
+        stale-pr-label: 'pr:stale'
+        exempt-issue-labels: 'status:confirmed'
+        exempt-pr-labels: 'status:confirmed'
+        close-issue-label: 'resolution:expired'
+        close-pr-label: 'pr:frozen ❄'
```
f1ames, line 23: I hope it will work :see\_no\_evil:

f1ames: Now, we have consistent PR labels too:

f1ames (changes requested):
@jacekbogdanski is it possible to configure GH user which will be commenting/closing issues/PRs? We have special bot user (used for example in presets update action - https://github.com/ckeditor/ckeditor4-presets/blob/93429c55c92436cea8a8ef533c9d6bcbd32e2ae0/.github/workflows/update.yml#L33-L34), which could be used here and probably have more permissions :thinking:

jacekbogdanski:
> @jacekbogdanski is it possible to configure GH user which will be commenting/closing issues/PRs? We have special bot user (used for example in presets update action - https://github.com/ckeditor/ckeditor4-presets/blob/93429c55c92436cea8a8ef533c9d6bcbd32e2ae0/.github/workflows/update.yml#L33-L34), which could be used here and probably have more permissions thinking

Not sure, I suppose if we would use a different token it would probably use token owner instead - https://github.com/ckeditor/ckeditor4/blob/b8df87015a9a7900bcb82134edc3390086328b51/.github/workflows/stalebot.yml#L13

But in that case, we will have to add this token into secrets for each repository - I don't know if secrets can be managed on the organization level. What permissions issue do you would like to solve with a different user?

f1ames (approved):
LGTM :+1: 

---

> Not sure, I suppose if we would use a different token it would probably use token owner instead -
> 
> https://github.com/ckeditor/ckeditor4/blob/b8df87015a9a7900bcb82134edc3390086328b51/.github/workflows/stalebot.yml#L13
> 
> But in that case, we will have to add this token into secrets for each repository - I don't know if secrets can be managed on the organization level. What permissions issue do you would like to solve with a different user?

I think there was some problem with issue marked with `stale` label manually. But as you mentioned it might by cumbersome when adding to other repos and we don't plan to use this label manually too, so let's go with the current approach and see if there are any issues.

PR #4277, opened 2020-09-15
Fix resize table with rowspan cells

### LINKED ISSUE #3961: Table resize leaves cell uneditable and leaves cursor in <-> resize mode
## Type of report

Bug

Merge cells from a html table-- and then have it lock the resulting cell as contenteditable=false unselectable=on.

<div data-cke-temp="1" contenteditable="false" unselectable="on" style="position: absolute; cursor: col-resize; opacity: 0; padding: 0px; background-color: rgb(0, 0, 68); background-image: none; border: 0px none; z-index: 10; width: 444.091px; height: 594.182px; left: 400px; top: 26.3636px; display: none;"></div>

If I remove the tableresize plugin, I can edit the cell, but the cursor still remains in resize mode.

NOTE: I was able to reproduce this by using "cell -> merge down" several times on an 18 row table. See uploaded screenshot

### Expected result

Should be able to edit content in the cell after cell merge

### Actual result

A purple-ish hover-over prevents from editing the cell.. and the screen has a resize arrow for a cursor.

image

## Other details

* Browser: … firefox or chrome
* OS: … macos/linux
* CKEditor version: … 4.13.1
* Installed CKEditor plugins: … table tools/table selection

jacekbogdanski:

### This PR contains

- [ ] Unit tests
- [x] Manual tests

```
* [#3961](https://github.com/ckeditor/ckeditor4/issues/3961): Table resize prevents editing merged cells.
```

## What changes did you make?

Rowspans are not taken into consideration when calculating resize pillars. To fix the issue, I'm using `tools.buildTableMap` helper, so if a row has fewer cells due to rowspan neighbor, it will be extended with shared cells.

Writing unit tests for this issue is a pain because the whole interaction is done via mouse. I'm not really sure if we will be able to write any sensible automatic test for that, so let's check if these changes are correct before jumping to unit tests as they seem hard to write (and re-write).

Closes #3961

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -3,6 +3,10 @@ CKEditor 4 Changelog
 
 ## CKEditor 4.15.1
 
+Fixed Issues:
+
+* [#3961](https://github.com/ckeditor/ckeditor4/issues/3961): Fixed: [Table Resize](https://ckeditor.com/cke4/addon/tableresize) prevents editing of merged cells.
+
 Other Changes:
 
 * [#4262](https://github.com/ckeditor/ckeditor4/issues/4262): Removed global reference to `stylesLoaded` variable. Thanks to [Levi Carter](https://github.com/swiftMessenger)!
--- a/core/tools.js
+++ b/core/tools.js
@@ -2199,6 +2199,27 @@
 				}
 
 				return false;
+			},
+
+			/**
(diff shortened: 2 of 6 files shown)


Comandeer (changes requested):
The fix seems to work in all browsers except IE8. It de facto hangs browser.

Comandeer (approved):
LGTM!

PR #4307, opened 2020-09-30
Fix: Autogrow plugin causing width of editor's container to be set to 0px

### LINKED ISSUE #4286: Autogrow plugin causing width of editor's container to be set to 0px
## Type of report

Bug

1. Navigate to CKEditor's autogrow plugin demo site: https://ckeditor.com/docs/ckeditor4/latest/examples/autogrow.html#!/guide/dev_autogrow
2. Remove border widths from the container div (div with id='cke_editor1')
3. Click into the editor 
4. Press Enter until the size of the editor grows (at least twice)

### Expected result

*The length of the editor should increase and the width should not change*
Screenshot:
image

### Actual result

*The width of the container (div with id='cke_editor1') is set to 0px and becomes unusable*
Screenshot:
image

## Debugging details
- This commit in version `4.14.1` has caused this issue to surface: https://github.com/ckeditor/ckeditor4/pull/4010/files?file-filters%5B%5D=.js
- More specifically, this line: https://github.com/ckeditor/ckeditor4/pull/4010/files?file-filters%5B%5D=.js#diff-d0c2cb93caeb33d1d59fd2506921ea4aR290
  - Previously, before `width = convertCssUnitToPx( width );` was introduced, width was an empty string `""` and so the next line `outer.setSize( 'width', width, true );` did not actually set the style because `width` was not a number.
  - Now that `width` is a number (0), it calculates the width size and adjusts based on padding. 
    - During this calculation, if there is some padding (say 1px all around), then `size` gets set to `0 - 2 = -2px` and `setStyle()` is called. `setStyle()` doesn't set a negative width so it's a no-op.
    - If there is no padding, then `size` gets set to `0px`. `setStyle()` is called and the width of the container is then set to `0px`.

## Other details

* Browser: Chrome
* OS: MacOS Catalina
* CKEditor version: `4.14.1` and `4.15.0`
* Installed CKEditor plugins: `autogrow`

hub33k:

Bug fix

### This PR contains

*   [x] Unit tests
*   [x] Manual tests

*   [x] PR is consistent with the code style guide

```
* [#4286](https://github.com/ckeditor/ckeditor4/issues/4286): validate if width is valid value in editor.resize function.
```

## What changes did you make?

Check if width has valid value.

Closes #4286.


core/creators/themedui.js
```diff
@@ -287,7 +287,9 @@ CKEDITOR.replaceClass = 'ckeditor';
 			outer = container;
 		}
 
-		width = convertCssUnitToPx( width );
+		if ( width ) {
```
f1ames, line 290: It will also skip conversion for 0. Since 0 can be used without units (`px` in this case) it will work fine (kind of as a side effect). However, lack of such explicitness may lead to difficulties in understanding of this code (and it's intentions) in the future. Therefore I would be for checking `width || width === 0`. WDYT?

tests/plugins/autogrow/autogrowwidth.js
```diff
@@ -0,0 +1,37 @@
...
+		// #4286
+		'test editor width': function() {
+			var editor = this.editor,
+				bot = this.editorBot;
+
+			var html = '';
+			for ( var i = 0; i < 6; i++ ) {
+				html += '<p>test ' + i + '</p>';
+			}
+
```
f1ames, line 25: Maybe we can get editor width here and check if it's the same after editor autogrows (as a second check after `> 0`)?

tests/plugins/autogrow/autogrowwidth.js
```diff
@@ -0,0 +1,37 @@
...
+	bender.test( {
+		init: function() {
+			// Remove root editor element border.
+			var editorRootElement = CKEDITOR.document.getById( 'cke_test_editor' );
+			editorRootElement.setStyle( 'border', 'none' );
+		},
+
+		// #4286
+		'test editor width': function() {
```
f1ames, line 17: Since there are no test for `autogrow` at all, could we add similar test as this one but without removing border (checking if width doesn't change and if height grows)?

tests/plugins/autogrow/autogrowwidth.js
f1ames: You could also check if height increased as a last check.

tests/plugins/autogrow/autogrowwidth.js
```diff
@@ -0,0 +1,37 @@
+/* bender-tags: editor, autogrow */
+/* bender-ckeditor-plugins: autogrow */
+
+( function() {
+	'use strict';
+
+	bender.editor = {};
```
f1ames, line 7: It might be goo to set editor height and width explicitly so it is not influenced by any other plugins (e.g. in full presets tests).

tests/plugins/autogrow/autogrowwidth.js
```diff
@@ -0,0 +1,37 @@
...
+				html += '<p>test ' + i + '</p>';
+			}
+
+			bot.setData( html, function() {
+				setTimeout( function() {
+					resume( function() {
+						var width = parseInt( editor.editable().getComputedStyle( 'width' ), 10 );
+						assert.isTrue( width > 0, true );
+					} );
+				}, 1000 );
```
f1ames, line 32: Do we need such a long timeout? You may try to listen to `autogrow` event (however it is fired before physical eidtor size changes and provides data only about height) or `afterCommandExec` for `autogrow` command with some low priority to check resulted width.

f1ames (changes requested):

grumd:
We also encountered this issue in our project that's using CKEditor. Looking forward to see this merged soon.

Is there a way to use an older version of CKEditor in the online builder? (https://ckeditor.com/cke4/builder)
It only gives an option of using the latest version that's currently bugged.

f1ames:
> Is there a way to use an older version of CKEditor in the online builder? (https://ckeditor.com/cke4/builder)  
> It only gives an option of using the latest version that's currently bugged.

@vokiel correct me if I'm wrong, but I don't think it's doable with online builder. @grumd you would need to create CKEditor 4 build "manually" (using [`ckeditor4-presets` repo](https://github.com/ckeditor/ckeditor4-presets)).

grumd:
@f1ames That repo seems to only support building the three default presets (basic/standard/full), but I found that it's possible to build ck4 code by checking out the branch/tag you need, and running `/dev/builder/build.sh` and using a custom `/dev/builder/build-config.js`. But I'd rather have the latest version so I'll wait a while for this PR first.

vokiel:
> I don't think it's doable with online builder.

True, the online builder works with the latest version only.

hub33k:
@grumd you can create customized presets by creating two files: `presets/$1-build-config.js` & `presets/$1-ckeditor-config.js` (replace `$1` preset name) - https://github.com/ckeditor/ckeditor4-presets/commit/2e89eb06fb7694768ee9e2fe56497036dda9952d.

tests/plugins/autogrow/autogrowwidth.js
hub33k: It seems that setting width explicitly does not break editor width. I'm for not setting editor size here. WDYT?

f1ames: Ok, so setting editor width explicitly is also a workaround for this issue :+1:

Let's leave it without explicit width so we are sure test checks this regression.

tests/plugins/autogrow/autogrow.js
```diff
@@ -0,0 +1,45 @@
...
+				html += '<p>test ' + i + '</p>';
+			}
+
+			bot.setData( html, function() {
+				editor.once( 'afterCommandExec', function() {
+					resume( function() {
+						var editorWidth = getEditorSize( editor ).width,
+							editorHeight = getEditorSize( editor ).height;
+
+						assert.isTrue( editorHeight > initialEditorHeight, true, 'is editor height bigger than initial editor height' );
```
f1ames, line 28:
```suggestion
                        assert.isTrue( editorHeight > initialEditorHeight, 'editor height should increase' );
```

tests/plugins/autogrow/autogrow.js
```diff
@@ -0,0 +1,45 @@
...
+						assert.areEqual( editorWidth, initialEditorWidth, 'does width of editor changes' );
```
f1ames, line 29:
```suggestion
						assert.areEqual( editorWidth, initialEditorWidth, 'editor width does not change' );
```

f1ames (changes requested):
Looks good :+1:

Only some polishing in unit tests needed - I don't see a reason to keep both unit tests in separate files so it will be good to merge them into one. See also review comments.

f1ames:
Rebased onto latest `master`.

f1ames (approved):
LGTM :+1:

grumd:
Great work! Any ETA on when a new release may happen with this PR included?

f1ames:
@grumd no definite ETA so far, but we are targeting beginning of December.

PR #4308, opened 2020-10-02
Tests for table selection in Safari

### LINKED ISSUE #4287: Failing tests/plugins/tableselection/integrations/tabletools/tabletools tests
## Type of report

Failing test

1. Open http://tests.ckeditor.test:1030/tests/plugins/tableselection/integrations/tabletools/tabletools#tests%2Fplugins%2Ftableselection%2Fintegrations%2Ftabletools%2Ftabletools%20test%20delete%20nested%20cells%20(classic)

### Expected result

Test passes

### Actual result

This test and its inline counterpart fails in Safari 14:
Screenshot 2020-09-18 at 14 56 43

## Other details

* Browser: Safari 14
* Installed CKEditor plugins: tableselection, tabletools

### LINKED ISSUE #4306: Can't select last column if there is a header in the table
## Type of report

Bug

1. Open https://jsfiddle.net/Comandeer/2841j0cw/
2. Try to select the last column

### Expected result

The last column can be selected.

### Actual result

The last column can't be selected or the selection is selecting the whole table.

## Other details

* Browser: Safari 14
* OS: macOS
* CKEditor version: 4.7.0+

Behaviour of the bug is different in different CKE4 versions:
* **4.7.0–4.12.1**: the table selection is not possible for the last column, the native selection is kicking in;
* **4.13.0+**: the table selection is activated, but it selects the first row of the table or the whole table – it can't be dragged down the column.

Comandeer:

Task

### This PR contains

- [x] Unit tests
- [x] Manual tests

Not needed

## What changes did you make?

I've ignored the unit test that fails in Safari and added a manual one to check if the upstream issue is still present.

Closes #4287.
Closes #4306.

### DIFF
--- a/tests/plugins/tableselection/manual/safaritheadlastcolumn.html
+++ b/tests/plugins/tableselection/manual/safaritheadlastcolumn.html
@@ -1,0 +1,61 @@
+<p><button id="select">Select the last column</button></p>
+
+<div id="editor">
+	<table border="1">
+		<thead>
+			<tr>
+				<th scope="col">Column 1</th>
+				<th scope="col">Column 2</th>
+				<th scope="col">Column 3</th>
+				<th scope="col">Column 4</th>
+			</tr>
+		</thead>
+		<tbody>
+			<tr>
+				<th scope="row">Cell 1.1</th>
+				<td>Cell 1.2</td>
+				<td>Cell 1.3</td>
+				<td>Cell 1.4</td>
+			</tr>
+			<tr>
+				<th scope="row">Cell 2.1</th>
+				<td>Cell 2.2</td>
+				<td>Cell 2.3</td>
+				<td>Cell 2.4</td>
+			</tr>
+			<tr>
+				<th scope="row">Cell 3.1</th>
+				<td>Cell 3.2</td>
+				<td>Cell 3.3</td>
+				<td>Cell 3.4</td>
+			</tr>
+			<tr>
+				<th scope="row">Cell 4.1</th>
+				<td>Cell 4.2</td>
+				<td>Cell 4.3</td>
+				<td>Cell 4.4</td>
+			</tr>
+			<tr>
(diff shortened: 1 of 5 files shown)


jacekbogdanski (commented):
Do the tests start failing from Safari 14? It seems to work when using Safari 13.1 and it's the latest version available on Browser Stack.

Comandeer:
Yes, they fail only on Safari 14.

jacekbogdanski:
Thanks, I'm removing myself from review as I don't have macOS to test it.

tests/plugins/tableselection/manual/safaritheadlastcolumn.html
```diff
@@ -0,0 +1,61 @@
...
+				<td>Cell 5.3</td>
+				<td>Cell 5.4</td>
+			</tr>
+		</tbody>
+	</table>
+</div>
+
+<script src="../_helpers/tableselection.js"></script>
+<script>
+	if ( !CKEDITOR.env.safari ) {
```
Dumluregn, line 50: As there is a part of test that requires mouse, we should either ignore the whole test on mobile or add a note that the second point shouldn't be executed there.

Dumluregn (changes requested):
The manual test in the current form fails. I think it will be misguiding during the testing, so I'd propose to switch the `Expected` and `Unexpected` results so that test passed and additionally add the explicit note that if test is passing then the bug is still present.

The other solution is adding a note that the test should fail instead, but I'd rather go with a positive way as in the long term it should be clearer to understand.

Also manual test should be ignored on mobile (see the comment).

Comandeer:
I've split the test into two: the one is for selecting the column with the button and the second one, ignored on mobile, to select with mouse.

Dumluregn (approved):
LGTM! 🎉

PR #4313, opened 2020-10-06
Change stale label for stalebot

jacekbogdanski:
Let's check if it will work correctly without `:` character. I see that closing labels has been normally added so it's probably safe to leave them for now.

Please, when merging this PR, also rename `status:stale` label to `stale`.


f1ames (approved):
LGTM :+1:

PR #4291, opened 2020-09-20
Improve importing images from Microsoft Word

Comandeer:
## What is the purpose of this pull request?

Bug fix

## Does your PR contain necessary tests?

All patches that change the editor code must include tests. You can always read more
on [PR testing](https://ckeditor.com/docs/ckeditor4/latest/guide/dev_contributing_code.html#tests),
[how to set the testing environment](https://ckeditor.com/docs/ckeditor4/latest/guide/dev_tests.html) and
[how to create tests](https://ckeditor.com/docs/ckeditor4/latest/guide/dev_tests.html#creating-your-own-test)
in the official CKEditor documentation.

### This PR contains

- [x] Unit tests
- [x] Manual tests

## Did you follow the CKEditor 4 code style guide?

Your code should follow the guidelines from the [CKEditor 4 code style guide](https://github.com/ckeditor/ckeditor4/blob/major/dev/docs/codestyle.md) which helps keep the entire codebase consistent.

- [x] PR is consistent with the code style guide

## What is the proposed changelog entry for this pull request?

```
Fixed Issues:

* [#2800](https://github.com/ckeditor/ckeditor4/issues/2800): Fixed: no images are imported from Microsoft Word if there is at least one image in unsupported format.

API Changes:

* [#3782](https://github.com/ckeditor/ckeditor4/issues/3782): Merge [`CKEDITOR.plugins.pastetool.filters.word.images`](https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_plugins_pastetools_filters_word_images.html) to [`CKEDITOR.plugins.pastetools.filters.image`](https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_plugins_pastetools_filters_image.html).
```

## What changes did you make?

Ok, this was a really tough one:

1. I've implemented a very primitive RTF parser to get rid of headers and footers before further parsing the document.
2. Now all images are extracted, even the ones with unknown formats. Thanks to that, we can at least partially render images in pasted content (ones in unknown formats are still broken, but the rest is rendered correctly). For supporting these unsupported formats see #4290.
3. Some images are inserted several times – in different formats (e.g. PNG & WMF) or for additional version for non-Word readers. The non-Word ones are dismissed and images with the same ids are not extracted if there is already an image data for the id.
4. Some images are WordArt shapes and they are not included in HTML source as `img` tags. Fortunately they are easily filtered by using `\defshp`.
[...]

PR #4339, opened 2020-10-19
Add support for Word's images pasting to Safari

### LINKED ISSUE #1134: Integration PFW Image with Safari

Comandeer:

New feature

### This PR contains

- [x] Unit tests
- [x] Manual tests

```
* [#1134](https://github.com/ckeditor/ckeditor4/issues/1134): [Safari] Fixed: [Paste from Word](https://ckeditor.com/cke4/addon/pastefromword) does not embed images.
```

## What changes did you make?

I've basically taken the contents from @msamsel's #1976 and integrated it with our `CKEDITOR.plugins.pastetools.filters.image` filter.

Closes #1134.


plugins/pastetools/filter/image.js
```diff
@@ -300,13 +362,60 @@
...
+	function getImageTypeFromHeader( bytesArray ) {
+		var header = '',
+			fileType = null,
+			bytesHeader = bytesArray.subarray( 0, 4 );
+
+		for ( var i = 0; i < bytesHeader.length; i++ ) {
+			header += bytesHeader[ i ].toString( 16 );
+		}
+
+		switch ( header ) {
```
jacekbogdanski, line 403: That's some dark magic 😄 I'm not sure if it should be named `header`, probably [File signature](https://en.wikipedia.org/wiki/File_signature) is more accurate here. It would be good to have real, existing images with these signatures, to confirm that the feature covers all expected image types. These can be checked e.g. here: https://www.filesignatures.net/index.php?search=FFD8FFE0&mode=SIG although I suppose it may be hard to get all `spf`, `jfif` etc extension examples and I'm not sure if Word is not converting them also into a more unified format.

plugins/ajax/plugin.js
```diff
@@ -128,12 +136,16 @@
 			 * @param {String} url The URL from which the data is loaded.
 			 * @param {Function} [callback] A callback function to be called on
 			 * data load. If not provided, the data will be loaded
-			 * synchronously.
-			 * @returns {String} The loaded data. For asynchronous requests, an
+			 * synchronously. Please notice that only text data might be loaded synchrnously.
+			 * @param {String} [responseType='text'] Defines type of returned data.
+			 * Currently supports: `text`, `xml`, `arraybuffer`. This parameter was added in 4.10.
```
jacekbogdanski, line 141:
```suggestion
			 * Currently supports: `text`, `xml`, `arraybuffer`. This parameter was added in 4.16.0.
```

plugins/ajax/plugin.js
```diff
@@ -128,12 +136,16 @@
 			 * @param {String} url The URL from which the data is loaded.
 			 * @param {Function} [callback] A callback function to be called on
 			 * data load. If not provided, the data will be loaded
-			 * synchronously.
-			 * @returns {String} The loaded data. For asynchronous requests, an
+			 * synchronously. Please notice that only text data might be loaded synchrnously.
```
jacekbogdanski, line 139:
```suggestion
			 * synchronously. Please notice that only text data might be loaded synchronously.
```

tests/plugins/ajax/ajax.js
```diff
@@ -179,6 +179,27 @@
 				} );
 			} );
 
+			wait();
+		},
+
+		// (#1134)
+		'test load async arraybuffer': function() {
```
jacekbogdanski, line 186: We are also missing `text` and `xml` types. I guess they are probably tested already as we are using them indirectly in ajax logic, but it would be good to check if API works correctly.

tests/plugins/pastefromword/_helpers/blob.js
```diff
@@ -0,0 +1,100 @@
...
...vULeoPXrMO5jbVotsiOQE8hGFh56FJohum5AZs/bIQsChGEhMkJwBZ1V/YLoUn3Qj0f/0dyaKHyUIA3ykHQTHyeGqLAI2Y9nnxk7YXRMqRhQBky6C/vCGSScquNkQIH6exlO5NSTYCeP30JHqMPq1sb3Wx9j1/dgS9Jh/p4lo7rk42ApBag9dHA+b1HWuokqsamwcYsmGV5NbKSgDXYQno95uNkhvNYoPBr62HQ2CG5KpJFi1cl2W0K+fxwfGqDiblIASPGpsYLsl7/6N+kLUHIGXMbR8gbKt6N47SWFdj5J45sjif8JedAKQEbeQc9KsP6VB1afC6aLgMvSKb3bLfAposb6i1wJkxR0ETJdWS6Kn/icPyrhVkogcgh9O7b8S+2bDxyVGF/2nN5PCPF8luKzMEICRsvO4i8vMZoFei7pw0tlWIEDaLsPYolN1MpghAaLiNiEPUoandlwSaBowQtoZzj7gku/NJAeYIQEp5jTuBKOHeSP+UbpWEWAlhW5bD57l/M2MWkwQgdLyEtQT0gGThWMYMWCYpIvzzh8UsET7z/tOkarpamJm3AF2G0U6jF2fEoiShazZVpvdul7AEuAldMEXvJoJZOpVCI2zeVJ3viao73qi83RuF5yOFcC7CHP0uiHVo7lCO8E8XMPXPb8KbeQKQovXVVkhe/Q7SNi1r0tugX3KA36y96D1/p7Ataxw0VvpF7Kwpdkb+1+ORuXM28o+NRWOd4Rut0ihfpLAppNPAFIN0lqqwIgjQBMb9S8Nxfe0a5H05sSmrw1/n0EQELt0mbNKwB6YGrazK6YVUYe9D2vhanx3QqKeh8f3gaGFql8xbw3UEkqII0GRIccIQrSMKz0X9f958i26aXiHpU3PAkm3wnnS0SaTLfqlXSPv7MmTtnoGy6/3b1GvdMw8BCz9GwOIY2Pllt7nOWoYiCdASRIq/W3ptoLDO3krYjaMA9kHpkv3jym70Q1lKMGpLnWBuVwmXIclC+2nM7AjWEiddx4ongC7DeL5+CDD7Gqif+ryUqQhwApiKoMLlOQEU7kBT1ecEMBVBhctzAijcgaaqzwlgKoIKl+cEULgDTVWfE8BUBBUuzwmgcAeaqj4ngKkIKlyeE0DhDjRVfU4AUxFUuDwngMIdaKr6nACmIqhweU4AhTvQVPX/B7+LJxLB2zHEAAAAAElFTkSuQmCC',
```
jacekbogdanski, line 34: Do you think it would be possible to use real images instead with more test cases?

tests/plugins/pastefromword/manual/pasteimagessafari.html
```diff
@@ -0,0 +1,7 @@
+<textarea id="editor" cols="30" rows="10"></textarea>
+<script>
+	if ( !CKEDITOR.env.safari ) {
+		bender.ignore();
+	}
```
jacekbogdanski, line 5: Maybe it makes sense to keep this test also for different browsers to make sure we didn't introduce regression?

tests/plugins/pastefromword/manual/pasteimagessafari.md
```diff
@@ -0,0 +1,20 @@
+@bender-tags: bug, 4.16.0, 1134
+@bender-ui: collapsed
+@bender-ckeditor-plugins: wysiwygarea, toolbar, undo, basicstyles, pastefromword, elementspath, image, format, sourcearea
+
+1. Open [attached Word document](_assets/Image_safari.docx).
```
jacekbogdanski, line 5: It would be good to add a more complicated document with more image sources, especially gif and different flavors of jpg.

jacekbogdanski (changes requested):
Works pretty nice! I see that this code had already some partial review in original PR, so there is not much left. For now, I've tested it with jpg, png, and gif. Please, see the review comments.

I also see that you marked this PR as a feature, although the original issue has been reported as a bug. I suppose that we could change it to feature as image support for OS X Word edition.

jacekbogdanski:
I also see that there are some jsduck errors 
image

Please, update docs to match the correct types.

Comandeer:
> Please, update docs to match the correct types.

Hmm, that can be a little problematic, as [`Uint8Array`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Uint8Array) is correct… Seems like jsduck is old enough to not recognize Typed Arrays :thinking:

plugins/pastetools/filter/image.js
```diff
@@ -300,13 +362,60 @@
...
+			case '89504e47':
+				fileType = 'image/png';
+				break;
+			case '47494638':
+				fileType = 'image/gif';
```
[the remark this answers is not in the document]
Comandeer: Wow, I didn't know that Word supports GIFs :sweat\_smile: I was pretty sure they are converted to PNGs.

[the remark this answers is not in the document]
Comandeer: Actually they are converted to PNG… So it works because [most browsers support APNG](https://caniuse.com/apng).

Safari in such a case use GIF format. It's strange that it isn't rendered correctly as we have GIF recognition :thinking:

tests/plugins/pastefromword/manual/pasteimagessafari.html
Comandeer: All other browsers are covered by unit tests for Paste Tools, Paste from Word and Paste from Libre Office, so I don't think it would be necessary.

plugins/pastetools/filter/image.js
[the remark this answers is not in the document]
Comandeer: Adding GIF mime type to array of supported image formats seems to do the trick.

tests/plugins/pastefromword/manual/pasteimagessafari.md
Comandeer: How can I produce different flavors of JPG? :thinking:

Comandeer: Actually we probably don't need to test different flavors of JPG, as all of its signatures starts with the same sequence: `FF D8 FF`. Checking if the image file starts with it should be enough for our case.

Comandeer:
Ok,

*   support for GIFs added by adding them to the array of supported image types,
*   tests with real images added,
*   I've also added throwing error when the image in blob is in unsupported format (reused the error introduced in #4291,
*   some refactoring.

tests/plugins/pastefromword/_helpers/blob.js
```diff
@@ -0,0 +1,113 @@
...
+		return new Blob( [ arrayBuffer ], { type: fileType } );
+	}
+
+	function simulatePasteBlob( editor, assertion, options ) {
+		var image = options && options.image || SAMPLE_PNG,
+			imageDataUrl = options && options.imageDataUrl || SAMPLE_PNG,
+			url = getObjectUrl( image, options && options.imageType ),
+			template = options && options.template || '<p{%CLASS%}>Hello<img style="height:200px; width:200px" src="{%URL%}" />world.</p>',
+			input = template.replace( /{%URL%}/g, url ).replace( /{%CLASS%}/g, ' class="MsoNormal"' ),
+			expected = template.replace( /{%URL%}/g, imageDataUrl ).replace( /{%CLASS%}/g, '' );
```
jacekbogdanski, line 41: The code is a bit unreadable with all default option values. Maybe we could simplify it with some merging logic?

```js
CKEDITOR.tools.object.merge( defaultOptions, options || {} );
```

plugins/pastetools/filter/image.js
```diff
@@ -300,13 +390,50 @@
...
+		}
+
+		return img.type ? 'data:' + img.type + ';base64,' + CKEDITOR.tools.convertBytesToBase64( data ) : null;
+	}
+
+	function convertBlobUrlToBase64( blobUrlSrc ) {
+		return new CKEDITOR.tools.promise( function( resolve ) {
+			CKEDITOR.ajax.load( blobUrlSrc, function( arrayBuffer ) {
+				var data = new Uint8Array( arrayBuffer ),
+					imageType = getImageTypeFromSignature( data.subarray( 0, 4 ) ),
```
jacekbogdanski, line 411: Shouldn't we pass here the whole `arrayBuffer`? It seems like taking the first 4 characters is already done in `getImageTypeFromSignature` function? In that case, this should be probably renamed again to just `getImageType`.

jacekbogdanski (changes requested):
> Hmm, that can be a little problematic, as Uint8Array is correct… Seems like jsduck is old enough to not recognize Typed Arrays 🤔

I see that `Blob` is correctly recognized, so it's indeed weird that jsduck is not able to read this type. Maybe there is some workaround for custom type? If not, I suppose we will have to rename it into `Object` to avoid building warnings? Although, it's pity going some weird workarounds because the tooling doesn't work correctly.

Currently, ajax methods are a bit mixed up. Previously, we had:
* `load` - loading with `text` response
* `loadXml` - loading with `xml` response

now we have:
* `load` - by default text, but also supports `xml` and `arraybuffer`
* `loadXml` - still supporting only `xml`

Maybe, we should add explicit `loadBinaryData/loadData`? It probably doesn't really make sense deprecating `loadXml` as explicitness of this method is actually nice. Most readable seems to have `load` (without default type), `loadText`, `loadXml`, `loadBinaryData` but it's a bit too late for that.

jacekbogdanski:
Could you also rebase PR onto the latest major? Seems like there are some conficts.

plugins/pastetools/filter/image.js
```diff
@@ -300,13 +390,50 @@
...
+			data = CKEDITOR.tools.convertHexStringToBytes( img.hex );
+		}
+
+		return img.type ? 'data:' + img.type + ';base64,' + CKEDITOR.tools.convertBytesToBase64( data ) : null;
+	}
+
+	function convertBlobUrlToBase64( blobUrlSrc ) {
+		return new CKEDITOR.tools.promise( function( resolve ) {
+			CKEDITOR.ajax.load( blobUrlSrc, function( arrayBuffer ) {
+				var data = new Uint8Array( arrayBuffer ),
```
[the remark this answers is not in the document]
Comandeer: Yes, in old IEs this code is just ignored.

Comandeer:
Ok, I've addressed issues found during the review.

> Maybe, we should add explicit `loadBinaryData/loadData`? It probably doesn't really make sense deprecating `loadXml` as explicitness of this method is actually nice. Most readable seems to have `load` (without default type), `loadText`, `loadXml`, `loadBinaryData` but it's a bit too late for that.

TBH I don't find it's too late. We can always have additional methods and `load` with a default response type.

I'll raise an issue for it.

jacekbogdanski (approved):
Good job!

Issue #4333, opened 2020-10-16
Color history incorrectly displays colors entered as name or in RGB format


Dumluregn:
## Type of report

Bug

## Provide detailed reproduction steps (if any)

1. Go to [CDN](http://cdn.ckeditor.com/4.15.0/full-all/samples/).
2. Select some text.
3. Open color dialog.
4. Type `red` in the input and click `Ok`.
5. Open color panel and examine the first color in color history.

### Expected result

There is a red color box.

### Actual result

Color box is transparent - it's color is `ED`.

## Other details

The same problem occurs if you type a color in the `RGB` format. The reason is color history assumes the first character in color name will be a `#` and removes it - no questions asked.

* Browser: each
* OS: macOS
* CKEditor version: starting with `4.15.0` and color history feature (#1795)
* Installed CKEditor plugins: `colordialog`

sculpt0r:
I'm not sure if we should just take care of `HEX` and `RGB` format?

Maybe we should validate input for common standards (`HEX`, `RGB`, `HSV`, others?).
Then, in the case of 'plain text', we could try to define allowed color names (red, green, etc), and look into it?

If there is no match - there is no color change after `OK` button click or just set it to `#000` or `#FFF`?

f1ames:
> I'm not sure if we should just take care of `HEX` and `RGB` format?
> 
> Maybe we should validate input for common standards (`HEX`, `RGB`, `HSV`, others?).  
> Then, in the case of 'plain text', we could try to define allowed color names (red, green, etc), and look into it?
> 
> If there is no match - there is no color change after `OK` button click or just set it to `#000` or `#FFF`?

Well, there are at least two cases here - one related to the mechanism which assumes that all colors inserted in color dialog are in HEX format (thus starting with `#`). This assumption results in first character from color code (no matter the format and character itself) being removed. AFAIU it means it happens for any color format and the first character is simply removed. And this is what this issue is about. This means we should try to handle only this case here.

The other case is related to `RGBA`/`HSLA` handling and reported in #4351 (not sure if you had this one in mind), but it should be handled as separate fix.

> Maybe we should validate input for common standards (`HEX`, `RGB`, `HSV`, others?).

I guess you may be looking into `major` branch and not `master`. We have recently added validation but the latest code is on `master` branch for now only. @sculpt0r if you are working on a bugfix, you should always branch and target to `master`, while only new features should branch and target `major` branch.

sculpt0r:
[...]

PR #4366, opened 2020-11-09
Implemented color normalizer

### LINKED ISSUE #4358: Implement Color normalizer
## Type of report

Feature request

*What is the expected behavior of the proposed feature?*

Create separate class `Color` (aka color normalizer). Constructor takes `string` as an argument -> color code.
Normalizer should:
- verify `colorCode` for different formats `HEX`, `RGB`, plain string (eg: `red`). Converts to default color (eg. #000) in case of invalid format or non-existing plain-text color
- provide interface to get any format from input data

Benefits:
- use `Color` object as argument between  functions, instead of `magic string` with always unknown content
- whenever need color string (for CSS etc) - we use explicit format like: `color.toHex()` or `color.toRGBA()`
- conversions should be quite comfortable for unit tests

Color normalizer should be provided with `core/tools` or separate class.

For more context of this feature, please look at comments at #4333

sculpt0r:

New feature: Implemented color normalizer. Refactor color operations in `CKEDITOR.tools`

### This PR contains

- [X] Unit tests
- [ ] Manual tests

```
* [#4358](https://github.com/ckeditor/ckeditor4/issues/4358): Implemented color normalizer.
```

## What changes did you make?

Added new class `Color` in `core` in `CKEDITOR.tools.style` namespace.

Added Color as a dependency for `tools`. Also `Color` depends on `tools`.

Move color operations, RegExps, and `Names Colors` into `Colors`. `tools` preserved old methods (for backward compatibility). Now old methods and properties are using `Color` object inside.

`Color` constructor has `compatibility` flag as second parameter. It can be skipped, so `Color` class should behave as expected - so only valid color codes can be received (with **default color: `#000000`**). If compatibility flag is passed, then if there is no match to `Named Colors`, `HEX`, `RGB` or `HSL` input string (formerly first constructor argument) is returned.

Due to circular dependency (`Color` needs `CKEDITOR.tools.createClass` and `tools` needs `Color` for backward compatibility):
- `Color` class wrap class creation inside function. Until first `Color` object is created - there is no need for `tools`
- `Color` and `tools` are checking if a namespace is already declared.
- `tools` namespace has now a more extensible approach, instead of 'declaration'. Before, `Color` was declared in `CKEDITOR.tools.style` namespace and after `tools` loaded - namespace was overwritten. A new namespace declaration was done for almost every object in `tools` so it could be a nice step to separate other classes from this swollen file.

Closes #4358


f1ames:
I have rebased and retargeted this PR on latest `major`. Strangely there were some commits which shouldn't be there at all so looks like some rebase issue :thinking: 



But fixed this also :+1:

sculpt0r:
Thank you for rebasing. 

I have some issues with solving `tools.js` diffs. 

I may be wrong - but after [fix commit](https://github.com/ckeditor/ckeditor4/pull/4366/commits/bbc6ca5e6b5bf2d2438cf25ce2d926879c19935e) there are two `zip` functions. 

One in the proper namespace `CKEDITOR.tools.array`: https://github.com/ckeditor/ckeditor4/blob/bbc6ca5e6b5bf2d2438cf25ce2d926879c19935e/core/tools.js#L2096

The second one is without any namespace: https://github.com/ckeditor/ckeditor4/blob/bbc6ca5e6b5bf2d2438cf25ce2d926879c19935e/core/tools.js#L2240

The only diff is in `@returns` section. `A two-dimensional array of object pairs.` - Letter `A` added for correctness.

core/color.js
```diff
@@ -0,0 +1,467 @@
...
+* @fileoverview Defines the {@link CKEDITOR.tools.style.Color} normalizer class
+* 		that parse color string code other formats.
+*/
+
+/**
+*Basic definitions
+*HEX 	 -> valid HEX Color Value with # value: #FFF or #FFFFFF
+*3-HEX -> HEX but with only 3 characters value: #FFF
+*6-HEX -> HEX but with exactly 6 characters value: #FFFFFF
+*/
```
f1ames, line 15: This comment shouldn't float around here like that. It could be moved to class definition/description or to relevant method docs.

f1ames: And the formatting is incorrect, it should be:

```
/**
 * Foo
 * Bar
 */
```

core/color.js
```diff
@@ -0,0 +1,467 @@
...
+
+					return this._.clampValueInRange( value, 0, 1 );
+				},
+				hslToRgb: function( hue, sat, light ) {
+
+					hue = this._.clampValueInRange( hue, 0, 360 );
+					sat = this._.normalizePercentValue( sat );
+					light = this._.normalizePercentValue( light );
+
+					//based on https://www.w3schools.com/lib/w3color.js
```
f1ames, line 152: I would say W3C is not really reliable source. Keep also in mind that borrowing code parts from other projects requires proper license and it's compatibility with CKEditor 4 license so we rarely do this. It's good to base on official specifications like w3.org or whatwg.org if it has necessary information or just common sources like in case of alpha blending algorithm below.

core/color.js
```diff
@@ -0,0 +1,467 @@
...
+					this._.hexColorCode = hexStringFromNamedColor || hexFromHexString || hexFromRgbOrHsl || this._.defaultHexColorCode;
+				},
+				rgbOrHslToHex: function( colorCode ) {
+					var colorFormat = colorCode.trim().slice( 0,3 ).toLowerCase();
+
+					if ( colorFormat !== 'rgb' && colorFormat !== 'hsl' ) {
+						return null;
+					}
+
+					//take values
```
f1ames, line 116: It's good to keep comments as full sentences. Also they should bring additional value over the code and be easy to understand and complement commented code. Which means they should be very explicit to help understand the code and not to clutter up the code.

tests/core/tools/color.js
```diff
@@ -0,0 +1,113 @@
...
+			};
+		},
+
+		'test color object creation': function() {
+			var colorObject = this.createColor( '' );
+
+			assert.isObject( colorObject );
+		},
+
+		'test color from invalid string color name return default color': function() {
```
f1ames, line 18:
```suggestion
		'test color from invalid string color name returns default color': function() {
```

Please remember about proper form :wink:

tests/core/tools/color.js
```diff
@@ -0,0 +1,113 @@
...
+			var validRgbString = 'hsl( 195, 100%, 50% )';//0-360 , 0-1, 0-1
+			var expectedHexCode  = '#00BFFF';
+			var colorObj = this.createColor( validRgbString );
+
+			var resultHex = colorObj.getHex();
+
+			assert.areSame( expectedHexCode, resultHex );
+
+		}
+	} );
```
f1ames, line 111: Aren't tests for HSLA, RGBA and 8-digit HEX missing? There is `blendAlphaColor` method in `Color` class so I would assume it is covered in the code (and should be covered in tests)? :thinking:

f1ames (changes requested):
## The code

I had to rethink the entire approach especially that we have an issue with circular deps here. It is caused by the fact that `Color` class tries to use `createClass` function from `tools` and `tools` refers to some method/props from `Color`. From what I see `createClass` doesn't give as many benefits (almost none I suppose) and the complications caused by using it are quite significant... So I would get rid of creating `Color` class this way and go with another standard way we use - see for example https://github.com/ckeditor/ckeditor4/blob/c01d652320e9220c3411a161ea9447130e83bbcd/core/dom/element.js#L11-L34

This would mean we can get rid of all the magic from https://github.com/ckeditor/ckeditor4/pull/4366/files#diff-82fb93ef0501268a8be9290add301a266e430ea74b7f96dd0d437b10e6183de9R444 and simplify `Color` class.

Also I'm not sure about changes in `tools.js`. Assuming we would go with the above proposal (so getting rid of circular dependency), tools itself would not require flattening entire `CKEDITOR.tools` object and could exists in the current form. It would only need to refer to some of `Color` props/methods (liked `namedColors`) but since there will be direct, one-way dependency it will no longer cause any issues.

Other things:

1. I would be for moving `colors.js` to `core/tools/color.js`.
1. Something strange is going on since I see `colornormalizer.js` and `color.js` which has the same (partially) code :thinking: Some rebase issue probably? I assume the `color.js` one is correct so will add my comments there. And as you also pointed out in https://github.com/ckeditor/ckeditor4/pull/4366#issuecomment-724565849 that `zip` function got duplicated during my rebase... :thinking: Could I ask you to do a clean up here?
1. Please see also review comments.

I hadn't dive deeper into the code since it will be good to first clean up a little and then we can polish the whole thing.

f1ames:
I feel like some clarifications are needed here :thinking::bulb:

The current approach assumes that `color util` are part of the `tools` but should be moved to a separate file to not bulk the `tools.js` file further and be easier to maintain. This implicates close coupling of those two and since we would like to move some parts it has kind of dependency issue. It could be solved in a way that `color` depends on `tools` (and so `tools` are loaded first) and then `color` class/util when loaded injects itself into tools namespace.

I had some doubts whether `color` should really depend on `tools` but since it extends it, it makes sense:bulb: I will rephrase some of the review comments which are not in line with the above to be consistent with this comment.

core/loader.js
```diff
@@ -83,12 +84,13 @@ if ( !CKEDITOR.loader ) {
...
 			'template': [],
-			'tools': [ 'env' ],
+			'tools': [ 'env', 'color' ],
 			'ui': [],
 			'creators/themedui': [],
 			'creators/inline': [],
 			'promise': [ 'tools' ],
-			'selection/optimization': [ 'selection' ]
+			'selection/optimization': [ 'selection' ],
+			'color': [ 'tools' ]
```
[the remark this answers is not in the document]
sculpt0r: - `tools` reversed to original from `major` branch & only `@deprecated` marks added.
- circular dependency removed: `tools` no longer depends on `color`.
- finally `color` class no longer use `tools.createClass`. I'm not sure about this construction right now. It's no longer problem to use `tools.createClass`, but I removed it anyway.

core/color.js
```diff
@@ -0,0 +1,467 @@
...
+	var colorClassDefinition = {
+			$: function( colorCode, compatibilityMode ) {
+				this._.compatibilityMode = compatibilityMode || false;
+				this._.originalColorCode = colorCode;
+
+				if ( this._.compatibilityMode ) {
+					this._.legacyParsing( colorCode );
+				} else {
+					this._.parseInput( colorCode );
+				}
```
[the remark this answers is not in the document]
sculpt0r: `CompatibilityMode` removed. I left only desired functionalities in `color`. Methods and properties marked as `@deprecated`.

core/color.js
```diff
@@ -0,0 +1,467 @@
+/**
+* @license Copyright (c) 2003-2020, CKSource - Frederico Knabben. All rights reserved.
+* For licensing, see LICENSE.md or https://ckeditor.com/legal/ckeditor-oss-license
+*/
+/**
+* @fileoverview Defines the {@link CKEDITOR.tools.style.Color} normalizer class
+* 		that parse color string code other formats.
```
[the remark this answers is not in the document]
sculpt0r: Fixed comment indent.

core/color.js
sculpt0r: Fixed comment formating. Finally, I moved it to the test file, because they are used in tests cases names.

core/color.js
sculpt0r: I rewrite `HSL` to `RGB` and backward based on the wiki description. Proper links provided in methods.

There are some outstanding specifications (like character case in `hex` color or return value in a case of invalid input string). I assume that this way could be more useful in code than implementing plain specification which may lead to many `undefined` checks.

core/color.js
sculpt0r: This comment was removed.

tests/core/tools/color.js
sculpt0r: Correct in all test case names.

tests/core/tools/color.js
sculpt0r: Yes, they were missing. Cause I want to clarify overall implementation before full functionality cover. It appears there were some big changes at the end.

Right now alpha blending is covered in tests together with `hsla`, `rgba` and `8-digit hex`

sculpt0r:
Realize that there is no API-doc in my PR. But maybe it will be good if I wait till all code will be accepted?

core/color.js
```diff
@@ -0,0 +1,432 @@
...
+ */
+/**
+ * @fileoverview Defines the {@link CKEDITOR.tools.style.color} normalizer class
+ * that parse color string code other formats.
+ */
+
+( function() {
+	'use strict';
+
+	CKEDITOR.tools.style.color = function( colorCode ) {
```
f1ames, line 13: I thought I have mentioned it earlier but couldn't find it :thinking: I thing it shouldn't be in `style` namespace since it's more generic. So I would be fro moving it to `CKEDITOR.tools.color` namespace.

So all my previous comments mentioning `CKEDITOR.tools.style.color` means `CKEDITOR.tools.color` color namespace too :see\_no\_evil: :smile:

core/color.js
```diff
@@ -0,0 +1,432 @@
...
+
+			return hslPrefix + '(' +
+			hsl[0] + ',' +
+			hsl[1] + '%,' +
+			hsl[2] + '%' +
+			alphaString +
+			')';
+		},
+		// Color list based on https://www.w3.org/TR/css-color-4/#named-colors.
+		namedColors: {
```
f1ames, line 279: This should be a static property defined as:

```javascript
CKEDITOR.tools.color.namedColors = ...;
```

core/color.js
f1ames: And since we get rid of circular dependency we can get back to `createClass` (as it is also good practice mentioned in our code style guide - https://github.com/ckeditor/ckeditor4/blob/major/dev/docs/codestyle.md#2226-constructor-declaration).

Sorry for the confusion since I mentioned earlier to get rid of it :thinking: :speak\_no\_evil:

core/color.js
```diff
@@ -0,0 +1,432 @@
...
+		this.originalColorCode = colorCode;
+		this.parseInput( colorCode );
+	};
+
+	CKEDITOR.tools.style.color.prototype = {
+		originalColorCode: '',
+		hexColorCode: '',
+		defaultHexColorCode: '#000000',
+		alpha: 1,
+		hex3charsRegExp: /#([0-9a-f]{3})/gi,
```
f1ames, line 23: This should be static property I suppose.

core/color.js
```diff
@@ -0,0 +1,432 @@
...
+
+	CKEDITOR.tools.style.color = function( colorCode ) {
+		this.originalColorCode = colorCode;
+		this.parseInput( colorCode );
+	};
+
+	CKEDITOR.tools.style.color.prototype = {
+		originalColorCode: '',
+		hexColorCode: '',
+		defaultHexColorCode: '#000000',
```
f1ames, line 21: Not sure if we need this :thinking: What would be the reason/case for returning default value if invalid color is passed? And even though if we need something like thath somewhere in the code it shouldn't be color class responsibility I suppose :thinking:

f1ames (changes requested):
Ok, starting to look good :+1: 

Apart from review comments, I would also suggest cleaning-up `color.js` a little:

> Realize that there is no API-doc in my PR. But maybe it will be good if I wait till all code will be accepted?

I think it will be good opportunity to clean up `color.js` class and make it more readable (and also easier to review :wink: ) so I think it's good idea to work on docs now. So:

* Please add API docs to `color.js` file. Please remember about descriptive docs and marking private methods with `@private` etc.
* Also I would suggest slight code restructuring - getters should go firts then setters and then all the related helpers.
* If you think any method should not be accessible in any way through API you can move it out from class scope as a local function (e.g. `clamp` function, percentage normalization, etc).

tests/core/tools/color.js
```diff
@@ -0,0 +1,264 @@
...
+ * HEX 	 -> valid HEX Color Value with # value: #FFF or #FFFFFF
+ * 3-HEX -> HEX but with only 3 characters value: #FFF
+ * 6-HEX -> HEX but with exactly 6 characters value: #FFFFFF
+ * 8-HEX -> HEX but with exactly 8 characters value: #FFFFFF00. Last two characters are for alpha
+ */
+( function() {
+	'use strict';
+	bender.test( {
+		setUp: function() {
+			this.createColor = function( colorCode ) {
```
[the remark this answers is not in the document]
sculpt0r: Honestly, I made `createColor` only for tests just to avoid typing `CKEDITOR.tools.color` everywhere. In case something change in `color` class, or namespace - there is only one place where changes need to be done.

Static creator vs `new`...
- there are approaches that you should never create any class with `new` keyword. Besides factory methods, which are implemented inside a class. So they are allowed to know the complex algorithm of creating this object. So whenever big preparations are needed before object creation, we would rather encapsulate this logic inside a class (with the static creator).
This helps avoid huge code changes in case of object construction change.  But I think it's more useful when you need to create many dependencies for object creation.
For such a plain object like ours `color`, one thing that could change is the parameters list - so it would require modifying the static creator function declaration anyway. No befits for me here.
- there was one moment when I think about it. We could have base color class and static factory method. Based on input it could return a specific color object. So we would have `ColorRgb`, `ColorHsl`, `ColorHex` - all of them extend `color`. Each of these classes would contain a parser just for its type. However, there would be many more combinations with conversion between colors objects. 
Probably there should be another class: `converter` which would take different color objects and then convert them? So anyway we would need to extract some basic logic from each class for such conversion. I think this approach does not suit me. Where parsers in separate classes would make more clean code, then `converter` would make it messier.

core/color.js
sculpt0r: Originally you said it should be in `CKEDITOR.tools.color`. It was on slack chat I think. Then `style` namespace was suggested by @jacekbogdanski in another private chat.

Anyway - I move it to `CKEDITOR.tools.color`

core/color.js
sculpt0r: I should write it before... There are my drivers:
- whenever we **expect** User input any color string - we could almost blindly(maybe after sanitization) pass it to `color` class.
- whenever we need pass color to any method - we pass `color` object. So there will be no confusion about what kind of `string` this may be. Should we sanitize it? Maybe validate it? Instead of checking a long chain of calls before our method (to find out what is the origin of this color) - we can just use `color` class interface.
- whenever we need in end-cases pass specific string (like style property value for `background-color`) - then we use proper formatter for that, like `getHex()`. It's very clear in that case what kind of string we expect in that case on output.

Default color:
- Based on previous assumptions - if `color` class **always** provide **valid** color code - there is no place for mistakes for programmers. It always gives us a valid color, so there will be no `background-color: sthWeird`.
- The only source of untrusty (as always) is User input. Or code where we create an object "from hand" - but for that case we have tests - right?
- It may be not the expected color when User misspells `rde` instead of `red` but it's always **valid**. Like in color history case (#4333). No matter of user input we display something. But for now it's something invalid. Maybe we should check input in color palette first - then if color is invalid, just drop it - that is right. But again - in case we miss such validation on any level - there is always **valid** color code at the end.

From the other side, I can imagine that instead of default color we could throw exception during `color` creation, and other parts of code will be responsible to create a valid `color` object in that case.

So from my point of view, I try to make it more like static typed. Even - to get rid of checking if a variable is a valid object or some null. There is in "Code refactor" by Martin Fowler idea about 'default` object instead of null returning. Which makes it another driver for me.

tests/core/tools/color.js
```diff
@@ -0,0 +1,264 @@
+/* bender-tags: editor */
+
+
+/**
+ * Basic definitions
+ * HEX 	 -> valid HEX Color Value with # value: #FFF or #FFFFFF
+ * 3-HEX -> HEX but with only 3 characters value: #FFF
+ * 6-HEX -> HEX but with exactly 6 characters value: #FFFFFF
+ * 8-HEX -> HEX but with exactly 8 characters value: #FFFFFF00. Last two characters are for alpha
+ */
```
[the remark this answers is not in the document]
sculpt0r: I try to move in a slightly different direction.
Most important are input and output digits in hexadecimal color code.
So I remove `HEX` from test cases names. In each test case, we know exactly how many digits are in input and output. 

There is no need to define so generic valid `HEX` term.
It was - before API was defined to return only `6-HEX` or `8-HEX`.

Hope it will be good.

sculpt0r:
I'm not sure about https://github.com/ckeditor/ckeditor4/blob/88adc86c9fe3622ab65bafa24d173139a85b9811/core/tools.js#L1078-L1084

Marked it as deprecated before, but now I'm not sure if I should move it to `CKEDITOR.tools.color`?

sculpt0r:
I made all the requested changes.
Please look at https://github.com/ckeditor/ckeditor4/pull/4366#discussion_r529240426 which explain why I decided to make very strict access level for most methods in `color`.

f1ames (commented):
As for the main part of the code (`color.js`) we agreed with @sculpt0r for a pair refactoring session to polish the solution together. Apart form that I'm leaving few comments related to docs and tests (which we will cover too).

## Documentation

There are some invalid API docs properties:

image

## Tests

See review comments.

tests/core/tools/_helpers/colorTools.js
```diff
@@ -0,0 +1,17 @@
+/* exported colorTools */
+
+var colorTools = ( function() {
+	function testColor( inputColorCode, expectedColorCode, getterMethod ) {
```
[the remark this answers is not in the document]
sculpt0r: I decided to use `testColorConversion`

tests/core/tools/color.js
```diff
@@ -0,0 +1,74 @@
...
+			};
+		},
+
+		'test color object creation': function() {
+			var colorObject = this.createColor( '' );
+
+			assert.isObject( colorObject );
+		},
+
+		'test color from invalid string color name returns default color': colorTools.testColor( 'NotValidColorName', '#000000', 'getHex' ),
```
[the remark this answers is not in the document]
sculpt0r: Second test checks not valid input string with `NotValidColorName` input.

I'll add tests for non-string values and invalid HEX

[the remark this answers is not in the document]
sculpt0r: Because of `Array` test case - now I'm wondering if we should create color from array `[ red, green, blue, alpha ]`. But there will be confusion if values are for rgb or hls...

core/color.js
```diff
@@ -0,0 +1,723 @@
...
+							number < 0 ? 255 : number;
+					}
+					return number;
+				}
+				red = normalizer( red );
+				green = normalizer( green );
+				blue = normalizer( blue );
+
+				return [ red, green, blue ];
+			},
```
f1ames, line 313: This method seems to be slightly confusing multi-purpose method. On one hand it just normalizes a number, but for some reason it does it for 3 numbers at once. It is clear in the `color` class context why, but at the same time it seems to be breaking separation of concerns :thinking: Additionally, it handles both integers and percentages differently and the normalization range (255) is hardcoded - both doesn't really arise from method name or its description. So here you have:

1. Normalization of percentage values based on given range (0 - 255).
1. Normalization of numbers form outside 0 - 255 range (not sure why greater numbers results in 0 why negative ones in 255 :thinking:).
1. It is done for 3 numbers at once.

So I think this could be split into smaller simpler methods/functions.

f1ames (changes requested):
I took this occasion to polish the code a little. Extracting and operating on RGBA color channels simplifies code visibly, but still there were some unnecessary conversions between formats in `parseInput()` method. Since we need color channels only, it makes more sense to extract them without converting all formats to HEX first (still we will need to have color channels conversion from HSLA since it uses different representation). And so my proposal is to just introduce extraction methods - I already prepared one which handles HEX - see https://github.com/ckeditor/ckeditor4/pull/4366/commits/6f52e756819d4e7b44ea6fddaf2d9896cb2ed9f8. RGBA and HSLA still needs to be covered.

I tried to split refactoring into separate logical parts (commits) so you can see what and why specific changes were applied. And so:

1. https://github.com/ckeditor/ckeditor4/pull/4366/commits/6f52e756819d4e7b44ea6fddaf2d9896cb2ed9f8 - as already mentioned switching from unnecessary conversions to RGBA channels extraction.
1. https://github.com/ckeditor/ckeditor4/pull/4366/commits/3a510383f4f98a235f7e9c089cb78342c8f6031c - simplified HEX getters. You have introduced higher order function (method to be precise) `getColorValue()` only to handle simple logic in getters. This significantly increases code complexity and doesn't really bring any benefit (apart from few lines of code less).
1. https://github.com/ckeditor/ckeditor4/pull/4366/commits/cc133298b1dd879f18a659e65255e15db822585d - just paying attention to details in API docs and minor rewordings. Please also keep in mind that all types in API docs type annotations should start with uppercase like `{String}` or `{Function}`.
1. https://github.com/ckeditor/ckeditor4/pull/4366/commits/fec25d5fd16144104cd96cfb51c457f54410c590 - moved `blendAlphaColor()` method outside of class scope. I'm not sure, on the second thought it seems all color related functions/methods should be color class private members :thinking: 
1. https://github.com/ckeditor/ckeditor4/pull/4366/commits/f5e360ed47e319bd9293b0f35408c82c4553b5a5 - simplified HEX3 to HEX6 conversion. If you have a method like this, which does one thing only and the input is always the same (sting with fixed length here) there is no reason to complicate things. Also since it's private method, you can see based on it's usage that it will always get valid input.
1. https://github.com/ckeditor/ckeditor4/pull/4366/commits/40b0748e748a1e980295c61fe24237e43145728d - since we are going to extract color channels from initial formats, there is no need for those conversion methods.
1. https://github.com/ckeditor/ckeditor4/pull/4366/commits/79e630fbc9ca44e995503932cc3736f182671398 - again attention to details, fixed camelCase for static properties and added proper spacing between API comments there.
1. https://github.com/ckeditor/ckeditor4/pull/4366/commits/413997d709d5fe7fa2ad073a6647782ea5f97fd6 - at this point I build CKEditor 4 docs locally to check how it looks, there was one missing `@private` annotation.

There is also some stuff mentioned in previous comments not yet covered:

*   https://github.com/ckeditor/ckeditor4/pull/4366#discussion_r532532316
*   https://github.com/ckeditor/ckeditor4/pull/4366#discussion_r532532584
* > I would be for moving colors.js to core/tools/color.js. (see https://github.com/ckeditor/ckeditor4/pull/4366#pullrequestreview-526957825)

**As for now, unit tests related to RGBA and HSLA handling are failing. This is due missing extraction methods for those formats. I would like to ask you @sculpt0r to add these methods, cleanup unnecessary stuff and polish API docs if necessary.**

core/color.js
sculpt0r: As we spoke: we get rid of optimistic input value parsing.
So only valid numbers will be here (percent and proper ranged numbers). 

I'll adjust tests and clean up this method.

f1ames (changes requested):
## Code

### Private methods order

The order of methods definitions still seems to be a little off. How it is now:

1. parseInput
1. matchStringToNamedColor
1. extractColorChannelsFromHex
1. extractColorChannelsFromRgba
1. validateRgbValues
1. validateAlphaValue
1. validateRgbChannelValue
1. extractColorChannelsFromHsla
1. validateHslValues
1. isValidPercentOrNormalizedPercent
1. validateValueInRange
1. validatePercent
1. extractDigitsFromPattern
1. hex3ToHex6
1. rgbToHsl
1. hslToRgb

If you take a look which methods call which internally you will see that the order should be slightly different to make it easier to read:

```javascript
parseInput: function() {},
// calls matchStringToNamedColor, extractColorChannelsFromHex, extractColorChannelsFromRgba, extractColorChannelsFromHsla

matchStringToNamedColor: function() {},

extractColorChannelsFromHex: function() {},
// calls hex3ToHex6

extractColorChannelsFromRgba: function() {},
// calls extractDigitsFromPattern

validateRgbValues: function() {},
// calls validateRgbChannelValue, validateAlphaValue

validateAlphaValue: function() {},
// calls isValidPercentOrNormalizedPercent

validateRgbChannelValue: function() {},
// calls validatePercent, validateValueInRange

extractColorChannelsFromHsla: function() {},
// calls extractDigitsFromPattern, hslToRgb

validateHslValues: function() {},
// calls validateValueInRange, isValidPercentOrNormalizedPercent, validateAlphaValue

isValidPercentOrNormalizedPercent: function() {},
// calls validatePercent, validateValueInRange

validateValueInRange: function() {},

validatePercent: function() {},

extractDigitsFromPattern: function() {},

hex3ToHex6: function() {},

rgbToHsl: function() {},

hslToRgb: function() {},
```

so you see it should be something like:

1. parseInput
1. matchStringToNamedColor
1. extractColorChannelsFromHex
1. extractColorChannelsFromRgba
1. extractColorChannelsFromHsla
1. hex3ToHex6
1. extractDigitsFromPattern
1. hslToRgb
1. validators...

### Validators

Let's go with the second round of pair coding :wink: 

## Tests

I have added 3 tests which should pass but are failing alpha channel in resulting RGBA ranges from 0 to 255 while it should be between 0 and 1 AFAIU. And also alpha channel in HEX format instead of `FF` is converted to `01` (see https://github.com/ckeditor/ckeditor4/pull/4366/commits/3b06417c84b88a215cd8e07467b1486a119101cc). 

Apart from the above, some tests are failing on IEs.

* [x] Chrome
* [x] Firefox
* [x] Safari
* [x] Edge
* [x] IE11 (18/36 failing)
* [x] IE10 (18/36 failing)
* [x] IE9 (18/36 failing)
* [x] IE8 (4/36 failing)
* [x] Chrome@Android
* [x] Safari@macOS

### IE11 (same for IE9-IE10)

image

### IE8

image


## Documentation

Sorry for the confusion regarding `@deprecated` annotation support in our docs. Using `@deprecated 4.16.0` works fine and is rendered correctly, we are already using it:

image

However, any further text is ignored. So **`@deprecated` annotation should contain version number**. All further description, e.g. what should be used instead should be moved to API description (as it is already done :+1:).

**Ok, I had already fixed that when doing minor rewording in docs** (see https://github.com/ckeditor/ckeditor4/pull/4366/commits/7f4b1f7f20a22ef66ad2368c0d113ebb5fa331ab) :+1:

core/tools/color.js
```diff
@@ -0,0 +1,871 @@
...
+
+			/**
+			 * Extracts RGBA channels from given RGBA string.
+			 *
+			 * @private
+			 * @param {String} colorCode RGB or RGBA color representation.
+			 */
+			extractColorChannelsFromRgba: function( colorCode ) {
+				var rgbMatch =  this._.extractDigitsFromPattern( colorCode, CKEDITOR.tools.color.rgbRegExp, this._.validateRgbValues );
+				return rgbMatch;
```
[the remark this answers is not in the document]
f1ames: Already covered.

core/tools/color.js
```diff
@@ -0,0 +1,871 @@
...
+						hexToValue( parts[ 5 ] + parts [ 6 ] ),
+						parts[ 7 ] && parts[ 8 ] ? hexToValue( parts[ 7 ] + parts[ 8 ] ) : 1
+					];
+				}
+
+				return null;
+			},
+
+			/**
+			 * Extracts RGBA channels from given RGBA string.
```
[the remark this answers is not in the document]
f1ames: Done :+1:

core/tools/color.js
```diff
@@ -0,0 +1,871 @@
...
+				wheat: '#F5DEB3',
+				white: '#FFFFFF',
+				whitesmoke: '#F5F5F5',
+				yellow: '#FFFF00',
+				yellowgreen: '#9ACD32'
+			}
+		}
+	} );
+
+	CKEDITOR.tools.color.MAX_RGB_CHANNEL_VALUE = 255;
```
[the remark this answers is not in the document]
f1ames: We have few more consts like that now, all should be moved to `statics` above.

core/tools/color.js
```diff
@@ -0,0 +1,871 @@
...
+			 */
+			getHsl: function() {
+				var color = blendAlphaColor( this._.red, this._.green, this._.blue, this._.alpha );
+
+				if ( this._.isValidColor ) {
+					var hsl = this._.rgbToHsl( color[ 0 ], color[ 1 ], color[ 2 ] );
+					return formatHslString( 'hsl', hsl[ 0 ], hsl[ 1 ], hsl[ 2 ] );
+				} else {
+					return this._.defaultValue;
+				}
```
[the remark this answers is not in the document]
f1ames: Done :+1:

f1ames:
@sculpt0r I have pushed changes we concluded yesterday together and also did some further refactoring. What we are still missing is some final polishing so from https://github.com/ckeditor/ckeditor4/pull/4366#pullrequestreview-546830688:

*   Methods order.
*   Failing tests on IEs.

Apart from that:

*   https://github.com/ckeditor/ckeditor4/pull/4366#discussion_r540050818
*   Adding API docs to new methods (I skipped some and left `TODO` there).
*   Moving colors related methods (so `blendAlpha` and `format*String`) as private color class methods.

sculpt0r:
I made corrects according to https://github.com/ckeditor/ckeditor4/pull/4366#issuecomment-742431904

> ### Methods order.

- Mainly covered in https://github.com/ckeditor/ckeditor4/pull/4366/commits/33992fa328530e9acfc34e75f0b4f41b847cbf43
- Also in https://github.com/ckeditor/ckeditor4/pull/4366/commits/527297b302c171675beed38c79d2894c7c3fb4e8, where methods were moved into the class private scope

> ### Failing tests on IEs.

- Partially fixed for IE > 8 in https://github.com/ckeditor/ckeditor4/pull/4366/commits/380f32a226ef7ce1130ce949c6c2bd57abd198c3 and https://github.com/ckeditor/ckeditor4/pull/4366/commits/8253d2d8679dcd369c3b94c46ec64037348018b9
I used global `parseFloat` instead of `Number.parseFloat` as those methods are equivalent. Also used global `isNaN` instead of `Number.isNaN`.
- For IE8 fix is https://github.com/ckeditor/ckeditor4/pull/4366/commits/a71fe560328c2a6b14c016c43cd6a7a56ecff635. Use `trim` from `CKEDITOR.tools` instead of native which is not supported at IE8.
- I run tests for IE 8-11 after the entire refactor and they still pass.

> ### I have added 3 tests which should pass but are failing alpha channel in resulting RGBA ranges from 0 to 255 while it should be between 0 and 1 AFAIU. And also alpha channel in HEX format instead of FF is converted to 01 (see 3b06417).

- Failing test fixed in https://github.com/ckeditor/ckeditor4/pull/4366/commits/8fea2d99b6dacd01c3a3a56a643ca24da044ca9c which was finally refactored by @f1ames in https://github.com/ckeditor/ckeditor4/pull/4366/commits/8107b83fba6a15761b0d6556eb156e7510a62ad0 Alpha wasn't normalized from `HEX` representation parsing.

- Other case fixed in https://github.com/ckeditor/ckeditor4/pull/4366/commits/1f12a6eff0d81f0c086745dea02109304672d153
Alpha for `HEX` representation needs to be re-ranged into 0-255 and was 0-1.

> ### https://github.com/ckeditor/ckeditor4/pull/4366#discussion_r540050818

- Floating consts moved to class scope in https://github.com/ckeditor/ckeditor4/pull/4366/commits/aaa6a4d60a751be8db3136272ef42de8758436f0

> ### Adding API docs to new methods (I skipped some and left TODO there).
- New comments added and add small corrects in: 
- while moving const to class scope: https://github.com/ckeditor/ckeditor4/pull/4366/commits/aaa6a4d60a751be8db3136272ef42de8758436f0
- https://github.com/ckeditor/ckeditor4/pull/4366/commits/f072e1d75b443b0e273de642616b93ccd44423fc
- https://github.com/ckeditor/ckeditor4/pull/4366/commits/217892285d2fcf77f3b62bb4ae9d29f202643e84
- https://github.com/ckeditor/ckeditor4/pull/4366/commits/d4dcb014ecf07295eb084fdaf0eb57dc770d906d
- https://github.com/ckeditor/ckeditor4/pull/4366/commits/216dc936accab69d6e816b5cb5997c8e70f9b642

> ### Moving colors related methods (so blendAlpha and format*String) as private color class methods.
- Methods moved in https://github.com/ckeditor/ckeditor4/pull/4366/commits/527297b302c171675beed38c79d2894c7c3fb4e8

Also, I thought that range inclusiveness (while checking if a variable is within range) is more obvious in the new form https://github.com/ckeditor/ckeditor4/pull/4366/commits/d17b51b2ca52e11c7eca1cd6bf52d2c422997a71

Found missed cases with invalid arguments number according to color code prefix (rgb & hsl). So I add two failing tests for it https://github.com/ckeditor/ckeditor4/pull/4366/commits/943dd7387cf52fbaf223989d8518cb18d5c9269b and then fixed it in https://github.com/ckeditor/ckeditor4/pull/4366/commits/e2b23c17271e4ee9052239c0f4e5ba7d11766860

There was a mismatch in test naming: sometimes `rgb` or `hsl` was lowercased and sometimes uppercased. So I made everything uppercase - cause it seems to be more readable in those long tests names: https://github.com/ckeditor/ckeditor4/pull/4366/commits/f46b988b349595b166a2e03145202a8a2aef9a50

Finally, I rebased on the newest `major`. As before - I had a problem with `core/tools.js`, but finally, only one fix was needed: https://github.com/ckeditor/ckeditor4/pull/4366/commits/59b65e3733f65a92fc41be392093a7a902cd335c

Now it's ready for another review.

f1ames (approved):
LGTM :+1: :tada: :cake: 

## Code

:+1: 

## Docs

Did some minor polishing only :+1:

As for `CKEDITOR.tools.convertRgbToHex()` function it cannot be entirely replaced by `color` class since it operates on styles string, for example:

```
background-color: rgb(255,255,255); text-align: center;
```

so it does a little more. I have removed deprecated annotation and rephrase `Note` there.

## Tests

* [x] IE8 - IE11
* [x] Chrome
* [x] Firefox
* [x] Safari

---

Apart from the above, this PR does not introduce any feature but extends public API. In general, such changes should be targeted to minor release (`master` branch). However, since we deprecate some API here too it means `major` release (`major` branch) :+1:

PR #4370, opened 2020-11-10
Fix: Editor Placeholder plugin throws an error during editor initialization when used with bbcode and fullpage

### LINKED ISSUE #4253: Editor Placeholder plugin throws an error during editor initialization when used with bbcode and fullpage
## Type of report

Bug

See https://github.com/ckeditor/ckeditor4/pull/4251#pullrequestreview-481525255.

1.  Open any empty editor instance with `bbcode` plugin and `fullPage: true` configuration set.

### Expected result

Editor initializes and works correctly.

### Actual result

Error is thrown (same as in #4249).

## Other details

Can be tested on https://codepen.io/f1ames/pen/oNLQvNe.

*   Browser: -
*   OS: -
*   CKEditor version: `4.15.0`
*   Installed CKEditor plugins: `editorplaceholder,bbcode`

sculpt0r:
Bug fix

### This PR contains

- [X] Unit tests
- [ ] Manual tests

```
* [#4253](https://github.com/ckeditor/ckeditor4/issues/4253): Prevent index-reading from undefined variable.
```

## What changes did you make?

Check `RegExp` match results before trying access via index.

Closes #4253

### DIFF
--- a/plugins/editorplaceholder/plugin.js
+++ b/plugins/editorplaceholder/plugin.js
@@ -74,9 +74,9 @@
 		if ( isFullPage ) {
 			var bodyDataMatched = data.match( fullPageRegex );
 
+			// Check if body element exists in editor HTML (#4253).
 			if ( bodyDataMatched && bodyDataMatched.length > 1 ) {
-				//found data inside `<body>` element
-				data = bodyDataMatched[1];
+				data = bodyDataMatched[ 1 ];
 			}
 		}
 
--- a/plugins/editorplaceholder/plugin.js
+++ b/plugins/editorplaceholder/plugin.js
@@ -72,7 +72,12 @@
 			data = editor.getData();
 
 		if ( isFullPage ) {
-			data = data.match( fullPageRegex )[ 1 ];
+			var bodyDataMatched = data.match( fullPageRegex );
+
+			if ( bodyDataMatched && bodyDataMatched.length > 1 ) {
+				//found data inside `<body>` element
+				data = bodyDataMatched[1];
+			}
 		}
 
 		return data.length === 0;
--- a/tests/plugins/editorplaceholder/editorplaceholder.js
+++ b/tests/plugins/editorplaceholder/editorplaceholder.js
@@ -220,8 +220,7 @@
 	};
 
 	// (#4253)
-	tests[ 'test placeholder loads correctly in full-page editor with bbcode plugin'] = function() {
-
+	tests[ 'test placeholder loads correctly in full-page editor with bbcode plugin' ] = function() {
 		consoleErrorStub = sinon.stub( console, 'error' );
 
 		var runtimeErrorListener = function() {
@@ -232,7 +231,6 @@
 
 		window.addEventListener( 'error', runtimeErrorListener );
 
-
 		bender.editorBot.create( {
 			name: 'bbcodeFailTest',
 			config: {
@@ -241,18 +239,80 @@
 				fullPage: true,
 				on: {
 					instanceReady: function() {
-									resume( function() {
-										assert.isFalse( consoleErrorStub.called, 'There should be no errors during initialization' );
-									} );
-								}
+						resume( function() {
+							assert.isFalse( consoleErrorStub.called, 'There should be no errors during initialization' );
+						} );
+					}
 				}
 			}
 		}, function( bot ) {
 
 		} );
 
-
 		wait();
+	};
+
+	// (#4253)
+	tests[ 'test placeholder loads correctly in full-page editor with bbcode plugin and initial content' ] = function() {
+		bender.editorBot.create( {
+			name: 'bbcodeFailTestContent',
+			config: {
+				extraPlugins: 'bbcode',
+				editorplaceholder: 'any',
+				fullPage: true,
+				startupData: '<p>Initialized content</p>'
+			}
+		}, function( bot ) {
+			var editor = bot.editor;
+
+			assert.isFalse( editor.editable().hasAttribute( 'data-cke-editorplaceholder' ) );
+		} );
+	};
+
+	// (#4253)
+	tests[ 'test placeholder is visible with data filter to no content' ] = function() {
+		bender.editorBot.create( {
+			name: 'placeholderFilterToNoContent',
+			config: {
+				editorplaceholder: 'any',
+				fullPage: true,
+				startupData: '<p>Initialized content</p>'
+			}
+		}, function( bot ) {
+			var editor = bot.editor;
+
+			editor.on( 'toDataFormat', function ( evt ) {
+				evt.data.dataValue = '';
+			}, null, null, 16 );
+
+			editor.fire( 'change' );
+
+			assert.isTrue( editor.editable().hasAttribute( 'data-cke-editorplaceholder' ) );
+		} );
+	};
+
+	// (#4253)
+	tests[ 'test placeholder not visible with data filter to preinitialized content' ] = function() {
+		var preinitializedContent = '<p>Initialized content</p>';
+
+		bender.editorBot.create( {
+			name: 'placeholderFilterPreinitialized',
+			config: {
+				editorplaceholder: 'any',
+				fullPage: true,
+				startupData:preinitializedContent
+			}
+		}, function( bot ) {
+			var editor = bot.editor;
+
+			editor.on( 'toDataFormat', function ( evt ) {
+				evt.data.dataValue = preinitializedContent;
+			}, null, null, 16 );
+
+			editor.fire( 'change' );
+
+			assert.isFalse( editor.editable().hasAttribute( 'data-cke-editorplaceholder' ) );
+		} );
 	};
 
 	bender.test( tests );
--- a/tests/plugins/editorplaceholder/editorplaceholder.js
+++ b/tests/plugins/editorplaceholder/editorplaceholder.js
@@ -26,9 +26,23 @@
 		}
 	};
 
+	var consoleErrorStub;
+	var runtimeErrorListener;
+
 	var tests = {
+
 		setUp: function() {
 			bender.tools.ignoreUnsupportedEnvironment( 'editorplaceholder' );
+		},
+
+		tearDown: function() {
+			if ( consoleErrorStub ) {
+				consoleErrorStub.restore();
+			}
+
+			if ( runtimeErrorListener ) {
+				window.removeEventListener( 'error', runtimeErrorListener );
+			}
 		},
 
 		'test getting data from editor': function( editor ) {
@@ -208,21 +222,16 @@
 	// (#4253)
 	tests[ 'test placeholder loads correctly in full-page editor with bbcode plugin'] = function() {
 
-		var errorLogStub = sinon.stub( console, 'error' );
+		consoleErrorStub = sinon.stub( console, 'error' );
 
-		var onErrorStub = function() {
+		var runtimeErrorListener = function() {
 			resume( function() {
 				assert.fail( 'There should be no RunTime errors!' );
 			} );
 		};
 
-		window.addEventListener( 'error', onErrorStub );
+		window.addEventListener( 'error', runtimeErrorListener );
 
-		var instanceReadyCallback = function( e ) {
-			resume( function() {
-				assert.isFalse( errorLogStub.called, 'There should be no errors during initialization' );
-			} );
-		};
 
 		bender.editorBot.create( {
 			name: 'bbcodeFailTest',
@@ -231,15 +240,17 @@
 				editorplaceholder: 'any',
 				fullPage: true,
 				on: {
-					instanceReady: instanceReadyCallback
+					instanceReady: function() {
+									resume( function() {
+										assert.isFalse( consoleErrorStub.called, 'There should be no errors during initialization' );
+									} );
+								}
 				}
 			}
 		}, function( bot ) {
 
 		} );
 
-		errorLogStub.restore();
-		window.removeEventListener( 'error', onErrorStub );
 
 		wait();
 	};
--- a/tests/plugins/editorplaceholder/editorplaceholder.js
+++ b/tests/plugins/editorplaceholder/editorplaceholder.js
@@ -26,22 +26,36 @@
 		}
 	};
 
-	var consoleErrorStub;
-	var runtimeErrorListener;
+	function addErrorAsserts( benderScope ) {
+		benderScope.consoleErrorStub = sinon.stub( console, 'error' );
+
+		benderScope.runtimeErrorListener = function() {
+			resume( function() {
+				assert.fail( 'There should be no RunTime errors!' );
+			} );
+		};
+
+		window.addEventListener( 'error', benderScope.runtimeErrorListener );
+	}
 
 	var tests = {
 
 		setUp: function() {
 			bender.tools.ignoreUnsupportedEnvironment( 'editorplaceholder' );
+
+			this.consoleErrorStub = null;
+			this.runtimeErrorListener = null;
 		},
 
 		tearDown: function() {
-			if ( consoleErrorStub ) {
-				consoleErrorStub.restore();
+			if ( this.consoleErrorStub ) {
+				console.log('cleared errStub');
+				this.consoleErrorStub.restore();
 			}
 
-			if ( runtimeErrorListener ) {
-				window.removeEventListener( 'error', runtimeErrorListener );
+			if ( this.runtimeErrorListener ) {
+				console.log('cleared window');
+				window.removeEventListener( 'error', this.runtimeErrorListener );
 			}
 		},
 
@@ -221,15 +235,7 @@
 
 	// (#4253)
 	tests[ 'test placeholder loads correctly in full-page editor with bbcode plugin' ] = function() {
-		consoleErrorStub = sinon.stub( console, 'error' );
-
-		var runtimeErrorListener = function() {
-			resume( function() {
-				assert.fail( 'There should be no RunTime errors!' );
-			} );
-		};
-
-		window.addEventListener( 'error', runtimeErrorListener );
+		addErrorAsserts( this );
 
 		bender.editorBot.create( {
 			name: 'bbcodeFailTest',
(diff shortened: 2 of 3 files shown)


plugins/editorplaceholder/plugin.js
```diff
@@ -72,7 +72,12 @@
 			data = editor.getData();
 
 		if ( isFullPage ) {
-			data = data.match( fullPageRegex )[ 1 ];
+			var bodyDataMatched = data.match( fullPageRegex );
+
+			if ( bodyDataMatched && bodyDataMatched.length > 1 ) {
+				//found data inside `<body>` element
+				data = bodyDataMatched[1];
```
f1ames, line 79:
```suggestion
				data = bodyDataMatched[ 1 ];
```

plugins/editorplaceholder/plugin.js
```diff
@@ -72,7 +72,12 @@
 			data = editor.getData();
 
 		if ( isFullPage ) {
-			data = data.match( fullPageRegex )[ 1 ];
+			var bodyDataMatched = data.match( fullPageRegex );
+
+			if ( bodyDataMatched && bodyDataMatched.length > 1 ) {
+				//found data inside `<body>` element
```
f1ames, line 78: I think this comment is not necessary here, but you may add reference to issue number above `if` like:

```javascript
// Check if body element exists in editor HTML (#4253).
if (...
```

tests/plugins/editorplaceholder/editorplaceholder.js
```diff
@@ -205,5 +205,20 @@
 		} );
 	};
 
+	// (#4253)
+	tests[ 'test placeholder loads correctly in full-page editor with bbcode plugin'] = function() {
```
f1ames, line 209: Well, this test fails without the fix, which is good, but not in a way it should:



The correct way would be to `spy/stub` console error with `sinon` to see if no errors where thrown - if error is thrown use `assert.fail( ... )` to fail the test and if editor fires [`instanceReady` event](https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_editor.html#event-instanceReady) test should be marked as passing with `assert.pass( ... )`.

tests/plugins/editorplaceholder/editorplaceholder.js
```diff
@@ -205,5 +205,20 @@
...
+				extraPlugins: 'bbcode',
+				editorplaceholder: 'any',
+				fullPage: true
+			}
+		}, function( bot ) {
+			//no errors during editor load - so test can pass
+			assert.isTrue( true );
+		} );
+	};
+
```
f1ames, line 222: I'm also missing second test covering case mentioned in https://github.com/ckeditor/ckeditor4/issues/4253#issuecomment-689774158. Please check [`toDataFormat` event](https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_editor.html#event-toDataFormat) to see how to manipulate editor data on output.

f1ames (changes requested):
Looks good, some polishing in tests needed :+1:

tests/plugins/editorplaceholder/editorplaceholder.js
sculpt0r: I add a test against content without body, but it appears `bbcode` plugin filter-out all non-body content.

So it looks like `bbcode` plugin with `config.fullPage: true;` has no sense.

plugins/editorplaceholder/plugin.js
sculpt0r: Fixed with: https://github.com/ckeditor/ckeditor4/pull/4370/commits/b3f52357deea82a0a3d8fe1c7af125175a39131a

plugins/editorplaceholder/plugin.js
sculpt0r: Fixed with: https://github.com/ckeditor/ckeditor4/pull/4370/commits/b3f52357deea82a0a3d8fe1c7af125175a39131a

tests/plugins/editorplaceholder/editorplaceholder.js
sculpt0r: Test changed with: https://github.com/ckeditor/ckeditor4/pull/4370/commits/742a64a46273fcf5caad5667764349e1b5a81f28

But for RunTime errors there is no `console.error` invokes. It looks like it in dev-console, but has no effect in code.
Finally I bind with `window.addEventListener( 'error', onErrorStub );` to catch RunTime error. Also checks for common `console.error` in case of other fails in future.

tests/plugins/editorplaceholder/editorplaceholder.js
f1ames: The case from https://github.com/ckeditor/ckeditor4/issues/4253#issuecomment-689774158 doesn't involve `bbcode` plugin, but the effect is the same as you mentioned - you get HTML without body when `fullpage` plugin is used. So it looks it even has more sense than `fullpage` + `bbcode` plugin :thinking:

tests/plugins/editorplaceholder/editorplaceholder.js
```diff
@@ -205,5 +205,44 @@
 		} );
 	};
 
+	// (#4253)
+	tests[ 'test placeholder loads correctly in full-page editor with bbcode plugin'] = function() {
+
+		var errorLogStub = sinon.stub( console, 'error' );
```
f1ames, line 211:
```suggestion
		var consoleErrorStub = sinon.stub( console, 'error' );
```

tests/plugins/editorplaceholder/editorplaceholder.js
```diff
@@ -205,5 +205,44 @@
...
+
+		var onErrorStub = function() {
```
f1ames, line 213:
```suggestion
		var runtimeErrorListener = function() {
```

This is not a stub but listener so should be named accordingly.

tests/plugins/editorplaceholder/editorplaceholder.js
```diff
@@ -205,5 +205,44 @@
...
+			} );
+		};
+
+		window.addEventListener( 'error', onErrorStub );
+
+		var instanceReadyCallback = function( e ) {
+			resume( function() {
+				assert.isFalse( errorLogStub.called, 'There should be no errors during initialization' );
+			} );
+		};
```
f1ames, line 225: This function could be inlined into `editor.create`.

tests/plugins/editorplaceholder/editorplaceholder.js
```diff
@@ -205,5 +205,44 @@
...
+				on: {
+					instanceReady: instanceReadyCallback
+				}
+			}
+		}, function( bot ) {
+
+		} );
+
+		errorLogStub.restore();
+		window.removeEventListener( 'error', onErrorStub );
```
f1ames, line 242: Well, if the test fails, this lines will not be executed. You can move stubs/listener removals to `tearDown` method like:

```javascript
if ( errorLogStub ) {
	errorLogStub.restore();
}
```

tests/plugins/editorplaceholder/editorplaceholder.js
```diff
@@ -205,5 +205,44 @@
...
+				extraPlugins: 'bbcode',
+				editorplaceholder: 'any',
+				fullPage: true,
+				on: {
+					instanceReady: instanceReadyCallback
+				}
+			}
+		}, function( bot ) {
+
+		} );
```
f1ames, line 239: It will be good to check if editor placeholder is present or not (you can see how it's done in other tests). However, due to issue with `bbcode` always returning empty space (see review description) it might be not possible for `bbcode` + `fullpage` combination. This means it will be good to add another test :wink: Probably the one mentioned in https://github.com/ckeditor/ckeditor4/pull/4370#discussion_r520687252 - you can just use fixed content like:

```
editor.on( 'toDataFormat', function( evt) {
	evt.data.dataValue = '<p>Foo Bar</p>'
}, null, null, 16 );
```

Also it's good to have two versions of such tests - one with empty editor, another one with editor preinitliazed with some content. This would mean we should add 4 tests:

*   `bbcode` + `fullpage` with no content (this one we already have)
*   `bbcode` + `fullpage` with preinitialized content
*   `toDataFormat` body filtering with no content
*   `toDataFormat` body filtering with preinitialized content

f1ames (changes requested):
Looks good, some polishing of unit tests needed :+1: 

---

There is also one issue we have stumbled upon earlier (see https://github.com/ckeditor/ckeditor4/pull/4251#pullrequestreview-481525255):

> Also for some reason for empty editor, `getData()` returns one empty space (so length 1 instead of 0 which doesn't help here). And yes, it is an empty space, so there is now way to distinguish if editor is empty or someone just typed space because `charCode` is the same.

So we can only fix error being thrown for `bbcode` + `fullpage` but editor placeholder will not be displayed anyway (that's how it works now, which is good) due to single `space` character always being returned. But as @sculpt0r mentioned:

> It appears `bbcode` plugin filter-out all non-body content.

This is very rare edge case so we can accept that.

tests/plugins/editorplaceholder/editorplaceholder.js
f1ames: Ok, makes sense :+1:

tests/plugins/editorplaceholder/editorplaceholder.js
f1ames: See https://github.com/ckeditor/ckeditor4/pull/4370#discussion_r524966468, it seems we will need more tests anyway... :smile:

tests/plugins/editorplaceholder/editorplaceholder.js
sculpt0r: Added test cases with [3d52ae18](https://github.com/ckeditor/ckeditor4/pull/4370/commits/d52ae18718fd2f0f89053aa85727ecdca42be7b9)

tests/plugins/editorplaceholder/editorplaceholder.js
```diff
@@ -205,5 +219,101 @@
...
+	};
+
+	// (#4253)
+	tests[ 'test placeholder is visible with data filter to no content' ] = function() {
+		bender.editorBot.create( {
+			name: 'placeholderFilterToNoContent',
+			config: {
+				editorplaceholder: 'any',
+				fullPage: true,
+				startupData: '<p>Initialized content</p>'
```
f1ames, line 279: I think there should be no startup data as this tests checks empty editor scenario.

tests/plugins/editorplaceholder/editorplaceholder.js
```diff
@@ -205,5 +219,101 @@
...
+			}
+		}, function( bot ) {
+			var editor = bot.editor;
+
+			editor.on( 'toDataFormat', function ( evt ) {
+				evt.data.dataValue = '';
+			}, null, null, 16 );
+
+			editor.fire( 'change' );
```
f1ames, line 288: Why `change` event needs to be fired here (and in the next test too)?

f1ames (changes requested):

tests/plugins/editorplaceholder/editorplaceholder.js
```diff
@@ -205,5 +219,101 @@
...
+		}, function( bot ) {
+
+		} );
+
+		wait();
+	};
+
+	// (#4253)
+	tests[ 'test placeholder loads correctly in full-page editor with bbcode plugin and initial content' ] = function() {
+		bender.editorBot.create( {
```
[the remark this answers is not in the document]
sculpt0r: Added log error checking for another test with pattern:
```
addErrorAsserts();

//editor creation

wait()
```
Works fine for `test placeholder loads correctly in full-page editor with bbcode plugin and initial content`.

Try the same with `test placeholder not visible with data filter to preinitialized content`. This time error is not handled.
But if I add `extraPlugins: 'bbcode'` - it works fine... The error is caught and test fail due to RuntimeError...  :thinking: 

With `bbcode` this error looks like error in console. But without this plugin it looks like casual `log`

image
Tried with another extraPlugins like: `htmlwriter`, `easyimage`, `devtools`, `image2` and `mathjax` - but non of them behaved like `bbcode`. So adding each as the only extraPlugin gives same result as without any plugin for this test.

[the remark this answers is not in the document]
sculpt0r: Finally, after talk with @f1ames there will be only one test for RunTime error check. Since it's the core sense of this issue.
The rest of them just check edge cases in data.

f1ames:
Rebased onto latest `master`.

tests/plugins/editorplaceholder/editorplaceholder.js
```diff
@@ -205,5 +233,93 @@
...
+			name: 'placeholderFilterToNoContent',
+			config: {
+				editorplaceholder: 'any',
+				fullPage: true
+			}
+		}, function( bot ) {
+			var editor = bot.editor;
+
+			editor.on( 'toDataFormat', function ( evt ) {
```
f1ames, line 290:
```suggestion
			editor.on( 'toDataFormat', function( evt ) {
```

Unnecessary space.

tests/plugins/editorplaceholder/editorplaceholder.js
```diff
@@ -205,5 +233,93 @@
...
+			name: 'placeholderFilterPreinitialized',
+			config: {
+				editorplaceholder: 'any',
+				fullPage: true,
+				startupData:preinitializedContent
+			}
+		}, function( bot ) {
+			var editor = bot.editor;
+
+			editor.on( 'toDataFormat', function ( evt ) {
```
f1ames, line 314:
```suggestion
			editor.on( 'toDataFormat', function( evt ) {
```

Unnecessary space.

tests/plugins/editorplaceholder/editorplaceholder.js
```diff
@@ -26,9 +26,37 @@
 		}
 	};
 
+	function addErrorAsserts( benderScope ) {
+		benderScope.consoleErrorStub = sinon.stub( console, 'error' );
+
+		benderScope.runtimeErrorListener = function() {
+			resume( function() {
+				assert.fail( 'There should be no RunTime errors!' );
```
f1ames, line 34:
```suggestion
				assert.fail( 'There should be no runtime errors!' );
```

tests/plugins/editorplaceholder/editorplaceholder.js
```diff
@@ -205,5 +233,93 @@
...
+		bender.editorBot.create( {
+			name: 'bbcodeFailTest',
+			config: {
+				extraPlugins: 'bbcode',
+				editorplaceholder: 'any',
+				fullPage: true,
+				on: {
+					instanceReady: function() {
+						resume( function() {
+							assert.isFalse( this.consoleErrorStub.called, 'There should be no errors during initialization' );
```
f1ames, line 249: Didn't notice before but I wonder why this assert is here and not in `bender.editorBot.create` callback?

tests/plugins/editorplaceholder/editorplaceholder.js
```diff
@@ -205,5 +233,93 @@
 		} );
 	};
 
+	// (#4253)
+	tests[ 'test placeholder loads correctly in full-page editor with bbcode plugin' ] = function() {
```
f1ames, line 237: I just noticed this test could be further simplified to something like:

```javascript
	// (#4253)
	tests[ 'test placeholder loads correctly in full-page editor with bbcode plugin' ] = function() {
		addErrorAsserts( this );

		bender.editorBot.create( {
			name: 'bbcodeFailTest',
			config: {
				extraPlugins: 'bbcode',
				editorplaceholder: 'any',
				fullPage: true
			}
		}, function( bot ) {
			// ...
			if ( bot.testCase.consoleErrorStub ) { // Make sure console stub is there (handle older IEs)
				assert.isFalse( bot.testCase.consoleErrorStub.called, 'There should be no errors during initialization' );
			}

			assert.pass( 'Error was not thrown' );
		} );
	};
```

Wait is not needed here, since `resume` in runtime error listener resumes `wait()` from `editorBot.create`.

f1ames (changes requested):
One test fails in IE9 (probably IE10 too but haven't checked):

image

The issue here is that, without dev console open, older IEs don't have `console` object available... So probably stubbing `console.log` for older IEs should be skipped. I think it's not a problem here since the error the test is checking for is runtime error so it still will be caught.

tests/plugins/editorplaceholder/editorplaceholder.js
f1ames: This comment is not relevant anymore see https://github.com/ckeditor/ckeditor4/pull/4370#discussion_r526898877.

tests/plugins/editorplaceholder/editorplaceholder.js
f1ames: Also you can get rid of additional listener as mentioned in https://github.com/ckeditor/ckeditor4/pull/4370#discussion_r526872559. And handle older IEs console error mentioned in https://github.com/ckeditor/ckeditor4/pull/4370#pullrequestreview-534402832.

tests/plugins/editorplaceholder/editorplaceholder.js
sculpt0r: >The correct way would be to spy/stub console error with sinon to see if no errors where thrown - if error is thrown use assert.fail( ... ) to fail the test and if editor fires instanceReady event test should be marked as passing with assert.pass( ... ).

Maybe I took your comment too literally.

sculpt0r:
> One test fails in IE9 (probably IE10 too but haven't checked):
> 
> image

First, I introduce pre-check with `if (window.console)` before stubbing console in the test. With the issue fix, all tests passed. Tested on IE (11, 10, 9), Chrome, FF, Safari.

 IE 8 is ignored for all `editorplaceholder` plugin.

Then, run a single runtime checking error test, but with old-broken logic in the plugin.
IE 11, IE 10 - test failed, but with proper error catch. So it's ok.
IE 9 - failed with theoretically proper error catch, but still with `console` errors
image

Even with `console` open, I've got

image

So I'm moving to skip `console` stubbing with pre-check and also `CKEDITOR.env.ie && CKEDITOR.env.version < 10`

tests/plugins/editorplaceholder/editorplaceholder.js
sculpt0r: `editor.fire( 'change' );` is invoked just to be sure, that the newly added `toDataFormat` filter will be applied before assertion. No matter what `startupData` is.

In this case there is no `startupData` anyway, but I want to be sure that case from [#4253 (
 comment)](https://github.com/ckeditor/ckeditor4/issues/4253#issuecomment-689774158) is covered.

f1ames: Ok, I see that without `change` fire this test does not fail without a fix, so it makes sense as you mentioned :+1:

f1ames (approved):
LGTM :+1: :tada: 

Agree that skipping runtime error checks for IE9 as mentioned in https://github.com/ckeditor/ckeditor4/pull/4370#issuecomment-730943094 makes sense :+1:

f1ames:
Rebased onto latest `master`.

PR #4376, opened 2020-11-17
Fix failing tests/plugins/autogrow/autogrow test

### LINKED ISSUE #4323: Failing tests/plugins/autogrow/autogrow test
## Type of report

Failing test

On some widths both autogrow tests randomly fails. Probably due to some toolbar elements :thinking: **It fails on CI (for [example here](https://travis-ci.org/github/ckeditor/ckeditor4-presets/jobs/740155838#L404)), but we had some difficulties to reproduce this when running locally. Might be resolution/OS related**.

### Expected result

Test passes

### Actual result

Test fails

image

## Other details

* Browser: Firefox, IE8
* OS: All
* CKEditor version: 4.15.1
* Installed CKEditor plugins: preset full - only release version

Dumluregn:

Fix failing test

### This PR contains

- [x] Unit tests
- [ ] Manual tests

Skip

## What changes did you make?

The reason for tests failure on Firefox (btw I also reproduced it on Chrome and Safari) were vertical scrollbars in two places: in the editor itself and in the browser. 
As for the first issue, there was a race condition (because the scrollbar was visible just for a moment), so to resolve it I changed the object that was measured: from `editor.getEditable()` to `editor.getResizable()` - the latter doesn't change it's value no matter if scrollbar is visible or not (the same was the reason I couldn't use `editor.getViewPaneSize()`).
The second problem was solved by extracting tests to separate files - now the editors content always fits the browser viewport, so scrollbar doesn't show up.
Since IE8 doesn't support the way the width or height are calculated and there was no reliable way to replace it, I ignored it.

Closes #4323.

### DIFF
--- a/tests/plugins/autogrow/_helpers/tools.js
+++ b/tests/plugins/autogrow/_helpers/tools.js
@@ -1,0 +1,14 @@
+/* exported autogrowTools */
+
+var autogrowTools = ( function() {
+	function getEditorSize( editor ) {
+		return {
+			width: parseInt( editor.getResizable().getComputedStyle( 'width' ), 10 ),
+			height: parseInt( editor.getResizable().getComputedStyle( 'height' ), 10 )
+		};
+	}
+
+	return {
+		getEditorSize: getEditorSize
+	};
+} )();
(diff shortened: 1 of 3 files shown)


f1ames (approved):
LGTM :+1: 

Did some minor refactoring to simplify tests. Let's wait for the CI to be sure and then I will merge it.

Dumluregn:
CI failed but because of some random test in `core`, so I've restarted it.

PR #4399, opened 2020-11-24
Create setup-workflows.yml

jacekbogdanski:


f1ames (approved):
LGTM :+1: 

---

This PR results in the following workflows being added to the repo:

* [stalebot](https://github.com/ckeditor/ckeditor4-workflows-common/blob/master/workflows/stalebot.yml)
* [update-deps](https://github.com/ckeditor/ckeditor4-workflows-common/blob/master/workflows/update-deps.yml)
* setup-workflows (the one added in this PR)

So to make sure if it will work I have checked if requirements for all workflows are met:

1. stalebot - **already in use here** :heavy_check_mark: 
1. update-deps
    * Is `package.json` there :heavy_check_mark:
        * Has prod deps? :-1: (will be handled correctly by workflow itself)
        * Has dep deps? :heavy_check_mark: 
1. Required secrets / env variables :heavy_check_mark: 

---

~Updating dev deps may be painful for the first time since some of them are "slightly" outdated so I'm not sure if we will be able to to it~ Doesn't look bad at all :crossed_fingers: 
image

PR #4407, opened 2020-11-27
Fix Samples for CKEDITOR.stylesSet object's methods are calls of CKEDITOR.plugin

### LINKED ISSUE #4281: Samples for CKEDITOR.stylesSet object's methods are calls of CKEDITOR.plugin
## Type of report

Docs

## Provide a description of requested docs changes

[Here](https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_stylesSet.html)

E.g. code sample for `CKEDITOR.stylesSet.add()` is `CKEDITOR.plugins.add( 'sample', { ... plugin definition ... } );`.

sculpt0r:

Bug fix

### This PR contains

- [ ] Unit tests
- [ ] Manual tests

```
Skip
```

## What changes did you make?

*Give an overview…*
- Added note in API comments, that `CKEDITOR.stylesSet` is instance of `CKEDITOR.resourceManager`
- Removed `@extends` annotation from `CKEDITOR.styleSet` since it's not the proper inheritance and to prevent documentation displaying all properties and methods from `CKEDITOR.resourceManager`
- Add additional API comment for `CKEDITOR.stylesSet#add` method since it's binded to `CKEDITOR.stylesSet` context and used to adding new stylessets

https://github.com/ckeditor/ckeditor4/issues/4281#issuecomment-734791127

Closes #4281 .

### DIFF
--- a/core/plugins.js
+++ b/core/plugins.js
@@ -11,7 +11,7 @@
 /**
  * Manages plugins registration and loading.
  *
- * **Note** CKEDITOR.plugins is instance of {@link CKEDITOR.resourceManager}
+ * **Note** This object is an instance of {@link CKEDITOR.resourceManager}.
  *
  * @class
  * @extends CKEDITOR.resourceManager
--- a/core/plugins.js
+++ b/core/plugins.js
@@ -11,6 +11,8 @@
 /**
  * Manages plugins registration and loading.
  *
+ * **Note** CKEDITOR.plugins is instance of {@link CKEDITOR.resourceManager}
+ *
  * @class
  * @extends CKEDITOR.resourceManager
  * @singleton
--- a/core/style.js
+++ b/core/style.js
@@ -1888,7 +1888,7 @@ CKEDITOR.styleCommand.prototype.exec = function( editor ) {
 /**
  * Manages styles registration and loading. See also {@link CKEDITOR.config#stylesSet}.
  *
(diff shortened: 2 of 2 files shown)


jacekbogdanski (changes requested):
From what I see we are also using the approach with extending `resourceManager` in `core/plugins.js`:

https://github.com/ckeditor/ckeditor4/blob/523755ace313afdfa6e86637ca390926de86dee0/core/plugins.js#L11-L18

Looking at the docs https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_stylesSet.html the API is totally correct and accessible from `stylesSet` instance. Note that e.g. we are using other methods, like `addExternal`:
https://github.com/ckeditor/ckeditor4/blob/523755ace313afdfa6e86637ca390926de86dee0/core/style.js#L1931-L1934 so overwriting only `add` docs is a bit too little.

I don't honestly see any problem with the current docs as they are correctly pointing to the API of `resourceManager` which is used as a class to the `stylesSet` object. It could be a bit better because `stylesSet` looks like a class, but it also tells that it's an singleton and shows correct return type in constructor:

image
image

Looking at the cons and pros of the current approach, we should rather only add additional information that this object is an instance of the `resourceManager`. It would be also good to add the same information for `core/plugins.js` in this PR.

jacekbogdanski (approved):
LGTM!

PR #4413, opened 2020-12-01
Fix incorrect high contrast mode detection

### LINKED ISSUE #4379: A11y: Buttons present inside editor section are not visible in High Contrast Black mode.
Any environment

When using high contrast black mode, some buttons are not visible due to Opacity property of ckeditor button class. (Currently 0.3 for disabled buttons)

image

​OS Details:​
Windows Server 2019 Version 1809(OS Build 17763.914)
​
Browser Details:​
Chromium Edge Version 81.0.416.58 (Official build) beta (64-bit)

​Repro Steps:​​
1.Turn on High Contrast and open any ckeditor instance, demo site would suffice.

7.Verify user is able to view the buttons inside editor section or not.
​​
Actual Result:​​
Buttons present inside editor section are not visible in High Contrast Black mode.
Expected Result:​​
Buttons present inside editor section should be visible in High Contrast Black mode.
User Impact:​​
Low vision users can not see all controls if few controls are not visible in high contrast mode.
​
MAS Reference:​​
https://microsoft.sharepoint.com/:w:/r/sites/accessibility/_layouts/15/Doc.aspx?sourcedoc=%7B0BF582E1-6800-4E43-8947-DA015AA28B2B%7D&file=MAS%201.4.3%20%E2%80%93%20Contrast%20(Minimum).docx&action=default&mobileredirect=true

Comandeer:

Bug fix

### This PR contains

- [] Unit tests
- [x] Manual tests

```
* [#4379](https://github.com/ckeditor/ckeditor4/issues/4379): [Edge] Fixed: incorrect detection of the high contrast mode.
```

## What changes did you make?

I've enabled detecting HC also in Webkit/Blink. Additionally added the manual test for https://dev.ckeditor.com/ticket/5429

I didn't create unit tests as HC mode is really hard to detect programatically.

Closes #4379.

### DIFF
--- a/tests/core/ckeditor/manual/hcdetection.md
+++ b/tests/core/ckeditor/manual/hcdetection.md
@@ -14,4 +14,13 @@
 	## Unexpected
 
 	Buttons in the UI are displayed with icons.
+3. Check the information above the editor.
+
+	## Expected
+
+	It says "HC is on".
+
+	## Unexpected
+
+	It says "HC is off".
 
--- a/tests/core/ckeditor/manual/hcdetection.md
+++ b/tests/core/ckeditor/manual/hcdetection.md
@@ -2,7 +2,7 @@
 @bender-ui: collapsed
 @bender-ckeditor-plugins: wysiwygarea, toolbar, undo, clipboard, basicstyles, about, stylescombo, format, link, image
 
-**Note**: this test is dedicated for OSes that **DOES** support high contrast mode.
+**Note**: this test is dedicated for OSes that **DOES** support high contrast mode (Windows mostly). You can see if HC Mode is correctly detected based on test indicator (`HC is on`).
 
 1. Enable high contrast mode in your OS.
 2. Check the editor.
@@ -14,13 +14,3 @@
 	## Unexpected
 
 	Buttons in the UI are displayed with icons.
-3. Check the information above the editor.
-
-	## Expected
-
-	It says "HC is on".
-
-	## Unexpected
-
-	It says "HC is off".
-
--- a/tests/core/ckeditor/manual/hcdetection.md
+++ b/tests/core/ckeditor/manual/hcdetection.md
@@ -2,16 +2,16 @@
 @bender-ui: collapsed
 @bender-ckeditor-plugins: wysiwygarea, toolbar, undo, clipboard, basicstyles, about, stylescombo, format, link, image
 
+**Note**: this test is dedicated for OSes that **DOES** support high contrast mode.
+
 1. Enable high contrast mode in your OS.
 2. Check the editor.
 
 	## Expected
 
-	* If the current OS and browser support high contrast: Buttons in the UI are replaced by text labels.
-	* If the current OS and browser don't support high contrast: Buttons in the UI are displayed with icons.
+	Buttons in the UI are replaced by text labels.
 
 	## Unexpected
 
-	* If the current OS and browser support high contrast: Buttons in the UI are displayed with icons.
-	* If the current OS and browser don't support high contrast: Buttons in the UI are replaced by text labels.
+	Buttons in the UI are displayed with icons.
 
--- a/tests/core/ckeditor/manual/hcdetection.md
+++ b/tests/core/ckeditor/manual/hcdetection.md
@@ -1,0 +1,17 @@
+@bender-tags: feature, 4.15.2, 4379
+@bender-ui: collapsed
+@bender-ckeditor-plugins: wysiwygarea, toolbar, undo, clipboard, basicstyles, about, stylescombo, format, link, image
+
+1. Enable high contrast mode in your OS.
+2. Check the editor.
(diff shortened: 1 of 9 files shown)


tests/core/ckeditor/manual/hcdetection.md
```diff
@@ -0,0 +1,17 @@
+@bender-tags: feature, 4.15.2, 4379
+@bender-ui: collapsed
+@bender-ckeditor-plugins: wysiwygarea, toolbar, undo, clipboard, basicstyles, about, stylescombo, format, link, image
+
+1. Enable high contrast mode in your OS.
```
f1ames, line 5: This is a little inconsistent with expected - this step assumes one can enable HC mode and Expected/Unexpected mentions cases where HC mode is not supported by OS which makes it somewhat confusing.

At least we need better description or split this into two parts (OS supporting HC mode and the ones without it).

f1ames (changes requested):
Ok, so I have tested it with HC mode on various browsers and OSes.

## Windows 10

### Edge, IE11, Firefox

Setting HC mode on Windows 10 works fine with Edge, IE11 and Firefox and it is correctly detected and handled by CKEditor 4.

### Chrome

Chrome doesn't really support HC mode out-of-the-box and recommends [HC mode extension](https://chrome.google.com/webstore/detail/high-contrast/djcfdncoelnlbldjfhinnjlhdjlikmph?hl=en), using Dark mode (with some dark themes) or going with experimental features.

With HC mode extension, the contrast is similar to native support but CKEditor 4 doesn't recognize it correctly (probably due to our detection mechanism).

Dark mode doesn't really change the way pages are displayed as long as they don't support it - so there is no effect for CKEditor 4 itself too.

The 3rd option is to use experimental feature so go to `chrome://flags` and enable `Forced colors` - this turns on HC mode support and works correctly the same wa as in Edge.

**I would say this is fine. We can think of extracting support for HC mode extension and dark mode as feature requests** :thinking:

## Windows 7

### IE8

Still works fine :+1: 

## Linux

On Linux (Mint) HC mode doesn't really change the way both Chrome and Firefox renders content (which probably means lack of support), but that's a separate issue we can look later on if needed.

## macOS

There is [Invert Colors option](https://support.apple.com/lv-lv/guide/mac-help/unac089/mac) but it's not exactly the same as Windows HC mode and will not be correctly detected by CKEditor 4.

## Mobiles

As for mobiles, both Android and iOS supports Invert Colors / Text Contrast features (Android [1](https://support.google.com/accessibility/android/answer/6151855?hl=en&ref_topic=9079043) and [2](https://support.google.com/accessibility/android/answer/6151800?hl=en&ref_topic=9079043), [iOS](https://support.apple.com/en-us/HT207025)), but from what I have checked (on Android) they are not detected by our detection mechanism since the styling (border color) may be different. **This is the same case most probably as for macOS and Linux**.

## Summing up

If you take a look at our [High Contrast documentation section](https://ckeditor.com/docs/ckeditor4/latest/guide/skin_sdk_hc.html) it states only Windows HC feature support which is in tact with the above. This PR will require slight adjustments there - to mention at least Edge support (and maybe Chrome experimental one too), @Comandeer could you take care of this too?

So the fix is fine, we only need to polish test slightly so they are not too confusing and are only run in OSes/browsers which are supported. Please see also review comments regarding test.

Comandeer:
Probably we will be able to introduce better support for Chrome and mobile browsers after rewriting the detection mechanism to use [`forced-colors` media query](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/forced-colors) by default and fall back to the current mechanism in case of older browsers. However, at the moment the support for `forced-colors` is non-existent/experimental.

tests/core/ckeditor/manual/hcdetection.md
Comandeer: Probably splitting test into two will be the most readable solution.

Comandeer:
Ok, I've updated tests and prepared PR with changes to documentation → ckeditor/ckeditor4-docs#340.

f1ames:
> Probably we will be able to introduce better support for Chrome and mobile browsers after rewriting the detection mechanism to use [`forced-colors` media query](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/forced-colors) by default and fall back to the current mechanism in case of older browsers. However, at the moment the support for `forced-colors` is non-existent/experimental.

@Comandeer do you think `forced-colors` is also used for mobiles/macOS? Or we will need yet different mechanism to detect it on macOS/iOS/Android? Anyway, @Comandeer could you extract it as a follow-up?

f1ames (approved):
LGTM :+1: 

---

I was thinking if we shouldn't automatically ignore manual tests based on `CKEDITOR.env.hc`, but then if this detection will not/stop working we won't see it as tests will be auto ignored. So keeping it as is will help in such cases :+1:

PR #4416, opened 2020-12-02
ContextMenu: Enable Native Spellcheck for mis-spelled words

### LINKED ISSUE #3336: Add config option: hold ctrl and right click for CKeditor contextmenu
## Type of report

Feature request

## Provide a description of the new feature

Add an additional configuration option for the context menu plugin. 

Currently, if you want to use the browsers context menu while using the contextmenu plugin you must hold down ctrl and right-click. The majority of our users want the browser context by default while our power users can still hold down ctrl to use the CKeditor contextmenu instead. 

E.g reverse the behaviour

- right-click = browser context menu
- ctrl right-click = CKeditor context menu

 We would like an additional configuration option that enables this switch.

seanmcne:
The documented feature request is here: https://github.com/ckeditor/ckeditor4/issues/3336
This is extending current functionality if disableNativeSpellChecker = false, then it should display the native context menu for mis-spelled words. This is not net new functionality (per-se) however a change to enable native context menu's ONLY when clicking on mis-spelled words. 

Yes, it contains a manual test because we were unable to accomplish _mis-spelled word_ automatic test validation, at this time we could only accomplish test validation/verification with manual tests. 

### This PR contains

- [x ] Unit tests (Existing from the context menu) 
- [x ] Manual tests

https://github.com/ckeditor/ckeditor4/issues/3336
```
* [#3336](https://github.com/ckeditor/ckeditor4/issues/3336
```

## What changes did you make?
Added an extra parameter to the context menu plugin and use it to switch what context menu is shown by default.  This is accomplished via setting disableNativeSpellChecker to false or enable the NativeSpellChecker for only right mouse clicks on mis-spelled words. 

Closes #3336.

### DIFF
--- a/plugins/contextmenu/plugin.js
+++ b/plugins/contextmenu/plugin.js
@@ -48,7 +48,7 @@ CKEDITOR.plugins.add( 'contextmenu', {
 				 * @param {Boolean} [nativeContextMenuOnCtrl] Whether to open native context menu if the
 				 * <kbd>Ctrl</kbd> key is held on opening the context menu. See {@link CKEDITOR.config#browserContextMenuOnCtrl}.
 				 */
-				addTarget: function( element, nativeContextMenuOnCtrl ) {
+				addTarget: function( element, contextMenuSwitchingEnabled, nativeContextMenuOnByDefault) {
 					var holdCtrlKey,
 						keystrokeActive;
 
@@ -59,8 +59,15 @@ CKEDITOR.plugins.add( 'contextmenu', {
 								// which make this property unreliable. (https://dev.ckeditor.com/ticket/4826)
 								( CKEDITOR.env.webkit ? holdCtrlKey : ( CKEDITOR.env.mac ? domEvent.$.metaKey : domEvent.$.ctrlKey ) );
 
-						if ( nativeContextMenuOnCtrl && isCtrlKeyDown ) {
+						if ( contextMenuSwitchingEnabled ) {
 							return;
+							if ( nativeContextMenuOnByDefault && !isCtrlKeyDown ) {
+								return;
+							}
+							
+							if ( !nativeContextMenuOnByDefault  && isCtrlKeyDown ) {
+								return;
+							}
 						}
 
 						// Cancel the browser context menu.
@@ -163,7 +170,7 @@ CKEDITOR.plugins.add( 'contextmenu', {
 		var contextMenu = editor.contextMenu = new CKEDITOR.plugins.contextMenu( editor );
 
 		editor.on( 'contentDom', function() {
-			contextMenu.addTarget( editor.editable(), editor.config.browserContextMenuOnCtrl !== false );
+			contextMenu.addTarget( editor.editable(), editor.config.contextMenuSwitchingEnabled !== false ,editor.config.nativeContextMenuOnByDefault == false );
 		} );
 
 		editor.addCommand( 'contextMenu', {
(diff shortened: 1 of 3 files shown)


plugins/contextmenu/plugin.js
```diff
@@ -59,8 +59,15 @@ CKEDITOR.plugins.add( 'contextmenu', {
...
-						if ( nativeContextMenuOnCtrl && isCtrlKeyDown ) {
+						if ( contextMenuSwitchingEnabled ) {
 							return;
+							if ( nativeContextMenuOnByDefault && !isCtrlKeyDown ) {
+								return;
+							}
+							
+							if ( !nativeContextMenuOnByDefault  && isCtrlKeyDown ) {
+								return;
+							}
```
f1ames, line 70: This is basically so called "dead code" - the `return` above will stop the execution so these `if`s are never reached. Which means it doesn't do anything :thinking:

plugins/contextmenu/plugin.js
```diff
@@ -163,7 +170,7 @@ CKEDITOR.plugins.add( 'contextmenu', {
 		var contextMenu = editor.contextMenu = new CKEDITOR.plugins.contextMenu( editor );
 
 		editor.on( 'contentDom', function() {
-			contextMenu.addTarget( editor.editable(), editor.config.browserContextMenuOnCtrl !== false );
+			contextMenu.addTarget( editor.editable(), editor.config.contextMenuSwitchingEnabled !== false ,editor.config.nativeContextMenuOnByDefault == false );
```
f1ames, line 173: This is a breaking change as `browserContextMenuOnCtrl` is not supported anymore. IN CKEditor 4 we try to keep backwards compatibility between versions so such change should introduce new config options along with the old ones (and not changing how they worked before).

tests/plugins/contextmenu/manual/enabledNativeSpellChecker.html
```diff
@@ -0,0 +1,16 @@
+<h2>Spell Check Manual Test - Right Click should use native browser spell check when disableNativeSpellChecker is false (native spell checker enabled).</h2>
+<textarea id="editor1">&lt;p&gt;The more a thing tends to be permanent, the more it tends to be lifeleess.&lt;/p&gt;</textarea>
+
+<script>
+	// No console support on mobiles (without connecting to another device).
```
f1ames, line 5: From what I understand this test doesn't require console so this comment is a bit confusing. Still as context menu is not really there on mobiles, I think the test itself still should be ignored there.

tests/plugins/contextmenu/manual/enabledNativeSpellChecker.md
```diff
@@ -0,0 +1,12 @@
...
+@bender-ckeditor-plugins: wysiwygarea, toolbar, elementspath, resize
+
+This test checks whether, when a word is misspelled, that when right clicking, suggestions will appear.
+
+1. Misspell word.
+2. Right click.
+3. Locate correctly spelled word.
+4. Choose correctly spelled word.
+
+**Expected result:** After choosing correctly spelled word, spell check suggestion disappears.
```
f1ames, line 12: This basically checks native browser behaviour (spellchecker suggestions in native browser menu) which we don't have full control over. It will be good to redefine this test so it checks CKEditor 4 behaviour only (e.g. simply if native context menu is shown).

f1ames (changes requested):
Hello @seanmcne, thank you for the contribution! :clap: 

> This is not net new functionality (per-se) however a change to enable native context menu's ONLY when clicking on mis-spelled words.

Can you elaborate more on this? AFAIU this would mean that native context menu will be shown only for misspelled words and CKEditor 4 one for the rest of the content/elements? However from what I see, now it completely disables CKEditor 4 context menu :thinking: 

You have introduced two new config options `contextMenuSwitchingEnabled` and `nativeContextMenuOnByDefault`. All configuration options should be documented the same as e.g. [`browserContextMenuOnCtrl`](https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_config.html#cfg-browserContextMenuOnCtrl).

Also all new configuration options and resulting behaviour should be covered with tests (unit or manuals or both if possible) for all scenarios - so for boolean flags when it's turn off and on.

@seanmcne feel free to check our [contribution guide](https://ckeditor.com/docs/ckeditor4/latest/guide/dev_contributing_code.html) or reach out with any questions regarding this PR.

PR #4440, opened 2020-12-15
Explicit Ajax helpers for new response types

### LINKED ISSUE #4394: Explicit Ajax helpers for new response types
## Type of report

Feature request

Recently we've introduced an additional `arraybuffer` response type to the ajax load method. Since we already have semantic `loadXml` method and now `load` accepts additional type argument, we should unify API by adding additional:

* `loadText`
* `loadBinary`

helpers. See https://github.com/ckeditor/ckeditor4/pull/4339#pullrequestreview-526245162 for more details.

sculpt0r:

New feature - according to #4394 

### This PR contains

- [X] Unit tests
- [ ] Manual tests

```
* [#4394 ](https://github.com/ckeditor/ckeditor4/issues/4394): Added ajax predefined loading methods for: binary - `loadBinary` and text - `loadText`.
```

## What changes did you make?

*Give an overview…*

Added two methods:
- loadText
- loadBinary
based on the current `load` method.

Rename ajax load methods in tests. For those called `load` with `rsponseType` set to 'text' or 'arraybuffer' - I replace them with newly created `loadText` and `loadBinary`. Since it's a new API with predefined `responseType` - tests can be based on the newest changes. Also, add issue markers for tests based on `loadText` and `loadBinary`.

Closes #4394

### DIFF
--- a/tests/plugins/ajax/ajax.js
+++ b/tests/plugins/ajax/ajax.js
@@ -234,15 +234,14 @@
 		},
 
 		// (#4394)
-		'test load async binary': function() {
+		'test load image async as binary gets ArrayBuffer object': function() {
 			setTimeout( function() {
 				CKEDITOR.ajax.loadBinary( '../../_assets/test_icon.png', callback );
 			}, 0 );
 			wait();
-
 			function callback( data ) {
 				resume( function() {
-					assert.areSame( new ArrayBuffer(), data, 'The loaded data doesn\'t match' );
+					assert.areSame( ArrayBuffer, data.constructor , 'The loaded data doesn\'t match' );
 				} );
 			}
 		}
--- a/tests/plugins/ajax/ajax.js
+++ b/tests/plugins/ajax/ajax.js
@@ -231,6 +231,20 @@
 					assert.areSame( 'Sample Text', data, 'The loaded data doesn\'t match' );
 				} );
 			}
+		},
+
+		// (#4394)
+		'test load async binary': function() {
+			setTimeout( function() {
+				CKEDITOR.ajax.loadBinary( '../../_assets/test_icon.png', callback );
+			}, 0 );
+			wait();
+
+			function callback( data ) {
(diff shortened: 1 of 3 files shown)


tests/plugins/ajax/ajax.js
```diff
@@ -182,7 +182,7 @@
 			wait();
 		},
 
-		// (#1134)
+		// (#1134) (#4394)
```
f1ames, line 185: I wonder if it wouldn't be better to duplicate those tests. Ones with generic `load` (so stays as is) and other ones with specific type method. I know at the moment `load( ..., ..., type )` simply acts as a proxy for `loadType( ..., ... )` but if any more complex logic would appear in the future inside `load` itself we will have it covered and it will be better at detecting regressions. WDYT?

f1ames (changes requested):
Just one clarification regarding unit tests.

tests/plugins/ajax/ajax.js
sculpt0r: Well, I just based on theoretical knowledge, that if we can cover some logic with some API - there is no need to test it again separately. It creates more tests to maintain. So if new complex logic appears in `load` we need to reflect it in more tests.

TBO - I know nothing about effective regression detection. But from my point of view - we probably catch it - because new API calls also fail for the same reasons why `load` might fail.

f1ames: Ok, makes sense for me :+1:

f1ames (approved):
LGTM :+1: 

---

One note regarding docs and custom @params - now you will see that building docs show a warning about "unknown type":

image

It doesn't change the resulting docs itself (it is still shown correctly AFAIS), however it's there to say something is missing and should be fixed most probably. For such occasions we have external types definition in docs JSDuck task. I have already covered that on major branch.

f1ames:
Rebased onto latest `major`.

Issue #4462, opened 2020-12-31
Support editor functions after reattach to DOM


sculpt0r:
## Type of report

Feature request

## Provide description of the new feature

Allow CKEditor works in a detached state.
We could support detaching and reattaching the editor (or its parent).

The main problem happens with the editable area based on `iframe`. Each time `iframe` is attached again - browser tries to reload it. Our `iframe` has an empty `src` attribute so it's impossible to just reload its content. We need to recreate it. It involves many bindings between references to documents and events.

Some integrations also may invoke the `load` event when data changes and editor is bonded to the data model.

Related to #4390

sculpt0r:
Related issue: ckeditor/ckeditor4-react#109

CKEditorBot:
Closed in https://github.com/ckeditor/ckeditor4/pull/4601

PR #4463, opened 2020-12-31
Prevent editor creation on detached element

### LINKED ISSUE #4461: Prevent editor creation on detached element
## Type of report

Feature request

Prevent editor creation while targeting element is detached from DOM.
Also, we could add tracking, whenever the element will be attached - then the editor will be automatically created.

Some frameworks may 'hide' targeting element by detaching it - then we get some unexpected errors (like in #4390 )

sculpt0r:

New feature

### This PR contains

- [X] Unit tests
- [X] Manual tests

```
* [#4461](https://github.com/ckeditor/ckeditor4/issues/4461): Delay editor creation if target element is detached.
```

## What changes did you make?

*Give an overview…*
- If `config.delayDetached` is `true` and target element isDetached: immediately return `null` and start checking in `1000`ms periods.
- If target element is attached -> create an editor instance and stop checking.
- Editor will be available e.g. inside `onInstanceReady` via  `this`

```js
CKEDITOR.replace( textArea, {
	delayDetached: true,
	on: {
		instanceReady: function(evt) {
			this.setData("Editor created");
		}
	}
} )
```

Closes #4461 .


core/creators/themedui.js
```diff
@@ -384,6 +388,20 @@ CKEDITOR.replaceClass = 'ckeditor';
 		return editor;
 	}
 
+	function delayCreation( element, config, data, mode ) {
+		if ( config.delayDetached && element.isDetached() ) {
```
f1ames, line 392: Maybe more meaningful name would be useful here, fro example `delayCreationOnDetachedElement`?

f1ames: This config option is also missing API docs.

core/creators/themedui.js
```diff
@@ -384,6 +388,20 @@ CKEDITOR.replaceClass = 'ckeditor';
...
+			var intervalId = setInterval( function() {
+				if ( !element.isDetached() ) {
+					createInstance( element, config, data, mode );
+
+					clearInterval( intervalId );
+				}
+			}, 1000 );
```
f1ames, line 399: I guess this interval could be configurable, so one can adjust it depending on specific needs. And 1000 seems like a lot so I would go with default value with something between 50 - 100 ms.

core/creators/themedui.js
```diff
@@ -384,6 +388,20 @@ CKEDITOR.replaceClass = 'ckeditor';
 		return editor;
 	}
 
+	function delayCreation( element, config, data, mode ) {
+		if ( config.delayDetached && element.isDetached() ) {
+			var intervalId = setInterval( function() {
+				if ( !element.isDetached() ) {
+					createInstance( element, config, data, mode );
+
+					clearInterval( intervalId );
```
f1ames, line 397: I guess it will be better to move `clearInterval()` call to top so there is no risk of calling `createInstance` more than one time in some strange race condition, small interval cases.

tests/core/creators/manual/detached.md
```diff
@@ -0,0 +1,9 @@
+@bender-tags: 4.16.0, bug, 4461
+@bender-ui: collapsed
+@bender-ckeditor-plugins: toolbar, wysiwygarea, sourcearea, undo, clipboard, basicstyles, elementspath
+
+1. Use toggle button to detach and attach editor back.
+
+**Expected** Editor data is the same as at the beginning. Editor data is editable.
+
+**Unexpected** Editor data is empty. Editor data is not editable.
```
f1ames, line 9: I guess it will fail due to #4462 when editor is detached/attached multiple times? Maybe the scenario should be that editor is "created" on detached not and then one can attach it using button. The expected beahviour should be that it is initialized and works correctly?

tests/core/creators/manual/detached.html
```diff
@@ -0,0 +1,39 @@
...
+
+
+<script>
+	var editorElement = document.getElementById("editor1");
+
+	var textArea = document.createElement("textarea");
+
+	editorElement.appendChild(textArea);
+
+	editorElement.parentNode.removeChild(editorElement);
```
f1ames, line 20: I'm not sure why you are using `textarea` and attaching editor as it's child :thinking: AFAIR `textarea` couldn't have other elements inside (well, they will be treated as text).

Maybe, simply detaching parent `div` (or editor element itself) would be enough?

f1ames (changes requested):
CI is red for this PR could you check what's going on?

Could you also provide unit tests which covers this case?

sculpt0r:
> CI is red for this PR could you check what's going on?
> 
> Could you also provide unit tests which covers this case?

Initially, it was red because I didn't check if the config exists before use it. I fixed this.  
  
Now it's only red for \`BROWSER=Firefox MOZ\_HEADLESS=1\`. One of them shows fail with pdf export and second `tests/plugins/pastefromlibreoffice/generated/imagesextraction`

I restart one of them to check if there is any randomness.

tests/core/creators/manual/detached.html
sculpt0r: Actually, I'm not attaching editor as a child of `textarea`. I add `textarea` as a child of editor `div`.  
  
Originally I tried to recreate the same structure as we can observe in angular, but you are right: 

> Maybe, simply detaching parent `div` (or editor element itself) would be enough?

It should be enough. I changed this.

f1ames:
> fail with pdf export

This is a known issue we have already reported so it's not caused by this PR.

> and `tests/plugins/pastefromlibreoffice/generated/imagesextraction`

And this is not so if it still fails after restart it means this PR may be the cause here.

tests/core/creators/manual/detached.md
sculpt0r: I correct the scenario description and editor creation code.

sculpt0r:
> fail with pdf export
> 
> This is a known issue we have already reported so it's not caused by this PR.
> 
> and `tests/plugins/pastefromlibreoffice/generated/imagesextraction`
> 
> And this is not so if it still fails after restart it means this PR may be the cause here.

The second run: shows only fail with pdf on single BROWSER=Firefox MOZ\_HEADLESS=1.  
  
I will make all correct and I'll back to verify CI

core/creators/themedui.js
sculpt0r: Moved :)

core/creators/themedui.js
sculpt0r: As we talked: I will add an additional callback. If user provides it, we call with editor creation function as an argument. In this way, user has more control over creation time and won't be affected by 'ticking' interval in background.  
  
If there will bo no callback: we fall back to our `setInterval` and we will set timer based on provided configuration or use the default value (50 - 100 ms) as @f1ames mentioned.

If there will be an incorrect callback provided - we use one of the standard errors described in the documentation.

core/creators/themedui.js
sculpt0r: I changed the function name.

sculpt0r: I added config API docs.

sculpt0r:
Also, I add new CKEditor error code, and made a ckeditor4-docs PR: ckeditor/ckeditor4-docs#344

sculpt0r:
CI is green (except pdf exporter). Ready for review.

sculpt0r:
Waiting for next planing, after release

sculpt0r:
Rebase onto newest major.

core/creators/themedui.js
```diff
@@ -384,6 +388,71 @@ CKEDITOR.replaceClass = 'ckeditor';
...
+	//		registerCallback:RegisterCallback
+	//		}
+	//	} );
+	// ```
+	//
+	// @param {CKEDITOR.element} element The DOM element.
+	// @param {Object} [config] The specific configuration to apply to this
+	// editor instance. Configuration set here will override the global CKEditor settings.
+	// @param {String} [data] Initial value for the instance.
+	// @param {Number} mode CKEditor mode.
```
Dumluregn, line 413: I think it's a string - https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_editor.html#property-mode.

core/creators/themedui.js
```diff
@@ -384,6 +388,71 @@ CKEDITOR.replaceClass = 'ckeditor';
...
+			var intervalId = setInterval( function() {
+				if ( !element.isDetached() ) {
+					clearInterval( intervalId );
+
+					createInstance( element, config, data, mode );
+				}
+			}, delay );
+		};
+
+		var registerPayload = function() {
```
Dumluregn, line 434: We prefer function declaration over function expression. This way you can keep the whole logic concise and easier to follow.

core/creators/themedui.js
```diff
@@ -384,6 +388,71 @@ CKEDITOR.replaceClass = 'ckeditor';
...
+	// @param {String} [data] Initial value for the instance.
+	// @param {Number} mode CKEditor mode.
+	function delayCreationOnDetachedElement( element, config, data, mode ) {
+		if ( !config || !config.delayDetached || !element.isDetached() ) {
+			return false;
+		}
+
+		var callback = CKEDITOR.config.registerCallback;
+
+		var payload = function() {
```
Dumluregn, line 421: We prefer function declaration over function expression. This way you can keep the whole logic concise and easier to follow.

core/creators/themedui.js
```diff
@@ -384,6 +388,71 @@ CKEDITOR.replaceClass = 'ckeditor';
...
+				CKEDITOR.error( 'invalid-callback', { callback: config.registerCallback } );
+			} else {
+				payload = registerPayload;
+				callback = config.registerCallback;
+			}
+		}
+
+		callback( payload );
+
+		return true;
```
Dumluregn, line 453: When we change `payload` and `registerPayload` to function declarations, this logic can be moved before them to make clearer what is the goal of this function.

core/creators/themedui.js
```diff
@@ -384,6 +388,71 @@ CKEDITOR.replaceClass = 'ckeditor';
...
+	// @param {Object} [config] The specific configuration to apply to this
+	// editor instance. Configuration set here will override the global CKEditor settings.
+	// @param {String} [data] Initial value for the instance.
+	// @param {Number} mode CKEditor mode.
+	function delayCreationOnDetachedElement( element, config, data, mode ) {
+		if ( !config || !config.delayDetached || !element.isDetached() ) {
+			return false;
+		}
+
+		var callback = CKEDITOR.config.registerCallback;
```
Dumluregn, line 419: I don't recommend assigning a value to a variable and overwriting it later as it is unclear if that is intended behaviour or mistake.

BTW, here we could have an early return and call the payload instead:

```js
if ( !config.registerCallback ) {
	CKEDITOR.config.registerCallback( payload );
	return true;
}
```

with a second path if config was provided without any overwriting, but it's much easier to spot when you move the declarations after the logic (see the comments concerning `payload` and `registerPayload`).

Note that theoretically we could even ditch the reference to config here and just have `payload()`, but since this config should have a default value anyway, we can as well use it here I guess 🤔

core/creators/themedui.js
```diff
@@ -384,6 +388,71 @@ CKEDITOR.replaceClass = 'ckeditor';
...
+
+					createInstance( element, config, data, mode );
+				}
+			}, delay );
+		};
+
+		var registerPayload = function() {
+			if ( !element.isDetached() ) {
+				createInstance( element, config, data, mode );
+			} else {
```
Dumluregn, line 437: This empty `else` doesn't look cool, if you want to keep a comment it can be moved before the whole `if` statement.

core/creators/themedui.js
```diff
@@ -384,6 +388,71 @@ CKEDITOR.replaceClass = 'ckeditor';
...
+		};
+
+		if ( config.registerCallback ) {
+			if ( typeof config.registerCallback !== 'function' ) {
+				CKEDITOR.error( 'invalid-callback', { callback: config.registerCallback } );
+			} else {
+				payload = registerPayload;
+				callback = config.registerCallback;
+			}
+		}
```
Dumluregn, line 449: This logic can be simplified if you implement an early return proposed in the other review comment (sorry but I can't link it during review).

core/creators/themedui.js
```diff
@@ -569,3 +638,48 @@ CKEDITOR.config.startupMode = 'wysiwyg';
...
+ *		};
+ *
+ *		// Possible use:
+ *		// Save creation callback.
+ *		// Call resumeEditorCreation whenever you choose.
+ *		config.registerCallback = function ( createEditor ) {
+ *			resumeEditorCreation = createEditor;
+ *		};
+ *
+ * @cfg {Function} [registerCallback=see example]
```
Dumluregn, line 680: Such a reference doesn't look great in docs TBH: https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_config.html#cfg-blockedKeystrokes. Let's pass an actual default here (if multiline comment doesn't work here, it's short enough to inline it - `function( callback ) { callback() };`).

core/creators/themedui.js
```diff
@@ -569,3 +638,48 @@ CKEDITOR.config.startupMode = 'wysiwyg';
...
+ * @member CKEDITOR.config
+ */
+CKEDITOR.config.registerCallback = function( func ) {
```
Dumluregn, line 683: At least `callback` instead of `func` would be more informative.

Dumluregn (changes requested):
I didn't take a closer look at tests yet, but I see there are similar styling issues as in #4481, so please revise the PR (especially tests) taking into account e.g.:
- whitespaces
- removing commented code
- grouping `var` statements wherever possible.

Besides that - see inline comments.

sculpt0r:
@Dumluregn 

Manual and automatic test corrected (Description, tags, code style, grouping variables, and whitespace).  
Also, correct all inline comments. I guess we can drop ckeditor/ckeditor4-docs#344 due to current logic implementation?

Ready for another review.

core/creators/themedui.js
```diff
@@ -384,6 +388,63 @@ CKEDITOR.replaceClass = 'ckeditor';
...
+
+		config.registerCallback( registerPayload );
+
+		return true;
+
+		function payload() {
+			var delay = parseFloat( config.delay );
+			delay = isNaN( delay ) ? CKEDITOR.config.delayDetachedFrequency : delay;
+
+			var intervalId = setInterval( function() {
```
Dumluregn, line 432: As we discussed in team talk, let's add a `CKEDITOR.warn()` message before the first interval and when the interval is removed (so editor is created). Please remember to test it (probably stubbing the `CKEDITOR.warn()` method with Sinon).

tests/core/creators/manual/detached.md
```diff
@@ -0,0 +1,16 @@
+@bender-tags: 4.17.0, bug, 4461
+@bender-ui: collapsed
+@bender-ckeditor-plugins: toolbar, wysiwygarea, sourcearea, undo, clipboard, basicstyles, elementspath
+
+1. Click first button to attach first editor target element to DOM.
+2. Click second button to attach second editor target element and create it from callback.
```
Dumluregn, line 6: Let's split this test into two since they are independent. We'll have smaller and easier to understand `.html` files then.

tests/core/creators/manual/detached.html
```diff
@@ -0,0 +1,66 @@
...
+
+	editorParent.removeChild( editorElement );
+
+	CKEDITOR.replace( editorElement, {
+		delayDetached: true,
+		on: {
+			instanceReady: function() {
+				this.setData( 'Editor created' );
+			}
+		}
```
Dumluregn, line 29: This code is not checked now, so we should either remove it or add to test verification steps.

tests/core/creators/manual/detached.html
```diff
@@ -0,0 +1,66 @@
...
+
+	CKEDITOR.replace( secondEditor, {
+		delayDetached: true,
+		delayDetachedFrequency: 50,
+		registerCallback: RegisterCallback,
+		on: {
+			instanceReady: function() {
+				this.setData( 'Editor created from config callback' );
+			}
+		}
```
Dumluregn, line 54: This code is not checked now, so we should either remove it or add to test verification steps.

tests/core/creators/detached.js
```diff
@@ -0,0 +1,81 @@
+/* bender-tags: editor, 4.17.0, bug, 4461*/
+/* bender-ckeditor-plugins: wysiwygarea */
+
+( function() {
+	'use strict';
+
+	var tests = {
+		editorElement: null,
+		editorParent: null,
```
Dumluregn, line 9: We don't seem to use it since in each test we create new variables.

core/creators/themedui.js
```diff
@@ -569,3 +630,43 @@ CKEDITOR.config.startupMode = 'wysiwyg';
...
+ * Verifying frequency (ms) to use with default {@link CKEDITOR.config#registerCallback callback}.
+ *
+ * Used to set `setInterval` which checks whenever element is attached to DOM again.
+ *
+ *		config.delayDetachedFrequency = 1000;
+ *
+ * @cfg {Number} [delayDetachedFrequency=50]
+ * @member CKEDITOR.config
+ */
+CKEDITOR.config.delayDetachedFrequency = 50;
```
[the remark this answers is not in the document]
Dumluregn: Not sure about the performance with this interval, isn't 50 a bit too often? Would 200 be ok or it would cause a visible lag?

tests/core/creators/manual/detached.md
```diff
@@ -0,0 +1,16 @@
+@bender-tags: 4.17.0, bug, 4461
```
Dumluregn, line 1: bug->feature 😉

Dumluregn (changes requested):
Most of the comments are just naming tips. As a general rule try to avoid giving generic names to variables, functions etc. if their behaviour is always the same - rather try to describe what they are doing.

I've also introduced `delayIfDetached` namespace for the config options to indicate they are basically one feature. 

As for the tests - please add unit tests checking all scenarios where parameters are set, but `delayCreationOnDetachedElement()` function returns `false` (so the editor should be normally instantiated). Also I'd like to add a manual test with some more crazy callback after creating editor (something like alert to the face) to ensure we can do more than just create editor then. I was also wondering if we can check somehow if the interval is running correctly (in intervals - not just once), but with only a single log at the first attempt I have no idea how to do it - you're welcome to propose something.

Also I've created a docs issue to cover the feature introduced in this PR - #4553.

core/creators/themedui.js
```diff
@@ -384,6 +388,63 @@ CKEDITOR.replaceClass = 'ckeditor';
...
+			var intervalId = setInterval( function() {
+				if ( !element.isDetached() ) {
+					clearInterval( intervalId );
+
+					createInstance( element, config, data, mode );
+				}
+			}, delay );
+		}
+
+		function registerPayload() {
```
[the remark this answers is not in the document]
sculpt0r: Based on [https://github.com/ckeditor/ckeditor4/pull/4463#discussion_r586367355](https://github.com/ckeditor/ckeditor4/pull/4463#discussion_r586367355) I removed this function, and make its definition anonymous as `config.registerCallback` a parameter. Now it only wraps `createInstance` with proper params, so maybe there is no need to even name it?

core/creators/themedui.js
```diff
@@ -384,6 +388,63 @@ CKEDITOR.replaceClass = 'ckeditor';
...
+		if ( !config.registerCallback ) {
+			CKEDITOR.config.registerCallback( payload );
+			return true;
+		}
+
+		config.registerCallback( registerPayload );
+
+		return true;
+
+		function payload() {
```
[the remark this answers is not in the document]
sculpt0r: I try with `intervalyAttemptInstanceCreation`. It's a little bit longer, but I think it informs about: interval nature of conditional (`attempt` part) instance creation.

core/creators/themedui.js
```diff
@@ -384,6 +388,63 @@ CKEDITOR.replaceClass = 'ckeditor';
...
+	//		delayDetachedFrequency: 50,
+	//		registerCallback:RegisterCallback
+	//		}
+	//	} );
+	// ```
+	//
+	// @param {CKEDITOR.element} element The DOM element.
+	// @param {Object} [config] The specific configuration to apply to this
+	// editor instance. Configuration set here will override the global CKEditor settings.
+	// @param {String} [data] Initial value for the instance.
```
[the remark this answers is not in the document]
sculpt0r: Those params reflect `createInstance` arguments list, so yes - they are optional. I wanted to preserve the same order as in, but you are right :) I moved `mode` to the second position.

core/creators/themedui.js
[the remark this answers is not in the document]
sculpt0r: As I tested it - even with `100` there is a visible pop of element content before the editor initializes. The approach here was like this:

*   if you care about performance: use the dedicated callback to create an editor exactly when it's needed.
*   if don't: it looks like the editor shows immediately (but you have ticking interval with short period 'in background')

core/creators/themedui.js
```diff
@@ -569,3 +630,43 @@ CKEDITOR.config.startupMode = 'wysiwyg';
...
+ *		// Save creation callback.
+ *		// Call resumeEditorCreation whenever you choose.
+ *		config.registerCallback = function( createEditor ) {
+ *			resumeEditorCreation = createEditor;
+ *		};
+ *
+ * @cfg {Function} [registerCallback = function( createEditor ){ resumeEditorCreation = createEditor; }]
+ * @member CKEDITOR.config
+ */
+CKEDITOR.config.registerCallback = function( callback ) {
```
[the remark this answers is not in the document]
sculpt0r: We started mixing 'camel case' with underscore convention. Maybe we should pass the entire object as config?

```
config.delayCreation = {
    frequency: 50,
    callback: function() {}
}
```

WDYT?

core/creators/themedui.js
```diff
@@ -384,6 +388,63 @@ CKEDITOR.replaceClass = 'ckeditor';
...
+				if ( !element.isDetached() ) {
+					clearInterval( intervalId );
+
+					createInstance( element, config, data, mode );
+				}
+			}, delay );
+		}
+
+		function registerPayload() {
+			if ( !element.isDetached() ) {
```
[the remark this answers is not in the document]
sculpt0r: Actually, it is another check inside a method that is invoked by the developer. So, you are right. In this case, we should move responsibility on a developer.

core/creators/themedui.js
sculpt0r: Probably we should replace the test message with a new error-code?

The editor displays a tip, that more information could be in docs and also warn the function itself to treat the first argument as an error code - not a message.

tests/core/creators/manual/detached.html
sculpt0r: Removed.

tests/core/creators/manual/detached.html
sculpt0r: https://github.com/ckeditor/ckeditor4/pull/4463#discussion_r588162780

sculpt0r:
> Also I'd like to add a manual test with some more crazy callback after creating editor (something like alert to the face) to ensure we can do more than just create editor then.

I'm not sure about which callback is mentioned here. There is no custom callback on editor creation. Even if it's created in the interval, we still detect editor finished initialization by its events (like `instanceReady`) to not break anything.  
The user provides a callback which is immediately called. We provide there, as argument, function, and we are saying here: just invoke this thing whenever target element is attached to DOM so we safely create an editor for you even if we suspend creation before. And this is a part of the developer responsible for invoking it in proper time (https://github.com/ckeditor/ckeditor4/pull/4463#discussion_r586367355)

The reasons for this a little bit confuse construction:

*   don't want to leak `createInstance` function outside since we provide `replace` as creation API (for blocktext editing)
*   don't want to return anything new from `replace` method (currently it is `editor` or `null`).

We are asking the User for callback, which we can use to provide our 'resumeEditorCreation' function.

sculpt0r:
> Also I'd like to add a manual test with some more crazy callback after creating editor (something like alert to the face) to ensure we can do more than just create editor then.

I'm not sure about which callback is mentioned here. There is no custom callback on editor creation. Even if it's created in the interval, we still detect editor finished initialization by its events (like `instanceReady`) to not break anything.  
The user provides a callback which is immediately called. We provide there, as argument, function, and we are saying here: just invoke this thing whenever target element is attached to DOM so we safely create an editor for you even if we suspend creation before. And this is a part of the developer responsible for invoking it in proper time (https://github.com/ckeditor/ckeditor4/pull/4463#discussion_r586367355)

The reasons for this a little bit confuse construction:

*   don't want to leak `createInstance` function outside since we provide `replace` as creation API (for blocktext editing)
*   don't want to return anything new from `replace` method (currently it is `editor` or `null`).

We are asking the User for callback, which we can use to provide our 'resumeEditorCreation' function.

sculpt0r:
All comments are corrected. Sometimes I add some comments to them.

I added a test for verifying `CKEDITOR.warn` and propose a test to verify if `setInterval` is called. In the current version, we spy on `editorElement` `isDetached` method since it is the only thing that is called in `setInterval`.

@Dumluregn: ready for another review.

core/creators/themedui.js
Dumluregn: Yes, I think it should be defined as a standard error and the delay set in config should be passed as additional data. IIRC you've already mentioned before that https://github.com/ckeditor/ckeditor4-docs/pull/344 should be closed due to the change of implementation, so I've created a separate issue for adding error codes for this PR (https://github.com/ckeditor/ckeditor4/issues/4564).

Please note that what you wrote about `CKEDITOR.warn()` also applies to the success info, so it should also be extracted as an error code (I've had some doubts since it's a positive feedback for user, but we also use those codes for the warnings and this can be treated as a warning, so I think we are fine here).

core/creators/themedui.js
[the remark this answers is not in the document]
Dumluregn: I'm not sure which lines are you referring to - my guess would be
```js
config.delayIfDetached_callback( function() {
			createInstance( element, config, data, mode );
		} );
```
but uses the config that we want to have 🤔  Could you clarify it please?

core/creators/themedui.js
```diff
@@ -384,6 +388,62 @@ CKEDITOR.replaceClass = 'ckeditor';
...
+
+		config.delayIfDetached_callback( function() {
+			createInstance( element, config, data, mode );
+		} );
+
+		return true;
+
+		function intervalyAttemptInstanceCreation() {
+			var delay = parseFloat( config.delay );
+			delay = isNaN( delay ) ? CKEDITOR.config.delayIfDetached_interval : delay;
```
Dumluregn, line 432: This is a similar case we had with ensuring that element is detached after calling callback - if the config option is passed in the wrong format (e.g. `hello` instead of a number), we hide an error and not inform the developer about it making it very difficult to debug. We could throw a warning, but we don't do this for any other config option so I wouldn't introduce it now. Instead we should allow the script to fail and log the default browser debug tracing. So the default option should only be used if user doesn't provide any value.

core/creators/themedui.js
```diff
@@ -384,6 +388,62 @@ CKEDITOR.replaceClass = 'ckeditor';
...
+
+			CKEDITOR.warn( 'Editor creation will be delayed. We will keep trying every ' + delay + 'ms until target element will be reattached to DOM.' );
+
+			var intervalId = setInterval( function() {
+				if ( !element.isDetached() ) {
+					clearInterval( intervalId );
+
+					CKEDITOR.warn( 'Editor instance will be created now. Target element is attached to DOM.' );
```
Dumluregn, line 440: As mentioned in https://github.com/ckeditor/ckeditor4/pull/4463/files#r590134485, this should be extracted as a defined error message.

core/creators/themedui.js
Dumluregn: Also as for naming - delay is one-time thing and here we will use it for repeated action, so let's stick to the name `interval`.

core/creators/themedui.js
```diff
@@ -569,3 +629,45 @@ CKEDITOR.config.startupMode = 'wysiwyg';
...
+ *		config.delayIfDetached_callback = function( createEditor ) {
+ *			resumeEditorCreation = createEditor;
+ *		};
+ *
+ * @cfg {Function} [delayIfDetached_callback = function( createEditor ){ resumeEditorCreation = createEditor; }]
+ * @member CKEDITOR.config
+ */
+CKEDITOR.config.delayIfDetached_callback = function( createEditorFunction ) {
+	createEditorFunction();
+};
```
Dumluregn, line 673: If we go with wrapping the options into an object, whole this section needs to be updated.

tests/core/creators/detached.html
```diff
@@ -0,0 +1,47 @@
...
+		<p>Lorem ipsum dolor sit amet 7.</p>
+	</div>
+</div>
+
+<div>
+	<div id="editor8">
+		<p>Lorem ipsum dolor sit amet 8.</p>
+	</div>
+</div>
+
```
Dumluregn, line 47:
```suggestion
```

tests/core/creators/detached.js
```diff
@@ -0,0 +1,150 @@
+/* bender-tags: editor, 4.17.0, feature, 4461*/
```
Dumluregn, line 1: We use version tagging only for manual tests.

tests/core/creators/detached.js
```diff
@@ -0,0 +1,150 @@
...
+/* bender-ckeditor-plugins: wysiwygarea */
+
+( function() {
+	'use strict';
+	var tests = {
+		'test delay editor creation if target element is detached': function() {
+			var editorElement = CKEDITOR.document.getById( 'editor1' ).remove();
+
+			var editor = CKEDITOR.replace( editorElement, {
```
Dumluregn, line 10: Please join the `var` statements.

tests/core/creators/detached.js
```diff
@@ -0,0 +1,150 @@
...
+				delayIfDetached: true
+			} );
+
+			assert.isNull( editor );
+		},
+
+		'test editor is created immediately on not detached element even with delay config': function() {
+			var editorElement = CKEDITOR.document.getById( 'editor5' );
+
+			var editor = CKEDITOR.replace( editorElement, {
```
Dumluregn, line 20: Please join the `var` statements.

tests/core/creators/detached.js
```diff
@@ -0,0 +1,150 @@
...
+				delayIfDetached_callback: function() {}
+			} );
+
+			assert.isNotNull( editor );
+		},
+
+		'test editor is created immediately on not detached element with delayIfDetached config set as false': function() {
+			var editorElement = CKEDITOR.document.getById( 'editor6' );
+
+			var editor = CKEDITOR.replace( editorElement, {
```
Dumluregn, line 31: Please join the `var` statements.

tests/core/creators/detached.js
```diff
@@ -0,0 +1,150 @@
...
+				delayIfDetached: false
+			} );
+
+			assert.isNotNull( editor );
+		},
+
+		'test editor without config is created immediately on not detached element': function() {
+			var editorElement = CKEDITOR.document.getById( 'editor7' );
+
+			var editor = CKEDITOR.replace( editorElement );
```
Dumluregn, line 41: Please join the `var` statements.

tests/core/creators/detached.js
```diff
@@ -0,0 +1,150 @@
...
+
+			var editor = CKEDITOR.replace( editorElement, {
+				delayIfDetached: true
+			} );
+
+			assert.isNull( editor );
+		},
+
+		'test editor is created immediately on not detached element even with delay config': function() {
+			var editorElement = CKEDITOR.document.getById( 'editor5' );
```
Dumluregn, line 18: Can we change the fixtures order to follow the tests order (so that we had `editor2` here instead of `editor5`)?

core/creators/themedui.js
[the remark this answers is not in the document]
sculpt0r: Those are the lines I was writing about. Because earlier you asked me to rename `registerPayload` function.  
In [https://github.com/ckeditor/ckeditor4/pull/4463#discussion_r586367355](https://github.com/ckeditor/ckeditor4/pull/4463#discussion_r586367355) we decide to remove `isDetached` verify before further execution. So the only line which remains is the one with `createInstance`. And since we only need it to be wrapped in another function, and we don't use it anywhere else - it could be just placed as an argument.  
  
So I decline your 'rename' because now, there is no name needed ;) :P

tests/core/creators/detached.js
```diff
@@ -0,0 +1,150 @@
...
+				delayIfDetached: true,
+				delayIfDetached_interval: 50
+			} );
+
+			CKEDITOR.tools.setTimeout( function() {
+				resume( function() {
+					assert.isTrue( spyIsDetached.callCount > 5 );
+					spyIsDetached.restore();
+				} );
+			}, 3000 );
```
Dumluregn, line 143: This is a clever approach and I like it 👍🏻  I only propose to reduce the timeout to not extend the test longer than we need it - just 2 calls should make us comfortable, so `1000` would be enough for that.

tests/core/creators/manual/detachedconfig.html
```diff
@@ -0,0 +1,31 @@
...
+<div>
+	<div id="editor">
+		<p>Lorem ipsum dolor sit amet.</p>
+	</div>
+</div>
+
+<script>
+	var editor = document.getElementById( 'editor' ),
+		secondAttachButton = CKEDITOR.document.getById( 'attachButton' ),
+		secondEditorParent = editor.parentNode,
```
Dumluregn, line 12: Now it's just `attachButton` and `editorParent` 🙂

core/creators/themedui.js
```diff
@@ -343,6 +343,10 @@ CKEDITOR.replaceClass = 'ckeditor';
 			return null;
 		}
 
+		if ( delayCreationOnDetachedElement( element, mode, config, data ) ) {
```
Dumluregn, line 346: Please add a ticket reference here.

Dumluregn (changes requested):
There is some logic streamlining and wrapping config into an object ahead as we discussed in comments. Other than that - just some styling and naming.

As a side note: we have two docs issues extracted as follow-ups: https://github.com/ckeditor/ckeditor4/issues/4553 and https://github.com/ckeditor/ckeditor4/issues/4564.

tests/core/creators/detached.js
sculpt0r: There are some exceptions (like `widgetselection`) ;p  but I remove it :)

Dumluregn: Exceptions confirm the rule :P In manuals we use version tags to filter tests during testing phase, but we run all the units anyway, so it is a waste of a few seconds 😜  And if someone wants to add another test to this file in different release, this tag won't be valid anymore (same for the ticket reference btw), while we never add more cases to existing manual tests so their tags don't change.

tests/core/creators/detached.js
sculpt0r: I go even with 200. With `50`ms interval, it is enough to make at least 2 calls. Also, there is also one `isDetached` call in `replace` method. Since it's a unit test - those `800`ms might be valuable in a bigger test run.  
  
Also, I have to fix one bleeding `setInterval` in `test delay editor creation if target element is detached`. So after the test verifies the result - I attach the editor again. If we don't do that - this spy is counting still 'ticking' interval from the previous test.

core/creators/themedui.js
sculpt0r: Finally, I decided no to use an object as an option. `Delayed creation` with default options would look like this:
```
CKEDITOR.options.delayIfDetached = {
    delayIfDetached: true
}
```
which looks very redundant.  We can always get rid of this inner boolean and check if there is an object with `interval` or `callback` inside. But this would mean we have an object just for sake of single property...

Summary: Object as an option was my proposal and now I want to withdraw it... It would be more consistent if we stick to the convention which you mentioned here: https://github.com/ckeditor/ckeditor4/pull/4463#discussion_r590165258

sculpt0r:
@Dumluregn - ready for another review:

*   Simplify execution logic in delayed creation.
*   Tests are reordered (first non-delayed creation, then all tests with active delayed creation option).
*   Fixed: infinite setInterval in one test (https://github.com/ckeditor/ckeditor4/pull/4463#discussion_r590758236).
*   Use editor error code in warnings.
*   Adjust timing in the test which verifies `setInterval` calls.
*   Separate config properties are in use, instead of the entire object (https://github.com/ckeditor/ckeditor4/pull/4463#discussion_r591095500).
*   Correct delayed creation option API docs.

tests/core/creators/detached.js
```diff
@@ -0,0 +1,156 @@
+/* bender-tags: editor, feature, 4461*/
```
Dumluregn, line 1:
```suggestion
/* bender-tags: editor, feature, 4461 */
```

tests/core/creators/manual/detachedconfig.html
```diff
@@ -0,0 +1,31 @@
+<button id="attachButton">Attach editor to parent</button>
```
Dumluregn, line 1: We are testing config callback here, so let's take it into account in test name.

Dumluregn (changes requested):
Looks good! 👍🏻   The fixes left are smaller and smaller. Besides the comments, could you also provide a test for a custom interval time? It may be either unit one or manual - maybe the latter would be better if we could do something like "start test, wait 5s and check if editor appeared" with interval of 5000ms? We just have to delay the attaching editor to DOM for a while to not create it immidiately.

Dumluregn:
Also please rebase the PR onto latest `major` since we merged #4539 yesterday.

sculpt0r:

*   rebased on newest major
*   added one manual test: `detachedcustominterval`
*   rename one test: `detachedconfig` now became `detachedcustomcallback`
*   made all small requested changes

core/creators/themedui.js
```diff
@@ -384,6 +389,61 @@ CKEDITOR.replaceClass = 'ckeditor';
...
+	// If `config.delayIfDetached_callback` is declared as function, it will be invoked with single argument:
+	// function, that should be called to create editor.
+	//
+	// Otherwise: periodically (with `setInterval`) run check if element is attached to DOM and create editor automatically.
+	//
+	// ```js
+	//	CKEDITOR.replace( detachedEditorElement, {
+	//		delayIfDetached: true,
+	//		delayIfDetached_interval: 50,
+	//		delayIfDetached_callback:RegisterCallback
```
Dumluregn, line 405:
```suggestion
	//		delayIfDetached_callback: registerCallback
```

core/creators/themedui.js
```diff
@@ -384,6 +389,61 @@ CKEDITOR.replaceClass = 'ckeditor';
...
+	//		}
+	//	} );
+	// ```
+	//
+	// @param {CKEDITOR.element} element The DOM element.
+	// @param {String} mode CKEditor mode.
```
Dumluregn, line 411:
```suggestion
	// @param {String} mode CKEditor mode (source/wysiwyg).
```

tests/core/creators/manual/detachedcustominterval.html
```diff
@@ -0,0 +1,26 @@
+<button id="startButton">Start Test</button>
```
Dumluregn, line 1: Test works good 👍🏻  I'd just add a visual timer counting down the time to create editor.

tests/core/creators/detached.js
```diff
@@ -0,0 +1,155 @@
...
+
+	var tests = {
+		'test editor is created immediately on not detached element even with delay config': function() {
+			var editorElement = CKEDITOR.document.getById( 'editor1' ),
+				editor = CKEDITOR.replace( editorElement, {
+					delayIfDetached: true,
+					delayIfDetached_callback: function() {}
+				} );
+
+			assert.isNotNull( editor );
```
Dumluregn, line 15: We should add a failure message to each assert - without them:
1) it's almost impossible to track down an error on CI (like now);
2) it takes more time to understand what is wrong when checking the test after some time.

Dumluregn (changes requested):
One of the new unit tests is failing on the build version (both on [Chrome](https://travis-ci.org/github/ckeditor/ckeditor4/jobs/762927389) and [FF](https://travis-ci.org/github/ckeditor/ckeditor4/jobs/762927391)), but we don't know which one - please take a look at this, together with a small improvements in tests (see below).

tests/core/creators/detached.js
sculpt0r: It's `CKEDITOR.warn`stubbed test... ;(

sculpt0r:

*   Added messages for assertions
*   Found failing test [https://github.com/ckeditor/ckeditor4/pull/4463#discussion_r594932561](https://github.com/ckeditor/ckeditor4/pull/4463#discussion_r594932561) and fix it. There was another warn about ckeditor pdf token, which makes three `CKEDITOR.warn` calls - and we expect just two of them. I rewrite the test (based on `core/log` test, so we are testing now `errorCode` in warn calls.
*   Added visual timer for manual test

tests/core/creators/manual/detachedcustominterval.html
```diff
@@ -0,0 +1,43 @@
...
+		<p>Lorem ipsum dolor sit amet.</p>
+	</div>
+</div>
+
+<script>
+	var editorElement = document.getElementById( 'editor' ),
+		startButton = CKEDITOR.document.getById( 'startButton' ),
+		timerElement = CKEDITOR.document.getById( 'testTimer' ),
+		editorParent = editorElement.parentNode,
+		waitTime = 5000,
```
Dumluregn, line 15: It works very nice. Let's reduce time to `3000` to speed up the test.

tests/core/creators/manual/detachedcustominterval.md
```diff
@@ -0,0 +1,14 @@
...
+1. Click the button to start the test.
+2. Wait five seconds for editor creation.
+
+**Expected**
+  * Editor is created.
+  * Editor data is editable.
+
+**Unexpected**
+  * Editor isn't created.
+  * Editor data is empty.
```
Dumluregn, line 14: I think we should have the same expectations as in other tests?

core/creators/themedui.js
```diff
@@ -384,6 +389,61 @@ CKEDITOR.replaceClass = 'ckeditor';
...
+			return false;
+		}
+
+		if ( config.delayIfDetached_callback ) {
+			config.delayIfDetached_callback( function() {
+				createInstance( element, config, data, mode );
+			} );
+
+			return true;
+		}
```
Dumluregn, line 426: In this scenario we don't inform the user that the editor creation was suspended 🤔  Let's add a `CKEDITOR.warn()` call here too. I was wondering about the consistency in both cases (callback vs interval) and I found out the best solution is adding a `mode` parameter to the  'editor-delayed-creation' error (we already have it), then in callback scenario have `mode: 'callback'` and in interval one `mode: 'interval - ' + interval + ' ms'`. The code error info is rather hard to use programatically so I don't think it's super useful and important to keep it a numeric value instead of such a string. WDYT?

core/creators/themedui.js
```diff
@@ -384,6 +389,61 @@ CKEDITOR.replaceClass = 'ckeditor';
...
+
+		var interval = config.delayIfDetached_interval === undefined ? CKEDITOR.config.delayIfDetached_interval : config.delayIfDetached_interval;
+
+		CKEDITOR.warn( 'editor-delayed-creation', {
+			interval: interval
```
Dumluregn, line 431: As mentioned before, I propose to change the data passed:
```suggestion
			mode: 'interval -  ' + interval + ' ms'
```

or inversely (so the result would be `interval - 50 ms` or `50 ms interval` - the first is more coherent with callback scenario, but the second one feels more natural, so I'm leaving the decision up to you).

Dumluregn (changes requested):
Almost there 🤞🏻

core/creators/themedui.js
sculpt0r: The first scenario indeed sounds more coherent as `mode` property value.

core/creators/themedui.js
sculpt0r: > In this scenario we don't inform the user that the editor creation was suspended

I assume, that we don't have to inform the user if a callback was provided intentionally, but a simple warn should be helpful in debugging. I'll add another message :) I think such distinction also would be useful. The `mode` option sounds reasonable. TBH I would separate mode and time value as another property, but as you said, it should be used only for information purposes - so I leave it as once :)

sculpt0r: Actually, I added one more warn ;p To preserve consistent behavior in `interval` and `callback` mode, I added `editor-delayed-creation-sucess` to function which is passed as an argument to the user callback. So now, we have a warning if creation is delayed (both cases) and a warning for successful creation (both cases).

sculpt0r:

*   Added steps in manual tests to verify CKEDITOR.warn for interval and callback options.
*   Added unit test to verify warns displayed with callback option. Also added mode verification.
*   Adjusted wait time in a manual test.
*   Correct Expected / Unexpected section in a manual test.

Dumluregn (changes requested):
I was already adding the changelog entry and I had an eureka moment... What about the `CKEDITOR.inline()` method? We didn't cover it at all and I think we should at least check if it works - if yes then add unit tests, and if no then let's check why and how much work it takes to fix it.

PS Please pull the latest branch.

sculpt0r:
The original issue is caused by iframe - so it doesn't affect `inline` editor. Right now, it is possible to create an editor on the detached element, and after reattach - everything works fine.  
 

Since it is a new feature and by default is optional - probably we should create the same functionality for inline editor... Give me a second ;p

sculpt0r:
@Dumluregn - ready :)

*   Really heavy reworked unit tests. Now each test is run with three methods: `inline`, `replace`, `appendTo`
*   Added two tests for `inlineAll` and `replaceAll`. Both are basing on document searching. If elements are detached - there is nothing to find - so nothing is created.
*   Added delay for `inline` editor creation.

Dumluregn (changes requested):
I really appreciate an effort you put into covering all the possible cases in unit tests 👍🏻  🙏🏻  Unfortunately there are plenty of reasons why you will hardly find any generated tests (like you proposed) in our codebase and they can all be shortly summed up in two terms: readability and maintainability.

With generated tests it is very difficult to understand what exactly is tested, which cases are covered (probably the answer should be "all of them!" but that doesn't say much after half a year), you don't know how many tests there actually are until you run them (since adding cases is scattered in a few places), and the file is still big right now and hard to understand for an outsider. And if we need to add a test only for one of the methods, it still has to be put aside of the automation, so we have a poisonous mixture 💀 

There is hope, however; we should extract helpers to a separate file (by convention it is one file with path [`/_helpers/tools.js`](https://github.com/ckeditor/ckeditor4/blob/major/tests/plugins/widget/_helpers/tools.js) per plugin folder) and create a separate file for each editor creation method. This way we keep the code itself DRY (as you intended) and have smaller, easier to understand and modify files 💪🏻 

As the unit tests will be further reworked and most of the comments I would left now will not be valid afterwards, I had only left a few that more general ones I thought may be useful in the future - they start with [EXTRAS] tag 🙂

core/creators/inline.js
```diff
@@ -35,11 +35,17 @@
 			return null;
 		}
 
-		var editor = new CKEDITOR.editor( instanceConfig, element, CKEDITOR.ELEMENT_MODE_INLINE ),
-			textarea = element.is( 'textarea' ) ? element : null;
+		var textarea = element.is( 'textarea' ) ? element : null,
+			editorData = textarea ? textarea.getValue() : element.getHtml();
```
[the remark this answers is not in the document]
sculpt0r: To be completely honest, at some point I need to pass `editorData` to `delayCreationOnDetachedElement`\- so I need `if` between var declarations, and need to extract `editorData`. After some changes, I left this extraction, so that both below `if...else` branches look more alike. But it will work in the old way...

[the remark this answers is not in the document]
Dumluregn: In this comment I've only meant that previously we just used `textarea.getValue()`, and now if it is not present we fall back to `element.getHtml()` instead - and my question was how did it work previously if the element was not a textarea?

core/creators/inline.js
```diff
@@ -156,6 +162,68 @@
...
+	//	CKEDITOR.inline( detachedEditorElement, {
+	//		delayIfDetached: true,
+	//		delayIfDetached_callback: registerCallback
+	//	} );
+	// ```
+	//
+	// @param {CKEDITOR.element} element The DOM element.
+	// @param {Object} [config] The specific configuration to apply to this
+	// editor instance. Configuration set here will override the global CKEditor settings.
+	function delayCreationOnDetachedElement( element, config ) {
```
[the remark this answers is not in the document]
sculpt0r: Extracted with documentation comments.

sculpt0r:
@Dumluregn - ready again.

*   added `_helpers/tools.js`with detached test cases (as we talked)
*    Split `detach` unit test on three separate tests (eg. `detachedinline`)

As you mentioned, `[EXTRAS]`comments should be not necessary due to the above changes.

Dumluregn (approved):
LGTM! 👍🏻  However as we've already discussed, I'd like to ask for another pair of 👀 . So requesting rereview from @f1ames (probably)!

sculpt0r:
Rebase on newest major

f1ames (changes requested):
Looking good :+1: 

I've added some suggestions regarding API docs and code here and there. Also it looks like there is an issue with indentation in manual tests (especially 3rd point):

image

sculpt0r:
Rebase onto newest major.

core/editor.js
```diff
@@ -1648,6 +1648,83 @@
...
+	 *		delayIfDetached_callback: registerCallback
+	 *	} );
+	 * ```
+	 *
+	 * ```js
+	 *	CKEDITOR.replace( detachedEditorElement, {
+	 *		delayIfDetached: true,
+	 *		delayIfDetached_callback: registerCallback
+	 *	} );
+	 * ```
```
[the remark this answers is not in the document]
sculpt0r: :see_no_evil: :repeat:

core/editor.js
```diff
@@ -2228,3 +2305,53 @@ CKEDITOR.ELEMENT_MODE_INLINE = 3;
  * @event contentDomInvalidated
  * @param {CKEDITOR.editor} editor This editor instance.
  */
+
+/**
+ * Allow delayed editor creation (either with callback or intervals) if target element is detached from DOM.
```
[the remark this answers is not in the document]
sculpt0r: Also, added links to referred options.

core/editor.js
```diff
@@ -2228,3 +2305,53 @@ CKEDITOR.ELEMENT_MODE_INLINE = 3;
...
+ *		// Create editor calling `resumeEditorCreation()` whenever you choose (e.g. on button click).
+ *		resumeEditorCreation();
+ *
+ * @since 4.17.0
+ * @cfg {Function} [delayIfDetached_callback = function( createEditor ){ resumeEditorCreation = createEditor; }]
+ * @member CKEDITOR.config
+ */
+CKEDITOR.config.delayIfDetached_callback = function( createEditorFunction ) {
+	createEditorFunction();
+};
```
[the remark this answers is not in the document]
sculpt0r: Well - interval is used by default. It is determined by this line. https://github.com/ckeditor/ckeditor4/blob/16d640b8f19b97efd9b8478453493b4f1b2f0d04/core/editor.js#L1690

We are checking `config.delayIfDetached_callback` directly passed from the User - not `CKEDITOR.config.delayIfDetached_callback`. 

Provided default value actually based on this code: https://github.com/ckeditor/ckeditor4/blob/16d640b8f19b97efd9b8478453493b4f1b2f0d04/core/editor.js#L1695-L1700

This is because the entire 'merge' of configs happens later in... the editor constructor, which we don't want to call if creation needs to be delayed. So the default config option is useless after delay check. OTOH we need to present default behaviour of this option to the User.

Maybe we should extract this 'merge' at the beginning of the entire process?  :thinking:

core/editor.js
```diff
@@ -2228,3 +2305,53 @@ CKEDITOR.ELEMENT_MODE_INLINE = 3;
...
+ * @since 4.17.0
+ * @cfg {Boolean} [delayIfDetached=false]
+ * @member CKEDITOR.config
+ */
+CKEDITOR.config.delayIfDetached = false;
+
+/**
+ * Function with single argument. As argument is passed another function that continues editor creation.
+ * Allows to store the editor creation function and invoke it whenever it is convenient instead of periodically
+ * check if element is reattached to DOM.
```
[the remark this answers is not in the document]
sculpt0r: > Btw. what will happen if the function is called but the editor element is still detached? I see from the code it will simply try to create editor instance. This is good (I mean the 3rd-party controls it), however it will be good to mention here too - that the function should be called only after editor element is reattached to DOM.

Generally, it will be the same situation as we have now. 

I will improve doc comments.

core/editor.js
```diff
@@ -1648,6 +1648,83 @@
 
 		return element;
 	};
+
+	/**
+	 * Delay editor creation if given element is detached from DOM.
+	 *
+	 * Requires `config.delayIfDetached` to be equal to `true`.
+	 *
+	 * If `config.delayIfDetached_callback` is declared as function, it will be invoked with a single argument:
```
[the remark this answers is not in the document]
sculpt0r: OK, so I decide to change docs instead of code logic. Actually, we don't verify other config options that much. We just trust that callback is a function.

sculpt0r:
@f1ames - ready for review.

I correct all your suggestions but also left a few explanations as answers. So please check them to verify if some approaches are fine :)

core/editor.js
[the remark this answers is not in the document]
f1ames: That's true, we usually trust that the provided value will be of valid type. So in this context we could mention it as a function I suppose. I will correct the docs here :+1:

core/editor.js
```diff
@@ -1648,6 +1648,83 @@
 
 		return element;
 	};
+
+	/**
+	 * Delay editor creation if given element is detached from DOM.
```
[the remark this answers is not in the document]
f1ames: > We mark it as private, but it is actually used as a public

Yeah, that's sometimes an issue that's not really private. But marking it with `@private` makes it more clear when reading API docs that one should not use this method.

> I'm wondering if `initializeDelayedEditorCreation` could be also split because we have there two exclusive logic

I would leave it as is.

core/editor.js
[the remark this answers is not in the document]
f1ames: I would say this is a logic flaw here. If you take a look at [`CKEDITOR.config`](https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_config.html) docs:

> Stores default configuration settings. Changes to this object are reflected in all editor instances, if not specified otherwise for a particular instance.

This means those lines declare default value which should be used if not override by the user. So the fact we use `config` does not work the same as the rest of the editor.

So:

* This declaration is default value. Which should be `null`/`undefined` here.
* We should check if `config.delayIfDetached_callback` was declared and use it instead of a default value - not sure if it will be better to move entire config merging logic here (probably it may break something :speak_no_evil:) or just check this specific value.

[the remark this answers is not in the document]
sculpt0r: > not sure if it will be better to move the entire config merging logic here (probably it may break something speak_no_evil)

Agree. It's very core-related stuff - so there may be various edge cases...

> We should check if config.delayIfDetached_callback was declared and use it instead of a default value

Just to be clear. We should change the default value to `undefined` & correct API docs comments to describe that we use this callback only if it's declared. Otherwise, we use our default logic - which is interval checks?

core/editor.js
```diff
@@ -1648,6 +1648,79 @@
...
+					CKEDITOR.warn( 'editor-delayed-creation-success', {
+						method: 'interval - ' + interval + ' ms'
+					} );
+				}
+			}, interval );
+		}
+	};
+
+	/**
+	 * Verify whether editor creation could be delayed.
```
f1ames, line 1711:
```suggestion
	 * Whether editor creation should be delayed.
```

core/editor.js
```diff
@@ -1648,6 +1648,79 @@
...
+	 *
+	 * @since 4.17.0
+	 * @static
+	 * @member CKEDITOR.editor
+	 * @param {CKEDITOR.element} element The DOM element on which editor should be initialized.
+	 * @param {Object} config The specific configuration.
```
f1ames, line 1717:
```suggestion
	 * @param {Object} config Editor configuration.
```

core/editor.js
```diff
@@ -1648,6 +1648,79 @@
...
+	 * @returns {Boolean} True if creation could delayed.
```
f1ames, line 1718:
```suggestion
	 * @returns {Boolean} True if creation should be delayed.
```

core/editor.js
```diff
@@ -1648,6 +1648,79 @@
 
 		return element;
 	};
+
+	/**
+	 * Initialize delayed editor creation based on provided configuration.
+	 *
+	 * If the `config.delayIfDetached_callback` function is declared, it will be invoked with a single argument:
```
f1ames, line 1655: Could you make it a link so it will link to config option in built API docs?

core/editor.js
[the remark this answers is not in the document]
f1ames: > Just to be clear. We should change the default value to `undefined` & correct API docs comments to describe that we use this callback only if it's declared. Otherwise, we use our default logic - which is interval checks?

I mean, if the callback is not declared it won't be use because there is nothing to use there - so it will work the same as other configuration options which don't have value by default. So yes, the value here should be `undefined` as we want interval to be used by default. And I think the docs are already fine now :+1:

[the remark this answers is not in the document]
f1ames: And as for merging, since we change logic here, it will be good to cover with unit test too.

f1ames (changes requested):
Ok, looking good :+1: 

Only minor corrections in API docs here and there. And the fix related to `CKEDITOR.config` mentioned earlier. Tests are looking good too :+1: 

I didn't go with full test runs on all browser yet since we need to fix `CKEDITOR.config` thing first.

---

One more thing, when you try to build docs:

```bash
cd ckeditor4-docs
grunt build-serve --clean
```

there are two warnings:

```
Warning: /home/cksource/Workspace/CKEditor4/ckeditor4/core/editor.js:1652: Unknown type CKEDITOR.element
Warning: /home/cksource/Workspace/CKEditor4/ckeditor4/core/editor.js:1710: Unknown type CKEDITOR.element
```

Since `CKEDITOR.element` is not a valid type, it should be `CKEDITOR.dom.element` instead.

sculpt0r:
@f1ames - ready for review

- corrected all inline comments,
- fixed documentation creation warnings (changed argument type to valid one)

In case of [first](https://github.com/ckeditor/ckeditor4/pull/4463#discussion_r614723114) comment, related with [the second](https://github.com/ckeditor/ckeditor4/pull/4463#discussion_r614723688) as we also talked 1on1:
- added merge function for delayed creation options
- now, default values declarations are valid from the code point of view
- I didn't add any test cases. Those you mentioned, are already covered in helpers.
  - So we have the default behaviour for `delayIfDetached`: `test_editor_is_created_immediately_on_not_detached_element_even_with_delay_config`, `test_editor_is_created_immediately_on_not_detached_element_with_delayIfDetached_config_set_as_false`, `test_editor_without_config_is_created_immediately_on_not_detached_element`, `test_delay_editor_creation_if_target_element_is_detached`
  - Default behaviour for callback declaration. So if there is no callback we use interval strategy: `test_delay_editor_creation_with_default_interval_strategy_until_target_element_attach_to_DOM` or callback whenever provided: `test_editor_creation_from_provided_callback`
  - Behaviour with custom interval time: `test_editor_interval_attempts_to_create_if_target_element_is_detached`

f1ames:
Rebased onto latest `major`.

tests/plugins/widget/insertparagraphnexttowidget.js
```diff
@@ -23,6 +23,12 @@
 		bogusDiv = '<div>&nbsp;</div>';
 
 	bender.test( {
+		_should: {
+			ignore: {
+				// #4563
+				'test press shift + enter with ckeditor.enter_br inserts <br> element after a widget': bender.env.ie && bender.env.version == '8'
+			}
+		},
```
f1ames, line 31: I wonder what is the origin of this change here? :thinking:

f1ames (changes requested):
Ok, almost there :+1: Just one last touch regarding manual tests on mobiles and we are done :crossed\_fingers: 

## Code

:+1:

## Docs

:+1:

## Manual tests

All passes :+1:

*   Chrome :heavy\_check\_mark:
*   Firefox :heavy\_check\_mark:
*   Safari :heavy\_check\_mark:
*   Edge :heavy\_check\_mark:
*   IE11 :heavy\_check\_mark:
*   IE10 (skipped for now due to VM issues)
*   IE9 (skipped for now due to VM issues)
*   IE8 :heavy\_check\_mark:
*   Android :heavy\_check\_mark:
*   iOS :heavy\_check\_mark:

However, we usually skip manual tests which requires looking into dev console on mobiles as it's kind of cumbersome. So current test we have should ignored on mobiles. As a 2nd step we could introduce similar manual tests for mobile which only checks if editor is usable after attaching (so without validating dev console output). WDYT?

## Unit tests

*   Chrome (full, CI) :heavy\_check\_mark:
*   Firefox (full, CI) :heavy\_check\_mark:
*   Safari (full) :heavy\_check\_mark:
*   Edge (only `tag:4461`) :heavy\_check\_mark:
*   IE11 (full) :heavy\_check\_mark:
*   IE10 (only `tag:4461`) (skipped for now due to VM issues)
*   IE9 (only `tag:4461`) (skipped for now due to VM issues)
*   IE8 (only `tag:4461`) :heavy\_check\_mark:
*   Android (only `tag:4461`) :heavy\_check\_mark:
*   iOS (only `tag:4461`) :heavy\_check\_mark:

## Follow-ups

There is one case I could think of which may give some troubles. Imagine editor element is detached and there is interval in the background checking it it can be initialized. Now, instead of creating editor instance, one would like to destroy editor or more like stop the interval since editor is no longer needed.

In other words, if there is a change in the DOM structure and editor element will never be reattached the interval will run indefinitely (until browser tab is closed/refreshed) which may cause memory leaks, especially in SPA context.

:point\_up: I guess it could be reported as follow-up just for the future reference.

sculpt0r:
> There is one case I could think of which may give some troubles. Imagine editor element is detached and there is interval in the background checking it it can be initialized. Now, instead of creating editor instance, one would like to destroy editor or more like stop the interval since editor is no longer needed

You can't destroy editor, because it's not created yet :D However, the second case with canceling interval sounds reasonable, especially as follow up :p ;) I thought about it earlier. First - I want to return the result of the delay from the creation method. But it complicates everything. Because then, creation methods will return `undefined`, editor instance, cancelation token or/and a callback. So this will be a mess...

After some thought, this requires another groundwork on the current approach. So maybe it will be necessary to provide another callback :see_no_evil: :thinking: We could call it to provide cancelation callback to the user? In the same way, we provide an editor creation function? Because we don't have any instance jet, so we can't call some global method to cancel this interval...

tests/plugins/widget/insertparagraphnexttowidget.js
sculpt0r: Well, we should ask @Dumluregn ;p
Commit related to this is described as  'Ignore test failing due to the other ticket.`.

Dumluregn: You got me, I've tried to smuggle it here...

Dumluregn: https://github.com/ckeditor/ckeditor4/issues/4563 - this is the ticket.

sculpt0r:
Well, I'm wondering: should we create separate manual tests just to skip the dev console part? Probably it's against convention, but could we just add some note, that we should skip the console part on mobiles in two tests? `detacheddefault` seems to be fine, also for mobile?

f1ames:
> Well, I'm wondering: should we create separate manual tests just to skip the dev console part? Probably it's against convention, but could we just add some note, that we should skip the console part on mobiles in two tests?

Those would be different tests (with different tests scenarios) for different envs so it makes sense to separate them instead of creating one test with multiple scenarios. The simpler the test is the easier to perform it - and one can focus on checking if it passes instead of trying to guess how to follow test steps.

> `detacheddefault` seems to be fine, also for mobile?

:+1:

sculpt0r:
@f1ames - ready for review.
- added two manual tests: `detachedcustomcallbackmobile` and `detachedcustominterval`. Both are ignored in the non-mobile environment - since the test cases are the same as `detachedcustomcallback` and `detachedcustominterval`. However, I added `message` field in those tests. Because on mobile (at least mine) some operations are so slow, that in the first place you can think that nothing happened. So this field informs about the current stage.
- For `detachedcustomcallback` and `detachedcustominterval` I added mobile ignore.

> I guess it could be reported as follow-up just for the future reference.

Please take a look at https://github.com/ckeditor/ckeditor4/pull/4463#issuecomment-823856686. Should I create follow up?

f1ames:
Rebased onto latest `major`.

f1ames (approved):
LGTM :tada: Good job @sculpt0r and @Dumluregn :clap: :clap: :clap: 

---

> *   added two manual tests: `detachedcustomcallbackmobile` and `detachedcustominterval`. Both are ignored in the non-mobile environment - since the test cases are the same as `detachedcustomcallback` and `detachedcustominterval`. However, I added `message` field in those tests. Because on mobile (at least mine) some operations are so slow, that in the first place you can think that nothing happened. So this field informs about the current stage.
> *   For `detachedcustomcallback` and `detachedcustominterval` I added mobile ignore.

:+1: 

> Please take a look at [#4463 (comment)](https://github.com/ckeditor/ckeditor4/pull/4463#issuecomment-823856686). Should I create follow up?

Yes, please :+1:

---

Rechecked tests on IE9 and IE10 - everything's green :heavy_check_mark:

f1ames:
Just waiting for the CI to finish after rebase and I'll be merging this PR :+1:

sculpt0r:
> Yes, please +1

Created: https://github.com/ckeditor/ckeditor4/issues/4641

Issue #3868, opened 2020-02-19
Failing test tests/plugins/balloontoolbar/positioning


Comandeer:
## Type of report

Failing test

## Provide detailed reproduction steps (if any)

1. Open http://tests.ckeditor.test:1030/tests/plugins/balloontoolbar/positioning#tests%2Fplugins%2Fballoontoolbar%2Fpositioning%20test%20panel%20refresh%20position%20(divarea)

### Expected result

Test passes

### Actual result

Test fails

## Other details

* Browser: IE11, Chrome (only if directly executed)
* CKEditor version: 4.14.0

Dumluregn:
Test started to fail on `major` branch also when run from dashboard, so it breaks CI now.

Dumluregn:
Actually, test mentioned in this issue works (run both from dashboard and directly) in Chrome and IE11-9 🤔  The ones that fail on CI are different:
```
http://localhost:1030/tests/plugins/balloontoolbar/positioning#tests%2Fplugins%2Fballoontoolbar%2Fpositioning%20test%20panel%20-%20out%20of%20view%20-%20bottom%20center%20(editor1)
http://localhost:1030/tests/plugins/balloontoolbar/positioning#tests%2Fplugins%2Fballoontoolbar%2Fpositioning%20test%20panel%20-%20out%20of%20view%20-%20hcenter%20top%20(editor1)
```

So to close this issue I'm creating PR unignoring this test, but in turn I'll create a new issue for the issues above.

Dumluregn:
After merging it let's update the list in https://github.com/ckeditor/ckeditor4/issues/3794.

Dumluregn:
New issue for failing CI is https://github.com/ckeditor/ckeditor4/issues/4589.

CKEditorBot:
Closed in https://github.com/ckeditor/ckeditor4/pull/4588

Issue #3419, opened 2019-09-16
Edge and/or IE sometimes throw a permission denied error


msamsel:
## Type of report

Bug

## Provide detailed reproduction steps (if any)

There are some workarounds in code, which prevent of throwing this error. Maybe there will be time to investigate what exactly happens here:

https://github.com/ckeditor/ckeditor4/blob/acf57ad52650fafc07df2e98886517e82fd585da/core/dom/domobject.js#L141-L147

https://github.com/ckeditor/ckeditor4/blob/acf57ad52650fafc07df2e98886517e82fd585da/core/editable.js#L544-L547

From a brief analysis, it looks like issue tends to appear when we accessing a property/value/etc. of native HTML element which is detached. 

### Expected result

no error or some better error handling/prevention

### Actual result

simple try…catch which prevents error occurrence.

## Other details

Issue extracted during #3128 

* Browser: Edge || IE
* OS: Windows
* CKEditor version: …
* Installed CKEditor plugins: …

sculpt0r:
Some references:
- possible reason and solution made by jQuery https://github.com/jquery/sizzle/blob/5b3048605655285a81b06fbe4f49f2a14a8d790f/src/sizzle.js#L472-L480

The `wysiwygarea` uses `iframe` which **probably** causes a lot of reloads.

PR #4481, opened 2021-01-13
Support editor functions after reattach to DOM

### LINKED ISSUE #4462: Support editor functions after reattach to DOM

sculpt0r:

New feature

### This PR contains

- [x] Unit tests
- [X] Manual tests

```
* [#4462](https://github.com/ckeditor/ckeditor4/issues/4462): Support editor functions after reattaching to DOM.
```

## What changes did you make?

*Give an overview…*

Closes #4462 .


sculpt0r:
@ckeditor/qa-team Could you please look at this?  
There is a main manual test at: `http://0.0.0.0:1030/tests/plugins/wysiwygarea/manual/detached`  
The main reason, why it may fail is an iframe that may receive `onload` event and then tries to recreate the edit area.  
  
Also, after recreation the edit area: random functions may fail, like selection etc.  
  
For sure: any IE will fail (including 11) - so there is no need to verify this browser for now. Edge should be fine.  
  
Please ask me anything if some more explanation is needed.

sculpt0r:
I go through many issues related to `Permission denied` IE and the best I've figured out for now simply ignore it also on IE. It is already ignored on Edge.  
  
Without an opened console, it should work also on IE.

LukaszGudel:
On Safari toggling the editor in some cases removes content and makes the editor unfocusable. 

### Steps to reproduce:

Scenario 1:

1.  Open the manual test in Safari.
2.  Double click "Toggle" button.

Scenario 2: 

1.  Open the manual test in Safari.
2.  Add some content.
3.  Double click "Source" button.
4.  Double click "Toggle" button.

Scenario 3:

1.  Open the manual test in Safari.
2.  Click on "Set Data" button.
3.  Double click "Toggle" button.

### Result:

The content of the editor is missing and the editor is unfocusable. Toggling the editor once again or toggling source mode is fixing the issue.

sculpt0r:
Rebase onto newest major.

LukaszGudel:
There still some weird things happening on Safari. Redoing steps from my previous comment results in blinking content and high CPU usage:

sculpt0r:
@LukaszGudel - thank you. This is something I just can't catch on BrowserStack...

LukaszGudel:
I've retested this feature on all browsers and everything seems to work correctly 👍

sculpt0r:
CI - green, QA -verified on various browsers, so do I.  
Code was cleaned, ready for the real review.

plugins/wysiwygarea/plugin.js
```diff
@@ -95,6 +95,40 @@
...
+						if ( iframe.$.onloadFromSetData ) {
+							iframe.$.onloadFromSetData = false;
+							return;
+						}
+
+						var cacheData = editor.getData( false );
+
+						// Remove current editable, but preserve iframe
+						editor.editable().saveIframe = true;
+						editor.editable( 0 );
```
Dumluregn, line 120: `editor.editable()` takes an element as an argument and while it works with `0`, it would be more fitting and intuitive to pass `null` here.

plugins/wysiwygarea/plugin.js
```diff
@@ -95,6 +95,40 @@
 
 					editor.editable( new framedWysiwyg( editor, iframe.$.contentWindow.document.body ) );
 					editor.setData( editor.getData( 1 ), callback );
+
+					iframe.$.onloadFromSetData = false;
+
+					editor.on( 'mode', backupOnLoad, { iframe: iframe, editor: editor, callback: callback } );
+				}
+
+				function backupOnLoad( evt ) {
```
Dumluregn, line 104: Please add a short comment with a ticket reference before this function following [the code guide](https://github.com/ckeditor/ckeditor4/blob/major/dev/docs/codestyle.md#210-comments). Also please adjust the other comments according to this (I mean mostly the point `Comment is a sentence - start it with capital leter, end with period.`).

plugins/wysiwygarea/plugin.js
```diff
@@ -472,6 +507,11 @@
 
 					var doc = this.getDocument();
 
+					// Prevents iframe onload event
+					// Onload invokes SetData, so it leads to infinite loop
```
Dumluregn, line 511: Please add a ticket reference in this comment.

tests/plugins/wysiwygarea/manual/detached.html
```diff
@@ -0,0 +1,48 @@
...
+
+<div>
+	<div id="editor">
+		<p>Lorem ipsum dolor sit amet.</p>
+	</div>
+</div>
+
+
+<div id="fakedId"></div>
+<div id="fakedId1"></div>
```
Dumluregn, line 12: AFAICS this `<div>` is not used anywhere, is it necessary?

tests/plugins/wysiwygarea/manual/detached.html
```diff
@@ -0,0 +1,48 @@
...
+
+
+<script>
+	var editorElement = document.getElementById("editor");
+
+	var textArea = document.createElement("textarea");
+
+	editorElement.appendChild(textArea);
```
Dumluregn, line 20: Is it the same as in https://github.com/ckeditor/ckeditor4/pull/4463#discussion_r551381519?

tests/plugins/wysiwygarea/manual/detached.html
```diff
@@ -0,0 +1,48 @@
...
+
+
+	var e = CKEDITOR.replace( textArea, {
+		on: {
+			instanceReady:function(evt){}
```
Dumluregn, line 25: Is this empty listener needed?

tests/plugins/wysiwygarea/manual/detached.html
```diff
@@ -0,0 +1,48 @@
...
+	</div>
+</div>
+
+
+<div id="fakedId"></div>
+<div id="fakedId1"></div>
+
+
+<script>
+	var editorElement = document.getElementById("editor");
```
Dumluregn, line 16: Inside `<script>` tags we use [single quotes](https://github.com/ckeditor/ckeditor4/blob/major/dev/docs/codestyle.md#224-quotes) (`'`) instead of double quotes. That is also relevant for other lines like 18, 30, 31 etc.

tests/plugins/wysiwygarea/manual/detached.html
```diff
@@ -0,0 +1,48 @@
...
+
+
+	var e = CKEDITOR.replace( textArea, {
+		on: {
+			instanceReady:function(evt){}
+		}
+	});
+	e.setData("<p>ole!</p>");
+
+	var btn = CKEDITOR.document.getById("toggleButton1");
```
Dumluregn, line 30: Maybe more distinguishable name like `toggleButton` or `toggleBtn`?

tests/plugins/wysiwygarea/manual/detached.html
```diff
@@ -0,0 +1,48 @@
...
+	var setBtn = CKEDITOR.document.getById("setButton");
```
Dumluregn, line 31: Please join `var` statements.

tests/plugins/wysiwygarea/manual/detached.html
Dumluregn: We are also missing the spaces after starting bracket and before closing one. Please fix it throughout the whole file.

tests/plugins/wysiwygarea/manual/detached.html
```diff
@@ -0,0 +1,48 @@
...
+
+<script>
+	var editorElement = document.getElementById("editor");
+
+	var textArea = document.createElement("textarea");
+
+	editorElement.appendChild(textArea);
+
+
+	var e = CKEDITOR.replace( textArea, {
```
Dumluregn, line 23: If possible, meaningful variable names are easier to understand when they are referred to later, so `editor` > `e`.

tests/plugins/wysiwygarea/manual/detached.html
```diff
@@ -0,0 +1,48 @@
...
+	e.setData("<p>ole!</p>");
+
+	var btn = CKEDITOR.document.getById("toggleButton1");
+	var setBtn = CKEDITOR.document.getById("setButton");
+
+	setBtn.on ('click', function() {
+		e.setData('any any any');
+	});
+
+	var t = false;
```
Dumluregn, line 37: Is that possible to give this variable a meaningful name for better intuition? Maybe `editorDetached`?

tests/plugins/wysiwygarea/manual/detached.html
```diff
@@ -0,0 +1,48 @@
...
+	btn.on( 'click', function() {
+		if(t){
+			document.getElementById("fakedId").appendChild(editorElement);
+		}
+		else {
+			editorElement.parentNode.removeChild(editorElement);
```
Dumluregn, line 43: I suppose you can also just [`editorElement.remove()`](https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_dom_element.html#method-remove) 🙂

tests/plugins/wysiwygarea/manual/detached.md
```diff
@@ -0,0 +1,23 @@
+@bender-tags: 4.16.0, bug, 4462, wysiwygarea
```
Dumluregn, line 1: Please bump version tags to `4.17.0` (🤞🏻 😄 ).

tests/plugins/wysiwygarea/manual/detached.md
```diff
@@ -0,0 +1,23 @@
...
+1. Open dev console
+2. Click "Toggle" button twice. To hide and show editor.
+
+	**Expected:** Editor shows with same data. Content is editable. Elementspath (at the bottom is updated).
+
+	**Unexpected:** Editor data is lost, content area is not editable.
+
+3. Write some text in editor.
+
+4. Repeat step 2.
```
Dumluregn, line 15: In general the `Expected/Unexpected` is not a part of a step (and can be easily omitted if tester isn't very focused 😜 ), so it's better to repeat them every time they should be checked (so after 2, 4, 6 and 8).

Dumluregn (changes requested):
Editor can be detached and reattached with the same content, also undo steps are preserved, also integrations with Angular, React and Vue still work, so 👍🏻 Especially that the fix doesn't look very obvious.

OTOH, in manual test if I detach editor and then click `setData()` button, I've got an error. So I'm wondering - is this expected (if yes then we should rename this issue to indicate that editor can be detached and reattached, but not suggest that it works when detached) or not (and we still have to fix it). I guess either of you @sculpt0r and @f1ames can specify this as you together decided to extract #4462 so please clarify what is the goal of this issue.

Did you encounter some issues trying to write unit tests? I think they would be useful here to easily check some cases with executing commands (TBH manual test could also be automated, but it is also useful as is since you can control more things about editor's behaviour).

As for the manual tests, I'd like to have another one checking if undo steps are preserved (they are now so 👍🏻 ). So the steps would be something like:

1. Type anything.
2. Change formatting.
3. Detach.
4. Reattach.
5. Check if there are two undo steps and you can go back to the beginning.

Besides that - please take a look at inline comments.

Dumluregn:
Some of the comments may be hidden like this:

Screenshot 2021-02-15 at 20 29 23

Please make sure you find them all 🔍

tests/plugins/wysiwygarea/manual/detached.html
sculpt0r: It's the same case. [https://github.com/ckeditor/ckeditor4/pull/4463#discussion_r551806398](https://github.com/ckeditor/ckeditor4/pull/4463#discussion_r551806398) :smile:

sculpt0r:
> OTOH, in manual test if I detach editor and then click `setData()` button, I've got an error.

That's something we know about. Truth is, the original issue should be called: `Support editor functions after reattach to DOM`. So I also renamed #4462 . We talked with @f1ames about supporting editor even when it's detached, but it's a really huge feature. It will require a lot of changes inside the core, which partially operates on `document` search. After detach it could be hard to workaround.   
This PR mainly should fix a few issues mentioned in #4462 .

sculpt0r:
@Dumluregn - CI is still green - so ready for another review :)

Dumluregn:
> We talked with @f1ames about supporting editor even when it's detached, but it's a really huge feature.

Yeah, I can imagine the amount of work to fix all issues with selection etc., so it makes sense :+1:

@sculpt0r could you please also refer to this part of [my comment](https://github.com/ckeditor/ckeditor4/pull/4481#pullrequestreview-590703328):

> Did you encounter some issues trying to write unit tests? I think they would be useful here to easily check some cases with executing commands (TBH manual test could also be automated, but it is also useful as is since you can control more things about editor's behaviour).

tests/plugins/wysiwygarea/manual/detached.html
```diff
@@ -0,0 +1,37 @@
...
+	editor.setData( '<p>ole!</p>' );
+
+	var toggleButton = CKEDITOR.document.getById( 'toggleButton' ),
+		setButton = CKEDITOR.document.getById( 'setButton' );
+
+	setButton.on ( 'click', function() {
+		editor.setData( 'any any any' );
+	});
+
+	var editorDetached = false;
```
Dumluregn, line 25: Now it's clear we could have just one `var` statement for all of those variables.

tests/plugins/wysiwygarea/manual/detached.html
```diff
@@ -0,0 +1,37 @@
...
+		editorContainer = document.getElementById( 'editorContainer' ),
+		editorContainerParent = editorContainer.parentNode;
+
+	var editor = CKEDITOR.replace( editorElement );
+	editor.setData( '<p>ole!</p>' );
+
+	var toggleButton = CKEDITOR.document.getById( 'toggleButton' ),
+		setButton = CKEDITOR.document.getById( 'setButton' );
+
+	setButton.on ( 'click', function() {
```
Dumluregn, line 21:
```suggestion
	setButton.on( 'click', function() {
```

tests/plugins/wysiwygarea/manual/detached.html
```diff
@@ -0,0 +1,37 @@
...
+		editor.setData( 'any any any' );
+	});
```
Dumluregn, line 23:
```suggestion
	} );
```

tests/plugins/wysiwygarea/manual/detached.html
```diff
@@ -0,0 +1,37 @@
...
+
+	var editorDetached = false;
+	toggleButton.on( 'click', function() {
+		if( editorDetached ){
```
Dumluregn, line 27:
```suggestion
		if ( editorDetached ) {
```

plugins/wysiwygarea/plugin.js
```diff
@@ -93,8 +93,44 @@
 						return;
 					}
 
-					editor.editable( new framedWysiwyg( editor, iframe.$.contentWindow.document.body ) );
+					editor.editable( new framedWysiwyg( editor, iframe.getFrameDocument().getBody() ) );
 					editor.setData( editor.getData( 1 ), callback );
+
+					iframe.$.onloadFromSetData = false;
+
+					editor.on( 'mode', backupOnLoad, { iframe: iframe, editor: editor, callback: callback } );
```
Dumluregn, line 101: For the future reference:
At the first glance it looks cumbersome, but I think it has to stay this way 🤔 
If we just call `backupOnLoad()` instead of adding listener, it breaks on toggling source mode.
If we don't pass the context, editor breaks after second toggle.
We could pass `this` instead, but we would need to dig for the other necessary constructs, so it seems pointless.

Dumluregn (changes requested):
The code needs only a little bit of polish, good job 👍🏻  We could use some more testing though.

As for the unit tests, we already talked and agreed that we want to have at least the basic scenarios covered.

As for the manuals, as I've mentioned at the end of [the previous review](https://github.com/ckeditor/ckeditor4/pull/4481#pullrequestreview-590703328), we need another test for `undo` and perhaps some other core integrations (like `elementspath`, formatting, links etc.). Besides that, we need to cover also other editor types (`divarea` and `inline`). I've briefly tested them and they seem to work, but still separate tests for each of them is a must (just copy `detached` test and change editor type - don't add more editors there as it already has a long test description). And also a test with two editors where we reattach one of them to check if the correct data is bound to it.

So to sum up, we want to have the following manual tests:
1. General test for reattaching and preserving data - already covered in `detached` 👍🏻 
2. Same test as 1 for divarea editor.
3. Same test as 1 for inline editor.
4. Test for `undo` integration with classic editor.
5. Test for two editors in the same page and reattaching one of them.
6. Open test (kind of monkey clicking) for the integration of at least part of the editor features.

If you think that some of those scenarios would be a good fit for a unit test instead, feel free to propose it.

plugins/wysiwygarea/plugin.js
```diff
@@ -93,8 +93,44 @@
 						return;
 					}
 
-					editor.editable( new framedWysiwyg( editor, iframe.$.contentWindow.document.body ) );
+					editor.editable( new framedWysiwyg( editor, iframe.getFrameDocument().getBody() ) );
 					editor.setData( editor.getData( 1 ), callback );
+
+					iframe.$.onloadFromSetData = false;
```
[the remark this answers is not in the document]
sculpt0r: > Let's encapsulate the manipulation of this property within a method to operate on the same abstraction level as elsewhere.

I agree. Setting single `flag` properties make code hard to read and understand :+1: 

> I'd see it as a method of a `CKEDITOR.dom.element` (we have already have e.g. [`getFrameElement()`](https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_dom_element.html#method-getFrameDocument) method which is exclusive for iframes, so I think it's not an issue).

I wouldn't go up that much in abstraction. Property `contextWindow` [has some sense](https://html.spec.whatwg.org/multipage/iframe-embed-object.html#dom-object-contentwindow) even for the non-iframe elements. This flag even for `iframe` itself has no meaning. Actually, it only means something in `wysiwygarea` plugin. Meaning context of this flag reaches both: plugin and class extending editable, so I would go with helper function like https://github.com/ckeditor/ckeditor4/blob/9d34608a06a79fdb490d5a3f25972ee9a4a650b5/plugins/wysiwygarea/plugin.js#L597

> So I'd suggest sth similar as for [`editor.editable()`](https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_editor.html#method-editable) method, so if called with no parameters it returns the current value, and if called with an argument (`true` or `false`) it sets the proper value

I'm not a big fan of one method which could be both: getter and setter - but since it's already a kind of convention in existing code, it sounds like a good idea.

> However it's difficult to easily explain it :thinking:WDYT about `shouldReload()` as the name for the method I've proposed above?

Actually, this flag prevents iframe body recreation. Reload itself is called anyway. The previous proposal sounds good, because indeed - it attaches reload to iframe. In this case, the long description would be: "let's save information that `SetData` was called - because we know that it leads to another `onload` invocation". It is hard to find out a good name for one method - since it will be called in three contexts:

*   inside `SetData` : sth similar to SetDataWasCalled( true )
*   inside this so-called `backup on load function`: WasSetDataJustCalledBeforeThisOnload()
*   inside both `onload` functions as a 'clean up' after editable element creation: SetDataWasCalled ( false ).

All those above bullets lead me to sth like: `shouldRecreateEditable()` && `shouldRecreateEditable( true / false )`WDYT?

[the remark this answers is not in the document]
Dumluregn: > Meaning context of this flag reaches both: plugin and class extending editable, so I would go with helper function

Agreed 👍🏻 

> I'm not a big fan of one method which could be both: getter and setter - but since it's already a kind of convention in existing code, it sounds like a good idea.

Me neither, I wouldn't propose it in the new project, so we are on the same page here 😄 

> All those above bullets lead me to sth like: `shouldRecreateEditable( iframe )` && `shouldRecreateEditable( iframe, true / false )` WDYT?

I like it, it's informative and concise at the same time, let's go with it 👍🏻

sculpt0r:
Added requested changes:

*   Manual detached test for `divarea` editor
*   Manual detached test for `inline` editor
*   Manual `undo` test: 78e97e8
*   Manual test for two editor instances (`wysiwygarea` plugin). I'm not sure what exactly we want to test here?
*   Open 'monkey clicking' test with some plugins that might be somehow affected by `editable` recreation. (3f09942)

> Did you encounter some issues trying to write unit tests? I think they would be useful here to easily check some cases with executing commands (TBH manual test could also be automated, but it is also useful as is since you can control more things about editor's behaviour).

I added an automatic test for the detached `wysiwygarea` editor. I have to use some tricky-reference to elements combination. Also, it's impossible to just use casual CKEditor API, because sometimes at API-level all data are correct, but not displayed to the user properly. That is why I go more into manual tests. Also, I have to add this ugly `setTimeout` before `iframe` data verification. If I skip `iframe` data is not ready after recreation. Also, I remember, that at the very beginning: there were situations that the editor contains proper data and all attributes - but is not editable.  
 

*   Fixed all code styles issues (also comments)  
      
    Ready for review.

Dumluregn:
Some random test failed, so I've restarted CI:

tests/core/creators/manual/detachedInline.md
```diff
@@ -0,0 +1,37 @@
+@bender-tags: 4.17.0, bug, 4462
+@bender-ui: collapsed
+@bender-ckeditor-plugins: toolbar,link,format,undo,floatingspace,sourcedialog
+
+
+1. Open dev console
+
+2. Click "Toggle" button twice. To hide and show editor.
+
+	**Expected:** Editor shows with same data. Content is editable. Elementspath (at the bottom is updated).
```
Dumluregn, line 10: `Elementspath` plugin is disabled for inline editor, so it should be removed from all `expected` clauses in this test.

tests/core/creators/manual/detachedInline.md
```diff
@@ -0,0 +1,37 @@
+@bender-tags: 4.17.0, bug, 4462
+@bender-ui: collapsed
+@bender-ckeditor-plugins: toolbar,link,format,undo,floatingspace,sourcedialog
+
+
+1. Open dev console
```
Dumluregn, line 6: We ask tester to open console, but don't verify it at any point. We should either move this as a note at the beginning of the test with annotation to watch console for errors, or add it as a check in all `expected` clauses.

Same for all other manual tests.

tests/plugins/divarea/manual/detached.md
```diff
@@ -0,0 +1,37 @@
+@bender-tags: 4.17.0, bug, 4462, divarea
+@bender-ui: collapsed
+@bender-ckeditor-plugins: divarea,toolbar,basicstyles,link,format,sourcearea,elementspath,undo
+
+
+1. Open dev console
+
+2. Click "Toggle" button twice. To hide and show editor.
+
+	**Expected:** Editor shows with same data. Content is editable. Elementspath (at the bottom is updated).
```
Dumluregn, line 10: When we have 3 or more things to check it's more readable in form of bullet points:
```suggestion
  **Expected:**

  * Editor shows with same data.
  * Content is editable.
  * Elementspath (at the bottom is updated).
```

Also I recommend using double spaces (`  `) instead of tabs as they allow for better control over the indentation afterwards.

Same for all other manual tests.

tests/plugins/divarea/manual/detached.md
```diff
@@ -0,0 +1,37 @@
...
+
+5. Click "Set data" button.
+
+6. Repeat step 2.
+
+	**Expected:** Editor shows with same data. Content is editable. Elementspath (at the bottom is updated).
+
+	**Unexpected:** Editor data is lost, content area is not editable.
+
+7. Switch to "Source mode". Write some additional text. Back from "Source mode".
```
Dumluregn, line 30: Test steps should be atomic - only one thing done in each one. Then they are easier to follow and more difficult to mess up. So here we have actually three steps:

1. Switch to source mode.
2. Write text.
3. Switch back.

Please divide it into smaller chunks.

tests/plugins/wysiwygarea/manual/detachedUndo.md
```diff
@@ -0,0 +1,14 @@
...
+@bender-ckeditor-plugins: wysiwygarea,toolbar,basicstyles,link,format,sourcearea,elementspath,undo
+
+
+1. Type anything in the editor.
+
+2. Change text formatting. For example to 'Heading 1'.
+
+3. Click Toggle button twice to detach and reattach editor.
+
+**Expected** There are two undo steps.
```
Dumluregn, line 12:
```suggestion
**Expected:** There are two undo steps.
```

tests/plugins/wysiwygarea/manual/detachedUndo.md
```diff
@@ -0,0 +1,14 @@
...
+
+**Unexpected** Number of undo steps is different than two.
```
Dumluregn, line 14:
```suggestion
**Unexpected:** Number of undo steps is different than two.
```

tests/plugins/divarea/manual/detached.md
```diff
@@ -0,0 +1,37 @@
...
+
+	**Expected:** Editor shows with same data. Content is editable. Elementspath (at the bottom is updated).
+
+	**Unexpected:** Editor data is lost, content area is not editable.
+
+3. Write some text in editor.
+
+4. Repeat step 2.
+
+	**Expected:** Editor shows with same data. Content is editable. Elementspath (at the bottom is updated).
```
Dumluregn, line 18: If you follow the test carefully, Elementspath will be blank at this point - it will be updated once you focus an editor. I wonder - was there any issue with this plugin during development? Or is there another reason why you wanted to include this check?

tests/plugins/divarea/manual/detached.md
```diff
@@ -0,0 +1,37 @@
+@bender-tags: 4.17.0, bug, 4462, divarea
```
Dumluregn, line 1: We use only the first three tags: version, issue type and issue number. It happened in the past we were adding here some other info like plugin name (I did it myself 🙈 ), but it doesn't bring any more value since the tests are already grouped by plugins. The only situation I can think of where it could be useful are integration tests between different plugins, but this is not the case here.

Same for all other tests.

plugins/wysiwygarea/plugin.js
```diff
@@ -522,6 +563,9 @@
 			},
 
 			detach: function() {
+				if ( this.saveIframe ) {
+					return;
+				}
```
Dumluregn, line 568:
```suggestion
				}

```

plugins/wysiwygarea/plugin.js
```diff
@@ -93,8 +93,43 @@
...
+							return;
+						}
+
+						var cacheData = editor.getData( false );
+
+						// Remove current editable, but preserve iframe.
+						editor.editable().saveIframe = true;
+						editor.editable( null );
+
+						var newEditable = new framedWysiwyg( editor, iframe.getFrameDocument().getBody() );
```
Dumluregn, line 123: According to [style guide](https://github.com/ckeditor/ckeditor4/blob/major/dev/docs/codestyle.md#2222-one-variable-per-scope) we should have here:

```suggestion
						var cacheData = editor.getData( false ),
							newEditable;

						// Remove current editable, but preserve iframe.
						editor.editable().saveIframe = true;
						editor.editable( null );

						newEditable = new framedWysiwyg( editor, iframe.getFrameDocument().getBody() );
```

I know it's an additional line of code, but we promote readability and consistency.

Dumluregn (changes requested):
We are closing to the end 👍🏻 

> Manual test for two editor instances (wysiwygarea plugin). I'm not sure what exactly we want to test here?

I'd like to check if with two editors the one we reattach has correct data and history (undo steps). So to test it we should do some different changes in both editors, then detach and reattach one and see if it has correct data.

Besides the inline comments - one general note: test files names should always be lowercase (to avoid some Bender issues).

tests/plugins/divarea/manual/detached.md
sculpt0r: I split it into three separate steps.

I've checked at the beginning other manual tests and they also use tabs. Also, find in our code style guide (but in notion)  
https://www.notion.so/Code-Style-cb28fa1931bf48b6988ea3bb3990f566#393baa32179d49ef9ad4f3be923afe85

sculpt0r: So we decide (@Dumluregn & @f1ames) to allow spaces in bender tests. It's the only way to properly format test descriptions in bender.

tests/plugins/divarea/manual/detached.md
sculpt0r: I've changed the test description to be more precise.  
  
This plugin was my first indicator that something is not working. At the very beginning when I restore the data - it appears properly in the content area. However, the bottom bar has remained blank. Even if the content was editable.

That directed me to use build-in mechanics and flows as in the original editor creation flow. There is a lot of bindings to many events. So it will be hard to manually recreate all of them. It was better to somehow repeat all bindings automatically.  
  
So editable content and working `elementspath` plugin were my mine manual indicators for this solution. Then unit tests (mostly jQuery adapters) indicate if my working solution remains stable with other parts of the editor.

tests/plugins/wysiwygarea/detached.js
```diff
@@ -0,0 +1,31 @@
...
+			}, function( bot ) {
+				var editorContainer = bot.editor.container,
+					editorContainerParent = editorContainer.getParent();
+
+				editorContainer.remove();
+				editorContainerParent.append( editorContainer );
+
+				var iframeElement = bot.editor.ui.space( 'contents' ).findOne( 'iframe' );
+
+				CKEDITOR.tools.setTimeout( function() {
```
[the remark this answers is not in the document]
sculpt0r: I tried to find any, but I couldn't ;(  
In the end, it may end with hacking each browser event call. Similar problem I have during development, but fortunately was able to find one :)  
  
This time, it's less messy to just wait a little bit.

sculpt0r:
@Dumluregn 

Ready for another review.  
  
I changed manual test descriptions a lot. IE failing test now is passing.

tests/plugins/wysiwygarea/manual/detachedundo.md
```diff
@@ -0,0 +1,14 @@
...
+@bender-ckeditor-plugins: wysiwygarea,toolbar,basicstyles,link,format,sourcearea,elementspath,undo
+
+
+1. Type anything in the editor.
+
+2. Change text formatting. For example to 'Heading 1'.
+
+3. Reattach editor with double click on "Toggle" button.
+
+**Expected:** There are two undo steps.
```
Dumluregn, line 12: It is possible to have more undo steps created in step 1. One step with changing formatting would be more error-proof.

tests/plugins/wysiwygarea/manual/detachedwithplugins.html
```diff
@@ -0,0 +1,42 @@
...
+<script>
+	embedTools.delayJsonp();
+
+	var editorElement = document.getElementById( 'editor' ),
+		editorContainer = document.getElementById( 'editorContainer' ),
+		editorContainerParent = editorContainer.parentNode,
+		editor = CKEDITOR.replace( editorElement, {
+			mathJaxLib: 'https://cdnjs.cloudflare.com/ajax/libs/mathjax/2.7.4/MathJax.js?config=TeX-AMS_HTML',
+			embed_provider: '//ckeditor.iframe.ly/api/oembed?url={url}&callback={callback}',
+			extraPlugins: 'autoembed'
```
Dumluregn, line 19: This plugin is already included in bender tags, so no need to add it here as well.

tests/plugins/wysiwygarea/manual/detachedwithplugins.md
```diff
@@ -0,0 +1,12 @@
+@bender-tags: 4.17.0, bug, 4462
+@bender-ui: collapsed
+@bender-include: ../../embedbase/_helpers/tools.js
+@bender-ckeditor-plugins: wysiwygarea,toolbar,basicstyles,link,format,sourcearea,elementspath,undo,image,editorplaceholder,about,resize,showblocks,docprops,dialog,colorbutton,mathjax,autolink,embed,autoembed
+
+**Note:** Open dev console to track eventual errors.
+
+1. Use each available plugin and toggle editor in random moments.
```
Dumluregn, line 8:
```suggestion
1. Use each available plugin and use toggle button to detach and reattach editor in random moments.
```

Dumluregn (changes requested):
Just some small corrections left (mostly in tests).

tests/plugins/wysiwygarea/manual/detachedtwoeditors.md
```diff
@@ -0,0 +1,25 @@
+@bender-tags: 4.17.0, bug, 4462
```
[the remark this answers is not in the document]
sculpt0r: Och! I'm not sure why this would happen, but I changed the scenario - so both editors have different content and edit history.

tests/plugins/divarea/manual/detached.md
```diff
@@ -0,0 +1,63 @@
...
+
+	**Expected:** Editor shows with the same data.
+
+	**Unexpected:** Editor data is lost.
+
+5. Click inside the content editing area.
+
+	**Expected:** Content is editable. Elementspath (bottom bar) updates.
+
+	**Unexpected:** Content is not editable. Elemetspath (bottom bar) remains empty.
```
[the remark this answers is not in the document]
sculpt0r: I'm not sure about this one. Generally speaking - if the content is not editable - verifying `elementspath` is pointless.

With this PR fix, `elementspath` indeed remains the same after reattach. There was a time during development - that this simply doesn't show anything even if the content was visible and editable. So, if after reattach & focus inside the editor it remains the same - that indicates that the plugin is working. Also, this scenario is quite long already - so maybe we could stay as it is?

sculpt0r:
Ready for another review - @Dumluregn 

  
I've rebased it on to the newest major.  
I corrected all your comments, besides one. Please take a look at https://github.com/ckeditor/ckeditor4/pull/4481#discussion_r586237435

tests/plugins/divarea/manual/detached.md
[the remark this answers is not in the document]
Dumluregn: Ok, in general I don't have a problem with checking elementspath at all - I think we just need to change the word `updates` as it may suggest that something should change. So instead let's just say that it shows correct path or sth like this.

tests/plugins/wysiwygarea/manual/detachedtwoeditors.md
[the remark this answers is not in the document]
Dumluregn: > I'm not sure why would this happen

We never know 😄  I thought about the situation where the detached instance of editor is in fact removed (or just lost reference to it) and after reattach we have the other existing editor duplicated instead.

Dumluregn (changes requested):
We are close, but not home yet as I thought before 🙁 

1. CI turned red - but it's also red on `major`. Caused by https://github.com/ckeditor/ckeditor4/issues/3868.
2. I've tested integration with all features and found one bug: undo steps are preserved, but redo - not. It's not a catastrophe, but please look into it and try to fix it (if it turns out complicated I think we can extract it as a follow-up).
3. Unit tests on IE are bleeding on my VMs - mostly because of `Permission denied` errors. They seem to occur a bit randomly (I had a lot errors in the first run that don't show up in the rerun), which is not a great news. Please check if you reproduce this behaviour and fix it as this is an actual blocker for this PR. Below I'm listing some tests that were problematic:

IE8:
```
tests/plugins/tabletools/allowedcontent
tests/tickets/4527/1
```

IE9:
```
tests/core/editable/inserthtmlstripped
tests/plugins/autocomplete/view
tests/plugins/colorbutton/customcolorsasync
tests/plugins/image2/image2
tests/plugins/mathjax/mathjax
tests/plugins/removeformat/plugin
tests/plugins/tabletools/allowedcontent
tests/tickets/4527/1
```

IE10 (stopped in a half):
```
tests/plugins/colorbutton/ariaattributes
tests/plugins/colorbutton/colorbutton
tests/plugins/colorbutton/contentcolors
```

IE11 - seems to be fine 🤔

Dumluregn:
Also please note I've added a commit, so be sure to pull the latest version.

sculpt0r:
I fix `redo` history. CI seems to be fine (besides those failing tests due to another issue).  
TBH I try to do sth with those IE 'permission denied' but couldn't figure out anything constructive. I've run tests on BrowserStack and my local VM's and actually couldn't receive 'permission denied'... I even try to make separate test just to force IE to this issue (loading lot's of iframes and trying to access their body) - but it gives me this error maybe two times in the beginning at then nothing. I try to run clean VM - and nothing. I must say, that there were a lot of random tests failing with status 'BROKEN' - but after rerun - they were fine :thinking: We have #3419 for this IE / Edge weird behaviour. If this is 'almost' finish. We could give this to final confirmation to @ckeditor/qa-team ?  
  
  
@Dumluregn it's ready for another review. We talked about this IE thing but there is no 'one-place' to simply put there `try...catch` So I leave it as it is.

Dumluregn (changes requested):
The tests are still failing for me 😞 
The biggest problem is with the following two of them:
IE8
They fail because of the syntax problem, so rerunning them doesn't help. Please @sculpt0r check what may be the reason and if the code may be improved to pass them. 

But besides that, on IE8, 9 and 10 I still have a lot of randomly failing tests because of "Permission denied" error (even after multiple reruns, no matter if I run them on my local VMs or BS) which were fine before this PR. However since @sculpt0r [isn't able to reproduce them](https://github.com/ckeditor/ckeditor4/pull/4481#issuecomment-793912837), I have to ask @f1ames for another pair of eyes - could you also run the tests on IEs and check if they work for you?

We discussed before that since this feature is mostly needed in our integrations, maybe we could disable it for IE<11 (and this is another question for @f1ames), but as I've mentioned before right now for me there are failing tests even on IE11 so we have to fix them first.

Other than that - just one small note.

sculpt0r:
@Dumluregn 

*   Fixed both tests (`4527/1` and `tabletools/allowedcontent`). First one use `/` in editor id, and the second one `:`and `,`. This line [https://github.com/ckeditor/ckeditor4/blob/5c7a433b6a38d2858d10949b328f7ce738b1028c/plugins/wysiwygarea/plugin.js#L515](https://github.com/ckeditor/ckeditor4/blob/5c7a433b6a38d2858d10949b328f7ce738b1028c/plugins/wysiwygarea/plugin.js#L515) force `querySelector` to be used on the native element. With those characters, it simply failed on IE. On IE8 it was just `Invalid Character` but IE11 shows Syntax Error... I remove the entire custom editor name from the first test and replace both characters with `_` in the second one
*   Added editor status definition

Dumluregn (approved):
PR looks good :+1:

@ckeditor/qa-team please take a look to confirm we are in the right spot here.

LukaszGudel:
This feature is looking good :+1: Everything seems to be working correctly.

One note tho, during some monkey clicking I somehow broke undo feature and there were some errors in the console but I could not find steps to recreate this. I do not have those errors copied 😞  Please, keep in mind to test undo with this feature during the testing phase.

Dumluregn:
Rebased onto latest `major`. Also I've added the note to observe the `undo` plugin to the manual test with plugins.

Dumluregn (changes requested):
@sculpt0r I hate to say it but... we still have work to do here...

I tried to reproduce the errors [mentioned](https://github.com/ckeditor/ckeditor4/pull/4481#issuecomment-802806541) by @LukaszGudel and I didn't, but found more two issues during the monkey clicking.

### Scenario 1)

1. Embed an image (using dialog or autoembed).
2. Toggle source mode **twice**.

The whole webpage freezes. And as this happens even without detaching the editor, we have to fix it.

### Scenario 2)

1. Insert image to editor.
2. Detach and reattach editor.

This creates one dummy undo step which doesn't change the content. I would say it isn't something critical and only happens after toggling the editor, so if the fix would take more than 1/2 MD then let's extract this as a separate issue.

sculpt0r:
> ### **Scenario 1)**

Looks like `fullPage` config set to `true` causing problems with recreating :(

sculpt0r:
Closing due to:

*   long unnecesary commits history
*   change resolution to MutationObserver

This solution works in IE \< 11, but was error-prone (due to relying on `load` events mixed with lots of flags). This feature is mainly required by integrations, so support IE11 is enough for that. The PR with new solution: #4601

Issue #4052, opened 2020-05-13
Accessibility: Editor Control Type Incorrect


lumills:
## Type of report

Bug

## Provide detailed reproduction steps (if any)
1) Open a ckeditor 4 instance
2) Turn screen reader on (NVDA works)
3) Navigate into the Editor itself
4) Control type should be announced

### Expected result

Control type should be meaningful to the user, such as 'Editor'

### Actual result

Control type is not informational (I believe it is announced as 'group')

## Other details

* Browser: Firefox
* CKEditor version: 4

f1ames:
Hi @lumills,

does it only concerns NVDA or any other screen reader too?

lumills:
Hi @f1ames ,

Bug and repro that I'm seeing appears to be NVDA specifically

Comandeer:
I can confirm that editor is presented in unhelpful way. It seems to be connected with a convoluted way in which we label the editor – see #2445, #2499 and #1904.

lumills:
Hey @f1ames and @Comandeer just wanted to check if there is any sort of eta here / confirmation that it is in the pipeline? Thanks in advance

f1ames:
Hello @lumills, unfortunately we have no ETA for this one. We have briefly analyzed this issue and related ones (see https://github.com/ckeditor/ckeditor4/issues/4052#issuecomment-632057850). This requires a solid rework on our side regarding how editor markup is created and so it's really hard to estimate how and when we can have a working solution.

Still, we are keeping an eye on it as accessibility is very important for us so there is a chance there will be a fix in a foreseeable future.

naksenzh:
Hello @f1ames, we are tracking this issue as a blocker for our accessibility bugs. We would appreciate a lot if you could share the updates, please.

f1ames:
Hello @naksenzh, as mentioned in https://github.com/ckeditor/ckeditor4/issues/4052#issuecomment-686342038, we have no clear ETA for this issue since it requires a significant amount of work in the editor itself. Any visible progress we will make on this issue (whatever positive or negative) will be posted here so you can track the status.

naksenzh:
Hi @f1ames uny update or progress on this by chance ?

f1ames:
Hello @naksenzh, unfortunately due to other priorities we were unable to push this issue forward for now.

CKEditorBot:
Closed in https://github.com/ckeditor/ckeditor4/pull/5146

PR #4535, opened 2021-02-15
Improve accessibility of the autocomplete plugin

### LINKED ISSUE #2199: Improve accessibility of the autocomplete plugin
## Type of report

Bug

Autocomplete accessibility markup should be reviewed. Items should gain proper roles etc.

sculpt0r:

### This PR contains

- [ ] Unit tests
- [ ] Manual tests

- [ ] PR is consistent with the code style guide

```
* [#<ISSUE_NUMBER>](https://github.com/ckeditor/ckeditor4/issues/<ISSUE_NUMBER>): Describe the purpose of the pull request in a few simple sentences.
```

## What changes did you make?

*Give an overview…*

Closes #2199 .

### DIFF
--- a/plugins/autocomplete/plugin.js
+++ b/plugins/autocomplete/plugin.js
@@ -609,7 +609,7 @@
 		 * @readonly
 		 * @property {CKEDITOR.template}
 		 */
-		this.itemTemplate = new CKEDITOR.template( '<li data-id="{id}">{name}</li>' );
+		this.itemTemplate = new CKEDITOR.template( '<li id="ac-i-{id}" data-id="{id}" role="option">{name}</li>' );
 
 		/**
 		 * The editor instance.
@@ -663,7 +663,7 @@
 		append: function() {
 			this.document = CKEDITOR.document;
 			this.element = this.createElement();
-
+// this.editor.editable().append(this.element);
 			this.document.getBody().append( this.element );
 		},
 
@@ -717,6 +717,9 @@
 		 */
 		close: function() {
 			this.element.removeClass( 'cke_autocomplete_opened' );
+			this.editor.editable().setAttributes( {
+				'aria-expanded': 'false'
+			} );
 		},
 
 		/**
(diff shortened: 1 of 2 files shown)


plugins/autocomplete/plugin.js
```diff
@@ -727,6 +730,18 @@
...
+				'aria-label': 'autocomp list',
+				'id': 'ac-listbox-list'
+			} );
+			this.editor.editable().setAttributes( {
+				'role': 'combobox',
+				'aria-autocomplete': 'list',
+				'aria-expanded': 'false',
+				'aria-controls': 'ac-listbox-list',
+				'aria-activedescendant': ''
+			} );
```
Comandeer, line 744: I'm afraid that this approach is a dead end as it changes the role of the editor. Now the whole editable area is presented to user as a combobox, which is very counter-intuitive. Editor's role should remain unchanged.

Comandeer (changes requested):
Both VoiceOver and JAWS can't read properly the options inside the listbox. VO ignores them completely and JAWS reads them as "blank".

There is also issue with changing editor's role. According to ARIA's spec, [`[aria-autocomplete]` can be used also with `[role=textbox]` elements](https://w3c.github.io/aria/#aria-autocomplete). That means that there is no need to change editor's role. However it could be blocked by #4052 – as at the moment we have some issues with correctly representing editor in various screen readers.

plugins/autocomplete/plugin.js
sculpt0r: @Comandeer My second thought was to change autocomplete plugin behavior. After the User provide opening characters (like `@` or `[[`) we could add an input to the autocomplete list, and allow to typing continuously filter in this input field. It could look like this:



But this is rather unexpected?

Comandeer: Yes, it seems like an overkill. And, despite all a11y challenges, it would be probably difficult to implement.

Probably adding only `[aria-controls]` and `[aria-autocomplete]` attributes to editable will be enough and issues with that should be resolved in #4052. The more important issue is the fact that screen readers are not able to read options in the listbox.

sculpt0r: I've made some sample [https://github.com/ckeditor/ckeditor4/issues/2199#issuecomment-801450208](https://github.com/ckeditor/ckeditor4/issues/2199#issuecomment-801450208) which shows that `not reading listbox options` is caused by different `iframes`. You can manipulate a little with this sample to put listbox inside / outside of the iframe. It is visible even without screen reader - in a11y tools for browsers. Whenever listbox can't be found (because it is in another iframe than combobox) - combobox element don't have any children (in a11y tree).  
  
We can't have listbox is the same iframe as editor (in `wysiwyg` mode) - because this iframe is actually our content (especially if `fullPage: true`). So we should mark this issue as \`wontfix\` (because we don't want to rewrite the iframe and the entire `wysiwyg` mode) / `upstream`(because it fails due to browser inability to find elements outside the same iframe) ?

Dumluregn:
https://github.com/ckeditor/ckeditor4/issues/2199#issuecomment-809371227

f1ames:
Let's freeze this PR for now due to [https://github.com/ckeditor/ckeditor4/issues/2199#issuecomment-818711934](https://github.com/ckeditor/ckeditor4/issues/2199#issuecomment-818711934) and the fact that we decided to cover only #4617 for now.

PR #4580, opened 2021-03-22
Enhance accessiblity of richcombo label

### LINKED ISSUE #4493: Dropdown accessibility label is not updated when combo value changes
## Type of report

Bug

Our `richcombo` plugin (and all other plugins which use it underneath) does not comply with the accessibility recommendation described in https://www.w3.org/WAI/WCAG21/Understanding/label-in-name.html - _Ensure that elements labelled through their content must have their visible text as part of their accessible name_.

This is caused by the fact that richcombo accessibility label is not updated when combo value changes so it always holds the initial lable.

1. Go to https://codepen.io/f1ames/pen/rNMdgYL?editors=1010 or any editor instance with `Font` plugin.
2. Open browser dev console and find combo label element `<span id="cke_15_label" class="cke_combo_label">Font</span>` (ids may vary).
3. Change `Font` dropdown value.

### Expected result

Combo label element text changes and contains name of the value selected.

### Actual result

Combo label element text does not change and doesn't reflect selected value.

## Other details

* Browser: All (Tested on Chrome)
* OS: All (Tested on Linux)
* CKEditor version: `4.10.0`+ (but it seems it always worked like that)
* Installed CKEditor plugins: `richcombo`

Comandeer:

Bug fix

### This PR contains

- [x] Unit tests
- [x] Manual tests

```
* [#4493](https://github.com/ckeditor/ckeditor4/issues/4493): Fixed: [Rich Combo](https://ckeditor.com/cke4/addon/richcombo) label does not reflects the current value of the combobox.
```

## What changes did you make?

I've introduced a new element containing the current value of the combobox and linked it alongside the label in combobox's `[aria-labelledby]` attribute.

Closes #4493.

### DIFF
--- a/plugins/richcombo/plugin.js
+++ b/plugins/richcombo/plugin.js
@@ -20,7 +20,7 @@ CKEDITOR.plugins.add( 'richcombo', {
 			( CKEDITOR.env.gecko && !CKEDITOR.env.hc ? '' : ' href="javascript:void(\'{titleJs}\')"' ) +
 			' hidefocus="true"' +
 			' role="button"' +
-			' aria-labelledby="{id}_label"' +
+			' aria-labelledby="{id}_label {id}_text"' +
 			' aria-haspopup="listbox"',
 		specialClickHandler = '';
 
--- a/plugins/richcombo/plugin.js
+++ b/plugins/richcombo/plugin.js
@@ -20,7 +20,7 @@ CKEDITOR.plugins.add( 'richcombo', {
 			( CKEDITOR.env.gecko && !CKEDITOR.env.hc ? '' : ' href="javascript:void(\'{titleJs}\')"' ) +
 			' hidefocus="true"' +
 			' role="button"' +
-			' aria-labelledby="{id}_label {id}_value_label"' +
+			' aria-labelledby="{id}_label"' +
 			' aria-haspopup="listbox"',
 		specialClickHandler = '';
 
@@ -45,7 +45,6 @@ CKEDITOR.plugins.add( 'richcombo', {
 		' onfocus="return CKEDITOR.tools.callFunction({focusFn},event);"' +
 		' onclick="' + specialClickHandler + 'CKEDITOR.tools.callFunction({clickFn},this);return false;">' +
 			'<span id="{id}_text" class="cke_combo_text cke_combo_inlinelabel">{label}</span>' +
-			'<span id="{id}_value_label" class="cke_combo_label"></span>' +
 			'<span class="cke_combo_open">' +
 				'<span class="cke_combo_arrow">' +
 				// BLACK DOWN-POINTING TRIANGLE
@@ -320,12 +319,19 @@ CKEDITOR.plugins.add( 'richcombo', {
 					textElement.setText( typeof text != 'undefined' ? text : value );
 				}
 
-				var valueLabel = typeof text != 'undefined' ? text : value,
-					valueLabelElement = this.document.getById( 'cke_' + this.id + '_value_label' ),
-					isEqualToName = valueLabel === this.label;
+				var newLabel = createLabel( typeof text != 'undefined' ? text : value, this.label ),
+					labelElement = this.document.getById( 'cke_' + this.id + '_label' );
 
-				if ( valueLabelElement && !isEqualToName ) {
-					valueLabelElement.setText( valueLabel );
+				if ( labelElement ) {
+					labelElement.setText( newLabel );
+				}
+
+				function createLabel( newLabel, initialLabel ) {
+					if ( newLabel === initialLabel ) {
+						return newLabel;
+					}
+
+					return newLabel + ', ' + initialLabel;
 				}
 			},
 
--- a/plugins/richcombo/plugin.js
+++ b/plugins/richcombo/plugin.js
@@ -20,7 +20,7 @@ CKEDITOR.plugins.add( 'richcombo', {
 			( CKEDITOR.env.gecko && !CKEDITOR.env.hc ? '' : ' href="javascript:void(\'{titleJs}\')"' ) +
 			' hidefocus="true"' +
 			' role="button"' +
-			' aria-labelledby="{id}_label {id}_text"' +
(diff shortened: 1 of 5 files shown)


plugins/richcombo/plugin.js
```diff
@@ -318,6 +319,14 @@ CKEDITOR.plugins.add( 'richcombo', {
 
 					textElement.setText( typeof text != 'undefined' ? text : value );
 				}
+
+				var valueLabel = typeof text != 'undefined' ? text : value,
+					valueLabelElement = this.document.getById( 'cke_' + this.id + '_value_label' ),
```
sculpt0r, line 324: There are now three places that use `_value_label`. From my point of view, it could be useful to create some helper function which returns a properly formatted id for this element. OTOH, it may end with a big amount of functions like this. However, to change behaviour that relays on template values you don't have to search and alter so many places.

Could we add this kind of 'getter' here?

plugins/richcombo/plugin.js
```diff
@@ -45,6 +45,7 @@ CKEDITOR.plugins.add( 'richcombo', {
 		' onfocus="return CKEDITOR.tools.callFunction({focusFn},event);"' +
 		' onclick="' + specialClickHandler + 'CKEDITOR.tools.callFunction({clickFn},this);return false;">' +
 			'<span id="{id}_text" class="cke_combo_text cke_combo_inlinelabel">{label}</span>' +
+			'<span id="{id}_value_label" class="cke_combo_label"></span>' +
```
sculpt0r, line 48: I suppose there is a reason why the currently selected value needs to be in a separate label? Is it some a11y `good practice`? If not, maybe just updating the initial label element could be a better idea? It takes less code in a template.

tests/plugins/richcombo/manual/arialabel.md
```diff
@@ -0,0 +1,28 @@
+@bender-ui: collapsed
+@bender-tags: 4493, 4.16.1, bug
+@bender-ckeditor-plugins: wysiwygarea, stylescombo, toolbar
+
+**Note:** This is the test for screen readers, so open one before testing.
+
+1. Press <kbd>Tab</kbd> key multiple times to move focus to the editor.
+2. Press <kbd>Alt+F10</kbd> to switch focus to the toolbar and navigate to the `Styles` combo.
```
sculpt0r, line 8: Since there is only one combo in the toolbar. We don't need additional navigation in this step. Simple `Alt+F10` keys press put us in a right place.

sculpt0r (changes requested):
The initial issue is about not updating the label element which is still not updating. However it is all about screen readers, so your solution seems to be fine. Especially that I tested it with NVDA & JAWS on FF and it looks (`sounds`) good to me.

However, could you take a look at my other comments? After those explanations, I think it will be good to merge.

plugins/richcombo/plugin.js
Comandeer: Actually the first idea was to add `_text` element as a second element referenced by `[aria-labelledby]` – as it would be much easier than introducing logic for changing `_label` element. Unfortunately, `_text` element contains the initial label if user does not select anything. Due to that I introduced `_value_label` element, creating even the more convoluted solution than the logic for changing `_label` element :see\_no\_evil:.

However I'm thinking about even different approach – using both `[aria-labelledby]` and `[aria-describedby]` to [cover more assistive technology than only screen readers](https://www.sarasoueidan.com/blog/accessible-text-labels/). Yet it needs some testing.

tests/plugins/richcombo/manual/arialabel.md
Comandeer: That's true for tests run on dev version of the editor, but on built version there will be more buttons. That's why I included more verbose instructions.

Comandeer:
I've modified the whole logic. Now the original label element is modified to include the value of the combo.

I've also modified the order of label to "\<selected value>, \<combo name>" instead of "\<combo name> \<selected value>".

plugins/richcombo/plugin.js
sculpt0r: I assume we stay with `[aria-labelledby]` only :)

tests/plugins/richcombo/richcombo.js
```diff
@@ -89,6 +93,62 @@ bender.test( {
...
+			expectedLabel = 'ONE, ' + initialLabel;
+
+		combo.createPanel( editor );
+		combo.setValue( 'one', 'ONE' );
+
+		assert.areEqual( expectedLabel, comboBtnLabel.getHtml() );
+	},
+
+	// (#4493)
+	'test richcombo label after returning to defaul value': function() {
```
sculpt0r, line 138: I would improve this test case name in the same way as the previous one. This one could be something like `test richcombo label contains default label after richcombo returs to default value`

tests/plugins/richcombo/richcombo.js
```diff
@@ -89,6 +93,62 @@ bender.test( {
...
+			comboBtn = CKEDITOR.document.findOne( comboBtnSelector + ' .cke_combo_button' ),
+			comboBtnLabelElementId = comboBtnSelector.substr( 1 ) + '_label';
+
+		combo.createPanel( editor );
+
+		assert.areEqual( comboBtnLabelElementId, comboBtn.getAttribute( 'aria-labelledby' ), 'Aria-labelledby attribute should point to element with the combo label.' );
+	},
+
+	// (#4493)
+	'test richcombo initial label': function() {
```
sculpt0r, line 110: Could we improve this test case name, so that it will express what we are expecting? I mean something like: `test created richcombo label contains initial label value`

sculpt0r (changes requested):
I've run unit tests for Edge, All IE, FF, Chrome, Safari, and all passed - so :+1:
 
My JAWS with FF is passing the manual test - also: :+1: 

I have only some suggestions for tests cases names. After that - I think we will be good to go :)

Comandeer:
I've updated tests names.

sculpt0r:
LGTM :tada:

sculpt0r (approved):

PR #4591, opened 2021-03-25
Fix for incorrect encoding of RGBA/HSLA colors in color dialog

### LINKED ISSUE #4351: RGBA/HSLA colors are incorrectly encoded
## Type of report

Bug

1. Open https://ckeditor.com/latest/samples
1. Select some text.
1. Open colordialog and add `hsla(150,50%,52%,0.2)` color.
1. Open colordialog again.

### Expected result

Color should be converted to the correct hashcode including the alpha channel. Note that on the HTML level we are still keeping the original value, so this conversion shouldn't introduce a regression.

It would be good to find out why we are actually unifying colors to hashcode and if it's not really needed, we could instead use the original color value for every color including rgb/hsl.

### Actual result

Color is converted to `#rgba(71, 194, 133, 0.2)`

## Other

The issue is also reproducible with rgba colors.

Comandeer:

Bug fix

### This PR contains

- [x] Unit tests
- [x] Manual tests

```
* [#4351](https://github.com/ckeditor/ckeditor4/issues/4351): Fixed: Incorrect values for RGBA/HSLA colors in [Color Dialog](https://ckeditor.com/cke4/addon/colordialog).
```

## What changes did you make?

I've unified functions used for color normalization in `colorbutton` plugin and added special handling for RGBA and HLSA colors.

Closes #4351.


Comandeer:
Rebased onto latest `master`.

tests/plugins/colordialog/_helpers/tools.js
```diff
@@ -15,15 +15,44 @@ function assertColor( editor, inputColor, outputColor ) {
...
+
+			dialog.setValueOf( 'picker', 'selectedColor', options.inputColor );
+			dialog.getButton( 'ok' ).click();
+
+			assertValueFromDialog( editor, options );
+		} );
+	} );
+
+	toolbarButton.click( editor );
+	openColorDialog( toolbarButton );
```
Dumluregn, line 35: We could replace some of one-liner functions here with helpers to keep everything on the same level of abstraction - we'd have more readable code like this: 

```suggestion
	openColorPanel( toolbarButton, editor );
	openColorDialog( toolbarButton );
	
	function openColorPanel( button, editor ) {
		button.click( editor );
	}
```

We could do the same with e.g. `dialog.setValueOf( 'picker', 'selectedColor', options.inputColor );` -> `setColorInDialog( options.inputColor );` etc. But I'm not sure if creating a bunch of one-liners for such a purpose is justified, and the need of adding `editor` as a second argument also discourages me. WDYT?

tests/plugins/colordialog/colordialog.js
```diff
@@ -66,8 +66,49 @@
...
 			openDialogManually( this.editor, '', '<h1>[<span style="background-color:#0000ff">Foo</span>bar]</h1>', 'BGColor' );
-		}
+		},
+
+		// (#4351)
+		'test converting RGBA to hex color': function() {
+			// IE8 doesn't set unsupported values.
+			if ( CKEDITOR.env.ie && CKEDITOR.env.version < 9 ) {
+				assert.ignore();
+			}
```
Dumluregn, line 76: Bender has a rarely used feature of [`_should: { ignore {} }`](https://github.com/cksource/ckeditor4-security-tests/blob/master/plugins/colorbutton/colorvalidation.js#L12). It is not super-useful because when you look at test itself you don't know if it's ignored or not, but since we have a few tests ignored for the same reason here, maybe we could use it here and only describe the reason one?

tests/plugins/colordialog/colordialog.js
```diff
@@ -66,8 +66,49 @@
...
+
+			assertSettingAndGettingColor( this.editor, {
+				inputColor: 'hsla(150,50%,52%,0.2)',
+				expectedColor: '#daf3e7',
+				button: 'TextColor'
+			} );
+		}
```
Dumluregn, line 111: I would also test the transparent color code case described by you in the comment:
```js
// Browsers tend to represent transparent color as rgba(0, 0, 0, 0), so we need to filter out this value.
```

tests/plugins/colordialog/manual/colorconvertion.md
```diff
@@ -0,0 +1,22 @@
+@bender-tags: bug, 4.16.1, 4351
+@bender-ui: collapsed
+@bender-ckeditor-plugins: wysiwygarea, toolbar, colorbutton, colordialog
+
+1. Select some text in the editor.
+2. Click "Text Color" button in the toolbar and choose "More Colors" option.
+3. Insert into color text field `hsla(150,50%,52%,0.2)` value.
+4. Confirm the color by clicking "Ok".
+5. Reopen color dialog by repeating steps 1-2.
+6. Check the value of the color text field
```
Dumluregn, line 10: Maybe let's check also if the color is still in the original format after canceling the dialog - so using source mode.

tests/plugins/colordialog/manual/colorconvertion.md
```diff
@@ -0,0 +1,22 @@
...
+	## Expected
+
+	* The value is equal to `#daf3e7`.
+
+	## Unexpected
+
+	* The value is equal to `#rgba(71, 194, 133, 0.2)`.
+7. Repeat the whole procedure for following values (input → expected output):
+
+	* `hsl(150,50%,52%)` → `#47c285`,
```
Dumluregn, line 21: Is it just an antiregression check?

Dumluregn (changes requested):
Fix is working and tests are passing 👍🏻 

I've left mostly some refactoring proposals, please take a look.

plugins/colorbutton/plugin.js
```diff
@@ -806,6 +793,33 @@
...
+
+		// For colors with alpha channel we need to use CKEDITOR.tools.color normalization (#4351).
+		if ( isAlphaColor && !isTransparentColor ) {
+			colorInstance = new CKEDITOR.tools.color( color );
+
+			return CKEDITOR.tools.normalizeHex( colorInstance.getHex() || '' ).replace( /#/g, '' );
+		}
+
+		// Replace 3-character hexadecimal notation with a 6-character hexadecimal notation (#1008).
+		return CKEDITOR.tools.normalizeHex( '#' + CKEDITOR.tools.convertRgbToHex( color || '' ) ).replace( /#/g, '' );
```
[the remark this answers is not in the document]
Comandeer: There are two main issues:

*   no support for hex-like values (which will be fixed by #4599),
*   incorrect handling of transparent values (although not sure why, as the second normalize also normalizes it to hex :thinking: ), breaking handling of "Automatic" value.

That's why, as long as we don't do #4592, we need two normalization flows :disappointed:

tests/plugins/colordialog/_helpers/tools.js
Comandeer: I really like this suggestion, but I have same doubts if we really need it…

But I'll do it anyway :smile:

tests/plugins/colordialog/manual/colorconvertion.md
Comandeer: Strange, I was pretty sure that `hsl` values are also affected. I'll delete that case.

plugins/colorbutton/plugin.js
```diff
@@ -750,7 +737,7 @@
 			},
 
 			normalizeColor: function( color ) {
-				return CKEDITOR.tools.normalizeHex( '#' + CKEDITOR.tools.convertRgbToHex( color || '' ) ).replace( /#/g, '' );
+				return normalizeColor( color );
```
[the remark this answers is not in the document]
Comandeer: That's a good idea.

tests/plugins/colordialog/colordialog.js
Comandeer: TBH it's such an obscure feature that I would stick with ignoring in tests themselves. It conveys more information to the developer.

Comandeer: OTOH ignores in every test add so many clutter that I eventually rewrote it to `_should.ignore`.

Comandeer:
Ok, I've fixed all issues with the test and refactored normalization method (by moving it to `ColorBox.normalizeColo`).

Dumluregn (approved):
Well done 🎉

PR #4601, opened 2021-04-01
Support editor functions after reattach to DOM

### LINKED ISSUE #4462: Support editor functions after reattach to DOM

sculpt0r:

New feature

### This PR contains

- [X] Unit tests
- [X] Manual tests

```
* [#4462](https://github.com/ckeditor/ckeditor4/issues/4462): Support editor functions after reattaching to DOM.
```

## What changes did you make?

- Used `MutationObserver` to detect whenever the editor is attached to DOM again.
- Based on that information, recreate the entire editable section. In the case of IE, I have to wait until `load` event. So only for IE, there is an extra flag used to mark, that editable should be recreated.
- Added extra flag which preserves `iframe` element during editable `detach` process. Used only internally in recreating.
- Added config option (`detachableParent`) which allows pass native DOM object which should be observed for detach / attach changes
- Added new `editor.status` (`recreating`) which prevents null references in recreating the process
- Added manual & unit tests

We won't be able to support IE below `11`, because of `MutationObserver` usage. However, it's mostly for integrations, which doesn't support IE below `11`.

Three units test are passing regardless of this fix:
- 'test reattached editor makes editor data empty with observed invalid DOM object'
- 'test reattached editor parent makes editor data empty with observed invalid DOM object'
- 'test reattached editor parent makes editor data empty with observed detached parent element'

They confirm if iframe reloading still makes the editor empty. Also, they show that an invalid configuration still leads to an empty editor.

Fix two tests that create query strings for `querySelector()` with restricted characters (IE):
- https://github.com/ckeditor/ckeditor4/pull/4601/files#diff-a15eeadc2752af4c0b9dcfc057fb5fe450dfb6cf9a672e4fc9baacaac9036903
- https://github.com/ckeditor/ckeditor4/pull/4601/files#diff-64582745fd35627ba47ea103cea4f90242ccede20a04a22c9c7311393b50dba7

Closes #4462

### DIFF
--- a/core/selection.js
+++ b/core/selection.js
@@ -1260,7 +1260,8 @@
 
 		// Editable element might be absent or editor might not be in a wysiwyg mode.
 		var editable = this.editable();
-		return editable && this.mode == 'wysiwyg' ? new CKEDITOR.dom.selection( editable ) : null;
+
+		return editable && this.mode == 'wysiwyg' && editable.editor.status !== 'recreating' ? new CKEDITOR.dom.selection( editable ) : null;
 	};
 
 	/**
--- a/core/selection.js
+++ b/core/selection.js
@@ -1250,7 +1250,7 @@
 	 * @method
 	 * @member CKEDITOR.editor
 	 * @param {Boolean} forceRealSelection Return real selection, instead of saved or fake one.
-	 * @returns {CKEDITOR.dom.selection} A selection object or null if not available for the moment.
+	 * @returns {CKEDITOR.dom.selection/null} A selection object or null if not available for the moment.
 	 */
 	CKEDITOR.editor.prototype.getSelection = function( forceRealSelection ) {
 
--- a/core/selection.js
+++ b/core/selection.js
@@ -1261,7 +1261,7 @@
 		// Editable element might be absent or editor might not be in a wysiwyg mode.
 		var editable = this.editable();
 
-		return editable && this.mode == 'wysiwyg' && editable.editor.status !== 'recreating' ? new CKEDITOR.dom.selection( editable ) : null;
+		return editable && this.mode == 'wysiwyg' && this.status !== 'recreating' ? new CKEDITOR.dom.selection( editable ) : null;
 	};
 
 	/**
--- a/plugins/wysiwygarea/plugin.js
+++ b/plugins/wysiwygarea/plugin.js
@@ -113,7 +113,7 @@
 								return;
 							}
 
-							var editorSelector = '#' + editor.element.getAttribute( 'id' ),
+							var editorSelector = '#' + editor.container.getId(),
 								editorElement;
 
 							CKEDITOR.tools.array.forEach( mutation.addedNodes, function( node ) {
@@ -121,7 +121,7 @@
 									return;
 								}
 
-								editorElement = node.querySelector( editorSelector );
+								editorElement = node.querySelector( editorSelector ) || node === editor.container.$;
 
 								if ( editorElement ) {
 									if ( CKEDITOR.env.ie ) {
@@ -134,7 +134,7 @@
 						} );
 					} );
 
-					mutObserver.observe( CKEDITOR.config.detachableParent, { childList: true, subtree: true } );
+					mutObserver.observe( editor.config.detachableParent, { childList: true, subtree: true } );
 				}
 
 				function attachIframeReloader( evt ) {
--- a/plugins/wysiwygarea/plugin.js
+++ b/plugins/wysiwygarea/plugin.js
@@ -804,8 +804,8 @@ CKEDITOR.config.disableNativeSpellChecker = true;
  */
 
 /**
- * Document observation starting point for mutation observer. Needed to detect when classic, wysiwygarea
- * editor reattaches to DOM to restore editor's previous state. Native DOM object.
+ * Native DOM element, a document observation starting point for mutation observer. Needed to detect when classic, wysiwygarea
+ * editor reattaches to DOM to restore editor's previous state.
  *
  * To recreate editor `wysiwygarea` iframe after editor was reattached to DOM:
  *
--- a/plugins/wysiwygarea/plugin.js
+++ b/plugins/wysiwygarea/plugin.js
@@ -54,7 +54,7 @@
 				var frameLabel = editor.title,
 					helpLabel = editor.fire( 'ariaEditorHelpLabel', {} ).label,
 					recreateEditable = false,
-					mutObserver;
+					mutationObserver;
 
 				if ( frameLabel ) {
 					if ( CKEDITOR.env.ie && helpLabel )
@@ -78,7 +78,7 @@
 						desc.remove();
 					}
 
-					mutObserver.disconnect();
+					mutationObserver.disconnect();
 				} );
 
 				iframe.setAttributes( {
@@ -114,34 +114,37 @@
 				}
 
 				function observeEditor() {
-					mutObserver = new MutationObserver( function( mutationsList ) {
-						CKEDITOR.tools.array.forEach( mutationsList, function( mutation ) {
-							if ( mutation.type !== 'childList' || mutation.addedNodes.length === 0 ) {
-								return;
-							}
-
-							var editorSelector = '#' + editor.container.getId(),
-								editorElement;
-
-							CKEDITOR.tools.array.forEach( mutation.addedNodes, function( node ) {
-								if ( !node.querySelector ) {
-									return;
-								}
-
-								editorElement = node.querySelector( editorSelector ) || node === editor.container.$;
-
-								if ( editorElement ) {
-									if ( CKEDITOR.env.ie && !CKEDITOR.env.edge ) {
-										recreateEditable = true;
-									} else {
-										recreate();
-									}
-								}
-							} );
-						} );
+					mutationObserver = new MutationObserver( function( mutationsList ) {
+						CKEDITOR.tools.array.forEach( mutationsList, verifyIfAddsNodesWithEditor );
 					} );
 
-					mutObserver.observe( editor.config.detachableParent, { childList: true, subtree: true } );
+					mutationObserver.observe( editor.config.detachableParent, { childList: true, subtree: true } );
+				}
+
+				function verifyIfAddsNodesWithEditor( mutation ) {
+					if ( mutation.type !== 'childList' || mutation.addedNodes.length === 0 ) {
+						return;
+					}
+
+					CKEDITOR.tools.array.forEach( mutation.addedNodes, checkIfAffectsEditor );
+				}
+
+				function checkIfAffectsEditor( node ) {
+					if ( !node.querySelector ) {
+						return;
+					}
+					var editorSelector = '#' + editor.container.getId(),
+						editorElement = node.querySelector( editorSelector ) || node === editor.container.$;
+
+					if ( !editorElement ) {
+						return;
+					}
+
+					if ( CKEDITOR.env.ie && !CKEDITOR.env.edge ) {
+						recreateEditable = true;
+					} else {
+						recreate();
+					}
 				}
 
 				function attachIframeReloader( evt ) {
--- a/plugins/wysiwygarea/plugin.js
+++ b/plugins/wysiwygarea/plugin.js
@@ -803,7 +803,7 @@ CKEDITOR.config.disableNativeSpellChecker = true;
  * Editor `wysiwygarea` iframe will be recreated whenever this element will be reattached to DOM.
  *
  * @since 4.17.0
- * @cfg {Object} [detachableParent=CKEDITOR.document.$]
+ * @cfg {HTMLElement} [detachableParent=CKEDITOR.document.$]
  * @member CKEDITOR.config
  */
 CKEDITOR.config.detachableParent = CKEDITOR.document.$;
--- a/plugins/wysiwygarea/plugin.js
+++ b/plugins/wysiwygarea/plugin.js
@@ -810,28 +810,30 @@ CKEDITOR.config.disableNativeSpellChecker = true;
  */
 
 /**
- * Document observation starting point. Native DOM object.
+ * Document observation starting point for mutation observer. Needed to detect if wysiwygarea editor is reattaching to DOM. Native DOM object.
  *
  * To recreate editor `wysiwygarea` iframe after reattach editor to DOM:
  *
- * * make sure detachable element is placed any nested depth under observable element.
- * * make sure editor is placed any nested depth under detachable element.
+ * * make sure detachable element is nested on any level under observable element.
+ * * make sure editor is nested on any level under detachable element.
  *
  * See overview of hierarchy below (HTML pseudocode):
  *
  * ```
  * <observable>
- *   <detachable> <!-- Any nesting depth -->
- *     <editor></editor> <!-- Any nesting depth -->
- *   </detachable>
+ *   <...>
+ *     <detachable>
+ *       <...>
+ *         <editor></editor>
+ *       </...>
+ *     </detachable>
+ *   <...>
  * </observable>
  * ```
  *
- * By default, the entire document is observed.
+ * By default, the entire document is observed. However, if you know exactly which element is detaching, you can choose its direct parent to increase performance a little.
  *
- * If you know exactly which element is detaching - choose its direct parent.
- *
- * Note, that if you choose element which is detaching. No changes will be detected.
+ * Note, that if you choose element which is detaching, no changes will be detected.
  *
  * @since 4.17.0
  * @cfg {HTMLElement} [observableParent=CKEDITOR.document.$]
--- a/plugins/wysiwygarea/plugin.js
+++ b/plugins/wysiwygarea/plugin.js
@@ -52,7 +52,9 @@
 					iframe.on( 'load', onLoad );
 
 				var frameLabel = editor.title,
-					helpLabel = editor.fire( 'ariaEditorHelpLabel', {} ).label;
+					helpLabel = editor.fire( 'ariaEditorHelpLabel', {} ).label,
+					recreateEditable = false,
+					mutObserver;
 
 				if ( frameLabel ) {
 					if ( CKEDITOR.env.ie && helpLabel )
@@ -72,8 +74,11 @@
 				// Remove the ARIA description.
 				editor.on( 'beforeModeUnload', function( evt ) {
 					evt.removeListener();
-					if ( desc )
+					if ( desc ) {
 						desc.remove();
+					}
+
+					mutObserver.disconnect();
 				} );
 
 				iframe.setAttributes( {
@@ -95,7 +100,69 @@
 
 					editor.editable( new framedWysiwyg( editor, iframe.$.contentWindow.document.body ) );
 					editor.setData( editor.getData( 1 ), callback );
+
+					// (#4462)
+					editor.on( 'mode', attachIframeReloader, { iframe: iframe, editor: editor, callback: callback } );
+					observeEditor();
 				}
+
+				function observeEditor() {
+					mutObserver = new MutationObserver( function( mutationsList ) {
+						CKEDITOR.tools.array.forEach( mutationsList, function( mutation ) {
+							if ( mutation.type !== 'childList' || mutation.addedNodes.length === 0 ) {
+								return;
+							}
+
+							var editorSelector = '#' + editor.element.getAttribute( 'id' ),
+								editorElement;
+
+							CKEDITOR.tools.array.forEach( mutation.addedNodes, function( node ) {
+								if ( !node.querySelector ) {
+									return;
+								}
+
+								editorElement = node.querySelector( editorSelector );
+
+								if ( editorElement ) {
+									if ( CKEDITOR.env.ie ) {
+										recreateEditable = true;
+									} else {
+										recreate();
+									}
+								}
+							} );
+						} );
+					} );
+
+					mutObserver.observe( CKEDITOR.document.$, { childList: true, subtree: true } );
+				}
+
+				function attachIframeReloader( evt ) {
+					evt && evt.removeListener();
+
+					iframe.on( 'load', function() {
+						if ( CKEDITOR.env.ie && recreateEditable ) {
+							recreateEditable = false;
+							recreate();
+						}
+					} );
+				}
+
+				function recreate() {
+					var cacheData = editor.getData( false ),
+						newEditable;
+
+					// Remove current editable, but preserve iframe.
+					editor.editable().preserveIframe = true;
+					editor.editable( null );
+
+					newEditable = new framedWysiwyg( editor, iframe.getFrameDocument().getBody() );
+					editor.editable( newEditable );
+
+					editor.status = 'recreating';
+					editor.setData( cacheData, callback );
+				}
+
 			} );
 		}
 	} );
@@ -333,6 +400,8 @@
 		base: CKEDITOR.editable,
 
 		proto: {
+			preserveIframe: false,
+
 			setData: function( data, isSnapshot ) {
 				var editor = this.editor;
 
@@ -522,6 +591,10 @@
 			},
 
 			detach: function() {
+				if ( this.preserveIframe ) {
+					return;
+				}
+
 				var editor = this.editor,
 					doc = editor.document,
 					iframe = editor.container.findOne( 'iframe.cke_wysiwyg_frame' ),
--- a/plugins/wysiwygarea/plugin.js
+++ b/plugins/wysiwygarea/plugin.js
@@ -105,7 +105,7 @@
 
 					// (#4462)
 					// Skip IE's below version 11. They don't support MutationObserver.
-					if ( !CKEDITOR.env.ie || isIE11 ) {
+					if ( CKEDITOR.env.ie && !isIE11 ) {
 						return;
 					}
 
--- a/plugins/wysiwygarea/plugin.js
+++ b/plugins/wysiwygarea/plugin.js
@@ -114,6 +114,10 @@
 						editor.on( 'mode', attachIframeReloader, { iframe: iframe, editor: editor, callback: callback } );
 					}
 
+					editor.on( 'mode', function() {
+						editor.status = 'ready';
+					} );
+
 					observeEditor();
 				}
 
@@ -175,7 +179,7 @@
 					editor.editable( newEditable );
 
 					editor.status = 'recreating';
-					editor.setData( cacheData, callback );
+					editor.setData( cacheData, { callback:  callback, internal: false, fireSnapshot: false } );
 				}
 
 			} );
--- a/plugins/wysiwygarea/plugin.js
+++ b/plugins/wysiwygarea/plugin.js
@@ -119,7 +119,7 @@
 						CKEDITOR.tools.array.forEach( mutationsList, verifyIfAddsNodesWithEditor );
 					} );
 
-					mutationObserver.observe( editor.config.detachableParent, { childList: true, subtree: true } );
+					mutationObserver.observe( editor.config.observableParent, { childList: true, subtree: true } );
 				}
 
 				function verifyIfAddsNodesWithEditor( mutation ) {
@@ -802,12 +802,31 @@ CKEDITOR.config.disableNativeSpellChecker = true;
  */
 
 /**
- * Editor detachable parent. Native DOM object.
+ * Document observation starting point. Native DOM object.
  *
- * Editor `wysiwygarea` iframe will be recreated whenever this element will be reattached to DOM.
+ * To recreate editor `wysiwygarea` iframe after reattach editor to DOM:
+ *
+ * * make sure detachable element is placed any nested depth under observable element.
+ * * make sure editor is placed any nested depth under detachable element.
+ *
+ * See overview of HTML hierarchy below:
+ *
+ * ```
+ * <observable>
+ *   <detachable> <!-- Any nesting depth -->
+ *     <editor></editor> <!-- Any nesting depth -->
+ *   </detachable>
+ * </observable>
+ * ```
+ *
+ * By default, the entire document is observed.
+ *
+ * If you know exactly which element is detaching - choose its direct parent.
+ *
+ * Note, that if you choose element which is detaching. No changes will be detected.
  *
  * @since 4.17.0
- * @cfg {HTMLElement} [detachableParent=CKEDITOR.document.$]
+ * @cfg {HTMLElement} [observableParent=CKEDITOR.document.$]
  * @member CKEDITOR.config
  */
-CKEDITOR.config.detachableParent = CKEDITOR.document.$;
+CKEDITOR.config.observableParent = CKEDITOR.document.$;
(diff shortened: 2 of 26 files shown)


sculpt0r:
@ckeditor/qa-team Hello there :wave:   
Could you look at this feature again? The previous solution (#4481) was discarded and we came up with something better.

**Note** that it won't work for IE below `11`.

There is a manual test with some additional plugins (`tests/plugins/wysiwygarea/manual/detachedwithplugins`) which may be helpful during some monkey clicking.

LukaszGudel:
I've tested this feature using provided manual test and everything seems to work correctly 👍  I did not find anything broken during my testing.

plugins/wysiwygarea/plugin.js
```diff
@@ -716,3 +796,14 @@ CKEDITOR.config.disableNativeSpellChecker = true;
...
  * @param {CKEDITOR.dom.element} data The element being added.
  */
+
+/**
+ * Editor detachable parent. Native DOM object.
+ *
+ * Editor `wysiwygarea` iframe will be recreated whenever this element will be reattached to DOM.
+ *
+ * @since 4.17.0
+ * @cfg {Object} [detachableParent=CKEDITOR.document.$]
```
Dumluregn, line 806: Well, it is an object in the JS understanding, but it doesn't distinguish itself from non-DOM objects, so I'd put [`HTMLElement`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement) here - it's interface and not exactly a type, but it's much more informative (it's the same case as with `Array` or `RegExp`).

tests/plugins/wysiwygarea/detached.js
```diff
@@ -0,0 +1,260 @@
...
+		return !( CKEDITOR.env.ie && !CKEDITOR.env.edge && CKEDITOR.env.version < 11 );
+	}
+
+	bender.test( {
+		'test reattached editor contains the same data with observed default dom object': function() {
+			if ( !isSupportedEnvironment() ) {
+				assert.ignore();
+			}
+
+			var startupData = '<p>CKEditor4</p>';
```
Dumluregn, line 18: AFAICS this data is the same for each test, so we could define it before and reuse it.

tests/plugins/wysiwygarea/detached.js
```diff
@@ -0,0 +1,260 @@
...
+				var editorContainer = bot.editor.container,
+					editorContainerParent = editorContainer.getParent();
+
+				editorContainer.remove();
+				editorContainerParent.append( editorContainer );
+
+				CKEDITOR.tools.setTimeout( function() {
+					resume( function() {
+						var iframeElement = bot.editor.ui.space( 'contents' ).findOne( 'iframe' ),
+							editorData = iframeElement.getFrameDocument().getBody().getHtml();
```
Dumluregn, line 33: Can't we just use `bot.editor.getData()`?

tests/plugins/wysiwygarea/detached.js
```diff
@@ -0,0 +1,260 @@
...
+				assert.ignore();
+			}
+
+			var startupData = '<p>CKEditor4</p>';
+
+			bender.editorBot.create( {
+				startupData: startupData,
+				name: 'editor' + new Date().getTime(),
+				config: {
+					extraAllowedContent: 'iframe'
```
Dumluregn, line 237: I didn't notice where we use it? Shouldn't we provide custom startupData containing iframe in this test?

Dumluregn: Actually I bet you wanted to use fixture with `editor5` here as it contains `iframe` and is not used anywhere?

Dumluregn (changes requested):
As for the solution, it looks much clearer now with `mutationObserver` (no custom attributes on iframe native element, yay!). It looks to work just like before 👍🏻 

Besides inline comments, I'd like to ask for explicit tests for the cases that were [failing in the previous version of the solution](https://github.com/ckeditor/ckeditor4/pull/4481#pullrequestreview-616530029), so the embed test and an image with undo test (and BTW I see the second one is still failing - could you relate to it since you didn't in the original PR?).

tests/plugins/wysiwygarea/detached.js
sculpt0r: Sadly - we cannot :cry: It involves internal logic and it returns 'valid' data even if the iframe itself reloads and has no content. So with `bot.editor.getData()` tests could pass or fail due to wrong reasons.

For example `test reattached editor contains the same data with observed default dom object` can pass even without the fix. OTOH test that expects empty editor fails, because of `getData()`.

plugins/wysiwygarea/plugin.js
```diff
@@ -93,9 +98,78 @@
 						return;
 					}
 
-					editor.editable( new framedWysiwyg( editor, iframe.$.contentWindow.document.body ) );
+					editor.editable( new framedWysiwyg( editor, iframe.getFrameDocument().getBody() ) );
 					editor.setData( editor.getData( 1 ), callback );
+
+					// (#4462)
+					var isIE11 = CKEDITOR.env.ie && !CKEDITOR.env.edge && CKEDITOR.env.version === 11;
+					if ( isIE11 ) {
```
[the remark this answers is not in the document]
sculpt0r: In IE11 we have to wait until `load` event, before `resume()` will be called. If we don't - there is an error:

https://github.com/ckeditor/ckeditor4/blob/7481c473b85b14f0bbfcd24fb5b368ebbc1fc3f9/core/dom/element.js#L1264-L1267

Which says: `SCRIPT5007: Unable to get property 'setAttribute' of undefined or null reference`. In this case, we try to set `contenteditable=true` on an invalid reference. So `this.$` is invalid and normally it should point to the body of the iframe. But iframe just lost its content, so there is no `body` inside. In other browsers there is a body - it's empty - but the primary document is available.

plugins/wysiwygarea/plugin.js
```diff
@@ -93,9 +98,78 @@
...
+						editor.on( 'mode', attachIframeReloader, { iframe: iframe, editor: editor, callback: callback } );
+					}
+
+					// All non IE browsers or IE11.
+					if ( !( CKEDITOR.env.ie && !CKEDITOR.env.edge ) || ( isIE11 ) ) {
```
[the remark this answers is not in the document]
sculpt0r: I thought about it, but double IE11 check looks not so good to me. But overall - indeed looks like one part of some fix :+1:

tests/plugins/wysiwygarea/detached.js
```diff
@@ -0,0 +1,260 @@
...
+	'use strict';
+
+	function isSupportedEnvironment() {
+		// Ignore all IE versions below 11.
+		return !( CKEDITOR.env.ie && !CKEDITOR.env.edge && CKEDITOR.env.version < 11 );
+	}
+
+	bender.test( {
+		'test reattached editor contains the same data with observed default dom object': function() {
+			if ( !isSupportedEnvironment() ) {
```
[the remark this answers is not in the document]
sculpt0r: :thinking: Verify environment explicit in each test looks quite redundant, OTOH - you don't have to scroll all of the tests to find if one of them is ignored. And - its hard to compare those long tests titles (at least for me ;p). So I will stay with the current solution.

tests/plugins/wysiwygarea/manual/detachedwithplugins.html
```diff
@@ -0,0 +1,43 @@
...
+			mathJaxLib: 'https://cdnjs.cloudflare.com/ajax/libs/mathjax/2.7.4/MathJax.js?config=TeX-AMS_HTML',
+			embed_provider: '//ckeditor.iframe.ly/api/oembed?url={url}&callback={callback}',
+			fullPage: true
+		} ),
+		toggleButton = CKEDITOR.document.getById( 'toggleButton' ),
+		setButton = CKEDITOR.document.getById( 'setButton' ),
+		editorDetached = false;
+
+	setButton.on( 'click', function() {
+		// debugger
```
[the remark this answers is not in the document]
sculpt0r: :see_no_evil:

plugins/wysiwygarea/plugin.js
```diff
@@ -716,3 +796,14 @@ CKEDITOR.config.disableNativeSpellChecker = true;
  * @param {CKEDITOR.editor} editor This editor instance.
  * @param {CKEDITOR.dom.element} data The element being added.
  */
+
+/**
+ * Editor detachable parent. Native DOM object.
+ *
+ * Editor `wysiwygarea` iframe will be recreated whenever this element will be reattached to DOM.
```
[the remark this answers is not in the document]
sculpt0r: I change the name to `observableParent`. Because we can't observe element which is detachable. If so, then no changes will be detected. We have to choose at least the parent of the detachable element.

sculpt0r:
@Dumluregn - ready for review.

- added missing tests from the previous PR
- correct inline comments

I try to fix this additional step in history - but without success... TBH I was able to fix it, but at the same time `undo` was broken if I toggle editor before insert image - so still wrong behaviour - even worse. As you write - could we extract it as another issue? But in this case, we will have one failing manual test.

plugins/wysiwygarea/plugin.js
```diff
@@ -93,9 +98,82 @@
...
+				function attachIframeReloader( evt ) {
+					evt && evt.removeListener();
+
+					iframe.on( 'load', function() {
+						if ( recreateEditable ) {
+							recreateEditable = false;
+							recreate();
+						}
+					} );
+				}
```
Dumluregn, line 160: This function is actually invoked before `observeEditor()` etc, so let's move the declaration higher to be easier to find.

plugins/wysiwygarea/plugin.js
```diff
@@ -93,9 +98,82 @@
...
+					}
+
+					CKEDITOR.tools.array.forEach( mutation.addedNodes, checkIfAffectsEditor );
 				}
+
+				function checkIfAffectsEditor( node ) {
+					if ( !node.querySelector ) {
+						return;
+					}
+					var editorSelector = '#' + editor.container.getId(),
```
Dumluregn, line 137:
```suggestion

					var editorSelector = '#' + editor.container.getId(),
```

plugins/wysiwygarea/plugin.js
```diff
@@ -93,9 +98,82 @@
...
+
+					// (#4462)
+					// Skip IE's below version 11. They don't support MutationObserver.
+					if ( !CKEDITOR.env.ie || isIE11 ) {
+						if ( isIE11 ) {
+							editor.on( 'mode', attachIframeReloader, { iframe: iframe, editor: editor, callback: callback } );
+						}
+
+						observeEditor();
+					}
```
Dumluregn, line 114: We could flatten it even more:
```suggestion
					if ( CKEDITOR.env.ie && !isIE11 ) {
						return;
					}
					
					if ( isIE11 ) {
						editor.on( 'mode', attachIframeReloader, { iframe: iframe, editor: editor, callback: callback } );
					}

					observeEditor();
```

But the conditions now are quite clear so we can also leave it as is if you want 🤔

plugins/wysiwygarea/plugin.js
```diff
@@ -716,3 +800,33 @@ CKEDITOR.config.disableNativeSpellChecker = true;
...
+
+/**
+ * Document observation starting point. Native DOM object.
+ *
+ * To recreate editor `wysiwygarea` iframe after reattach editor to DOM:
+ *
+ * * make sure detachable element is placed any nested depth under observable element.
+ * * make sure editor is placed any nested depth under detachable element.
+ *
+ * See overview of HTML hierarchy below:
```
Dumluregn, line 812: I think it's important to mention that it's just a pseudocode and not actual HTML tags 🙃

Dumluregn (changes requested):
Just some small refactors left.

Besides that - on IE11 I've encountered an error after each toggle:
Screenshot 2021-04-15 at 15 48 38

Looks like Bender issue TBH 🤔  could you verify if it's reproducible outside of testing environment? I didn't experience any influence on editor (all tests pass), but in manual descriptions we write that any error causes the test to fail, so we should take some action.

plugins/wysiwygarea/plugin.js
```diff
@@ -716,3 +800,33 @@ CKEDITOR.config.disableNativeSpellChecker = true;
...
  * @param {CKEDITOR.dom.element} data The element being added.
  */
+
+/**
+ * Document observation starting point. Native DOM object.
+ *
+ * To recreate editor `wysiwygarea` iframe after reattach editor to DOM:
+ *
+ * * make sure detachable element is placed any nested depth under observable element.
+ * * make sure editor is placed any nested depth under detachable element.
```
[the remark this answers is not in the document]
sculpt0r: I meant that those elements don't have to be direct children. So your observable could be a top-level `div` element and the editor itself could be placed deep down under many others elements.

sculpt0r:
Rebase on newest major

plugins/wysiwygarea/plugin.js
sculpt0r: Well, it looks even better - so yes - lets make it flat :D

sculpt0r:
> Besides that - on IE11 I've encountered an error after each toggle:
> Screenshot 2021-04-15 at 15 48 38

Definitely, it's related to our solution. Because in unit tests which expect empty editor (so didn't call restore logic) this issue doesn't appear for me in IE11. All other units from `detached` throw this error... :thinking:

sculpt0r:
@Dumluregn - ready for review.
- rebased on the newest major
- correct all inline comments
- Suppress IE11 error - please take a look at https://github.com/ckeditor/ckeditor4/pull/4601#issuecomment-824855508

Dumluregn (changes requested):
I've encountered an issue with undo history which occurs on all browsers and is rather simple to reproduce, so we have to take care of it.

1. Open any manual test.
2. Toggle source mode.
3. Toggle editor (detach+reattach).
4. Use `bold` button.

Editor doesn't record any undo step. Even if it looks like it does (after you type anything), you can't access it. 

Other than that - mostly small ammendments.

sculpt0r:
Rebase on newest major.

tests/plugins/wysiwygarea/manual/detachedembededimage.md
```diff
@@ -0,0 +1,13 @@
...
+@bender-ui: collapsed
+@bender-include: ../../embedbase/_helpers/tools.js
+@bender-ckeditor-plugins: wysiwygarea,toolbar,sourcearea,elementspath,undo,embed
+
+**Note:** Open dev console to track possible errors. If any occurs, test failed.
+
+1. Use 'Insert Media Embed' and put there any image address. For example google logo.
+2. Reattach editor with double click on "Toggle" button.
+
+	**Expected:** There is exactly one undo step.
```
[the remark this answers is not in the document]
sculpt0r: Yes. Still we have a bug here: https://github.com/ckeditor/ckeditor4/pull/4601#issuecomment-818285341

sculpt0r:
> I've encountered an issue with undo history which occurs on all browsers and is rather simple to reproduce, so we have to take care of it.
> 
> 1. Open any manual test.
> 2. Toggle source mode.
> 3. Toggle editor (detach+reattach).
> 4. Use `bold` button.
> 
> Editor doesn't record any undo step. Even if it looks like it does (after you type anything), you can't access it.

I've got this one! :D

sculpt0r:
Probably have some explanation for doubled history, but I'll put in in extracted issue.

tests/plugins/wysiwygarea/detached.js
```diff
@@ -0,0 +1,206 @@
...
+		'test reattached editor restores editor data with editor parent observed': function() {
+			if ( !isSupportedEnvironment() ) {
+				assert.ignore();
+			}
+
+			var observableParent = CKEDITOR.document.getById( 'editorDetachableParent' );
+
+			bender.editorBot.create( {
+				name: 'editor',
+				startupData: startupData,
```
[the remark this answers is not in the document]
sculpt0r: We use `startupData` in assertions. That means, if we put it in `.html` file, we need to add additional data extraction from HTML. I would stay with the current solution. Actually, current HTML is provided for hierarchy visualization.

sculpt0r:
@Dumluregn - ready, steady. review! :D
- Sorry for 📄the wall of text below...
- Correct all inline comments, besides https://github.com/ckeditor/ckeditor4/pull/4601#discussion_r619112750. Please take a look at my answer: https://github.com/ckeditor/ckeditor4/pull/4601#discussion_r620910525
- In the case of [broken undo history](https://github.com/ckeditor/ckeditor4/pull/4601#discussion_r619119094), I extracted a separate issue: https://github.com/ckeditor/ckeditor4/issues/4644, correct manual test scenario - so it passes at the current state and I add a note about this issue in the test.
- I moved to modern `editor.setData` syntax: https://github.com/ckeditor/ckeditor4/blob/0c132bfc377f7d051c913f327519f8ff815d8bce/plugins/wysiwygarea/plugin.js#L182. However, I switch default `fireSnapshot` to `false` - so no additional snapshots would be taken during recreation. undo plugin still fails, but we don't need those snapshots at recreating time.

- The most important, but also most complex: https://github.com/ckeditor/ckeditor4/pull/4601#pullrequestreview-643147098. The reason for that was invalid `editor.status`.
  - Just before `setData` https://github.com/ckeditor/ckeditor4/blob/0c132bfc377f7d051c913f327519f8ff815d8bce/plugins/wysiwygarea/plugin.js#L181-L182 we are changing it to `recreating`. It is because undo plugin early returns from saving snapshots with status different than `ready` (BTW, I try with `lockSnapshot` event, which causes below `this.locked` to be true, but this makes the entire data gone...) https://github.com/ckeditor/ckeditor4/blob/0c132bfc377f7d051c913f327519f8ff815d8bce/plugins/undo/plugin.js#L372 We are not setting status to `ready` after `SetData` in wysiwygmode plugin. Because when the editor is created, the callback is passed and stays as context for our `recreate` function. So each `recreate` uses a callback which sets editor status back to `ready`. https://github.com/ckeditor/ckeditor4/blob/0c132bfc377f7d051c913f327519f8ff815d8bce/core/creators/themedui.js#L378-L386 However, if we are getting back from source mode, there is another `setMode` invoked without any callback. Now our `recreate` doesn't has a previous callback context, so nothing changes the status back. https://github.com/ckeditor/ckeditor4/blob/0c132bfc377f7d051c913f327519f8ff815d8bce/plugins/sourcearea/plugin.js#L145 Also we are not taking status back in `recreate` right after the `editor.setData` becuase it will be to early. In such case we are getting back to not working editor. `editor.status = 'ready'` which comes from `themedui.js` is set in timeout with zero-time, so after all stack invocations. Based on that, I decide to bind again to `mode` event https://github.com/ckeditor/ckeditor4/blob/0c132bfc377f7d051c913f327519f8ff815d8bce/plugins/wysiwygarea/plugin.js#L117-L119 and change status to `ready` again. This `mode` event is fired in a similar way that editor creation, but it is placed in a method which is called to change mode: https://github.com/ckeditor/ckeditor4/blob/0c132bfc377f7d051c913f327519f8ff815d8bce/core/creators/themedui.js#L245.

tests/plugins/wysiwygarea/detached.js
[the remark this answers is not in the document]
Dumluregn: Yeah, we have to keep `startupData` variable for assertions, but I don't see why we'd need to extract data from HTML?

plugins/wysiwygarea/plugin.js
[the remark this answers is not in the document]
Dumluregn: Ok, then I guess `is nested on any level` should be a little easier to understand 🤔

plugins/wysiwygarea/plugin.js
```diff
@@ -716,3 +808,33 @@ CKEDITOR.config.disableNativeSpellChecker = true;
...
+ *
+ * See overview of hierarchy below (HTML pseudocode):
+ *
+ * ```
+ * <observable>
+ *   <detachable> <!-- Any nesting depth -->
+ *     <editor></editor> <!-- Any nesting depth -->
+ *   </detachable>
+ * </observable>
```
Dumluregn, line 827: Maybe adding `...` instead of the comment?

```suggestion
 * <observable>
 *    <...>
 *     <detachable>
 *       <...>
 *         <editor></editor>
 *       </...>
 *     </detachable>
 *    </...>
 * </observable>
```

plugins/wysiwygarea/plugin.js
```diff
@@ -716,3 +808,33 @@ CKEDITOR.config.disableNativeSpellChecker = true;
...
+ * ```
+ *
+ * By default, the entire document is observed.
+ *
+ * If you know exactly which element is detaching - choose its direct parent.
```
Dumluregn, line 832:
```suggestion
 * By default, the entire document is observed. However, if you know exactly which element is detaching, you can choose its direct parent to increase performance a little.
```

plugins/wysiwygarea/plugin.js
```diff
@@ -716,3 +808,33 @@ CKEDITOR.config.disableNativeSpellChecker = true;
...
+ *
+ * Note, that if you choose element which is detaching. No changes will be detected.
```
Dumluregn, line 834:
```suggestion
 * Note, that if you choose element which is detaching, no changes will be detected.
```

plugins/wysiwygarea/plugin.js
```diff
@@ -716,3 +808,33 @@ CKEDITOR.config.disableNativeSpellChecker = true;
  * @param {CKEDITOR.editor} editor This editor instance.
  * @param {CKEDITOR.dom.element} data The element being added.
  */
+
+/**
+ * Document observation starting point. Native DOM object.
```
Dumluregn, line 813: The first line should say something about the purpose of this option, so I'd add a few words like `mutation observer` and `detach` 😛

Dumluregn (changes requested):
> Based on that, I decide to bind again to mode event

Works fine now 👍🏻 

The PR looks good... enough to start renamings and polishing docs 📝 🔥 But it shouldn't take much effort.

Also please add a manual test with the scenario that was failing before (https://github.com/ckeditor/ckeditor4/pull/4601#pullrequestreview-643147098). And I've commented here https://github.com/ckeditor/ckeditor4/pull/4601#discussion_r621319217 (linking since it's easy to omit).

tests/plugins/wysiwygarea/detached.js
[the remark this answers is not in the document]
Dumluregn: I agreed with @sculpt0r that having the data in variable is less prone to produce errors in case of changes in the expected data or test structure, so we'll leave it as is 👍🏻

sculpt0r:
- use proper name for `setData` call. I just change the name. It still has the same default value (which evaluate as `false`). So technically we could skip it, but I want to preserve it in this case, since there is a lot going on with `undoplugin` and we still have one extracted issue about it.
- Correct API doc comments and other smaller things according to your suggestions
- Added a new manual test (as you requested) with a previously failing case. I remove the note with the console from there because there is no need for additional checks here. Results are visible right in editor UI.

Dumluregn:
Rebased onto latest `major`, PR is almost ready for me. One last thing I've noticed is duplicated undo step (again), but this time on FF only:

1) Apply bold to text in editor
2) Detach and reattach editor
3) Check undo steps; should be 1, but there are 2.

Since this doesn't break anything and is browser specific, please take a look at it but don't spend more than 1 hour trying to find the cause and if you don't manage, we'll extract an issue. 

> Besides that, my sonarqube plugin shows that this invocation doesn't need an argument. But since there is a lot going on, I don't touch it yet. Do you think we could remove this argument now?

Normally I'd say it's better not to touch it, but here it looks like a very simple situation where the function was nested to use the scoped `editor` variable but it wasn't removed from arguments list, so I did remove it 👍🏻 

No more comments from me 👍🏻

Dumluregn (changes requested):
Rebased onto latest `major`, PR is almost ready for me. One last thing I've noticed is duplicated undo step (again), but this time on FF only:

1) Apply bold to text in editor
2) Detach and reattach editor
3) Check undo steps; should be 1, but there are 2.

Since this doesn't break anything and is browser specific, please take a look at it but don't spend more than 1 hour trying to find the cause and if you don't manage, we'll extract an issue. 

> Besides that, my sonarqube plugin shows that this invocation doesn't need an argument. But since there is a lot going on, I don't touch it yet. Do you think we could remove this argument now?

Normally I'd say it's better not to touch it, but here it looks like a very simple situation where the function was nested to use the scoped `editor` variable but it wasn't removed from arguments list, so I did remove it 👍🏻 

No more comments from me 👍🏻

sculpt0r:
@Dumluregn - done ;)

I track a little for this bug, but finally decided to extract it (https://github.com/ckeditor/ckeditor4/issues/4667) and describe what I found. And as we both know, there is no easy and obvious situation here. What is more, this happens not only for FF for me.

Thank you for your corrects and for the fix with this unused function argument.

Dumluregn (approved):
LGTM! 🎉 

Your turn, @f1ames ... 😄

f1ames:
> Your turn, @f1ames ...

I think I will ask @Comandeer to take this review :smile:

Comandeer:
Rebased onto latest `major`.

tests/plugins/wysiwygarea/manual/detachedembededimagesource.html
```diff
@@ -0,0 +1,26 @@
+<p>Google logo:</p>
+
+<pre>https://www.google.com/images/branding/googlelogo/1x/googlelogo_color_272x92dp.png</pre>
```
Comandeer, line 3: I'm not sure if we should use some resource that we don't have any control over.

tests/plugins/wysiwygarea/manual/detachedtwoeditors.html
```diff
@@ -0,0 +1,38 @@
+<button id="toggleButton">Toggle</button>
```
Comandeer, line 1: I'd move this button directly above detachable editor. It seems more intuitive for me. Additionally, currently I need to scroll page up to use the button after interacting with the second editor.

tests/tickets/4527/1.js
```diff
@@ -7,7 +7,6 @@
 	bender.editor = {
 		editorName: 'editor1',
 		startupData: '',
-		name: document.title,
```
Comandeer, line 10: Why is this change introduced?

Comandeer (changes requested):
I've discovered some weird bug connected with undo:

1. Open http://localhost:1030/tests/plugins/wysiwygarea/manual/detachedsourceundo
2. Bold text.
3. Undo.
4. Press "Toggle" button twice.
5. Focus the editable.
6. Switch to source mode.
7. Press "Toggle" button twice.
8. Switch back to WYSIWYG mode.

The redo step is lost.

There aren't also any unit tests for undo after reattaching.

sculpt0r:
Rebase on newest major.

tests/tickets/4527/1.js
sculpt0r: This one leads to errors in IE. This name was used inside `querySelector`. Unfortunately in IE, there were some restricted characters in the selector. The same goes for https://github.com/ckeditor/ckeditor4/pull/4601/files#diff-a15eeadc2752af4c0b9dcfc057fb5fe450dfb6cf9a672e4fc9baacaac9036903

So both tests start to fail because of this.

sculpt0r:
@Comandeer - ready for review
- correct all your inline comments. 
- add an explanation for your question [here](https://github.com/ckeditor/ckeditor4/pull/4601#discussion_r628106744)

plugins/wysiwygarea/plugin.js
```diff
@@ -93,9 +103,79 @@
...
+				function checkIfAffectsEditor( node ) {
+					if ( !node.contains || !node.contains( editor.container.$ ) ) {
+						return;
+					}
+
+					if ( isIE11 ) {
+						recreateEditable = true;
+					} else {
+						recreate();
+					}
```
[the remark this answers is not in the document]
sculpt0r: Well, that indicates that `recreate()` returns some meaningful value - and it's not. We know that will be undefined but if you read it looks like recreation could end with success or failure.
So to make consensus, maybe:

```js
if ( !isIE11 ) {
  recreate();
  return;
}

recreateEditable = true;
```

?

Comandeer (changes requested):
Looks good! But I see that some error appeared in IE11 due to new tests:
Permission denied error before redo test
Could you check it?

sculpt0r:
@Comandeer - ready for review
I found that access to document while editable is recreating caused this issue. So I simply prevent that and return null: https://github.com/ckeditor/ckeditor4/pull/4601/commits/b9544f37196ef34904b2468b5c23ad7b19f634f8

tests/plugins/wysiwygarea/detached.js
```diff
@@ -0,0 +1,263 @@
...
+				editorContainer.remove();
+				editorContainerParent.append( editorContainer );
+
+				assertReattachedData( bot, startupData, 'Reattached editor should have the same data with iframe. Observing default object.' );
+
+				wait();
+			} );
+		},
+
+		'test reattached editor makes editor data empty with observed invalid DOM object': function() {
```
Comandeer, line 188: This and following tests fail on IE8.

core/selection.js
```diff
@@ -1260,7 +1260,8 @@
 
 		// Editable element might be absent or editor might not be in a wysiwyg mode.
 		var editable = this.editable();
-		return editable && this.mode == 'wysiwyg' ? new CKEDITOR.dom.selection( editable ) : null;
+
+		return editable && this.mode == 'wysiwyg' && editable.editor.status !== 'recreating' ? new CKEDITOR.dom.selection( editable ) : null;
```
Comandeer, line 1264: It would be nice to add some test to really see if `null` is returned when editor is recreating.

Comandeer (changes requested):
Almost there… See comments.

tests/plugins/wysiwygarea/detached.js
sculpt0r: It's funny - they failed for me only with opened dev console :thinking: And I can run them with dev console -> fail. Then close this console and they pass. So it doesn't look like cache :thinking:

Comandeer: Strange. For me they fail even without console.

sculpt0r:

- Added unit test for getSelection()
- Fix failing tests. They failed because the previous test has still an active observer (the default observed element is the entire document).  So now - the editor is destroyed at the end of the test.

Comandeer (approved):
LGTM! 🎉 

I've just updated API docs for `CKEDITOR.editor#getSelection()` (it seems that we have them wrong for years 🙈) and simplify it a little bit (replaced `editable.editor.status` with `this.status`).

I've also added additional changelog entry for the change in `CKEDITOR.editor#getSelection()` under API Changes.

PR #4623, opened 2021-04-15
Fix for incorrect hover styles for buttons on mobile devices

### LINKED ISSUE #4611: CKEditor 4 mobile - buttons remain highlighted after being deselected
## Type of report

Bug

1. Open any editor sample on iOS or Android - https://ckeditor.com/docs/ckeditor4/latest/examples/standardpreset.html
2. Tap on one of basic styles button like BOLD
3. Start typing
4. Tap on the BOLD button again to disable bold style
5. Type few more characters

### Expected result

The BOLD button should't be focused.

### Actual result

The BOLD button is focused but not active which could be confusing to some of users especially that in desktop buttons aren't focused.

obraz
 - focused and active

obraz
 - focused and inactive. This is what you can see after disabling BOLD style on mobile.

obraz
- not focused and inactive. This is what you can see after disabling BOLD style on desktop.

## Other details

* Browser: Any
* OS: Android, iOS
* CKEditor version: 4.0
* Installed CKEditor plugins: BasicStyles

Comandeer:

Bug fix

### This PR contains

- [ ] Unit tests
- [x] Manual tests

```
* [#4611](https://github.com/ckeditor/ckeditor4/issues/4611): Fixed: incorrect hover styles for buttons in the toolbar on mobile devices.
```

## What changes did you make?

I've used [`hover` media query](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/hover) to revert styles on `:hover` to the original ones. I've also restored `:active` styles after that.

I've changed styles for both buttons and rich combos.

Closes #4611.


f1ames (changes requested):
TBH I can't recall many PRs not involving any JS code changes, this one looks nice in this regard :+1: :smile: 

On iOS when you long tap buttons it focuses editable area if editor is not focused yet. And when it's already focused it does nothing (from what I see it's just the way it works there). But for regular tap it's visible there is a transition between hover and active styles. I wonder if we shouldn't adjust the scenario slightly to be not confuisng :thinking: 

On Android device I have tested with (Android 6.0.1, Chrome 89.0.4389.105) it still works the same as described in #4611:

*   Short tap activates the button (with active styles which is fine), but another tab deactivates it with hover styles still visible (one have to tap outside of the editor to make hover styles disappear).
*   Long tap activates hover styles but they do not disappear after releasing the finger.

---

I would be for splitting manual tests to desktop/mobile separately. On smaller screens desktop scenario is hardly visible which may be misleading:



WDYT?

The 2nd thing - I see styles were also updated for `richcombo`. Shouldn't we have that covered in a test scenario too?

Comandeer:
> On Android device I have tested with (Android 6.0.1, Chrome 89.0.4389.105) it still works the same as described in [#4611](https://github.com/ckeditor/ckeditor4/issues/4611):

Did you test in incognito mode? :thinking: It works correctly on my phone with Android, so maybe you hit some issues with cache.

As for iOS, I didn't have access to any device and I haven't tested it. I'll need to revisit it later.

f1ames:
> > On Android device I have tested with (Android 6.0.1, Chrome 89.0.4389.105) it still works the same as described in [#4611](https://github.com/ckeditor/ckeditor4/issues/4611):
> 
> Did you test in incognito mode?  It works correctly on my phone with Android, so maybe you hit some issues with cache.

Rechecked and I have the same issue there... :thinking: 

> As for iOS, I didn't have access to any device and I haven't tested it. I'll need to revisit it later.

I mean it seems to be working fine but slightly different than on Android device (long tap does different thing) so maybe rewording test steps would be enough :thinking: Because for the initial scenario described in #4611 the behavior is now correct.

Comandeer:
I've updated manual tests to separate mobile and desktop ones. I've also slightly adjusted mobile ones description to include possible iOS behaviour (yet I don't know how accurate the description is as I don't have access to any iOS device).

All tests are now moved to `/tests/plugins/toolbar/manual/hover` as there are 12 files in total :see\_no\_evil:

f1ames:
Whooops... rebased this PR onto latest major instead of #4461 :facepalm: :see\_no\_evil: Reverting in progress...

f1ames:
Thanks @Comandeer :+1:

f1ames:
Rebased onto latest `master`.

f1ames (changes requested):
Seems like Kama skin does not provide active styles for richcombo:

image

While for moono / moono lisa it looks like:

image

And also Kama does not provide hover styles for active rich combos... This makes this test scenario hard to describe and follow :thinking: I updated the test scenario but I'm not really happy with how complex it got... @Comandeer maybe you would have a better idea here?

I think it would be better like:

1. Check if buttons have hover styles.
2. Check if clicking on buttons activate their active state.
3. Check if active buttons have hover styles.
4. Repeat the procedure for rich combos.
5. Repeat the procedure for the second editor.

Then it will be easier to describe expected/unexpected for each step and notes could be placed in better context. WDYT?

---

As for iOS, long tap focuses the editor (if it's not focused yet) or does nothing if editor is already focused (during the tap, after releasing the button is active - so it works the sam way as regular tap). Hover styles could be seen after releasing the tap, before active styles are added.

So this scenario is kind of confusing / not true... I'm not sure... do you think we should have separate tests for Android and iOS here? Putting everything in a single tests scenario makes it quite complex :thinking: 

@Comandeer could you try to polish it a bit more?

Comandeer:
Ok, so I've completely redesigned tests. There are now divided by:

*   the environment (desktop, Android, iOS),
*   the skin (Moono-Lisa, Moono, Kama),
*   the type of UI element (button, rich combo).

So now we have 18 tests in total :see\_no\_evil: But they should be much more readable and understandable.

I've also added stricter checking if the test should be run in the current env. It differentiates iOS and Android now.

f1ames (approved):
> So now we have 18 tests in total But they should be much more readable and understandable.
> 
> I've also added stricter checking if the test should be run in the current env. It differentiates iOS and Android now.

A lot of tests and a bit of duplication but indeed seems like the most reasonable approach here :+1: I only fixed spacing in tests (since `tab`s doesn't work well with nested "Expected" blocks...).

LGTM :+1: :tada:

PR #4627, opened 2021-04-19
Fix formatting when selecting new empty lines

### LINKED ISSUE #4626: Selecting an empty lines after pressing enter twice loses the formatting
## Type of report

Bug

1. Start with a clean ckeditor
2. Set formatting to: normal paragraph, bold, italic
3. Type "foo"
4. Press enter
5. Press enter again
6. Move the selection to the empty line (with mouse or with the left key)
7. Type "bar"

### Expected result

Both "foo" and "bar" are formatted

### Actual result

"bar" is not formatted

## Other details

* Browser: Opera / Chromium
* OS: Windows 10
* CKEditor version: 4.16
* Installed CKEditor plugins: enterkey

corebonts:

Bug fix

### This PR contains

- [x] Unit tests
- [x] Manual tests

```
* [#4626](https://github.com/ckeditor/ckeditor4/issues/4626): Fix formatting when selecting new empty lines.
```

## What changes did you make?

When enter is pressed, the _enterkey_ plugin clones the formatting nodes (span/b/i).
It also adds a `<br>` bogus so the line can be selected even if it's empty.

The problem was that the bogus was added after the formatting nodes, not inside them.
So instead of this:
`<p><b><br></b></p>`
it looked like this:
`<p><b></b><br></p>`

Closes #4626 .

### DIFF
--- a/plugins/enterkey/plugin.js
+++ b/plugins/enterkey/plugin.js
@@ -388,8 +388,8 @@
 				// Recreate the inline elements tree, which was available
 				// before hitting enter, so the same styles will be available in
 				// the new block.
-				var elementPath = splitInfo.elementPath;
-				var bogusHolder = newBlock;
+				var elementPath = splitInfo.elementPath,
+					bogusHolder = newBlock;
 				if ( elementPath ) {
 					for ( var i = 0, len = elementPath.elements.length; i < len; i++ ) {
 						var element = elementPath.elements[ i ];
@@ -401,7 +401,7 @@
 							element = element.clone();
 							newBlock.moveChildren( element );
 							newBlock.append( element );
-							if (bogusHolder === newBlock) {
+							if ( bogusHolder === newBlock ) {
 								bogusHolder = element;
 							}
 						}
--- a/plugins/enterkey/plugin.js
+++ b/plugins/enterkey/plugin.js
@@ -389,6 +389,7 @@
 				// before hitting enter, so the same styles will be available in
 				// the new block.
 				var elementPath = splitInfo.elementPath;
+				var bogusHolder = newBlock;
 				if ( elementPath ) {
 					for ( var i = 0, len = elementPath.elements.length; i < len; i++ ) {
 						var element = elementPath.elements[ i ];
@@ -400,11 +401,14 @@
 							element = element.clone();
 							newBlock.moveChildren( element );
 							newBlock.append( element );
+							if (bogusHolder === newBlock) {
+								bogusHolder = element;
+							}
 						}
 					}
 				}
 
-				newBlock.appendBogus();
+				bogusHolder.appendBogus();
 
 				if ( !newBlock.getParent() )
 					range.insertNode( newBlock );
--- a/tests/plugins/enterkey/enterkey.js
+++ b/tests/plugins/enterkey/enterkey.js
@@ -96,6 +96,25 @@
 			assert.areSame( '<p><b><i>foo</i></b></p><p><b><i>bar</i></b></p>', bot.getData( false, true ) );
 		},
 
+		// https://github.com/ckeditor/ckeditor4/issues/4626: Selecting an empty line after pressing enter twice loses the formatting
+		'test enter key twice then select empty line': function() {
+			// The bogus should be in the right place for empty lines, otherwise the selection will be
+			// outside the formatted range.
+			var bot = this.editorBots.editor,
+				editor = bot.editor,
+				editable = editor.editable();
+
+			bot.setHtmlWithSelection( '<p><b><i>foo^</i></b></p>' );
+			bot.execCommand( 'enter' );
+			bot.execCommand( 'enter' );
+
+			var selectionAtTheEndOfEmptyLine = new CKEDITOR.dom.range( editor.document );
+			selectionAtTheEndOfEmptyLine.moveToElementEditablePosition( editable.getChild(1), true );
+			editor.getSelection().selectRanges( [ selectionAtTheEndOfEmptyLine ] );
+			editor.insertText( 'bar' );
+			assert.areSame( '<p><b><i>foo</i></b></p><p><b><i>bar</i></b></p><p>&nbsp;</p>', bot.getData( false, true ) );
+		},
+
 		// https://dev.ckeditor.com/ticket/7946 TODO: Add editor doc quirks mode tests.
 		'test enter key key scrolls document': function() {
 			// On iPads, behavior of scrollTop, scrollHeight and clientHeight is a bit unexpected.
--- a/tests/plugins/enterkey/enterkey.js
+++ b/tests/plugins/enterkey/enterkey.js
@@ -96,27 +96,6 @@
 			assert.areSame( '<p><b><i>foo</i></b></p><p><b><i>bar</i></b></p>', bot.getData( false, true ) );
 		},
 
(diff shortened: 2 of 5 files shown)


plugins/enterkey/plugin.js
```diff
@@ -389,6 +389,7 @@
 				// before hitting enter, so the same styles will be available in
 				// the new block.
 				var elementPath = splitInfo.elementPath;
+				var bogusHolder = newBlock;
```
Dumluregn, line 392:
```suggestion
				var elementPath = splitInfo.elementPath,
					bogusHolder = newBlock;
```

plugins/enterkey/plugin.js
```diff
@@ -400,11 +401,14 @@
 							element = element.clone();
 							newBlock.moveChildren( element );
 							newBlock.append( element );
+							if (bogusHolder === newBlock) {
```
Dumluregn, line 404:
```suggestion
							if ( bogusHolder === newBlock ) {
```

tests/plugins/enterkey/enterkey.js
```diff
@@ -96,6 +96,25 @@
...
+
+			bot.setHtmlWithSelection( '<p><b><i>foo^</i></b></p>' );
+			bot.execCommand( 'enter' );
+			bot.execCommand( 'enter' );
+
+			var selectionAtTheEndOfEmptyLine = new CKEDITOR.dom.range( editor.document );
+			selectionAtTheEndOfEmptyLine.moveToElementEditablePosition( editable.getChild(1), true );
+			editor.getSelection().selectRanges( [ selectionAtTheEndOfEmptyLine ] );
+			editor.insertText( 'bar' );
+			assert.areSame( '<p><b><i>foo</i></b></p><p><b><i>bar</i></b></p><p>&nbsp;</p>', bot.getData( false, true ) );
```
Dumluregn, line 115:
```suggestion
			var selectionAtTheEndOfEmptyLine = new CKEDITOR.dom.range( editor.document );

			selectionAtTheEndOfEmptyLine.moveToElementEditablePosition( editable.getChild( 1 ), true );
			editor.getSelection().selectRanges( [ selectionAtTheEndOfEmptyLine ] );
			editor.insertText( 'bar' );

			assert.areSame( '<p><b><i>foo</i></b></p><p><b><i>bar</i></b></p><p>&nbsp;</p>', bot.getData( false, true ) );
```

tests/plugins/enterkey/manual/emptylineselection.md
```diff
@@ -0,0 +1,24 @@
+@bender-tags: 4.16, bug, 4626
```
Dumluregn, line 1:
```suggestion
@bender-tags: 4.16.1, bug, 4626
```

Dumluregn (changes requested):
Thank you for the contribution! The fix works well - it solves the issue and doesn't introduce any regression. All tests are green, also on IEs, so well done 👍🏻 

We should however put some more effort to unit testing this change. There are two aspects to take into account:
* different [`CKEDITOR.config.enterMode`](https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_config.html#cfg-enterMode) settings - the fix is tested in the default config, so with `ENTER_P` mode. We should also add tests to check if it works well in `ENTER_BR` and `ENTER_DIV` modes.
* different way of moving focus to the second line - we should test (at least) 4 of them:
  * arrow left
  * arrow up
  * backspace
  * selection (you've already covered this one)

So we need to add unit tests for each combination of those two factors. For the keystroke tests you should emulate `keydown` event (see [an example](https://github.com/ckeditor/ckeditor4/blob/major/tests/core/selection/selection.js#L838)).

Also, please make sure you follow our [Code Guide](https://github.com/ckeditor/ckeditor4/blob/major/dev/docs/codestyle.md) - I've added inline comments where the style should be changed.

corebonts:
I will try do do the unit tests soon.

corebonts:
@Dumluregn I managed to write unit test for backspace (keycode: 8), but I cannot simulate arrow keys with this snippet:
`editable.fire( 'keydown', new CKEDITOR.dom.event( { keyCode: 37 } ) ); // LEFT ARROW`

Isn't it the case that keystrokes that are handled natively by the browser cannot be simulated?

Dumluregn:
Oh, right. So this has to be covered in manual tests unfortunately.

corebonts:
Then I will do it like that.

Dumluregn:
@corebonts is PR ready for next review? If so, please mark it as such by clicking the button in the `Reviewers` section.

tests/plugins/enterkey/manual/emptylineselection.md
```diff
@@ -0,0 +1,24 @@
...
+1. Place a cursor right after 'foo' in either of the editors
+
+```
+foo
+   ^ - put selection here
+```
+
+2. Press `Enter` twice.
+
+3. Press `Left` key or `Up` key or `Backspace` to move the cursor to the empty line
```
Dumluregn, line 14: Manual tests have to be exhaustive, ie. we want to check all these options - not just one. So at first we want to press `left` and check if styles are preserved, then go back, press `up` and check, and the third time check with `backspace` - instead of randomising test and checking only one option.

tests/plugins/enterkey/format-after-enterkey.js
```diff
@@ -0,0 +1,102 @@
...
+			// Backspace is not handled by the editor itself in BR mode, so it cannot be simulated
+			'test enter key twice then select backspace with BR mode': true
+		}
+	},
+
+	// https://github.com/ckeditor/ckeditor4/issues/4626: Selecting an empty line after pressing enter twice loses the formatting
+	'test enter key twice then select empty line with P mode': doubleEnter( 'editorP', INITIALCONTENT_P, selectElement( secondLineInBlockMode ), '<p><b><i>foo</i></b></p><p><b><i>bar</i></b></p><p>&nbsp;</p>' ),
+	'test enter key twice then select empty line with DIV mode': doubleEnter( 'editorDiv', INITIALCONTENT_DIV, selectElement( secondLineInBlockMode ), '<div><b><i>foo</i></b></div><div><b><i>bar</i></b></div><div>&nbsp;</div>' ),
+	'test enter key twice then select empty line with BR mode': doubleEnter( 'editorBr', INITIALCONTENT_BR, selectElement( secondLineInBrMode ), '<b><i>foo<br />bar</i></b><br />&nbsp;' ),
+	'test enter key twice then select backspace with P mode': doubleEnter( 'editorP', INITIALCONTENT_P, pressBackspace( ), '<p><b><i>foo</i></b></p><p><b><i>bar</i></b></p>' ),
```
Dumluregn, line 47: In empty parenthesis we don't put a space - please remove it also in other spots.

Dumluregn (changes requested):
AFAICS current solution kind of reverses the original issue instead of solving it - try the following:

1. Format the line with bold.
2. Press enter twice.
3. **Deactivate bold on the second empty line.**
4. Move selection to the first empty line.
5. Go back to the second empty line (the one where we disabled formatting).

Expected: First line has formatting enabled and the second one doesn't.
Actual: Now when you move from formatted line to unformatted, the formatting is preserved.

Please try to solve this issue and also add test case for it (since it involves keyboard navigation I think the manual one).

corebonts:
I have to put this aside a bit because my other stuffs. Will consider your comments later.

corebonts:
I'm still busy with other stuff but I want to handle this.

PR #4633, opened 2021-04-20
Fix incorrect HSL/HSLA conversion

### LINKED ISSUE #4597: Incorrect conversion of HSL/HSLA values
## Type of report

Bug

1. Open browser console.
2. Go to https://jsfiddle.net/7Ljwqmtz/1/
3. Check console.

### Expected result

Colors in the same console row should be the same.

### Actual result

Colors are different. The second parameter changed from `50%` to `0%`.

## Other details

* CKEditor version: 4.16.0

Comandeer:

Bug fix

### This PR contains

- [x] Unit tests
- [ ] Manual tests

```
* [#4597](https://github.com/ckeditor/ckeditor4/issues/4597): Fixed: incorrect color conversion for HSL/HSLA values.
```

## What changes did you make?

Due to the issue mentioned in https://github.com/ckeditor/ckeditor4/issues/4597#issuecomment-823094637 I've introduced additional data saved for HSL colors (hue, saturation and lightness) alongside the concept of color type.

Additionally I added rounding of lightness value and fixed the issue with saturation being always 0 or 1, due to incorrect rounding.

Closes #4597.

Note: this PR is a bug fix, however it's targeted to `major` as already merged changes (#4599) introduced some tests that have incorrect values and this PR fixes it.

### DIFF
--- a/core/tools/color.js
+++ b/core/tools/color.js
@@ -181,15 +181,27 @@
 			/**
 			 * Original color type.
 			 *
+			 * Available types:
+			 *
+			 * ```
+			 * +------+---------------+---------------------------------------+
+			 * | Type | Integer value |                Constant               |
+			 * +------+---------------+---------------------------------------+
+			 * |  RGB |       1       | {@link CKEDITOR.tools.color.TYPE_RGB} |
+			 * +------+---------------+---------------------------------------+
+			 * |  HSL |       2       | {@link CKEDITOR.tools.color.TYPE_HSL} |
+			 * +------+---------------+---------------------------------------+
+			 * ```
+			 *
 			 * @private
 			 * @property {Number}
 			 */
 			type: 0,
 
 			/**
-			 * Hue value.
+			 * Hue value. Ranges between 0-360 (inclusive).
 			 *
-			 * Used in HSL colors. Ranges between 0-360 (inclusive).
+			 * Used in HSL colors.
 			 *
 			 * @private
 			 * @property {Number}
@@ -199,6 +211,8 @@
 			/**
 			 * Saturation value. Ranges between 0-100 (inclusive).
 			 *
+			 * The value of `0` means that the color will be a shade of gray.
+			 *
 			 * Used in HSL colors.
 			 *
 			 * @private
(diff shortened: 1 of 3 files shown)


core/tools/color.js
```diff
@@ -358,20 +417,21 @@
...
 				}
 
-				return [
-					hexToNumber( parts[ 1 ] + parts[ 2 ] ),
-					hexToNumber( parts[ 3 ] + parts[ 4 ] ),
-					hexToNumber( parts[ 5 ] + parts [ 6 ] ),
-					alpha
-				];
+				return {
+					type: CKEDITOR.tools.color.TYPE_RGB,
```
Dumluregn, line 421: It looks a bit awkward that we return `TYPE_RGB` instead of `TYPE_HEX` here. I know that functionally it won't make any difference, but maybe we could distinguish them for clarity?

Dumluregn: Unless it actually converts the hex color to RGB (and I guess it does) 🤔

tests/core/tools/color.js
```diff
@@ -173,7 +167,56 @@
...
+
+			assert.areSame( CKEDITOR.tools.color.TYPE_RGB, color._.type, 'Color type is correctly set to RGB' );
+		},
+
+		// (#4597)
+		'test saving data about color type for hex color': function() {
+			var input = '#FF00FF',
+				color = new CKEDITOR.tools.color( input );
+
+			assert.areSame( CKEDITOR.tools.color.TYPE_RGB, color._.type, 'Color type is correctly set to RGB' );
```
Dumluregn, line 218: In these test we should also test channels (there are a lot of tests above but they are only testing the whole conversion, not the part with channel extraction like the `test saving data about color type, hue, saturation and ligthness for HSL color` test for HSL).

Dumluregn (changes requested):
Just some small fixes needed here. I like the conversion of channels container from array to object 👍🏻

core/tools/color.js
Comandeer: Technically all colors are converted to RGB. However hex is only the other representation of RGB color, while HSL is a completely different color format. Because of that I've introduced only two types – RGB (for both representations, `rgb()` and hex ones) and HSL.

Comandeer:
Ok, I've updated API docs and slightly modified tests to also assert channels.

core/tools/color.js
Dumluregn: Ok, makes sense.

Dumluregn:
Rebased onto latest `major`.

Dumluregn (approved):
LGTM! 🎉 

I've only fixed tests error messages and added test case for hex code without `#`.

PR #4635, opened 2021-04-20
Fix incorrect handling of HSL/HSLA values

### LINKED ISSUE #4596: Incorrect handling of HSL/HSLA values
## Type of report

Bug

1. Open browser console.
2. Go to https://jsfiddle.net/7Ljwqmtz/
3. Check the console.

### Expected result

There should be `false` in the console, as the color is incorrect.

### Actual result

There is `true`

## Other details

According to the specification, [HSL format accept only percentage as second and third parameter](https://drafts.csswg.org/css-color/#the-hsl-notation).

* CKEditor version: 4.16.0

Comandeer:

Bug fix

### This PR contains

- [x] Unit tests
- [ ] Manual tests

```
* [#4596](https://github.com/ckeditor/ckeditor4/issues/4596): Fixed: incorrect handling of HSL/HSLA values.
```

## What changes did you make?

During fixing the issue with accepting non-percentage values, I've discovered that [hue can be passed also as degrees](https://drafts.csswg.org/css-color/#typedef-hue). So I've decided to fix it as well.

And I thought that the good idea would be to fix it by enhancing our regex. Oh, how naive I was…! Yet, I've managed to enhance our regex. And added some docs to it for the next naive one that would think the same thing as me.

Closes #4596.

Note: this PR is a bug fix, but it's targeted to `major` as already merged changes (especially #4599) heavily influenced the proposed solution.


Dumluregn:
Rebased onto latest `major`.

Dumluregn (changes requested):
The fix looks good 👍🏻 Two things left to do:
1) We need to update docs:
* hue - can accept `deg`
* saturation and lightness - only take `%` values
* alpha - can accept both fraction and percentage.

2) File with tests grew too much and becomes hard to maintain, I think it's time to split it into a few smaller ones. My proposal is based on the type of tests, so three files (for now):

* color conversion
* color channels
* color format

  Still, feel free to propose another division if you find out anything reasonable.

Comandeer:
I've divided tests into three files:

*   `color.js` – more general tests (like invalid formats etc.),
*   `conversion.js` – tests connected with conversion,
*   `channels.js` – test connected with validating extracted channels.

I've also moved them to the new dir, `/tests/core/tools/color`.

> We need to update docs

I'm not sure which docs you're referring to. If you mean API docs, then IMO they're correct. We state in them that we support RGB and HSL formats – and that's still true (even more true than before, as our understanding of these formats was different than what it's in the spec).

Dumluregn (approved):
LGTM! 🎉 

> I've divided tests into three files:

👍🏻 

> I'm not sure which docs you're referring to.

Took me a while, but now I see that even if e.g. saturation is passed in `%`, we save it in channel as integer value (same for lightness etc.). So you're right that docs are fine.

Issue #4604, opened 2021-04-02
Enhance support for getting types of pasted data


Comandeer:
## Type of report

Feature request

## Provide description of the new feature

Currently we [fetch types directly from native `DataTransfer`](https://github.com/ckeditor/ckeditor4/blob/05e2b1115beb0885a637adb81e7ed091b47a2d8c/plugins/clipboard/plugin.js#L2735-L2741). However in many situations it won't work (as the data inside `DataTransfer` is available synchronously only inside `paste` event handler). It could lead to a situation in which `CKEDITOR.plugins.clipboard.dataTransfer` object is created with a data, but without a type, e.g. when user pastes image (the only available type then is `Files`).

We [cache data with its type inside `CKEDITOR.plugins.clipboard.dataTransfer#cacheData()`](https://github.com/ckeditor/ckeditor4/blob/05e2b1115beb0885a637adb81e7ed091b47a2d8c/plugins/clipboard/plugin.js#L2599-L2613) to be able to get it asynchronously later, but it does not cache `Files` type (as this type does not provide any data; it only indicates that only files were pasted).

Therefore I propose that:

* types should also be cached (as `_.types`) so we will be able to fetch them later,
* (optionally) indicator for pasted files only should be introduced (it would replace [`isFileData` private function](https://github.com/ckeditor/ckeditor4/blob/05e2b1115beb0885a637adb81e7ed091b47a2d8c/plugins/clipboard/plugin.js#L195-L205) – or at least reduce its complexity).

CKEditorBot:
Closed in https://github.com/ckeditor/ckeditor4/pull/4619

PR #4662, opened 2021-04-30
Fix for regression with pasting images in Firefox

### LINKED ISSUE #3757: Image copied in clipboard is not insterted as base64 when not using uploadimage plugin

Feature request

1. We are not using uploadimage, nor easyimage
2. Press print screen
3. Paste with Ctrl+V in ckeditor

### Expected result

Image must be inserted as base64 like as when I copy/paste a file image

### Actual result

Nothing when uploadimage not used

## Other details

* Browser: Firefox 71
* OS: Windows
* CKEditor version: 4.13.1
* Installed CKEditor plugins: base64image and no  uploadimage, nor easyimage

Comandeer:

Bug fix

### This PR contains

- [ ] Unit tests
- [x] Manual tests

```
* [#3757](https://github.com/ckeditor/ckeditor4/issues/3757): [Firefox] Fixed: pasted images are not inserted as Base64-encoded images.
```

## What changes did you make?

It seems that fix for #4604 also fixed this issue. I've just slightly adjusted the solution to use the new `CKEDITOR.plugins.clipboard.dataTransfer#isFileTransfer()` method. I've also updated manual test. I didn't add new unit tests as it seems that we are covered enough in this matter.

Closes #3757.

### DIFF
--- a/plugins/clipboard/plugin.js
+++ b/plugins/clipboard/plugin.js
@@ -197,11 +197,7 @@
 					return false;
 				}
 
-				var types = dataTransfer.getTypes(),
-					isFileOnly = types.length === 1 && types[ 0 ] === 'Files',
-					containsFile = dataTransfer.getFilesCount() === 1;
-
-				return isFileOnly && containsFile;
+				return dataTransfer.isFileTransfer();
 			}
 
 			editor.on( 'paste', function( evt ) {
--- a/plugins/clipboard/plugin.js
+++ b/plugins/clipboard/plugin.js
@@ -197,7 +197,7 @@
 					return false;
 				}
 
-				return dataTransfer.isFileTransfer();
+				return dataTransfer.isFileTransfer() && dataTransfer.getFilesCount() === 1;
 			}
 
 			editor.on( 'paste', function( evt ) {
--- a/tests/plugins/clipboard/manual/pasteimage.md
+++ b/tests/plugins/clipboard/manual/pasteimage.md
@@ -1,10 +1,10 @@
 @bender-ui: collapsed
-@bender-tags: 4.6.2, bug, clipboard
+@bender-tags: 4.6.2, 4.17.0, bug, clipboard, 3757
 @bender-ckeditor-plugins: wysiwygarea, toolbar, undo, basicstyles, image, clipboard, sourcearea
 
 ## Scenario:
 
  * Paste image from your clipboard into the editor.
- 
+
 **Expected**: Image is pasted as base64 encoded file.
 
--- a/tests/plugins/clipboard/manual/pasteimage.md
+++ b/tests/plugins/clipboard/manual/pasteimage.md
@@ -1,10 +1,16 @@
 @bender-ui: collapsed
-@bender-tags: 4.6.2, bug, clipboard
+@bender-tags: 4.6.2, 4.17.0, bug, clipboard, 3757
 @bender-ckeditor-plugins: wysiwygarea, toolbar, undo, basicstyles, image, clipboard, sourcearea
 
 ## Scenario:
 
- * Paste image from your clipboard into the editor.
- 
-**Expected**: Image is pasted as base64 encoded file.
+1. Paste image from your clipboard into the editor.
+
+	### Expected result
+
+	* Image is pasted as base64 encoded file.
+
+	### Unexpected result
+
+	* Nothing happens.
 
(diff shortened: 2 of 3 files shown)


plugins/clipboard/plugin.js
```diff
@@ -197,11 +197,7 @@
 					return false;
 				}
 
-				var types = dataTransfer.getTypes(),
-					isFileOnly = types.length === 1 && types[ 0 ] === 'Files',
-					containsFile = dataTransfer.getFilesCount() === 1;
```
sculpt0r, line 202: Verifying if there is only a single file in dataTransfer is skipped now. I assume if we have anything in the clipboard and it is `file` type then we can take something at `0` index. And it is impossible to have more than one thing in the clipboard pasted, so this check seems to be unnecessary. Was that your way of thinking about it?

tests/plugins/clipboard/manual/pasteimage.md
```diff
@@ -1,10 +1,16 @@
 @bender-ui: collapsed
-@bender-tags: 4.6.2, bug, clipboard
+@bender-tags: 4.6.2, 4.17.0, bug, clipboard, 3757
 @bender-ckeditor-plugins: wysiwygarea, toolbar, undo, basicstyles, image, clipboard, sourcearea
 
 ## Scenario:
 
- * Paste image from your clipboard into the editor.
```
sculpt0r, line 7: > I didn't add new unit tests as it seems that we are covered enough in this matter.

Well it is more related to https://github.com/ckeditor/ckeditor4/pull/4619, and maybe it is intended, but: 
1. right-click at any internet graphic and use the "copy image" option (name probably depends on browser)
2. try to "Paste image from your clipboard" image is pasted as URL, not base64 :thinking: 

I find that there is an additional type in types when `copy/paste` like that and it is "text/html" but also "Files" is present there. Is it some edge case or I miss some knowledge?

sculpt0r (changes requested):
Looks LGTM, but I need just two explanations - to fully understand if everything is working fine. Please take a look at my comments.

tests/plugins/clipboard/manual/pasteimage.md
Comandeer: That's actually an interesting question :thinking: I don't really know, what is an intended behaviour. But I suspect if the clipboard contains the copied images as HTML `img` element, there is no need to convert it into Base64.

plugins/clipboard/plugin.js
Comandeer: Well, it is possible (e.g. Office always pastes HTML alongside RTF and screenshot). It's also possible to create such a clipboard data synthetically. There is also a case of drag & drop, when user can drop several files at once.

But at the moment we support pasting only one image at the time and we don't support d&d. And these two things are actually nice features. I'll create issues for them.

Back to the question: yup, as long as we support only one image, IMO we should be safe with the current condition.

Comandeer: On the second thought I've readded it just in case (as we never check if there is acutally a file inside clipboard, we just assume so). Better safe than sorry.

tests/plugins/clipboard/manual/pasteimage.md
```diff
@@ -1,10 +1,16 @@
...
- 
-**Expected**: Image is pasted as base64 encoded file.
+1. Paste image from your clipboard into the editor.
```
sculpt0r, line 7: Our code style guide defines Expected and Unexpected as:
```md
**Expected**
```
So, without result word and just as bolded text, not the header. I know we have tests where there are different styles, but we should stick to the rules if there is no important reason to do it differently.

Also, please notice that with your style, there is an incorrect indentation for `Unexpected result`:
image

Please, correct it according to: [code style rules](https://github.com/ckeditor/ckeditor4/blob/major/dev/docs/codestyle.md#31-manual-tests).

sculpt0r (changes requested):
@Comandeer - thanks for the explanations and the commit with 'safety-check'.

Please correct the manual test scenario, according to the comment and we are good to go :)

Comandeer:
Ok, I've reverted changes to the manual test and just updated the tags.

sculpt0r:
@Comandeer - thank you for fixing the manual test :)

LGTM :tada: :+1: 

I've changed a little changelog entry :)

sculpt0r (approved):

PR #4663, opened 2021-04-30
Fix for colon in ID is not escaped since IE doesn't support CSS escape

limingli0707:

Bug fix.
issue here: https://github.com/ckeditor/ckeditor4/issues/4664

### This PR contains

- [x] Unit tests
- [x] Manual tests

```
* [#<4664>](https://github.com/ckeditor/ckeditor4/issues/<4664>): [IE] : escape colon in id selector in IE
```

## What changes did you make?

Minor change in the tool.js > escapeCss(). When CSS.escape is not available (in IE), we need to escape colon otherwise it will cause javascript err.

Closes #<4664>.

### DIFF
--- a/tests/core/tools.js
+++ b/tests/core/tools.js
@@ -860,7 +860,7 @@
 			var escapedSelector = CKEDITOR.tools.escapeCss( selector );
 
 			// Check starts-with-number selector.
-			assert.areSame( escapedSelector, 'abd\\:efg', 'has-colon selector' );
+			assert.areSame( escapedSelector, 'abc\\:efg', 'has-colon selector' );
 		},
 
 		'test escapeCss - standard selector': function() {
--- a/tests/core/tools.js
+++ b/tests/core/tools.js
@@ -855,6 +855,14 @@
 			assert.areSame( escapedSelector, '\\30 ', 'only-one-number selector' );
 		},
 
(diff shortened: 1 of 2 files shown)


tests/core/tools.js
```diff
@@ -855,6 +855,14 @@
 			assert.areSame( escapedSelector, '\\30 ', 'only-one-number selector' );
 		},
 
+		'test escapeCss - has-colon selector': function() {
```
AdamTReineke, line 858: @limingli0707 This won't verify your added code if this runs in an environment where `CSS.escape` is defined. You would have to mock `CSS.escape` to be falsey during the test so it hits your code path.

limingli0707: Thanks @AdamTReineke. I just checked the test here, i think CSS.escape is not defined when the tests are run. They don't really test this part I think ```			if ( window.CSS && CSS.escape ) {
				return CSS.escape( selector );
			}```
The other test case` test escapeCss - starts-with-number selector` for example. I'm following the same. But you are right, ideally the test should cover all the cases including the `CSS.escape is defined`

limingli0707:
@ckedi

limingli0707:
closing this PR since this is target to major. https://github.com/ckeditor/ckeditor4/pull/4665 is the same one which targets to master (minor release)

Issue #4664, opened 2021-04-30
ID with colon causes JavaScript runtime err in IE. 


limingli0707:
## Type of report

Bug

## Provide detailed reproduction steps (if any)

Almost the same bug as https://github.com/ckeditor/ckeditor4/issues/681. Just in our case the character needs to be escaped is colon.
### Expected result

*What is the expected result of the above steps?*
no javascript syntax err

### Actual result

*What is the actual result of the above steps?*
has javascript syntax err
## Other details

Almost the same bug as https://github.com/ckeditor/ckeditor4/issues/681
The problem is in the implementation of CKEDITOR.tools.escapeCss() which doesn’t return the expected result in IE11 because IE11 doesn't support CSS.escape(). So we need to escape characters in IE manually, in our case our id could have colons. so the fix is addressing the issue by escaping colon when CSS.escape is not available.

* Browser: FF, IE 11, Edge, Chrome (although only IE gives exception)
* OS: windos macOS
* CKEditor version: 4.14 (i think it happens in the latest version as well)
* Installed CKEditor plugins: …

limingli0707:
FYI, I'm putting a fix here: https://github.com/ckeditor/ckeditor4/pull/4666. Thanks!

sculpt0r:
Hi @limingli0707 ,
Thank you for the report. I have to mark it as a duplicate and close this issue.

Please track #681 for updates.

PR #4665, opened 2021-04-30
Fix for colon in ID is not escaped since IE doesn't support CSS escape

limingli0707:

Bug fix.
issue here: https://github.com/ckeditor/ckeditor4/issues/4664

### This PR contains

- [x] Unit tests
- [  ] Manual tests

```
* [#<4664>](https://github.com/ckeditor/ckeditor4/issues/<4664>): [IE] : escape colon in id selector in IE
```

## What changes did you make?

Minor change in the tool.js > escapeCss(). When CSS.escape is not available (in IE), we need to escape colon otherwise it will cause javascript err. I saw we already did very similar escape for leading numbers due to the incomplete implementation of CKEDITOR.tools.escapeCss(). I'm just following the same pattern to escape colon.

Closes #<4664>.

### DIFF
--- a/core/tools.js
+++ b/core/tools.js
@@ -1552,7 +1552,8 @@
 				return '\\3' + selector.charAt( 0 ) + ' ' + selector.substring( 1, selector.length );
 			}
 
-			return selector;
+			//escape colon
+			return selector.replace(/:/g, "\\:");
 		},
 
 		/**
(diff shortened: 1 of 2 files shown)


limingli0707:
the change is exactly the same as https://github.com/ckeditor/ckeditor4/pull/4663. Only difference is this is for minor release, but the other is for the major release.
Should I just keep this one?

Issue #681, opened 2017-07-26
table elements (td, tr, th, ..) with an id that starts with dot (.) causes JavaScript runtime err


cschettle:
## Are you reporting a feature or a bug?
Bug

## [Check if the issue is already reported](http://docs.ckeditor.com/#!/guide/dev_issues_readme-section-avoid-duplicates)

*Put all reference links here…*

## Provide detailed reproduction steps (if any)

1. In the html for a table add an id to a one of the table cells that starts with dot.  for example make the td element have an id=".amt"
2. Run with IE 11 and press F12 to get to the debugger
3. Select the table cell in the editor
4.  You will see the Syntax error (in the console window) (see attached) IE11_syntax_err.PNG

### Expected result

*What is the expected result of the above steps?*
The table cell is selected (highlighted)

### Actual result

*What is the actual result of the above steps?*
Only IE 11 throws this exception :
 0x800a139e - JavaScript runtime error: SyntaxError
in ckeditor.js (minified version) or in dom\element.js in the source version
in findOne (or also find) at line 4024
## Other details
The syntax error is a result of a bad CSS selector being returned from 
getContextualizedSelector( this, selector )  (see element.js) While FF and Chrome do not throw an exception, the selector does would not work properly.  The selector produced using the attached table.htm file (and selecting any cell with an id=.amt.3 is
    #.amt.3 td, #.amt.3 th, #.amt.3 tr, #.amt.3 tbody, #.amt.3 table

You can test this selector using an online tool like https://try.jsoup.org/
I have attached the following files:
table.txt  html file that contains table with cells that have id's with dot
IE11_syntax_err.PNG shows err in IE11 dev tools with ckeditor minified version
IE11_syntax_err_src.PNG shows err in IE11 dev tools with ckeditor source
bad_selector.PNG shows value of selector returned from getContextualizedSelector which is called from find and findOne in dom\element.js
[table.txt](https://github.com/ckeditor/ckeditor-dev/files/1181038/table.txt)
![bad_selector](https://user-images.githubusercontent.com/13666205/28688600-3ce42320-72c7-11e7-8244-82d2abe26302.PNG)
![ie11_syntax_err](https://user-images.githubusercontent.com/13666205/28688605-42183eb2-72c7-11e7-81a8-844ea2869a3e.PNG)
![ie11_syntax_err_src](https://user-images.githubusercontent.com/13666205/28688610-459281ec-72c7-11e7-89a5-86987c491fd1.PNG)

[...]

Issue #641, opened 2017-07-12
UploadImage Plugin Widgets not working in IE, Opera, Safari, PhantomJS


b2m:
## Are you reporting a feature or a bug?

Bug

## [Check if the issue is already reported](http://docs.ckeditor.com/#!/guide/dev_issues_readme-section-avoid-duplicates)

None found

## Provide detailed reproduction steps (if any)

1. Use the Inline Editor (http://sdk.ckeditor.com/samples/inline.html).
2. Use an id with **commas** for the editor div (`id="introduction,first"`, valid in HTML 5 https://www.w3.org/TR/html5/dom.html#the-id-attribute).
3. Try uploading an image using Drag & Drop or via paste with embeded base64 image.

### _Reproduction steps in CKEditor test:_

1. Use test: `/tests/plugins/uploadwidget/manual/image.md` on **Safari** or other mentioned browser.
2. Edit file: `/tests/plugins/uploadwidget/manual/__template__.html` with followed things:
  * HTML: change `textarea` into `div` and add `contenteditbale="true"`
  * HTML: change `id` to `edito,r1` (something with comma)
  * SCRIPT: change `CKEDITOR.replace` into `CKEDITOR.inline`, add in config `extraPlugin: floatingspace`, modify `name` to `edito,r1`.
3. Try drag and drop image into editor.
4. Optional, if you wish to test it on Chrome: overwrite `CSS.escape`. Before initializing of editor in point 2 add `CSS.escape = undefiend`

### Expected result

Image is uploaded and inserted into editor.

### Actual result

Image is uploaded but not inserted into editor.

## Other details
 
* Browser: IE, Opera, Safari, PhantomJS
* CKEditor version: 4.5 and http://sdk.ckeditor.com/samples/inline.html
* Plugins: uploadimage

## Deeper analysis

The reason is the usage of the experimental *window.CSS.escape* feature (https://developer.mozilla.org/en-US/docs/Web/API/CSS/escape) that is used to generate CSS selectors to identify CKEditor Widgets. You might enhance the alternative escaping in CKEDITOR.tools.escapeCSS` to also escape commas.

msamsel:
Hi,

I'm not able to reproduce problematic situation. I test it on Safari on MacOS with this code: https://codepen.io/anon/pen/JJxNeE?editors=1010. I use image widget to insert picture to editor. And I'm able to resize it properly.
[...]

PR #4666, opened 2021-05-03
Fix element id is not escaped properly in IE

limingli0707:

Bug fix.
issue here: https://github.com/ckeditor/ckeditor4/issues/681

### This PR contains

- [x] Unit tests
- [x] Manual tests

```
* [#<4664>](https://github.com/ckeditor/ckeditor4/issues/681): [IE] : escape colon in id selector in IE
```

## What changes did you make?

Minor change in the tool.js > escapeCss(). When CSS.escape is not available (in IE), we need to escape colon otherwise it will cause javascript err. I saw we already did very similar escape for leading numbers due to the incomplete implementation of CKEDITOR.tools.escapeCss(). I'm just following the same pattern to escape colon.

Closes https://github.com/ckeditor/ckeditor4/issues/681
Closes https://github.com/ckeditor/ckeditor4/issues/641


limingli0707:
This is a known issue, ppl reported before: https://github.com/ckeditor/ckeditor4/issues/681. For us, the ID could have colon so we need to improve the escapeCSS function.
I also make the PR against the master branch for the upcoming minor release to get it fixed sooner :)
Thanks!

sculpt0r:
Thank you :)

limingli0707:
@sculpt0r I have updated the PR with a general solution. Actually that is the solution that MDN suggested too https://developer.mozilla.org/en-US/docs/Web/API/CSS/escape
Please let me know if you have any questions, i think it should fix all the IE issues :)

sculpt0r (changes requested):
Hi @limingli0707 
We are happy that you want to contribute to CKEditor4 and help to make it even better. Thanks for this update.

We should start formalities. They are also important for us, to make sure everything is in right place and has a purpose. Also, sorry for the long list, but we need to be sure everything is working as expected and it won't fail anymore. I think it is also important for you as a developer.

- Please update the reference ticket, so now closing section will point to the original issue: https://github.com/ckeditor/ckeditor4/issues/681
- Please update the proposed changelog, so it will match wit the resolution
- I see that you used polyfill copied from https://github.com/mathiasbynens/CSS.escape/blob/master/css.escape.js It is important because we need to remember to be fully transparent with licenses. This one is based on MIT so, I assume that will be no problem, but please @f1ames - could you say a word about this?

As for the code:
- I know you adjust the new unit test naming to the existing convention in `tools.js`. However, as for the test naming, please stick to such convention: `what is testing, in what circumstances, what is expected`. It helps a lot to understand such a test whenever it fails. So for this particular case, it could be: `test escapeCss escaped colon in the css selector`. Feel free to propose anything that comes to your mind and will match with this convention.
- Please add at least one unit test that verifies selector with the dot `.` character, as it is described in the original issue.
- Please add other unit tests that confirm, that the current flow is working properly with the specification (https://drafts.csswg.org/cssom/#serialize-an-identifier) even if this is still a draft.
- It will be nice to have a manual test with a test scenario like in the original issue. Could you also add one?
- I know it copied as it is from quite a reliable source, but since it could be integrated into our code base, please spend a while refactoring it, to match our [code style](https://github.com/ckeditor/ckeditor4/blob/major/dev/docs/codestyle.md). It is important to keep all code base consistent as much as possible.

limingli0707:
Hi @sculpt0r Thanks for your input! I have addressed the comments, would you like to take a look at it again? Thanks!

sculpt0r:
Rebase on newest master

core/tools.js
```diff
@@ -38,6 +38,81 @@
...
+				// (U+FFFD).
+				if ( codeUnit == 0x0000 ) {
+					result += '\uFFFD';
+					continue;
+				}
+
+				if (
+					// If the character is in the range [\1-\1F] (U+0001 to U+001F) or is
+					// U+007F, […]
+					(codeUnit >= 0x0001 && codeUnit <= 0x001F) || codeUnit == 0x007F ||
```
sculpt0r, line 66: All of those range checks could be replaced with extracted helper function inside this function scope.
```js
function isInRange(value, min, max) {
				return value <= min && value >= max;
			}
```

Then all if statements could be simplified to `isInRange( codeUnit, 0x0030, 0x0039 )`. Of course with proper values, so in this case it will be:
```js
( isInRange( codeUnit, 0x0001, 0x001F  ) ) || codeUnit == 0x007F ||
```

After this change, it will be also easier to simplify comments and make the code a little bit more compact.

Finally it could be something like this:

```js
if (
	// If the character is in the range [\1-\1F] or is U+007F
	isInRange( codeUnit, 0x0001, 0x001F ) || codeUnit == 0x007F ||
	// If the character is the first character and is in the range [0-9]
	( index == 0 && isInRange( codeUnit, 0x0030, 0x0039 ) ) ||
	// If the character is the second character and is in the range [0-9]
	// and the first character is a `-` (U+002D)
	( index == 1 && isInRange( codeUnit, 0x0030, 0x0039 ) && firstCodeUnit == 0x002D )
) {
	// https://drafts.csswg.org/cssom/#escape-a-character-as-code-point
	result += '\\' + codeUnit.toString( 16 ) + ' ';
	continue;
}
```

or even without comments if we use more meaningful const values

```js
if ( isInRange( codeUnit, UTF16.SOH, UTF16.SEPARATOR ) || codeUnit == UTF16.DEL ||
	( index == 0 && isInRange( codeUnit, UTF16.ZERO, UTF16.NINE ) ) ||
	( index == 1 && isInRange( codeUnit, UTF16.ZERO, UTF16.NINE ) && firstCodeUnit == UTF16.HYPHEN )
) {
	// https://drafts.csswg.org/cssom/#escape-a-character-as-code-point
	result += '\\' + codeUnit.toString( 16 ) + ' ';
	continue;
}
```
Although I'm not so happy with const names for `0x0001` and `0x001F`. If you have any better idea or think they should stay as they are - please express that in the next commits.

Maybe still not great, but more readable.

core/tools.js
```diff
@@ -38,6 +38,81 @@
...
+					// there is no second character, […]
+					index == 0 &&
+					length == 1 &&
+					codeUnit == 0x002D
+				) {
+					result += '\\' + string.charAt(index);
+					continue;
+				}
+
+				// If the character is not handled by one of the above rules and is
```
sculpt0r, line 94: After introduce const values for UTF-16 values and helper function, comments could be simplified a lot. Please, consider even remove some of them.

tests/core/tools.js
```diff
@@ -855,6 +855,47 @@
 			assert.areSame( escapedSelector, '\\30 ', 'only-one-number selector' );
 		},
 
+		'test escapeCss - escaped colon in the css selector': function() {
```
sculpt0r, line 858: We usually put reference to an issue which causes us to add new tests:
```suggestion
		// (#681)
		'test escapeCss - escaped colon in the css selector': function() {
```

Please add such comment to all new tests.

tests/core/tools/manual/escapeCss.html
```diff
@@ -0,0 +1,52 @@
+<textarea id="editor1">
```
sculpt0r, line 1: Please lowercase file name. I know that camel case convention is useful, but there are edge cases where our test runner can have problem with such files. Because of that, its more like informal convention, that all test files are named lower cased.

tests/core/tools/manual/escapeCss.md
```diff
@@ -0,0 +1,11 @@
+@bender-tags: 4.16.1, bug, 681
```
sculpt0r, line 1: Please lowercase file name. I know that camel case convention is useful, but there are edge cases where our test runner can have problem with such files. Because of that, its more like informal convention, that all test files are named lower cased.

limingli0707:
Hi @sculpt0r, I just rebased the latest master. Btw I noticed that you just linked the issue 4681, i think the issue is 681. Thanks a lot!

sculpt0r (changes requested):
I added a few comments.
Besides those, I'm glad that you add some extra unit tests for edge cases :+1: 

Please, don't commit anything new while you are not assigned to PR. It leads to the situation when I'm not sure if the code doesn't change after a while... Also, be sure that you are finished when you ask about another review. Be patient. It's better to take a break and return to PR after a day, than sending unpolished changes or adding changes while someone else is reviewing.

So, be sure you are assigned to PR until your work is done :)

Overall, very good job :+1: We just need to polish this solution.

limingli0707:
Hi @sculpt0r Thank you very much for your input! I am trying to assign myself to the PR but I am not able to see the action/option. I am guessing I might miss some permission to do that. Do you have any thoughts?

sculpt0r:
Hi @limingli0707 
If you don't see `As

sculpt0r:
OK - seems that this option is permitted for people outside the team. So I will assign you manually each time I send a review, and assign myself after your send a review request and I started to work on it.

limingli0707:
Hi @sculpt0r, thanks a lot for your suggestion!!  I have addressed the code review comments and now it's much more readable and clean. Thanks!!

Please take a look when you have a chance.

sculpt0r:
Rebase & fix doubled entry in CHANGELOG due to my rebasing.

core/tools.js
```diff
@@ -38,6 +38,78 @@
...
+
+			while ( ++index < selectorLength ) {
+				codeUnit = stringifiedSelector.charCodeAt( index );
+
+				if ( codeUnit == UTF16.NULL ) {
+					result += '\uFFFD';
+					continue;
+				}
+
+				if (
```
sculpt0r, line 80: Good job! Let's make all of new line conditions with the same tab indent, and extract one comparision to upper line:
```js
if ( codeUnit == UTF16.DELETE ||
	isInRange( codeUnit, UTF16.START_OF_HEADING, UTF16.UNIT_SEPARATOR ) ||
	( index == 0 && isInRange( codeUnit, UTF16.DIGIT_ZERO, UTF16.DIGIT_NINE ) ) ||
	( index == 1 && isInRange( codeUnit, UTF16.DIGIT_ZERO, UTF16.DIGIT_NINE ) && firstCodeUnit == UTF16.HYPHEN_MINUS )
) {
```

core/tools.js
```diff
@@ -38,6 +38,78 @@
...
+					HYPHEN_MINUS: 0x002D,
+					LOW_LINE: 0x005F,
+					START_OF_LATIN_1_SUPPLEMENT: 0x0080,
+					CAPITAL_LETTER_A: 0x0041,
+					CAPITAL_LETTER_Z: 0x005A,
+					SMALL_LETTER_A: 0x0061,
+					SMALL_LETTER_Z: 0x007A
+				};
+
+			// Helper function to check if value is in the range
```
sculpt0r, line 67:
```suggestion
			// Helper function to check if value is in the range.
```

core/tools.js
```diff
@@ -1546,13 +1618,7 @@
...
 				return CSS.escape( selector );
 			}
-
-			// Simple leading digit escape.
-			if ( !isNaN( parseInt( selector.charAt( 0 ), 10 ) ) ) {
-				return '\\3' + selector.charAt( 0 ) + ' ' + selector.substring( 1, selector.length );
-			}
-
-			return selector;
+			return cssEscapeForIE(selector);
```
sculpt0r, line 1621:
```suggestion

			return cssEscapeForIE( selector );
```

tests/core/tools.js
```diff
@@ -855,6 +855,52 @@
 			assert.areSame( escapedSelector, '\\30 ', 'only-one-number selector' );
 		},
 
+		// (#681)
+		'test escapeCss - escaped colon in the css selector': function() {
+			var selector = 'abc:def';
```
sculpt0r, line 860: Please group all var declarations in unit tests - the same you have done in `tools.js`.

tests/core/tools/manual/escapecss.md
```diff
@@ -0,0 +1,11 @@
+@bender-tags: 4.16.2, bug, 681
+
+1. Select the table cell in IE and open developer console.
+
+   ## Expected
```
sculpt0r, line 5: Our [code style](https://github.com/ckeditor/ckeditor4/blob/major/dev/docs/codestyle.md#31-manual-tests) define `Expected` and `Unexpected` as bolded values - not headers.

limingli0707:
Thanks @sculpt0r ! I just rebased and fixed the doubled entry in CHANGELOG. :) 
I just noticed that we had a release 14.6.1 last week, just want to double-check will the fix be merged to 14.6.2 instead? 
I saw we normally release every 2-3 months, I am assuming 14.6.2 is around August?
Thanks a lot!!

tests/core/tools/manual/escapecss.md
```diff
@@ -0,0 +1,11 @@
+@bender-tags: 4.16.2, bug, 681
+
+1. Select the table cell in IE and open developer console.
```
sculpt0r, line 3: Please, separate this step on two separate one:
- opening dev console
- selecting cells

If we want to restrict this test only to IE, all other browsers should be ignored in `html` file. For example, it is done like:
https://github.com/ckeditor/ckeditor4/blob/9c09a211f287e135223d4d2056959106d389c84b/tests/core/ckeditor/manual/hcdetection.html#L24-L26

Please adjust `html` file of this test to ignore non-IE browsers. Also, you can skip IE part in the scenario step description.

tests/core/tools/manual/escapecss.html
```diff
@@ -0,0 +1,52 @@
...
+      <td id=".trail.2" nowrap="true" style="width: 1%; font-family: &quot;Times New Roman&quot;, Times, serif; font-size: 10pt; margin-left: 0pt;">&nbsp;</td>
+      <td id=".lead.3" style="width: 1%; font-family: &quot;Times New Roman&quot;, Times, serif; font-size: 10pt;">&nbsp;</td>
+      <td id=".symb.3" style="width: 1%; font-family: &quot;Times New Roman&quot;, Times, serif; font-size: 10pt;">&nbsp;</td>
+      <td id=".amt.3" style="width: 12%; text-align: right; font-family: &quot;Times New Roman&quot;, Times, serif; font-size: 10pt; margin-left: 0pt;"><em data-td-fact-id="17716073" style="font-style: normal; font-weight: inherit;">171,362</em></td>
+      <td id=".trail.3" nowrap="true" style="width: 1%; font-family: &quot;Times New Roman&quot;, Times, serif; font-size: 10pt; margin-left: 0pt;">&nbsp;</td>
+    </tr>
+  </table>
+</textarea>
+<script>
+	CKEDITOR.replace('editor1', {
```
sculpt0r, line 49:
```suggestion
	CKEDITOR.replace( 'editor1', {
```

tests/core/tools/manual/escapecss.html
```diff
@@ -0,0 +1,52 @@
...
+		allowedContent: true
+	});
```
sculpt0r, line 51:
```suggestion
	} );
```

sculpt0r (changes requested):
Looks good :+1: Thank you for the fast resposnse.

I left a few suggestions and inline comments - for code style mainly.

Also, can you confirm that the manual test is working correctly for you?

sculpt0r:
> I just noticed that we had a release 14.6.1 last week, just want to double-check will the fix be merged to 14.6.2 instead?
> I saw we normally release every 2-3 months, I am assuming 14.6.2 is around August?

Yes, if we complete this PR before another release - it will be `4.16.2`. The timing is more or less regularly and sometimes depends on how many issues we fixed.

limingli0707:
Hi @sculpt0r, thank you so much for your review! Yes, the manual test is working correctly for me. I have addressed the code review comments. Would you like to take a look again? Thanks!!

sculpt0r (changes requested):
Hi, @limingli0707
Good job with all changes. I've added some polishing comments. Looks like, after those - it could be merged :+1: 

Beside that. Could you create another manual test that covers the case from https://github.com/ckeditor/ckeditor4/issues/641? It appears that your PR also fix this issue :+1: 

Also, please make a `pull` because I rebased this PR on the newest `master` branch.

limingli0707:
Thank you so much @sculpt0r . I have learned a lot from you!! 
I have addressed all the comments and please take a look again when you have a chance.
Regarding the other issue: #641 , I am not able to reproduce it. And I checked that the CSS.escape is supported in Safari. I saw the issue opened 4 years ago so I'm guessing maybe it was supported at that time. :)

limingli0707:
Sorry about pushing some extra commits after requesting a review. I just realized that some commits didn't get pushed correctly. Hope you don't mind :)

sculpt0r (approved):
Hi @limingli0707 
LGTM :tada: :+1: 

Everything looks good now. Please take a look at some final steps I've made:
- Licence update - because we are basing on someone else solution: https://github.com/ckeditor/ckeditor4/pull/4666/commits/a8399c30fb46e7febc690f425b62128cff447088
- I added your nick to the changelog :wink:  : https://github.com/ckeditor/ckeditor4/pull/4666/commits/869ddbfba300b202319af82fdb38629dc3f827f2
- but also added one entry that was lost after our rebases: https://github.com/ckeditor/ckeditor4/pull/4666/commits/869ddbfba300b202319af82fdb38629dc3f827f2
- Small correction for `if` statement: https://github.com/ckeditor/ckeditor4/pull/4666/commits/f447cb4639240401fc92f0565e8eff6b174a7e1a

However, I must say, that the manual tests weren't so good. Especially the second one added later. They don't verify the issues at all, because they didn't run. There were missed plugins, steps that didn't reproduce original issue scenarios, expected messages that weren't adequate to real observations. Please take a look at them now: https://github.com/ckeditor/ckeditor4/pull/4666/commits/7ee7ebd8b4aee4d335aa94f6d530be63fddce036 and https://github.com/ckeditor/ckeditor4/pull/4666/commits/c257a73b63c30c9c4734529a3b32aa0d12d347ec

Overall: very good job! :+1:  Thank you for your contribution, time and involvement. Also, your responses were so fast that sometimes it was a pleasure to have such fluent cooperation.

limingli0707:
Thank you so much for your help and guidance @sculpt0r ! It was a pleasure to work with you on this RP and I have learned a lot!

PR #4688, opened 2021-05-10
Fix for broken dialogs after opening the same one twice

### LINKED ISSUE #3638: Trying to open the dialog for the second time when it's already opened causes dialog cover to stay after dialog closing
## Type of report

Bug

1. Go to [CDN](http://cdn.ckeditor.com/4.13.0/full-all/samples/)
2. Execute `CKEDITOR.instances.editor.execCommand( 'link' )` twice 
3. Close the dialog

### Expected result

Dialog is closed and editor is available

### Actual result

Dialog cover stays after closing dialog, editor is unavailable

## Other details

This behaviour was introduced in #2413. Before it was also broken, but in different way - after calling dialog for the second time it has been hiding behind cover.

* Browser: all
* OS: macOS
* CKEditor version: `4.13.0`
* Installed CKEditor plugins: `dialog`

Comandeer:

Bug fix

### This PR contains

- [x] Unit tests
- [x] Manual tests

```
* [#3638](https://github.com/ckeditor/ckeditor4/issues/3638): Fixed: opening the same dialog twice causes it to become hidden under the dialog's page cover.
```

## What changes did you make?

I've just added a simple check if the current dialog (`CKEDITOR.dialog._.currentTop`) is the same as the one being opened. If it's the case, the positioning logic is skipped.

Additionally I've polished some tests to always close dialogs.

Closes #3638.

### DIFF
--- a/plugins/dialog/plugin.js
+++ b/plugins/dialog/plugin.js
@@ -905,7 +905,8 @@ CKEDITOR.DIALOG_STATE_BUSY = 2;
 				CKEDITOR.dialog._.currentTop = this;
 				this._.parentDialog = null;
 				showCover( this._.editor );
-			} else {
+			} else if ( CKEDITOR.dialog._.currentTop !== this ) {
+				// Do the magic with z-indexes only if the current dialog is not already visible (#3638).
 				this._.parentDialog = CKEDITOR.dialog._.currentTop;
 
 				var parentElement = this._.parentDialog.getElement().getFirst();
--- a/plugins/dialog/plugin.js
+++ b/plugins/dialog/plugin.js
@@ -906,7 +906,7 @@ CKEDITOR.DIALOG_STATE_BUSY = 2;
 				this._.parentDialog = null;
 				showCover( this._.editor );
 			} else if ( CKEDITOR.dialog._.currentTop !== this ) {
-				// Do the magic with z-indexes only if the current dialog is not already visible (#3638).
+				// Reposition the new dialog only if the current dialog is not already on the top (#3638).
 				this._.parentDialog = CKEDITOR.dialog._.currentTop;
 
 				var parentElement = this._.parentDialog.getElement().getFirst();
--- a/tests/plugins/dialog/plugin.js
+++ b/tests/plugins/dialog/plugin.js
@@ -81,6 +81,16 @@
 	bender.editor = {};
 
 	bender.test( {
+		tearDown: function() {
+			var dialog = CKEDITOR.dialog.getCurrent();
+
+			while ( dialog ) {
+				dialog.hide();
+
+				dialog = CKEDITOR.dialog.getCurrent();
+			}
+		},
+
 		// (#4262)
 		'test stylesLoaded is not polluting global context': function() {
 			assert.isUndefined( window.stylesLoaded );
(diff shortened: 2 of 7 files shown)


sculpt0r:
Rebased on the newest master

tests/plugins/dialog/plugin.js
```diff
@@ -81,6 +81,16 @@
 	bender.editor = {};
 
 	bender.test( {
+		tearDown: function() {
+			var dialog = CKEDITOR.dialog.getCurrent();
```
sculpt0r, line 85: Since this tearDown looks the same as the one used in another test (`positioning`) and we already have `_helpers` for dialog tests, maybe we could move this code to a separate function with proper naming, like `hideAllDialogs`?

tests/plugins/dialog/positioning.js
```diff
@@ -13,8 +13,11 @@
 		},
 		tearDown: function() {
 			var dialog = CKEDITOR.dialog.getCurrent();
-			if ( dialog ) {
+
+			while ( dialog ) {
```
sculpt0r, line 17: Please take a look at comment to `tearDown` in `tests/plugins/dialog/plugins.js`

tests/plugins/dialog/positioning.js
```diff
@@ -218,8 +221,39 @@
...
 
+			editorBot.dialog( 'link', function() {
+				editorBot.dialog( 'link', function( dialog ) {
+					var cover = CKEDITOR.document.findOne( '.cke_dialog_background_cover' ),
+						coverZIndex = parseInt( cover.getStyle( 'z-index' ), 10 ),
+						dialogZIndex = parseInt( dialog.getElement().getStyle( 'z-index' ), 10 );
+
+					dialog.getButton( 'cancel' ).click();
+
+					assert.isTrue( dialogZIndex > coverZIndex, 'Dialog should be above the cover' );
```
sculpt0r, line 238: We don't have anything similar to [chai isAbove](https://www.chaijs.com/api/assert/#method_isabove), so maybe we could at least give both zIndexes values to the assert message? In case of fail there will be already some results visible besides `expected true but was false`.

plugins/dialog/plugin.js
```diff
@@ -905,7 +905,8 @@ CKEDITOR.DIALOG_STATE_BUSY = 2;
 				CKEDITOR.dialog._.currentTop = this;
 				this._.parentDialog = null;
 				showCover( this._.editor );
-			} else {
+			} else if ( CKEDITOR.dialog._.currentTop !== this ) {
+				// Do the magic with z-indexes only if the current dialog is not already visible (#3638).
```
sculpt0r, line 909: I'm not 100% sure about this comment. I mean the part with zIndex. The entire `else...if` is prevented `if the current dialog is not already visible`. So besides zIndex operation we also have assignments to `parentDialog` and `currentTop`.

We could leave just the part with issue reference at the top, and move the zIndex part just before zIndex operation?

sculpt0r (changes requested):
Looks good and working fine. Just a few small suggestions from me.

Also, I made a small fix for white characters in `hidingcover.md` test and rebased - so make sure you pull all updates.

Comandeer:
I've updated tests according to the comments and slightly rephrase the comment inside the code – it should be more suitable now.

sculpt0r (approved):
LGTM :tada: 

Thank you for the corrections.

I made a single simple change myself: I bumped the CKEditor4 version in the manual test tag. Since the master branch is already frozen, we have no chance to release it with `4.16.1`.

Changelog added.

Waiting with merge till master will be available.

sculpt0r:
Blocked merge by frozen master branch

sculpt0r:
Rebased temporary on `next`.

PR #4718, opened 2021-05-18
Unify color handling

### LINKED ISSUE #4333: Color history incorrectly displays colors entered as name or in RGB format

### LINKED ISSUE #4592: Unify color handling
## Type of report

Task

See https://github.com/ckeditor/ckeditor4/issues/4351#issuecomment-805062200. Currently we are handling colors in a very convoluted way, with normalizing hexes without hashes and then adding hashes in several places. Probably we should rewrite the whole color handling logic to use `CKEDITOR.tools.color`.

## Other details

* Installed CKEditor plugins: colorbutton, colordialog

### LINKED ISSUE #4726: Ability to get the short hex from `CKEDITOR.tools.color` instance
## Type of report

Feature request

Currently [`CKEDITOR.tools.color#getHex()`](https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_tools_color.html#method-getHex) does not allow getting short version of hex.

We can introduce `shorten` parameter that would indicate that the short hex is wanted, e.g.

```javascript
var color = new CKEDITOR.tools.color( '#000000' );

console.log( color.getHex( true ) ); // '#000'
```

In case of the non-shortenable hexes, it should return the full hex:

```javascript
var color = new CKEDITOR.tools.color( '#CDAB65' );

console.log( color.getHex( true ) ); // '#CDAB65'
```

It should be also introduced for [`CKEDITOR.tools.color#getHexWithAlpha()`](https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_tools_color.html#method-getHexWithAlpha).

Comandeer:

Task

### This PR contains

- [x] Unit tests
- [x] Manual tests

```
Fixed Issues:

* [#4333](https://github.com/ckeditor/ckeditor4/issues/4333): Fixed: [Color Button](https://ckeditor.com/cke4/addon/colorbutton) history does not render correctly colors passed in non-hex format.

API Changes:

* [#4726](https://github.com/ckeditor/ckeditor4/issues/4726): Added `options` parameter to both [`CKEDITOR.tools.color#getHex()`](https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_tools_color.html#method-getHex) and [`CKEDITOR.tools.color#getHexWithAlpha()`](https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_tools_color.html#method-getHexWithAlpha) methods with `options.shorten` property to allow getting short hex from the given color.
```

## What changes did you make?

I've used `CKEDITOR.tools.color` wherever it was possible. Some places still need some love and refactor, but I'm not sure if it's worth it.

I've also introduced the ability to get short hex from `CKEDITOR.tools.color#getHex()` and `CKEDITOR.tools.color#getHexWithAlpha()` via a new parameter, `shorten`. However I'm not sure if it wouldn't be better to introduce `options` parameter. It would be more flexible 🤔

I've also changed the way in which colors in Color History are rendered.

Closes #4592.
Closes #4726.
Closes #4333.

### DIFF
--- a/plugins/colorbutton/plugin.js
+++ b/plugins/colorbutton/plugin.js
@@ -311,7 +311,7 @@
 								}
 							}, null, colorData );
 						} else {
-							setColor( color && '#' + color, colorName, history );
+							setColor( color && color, colorName, history );
 						}
 
 						// The colors may be duplicated in both default palette and color history. If user reopens panel
--- a/plugins/colorbutton/plugin.js
+++ b/plugins/colorbutton/plugin.js
@@ -529,7 +529,7 @@
 
 				// Replace 3-character hexadecimal notation with a 6-character hexadecimal notation (#1008).
 				// It also covers other cases, like colors with alpha channel (#4351).
-				return CKEDITOR.tools.normalizeHex( colorInstance.getHex() || '' );
+				return colorInstance.getHex() || '';
 			},
 
 			getAppliableColor: function( color ) {
--- a/plugins/colorbutton/plugin.js
+++ b/plugins/colorbutton/plugin.js
@@ -478,13 +478,15 @@
 	} );
 	ColorBox = CKEDITOR.tools.createClass( {
 		$: function( editor, colorData, clickFn ) {
-			var initialValue = colorData.color.getInitialValue();
+			var isValidColor = colorData.color._.isValidColor,
+				defaultLabel = colorData.color.getInitialValue();
 
 			this.$ = new CKEDITOR.dom.element( 'td' );
 
 			this.color = colorData.color;
 			this.clickFn = clickFn;
-			this.label = colorData.label || ColorBox.getColorName( editor, this.color ) || initialValue;
+			this.label = isValidColor ?
+				( colorData.label || ColorBox.getColorName( editor, this.color ) || defaultLabel ) : '';
 
 			this.setHtml();
 		},
@@ -530,6 +532,10 @@
 			},
 
 			getAppliableColor: function( color ) {
+				if ( !color._.isValidColor ) {
+					return '';
+				}
+
 				var initialValue = color.getInitialValue(),
 					isHexLike = initialValue.toLowerCase() ===
 						color.getHex( initialValue.length === 3 ).toLowerCase().substr( 1 ),
@@ -549,8 +555,8 @@
 			},
 
 			setHtml: function() {
-				var colorStr = this.color.getInitialValue(),
-					appliableColor = ColorBox.getAppliableColor( this.color );
+				var appliableColor = ColorBox.getAppliableColor( this.color ),
+					colorStr = appliableColor ? this.color.getInitialValue() : '';
 
 				this.getElement().setHtml( '<a class="cke_colorbox" _cke_focus=1 hidefocus=true' +
 						' title="' + this.label + '"' +
--- a/plugins/colorbutton/plugin.js
+++ b/plugins/colorbutton/plugin.js
@@ -386,7 +386,8 @@
 						colorName = parts[ 0 ],
 						colorCode = parts[ 1 ] || colorName,
 						colorLabel = parts[ 1 ] ? colorName : undefined,
-						box = new ColorBox( editor, { color: colorCode, label: colorLabel }, clickFn );
+						color = new CKEDITOR.tools.color( colorCode, colorCode ),
+						box = new ColorBox( editor, { color: color, label: colorLabel }, clickFn );
 
 					box.setPositionIndex( startingPosition + i, total );
 					output.push( box.getHtml() );
@@ -477,11 +478,13 @@
 	} );
 	ColorBox = CKEDITOR.tools.createClass( {
 		$: function( editor, colorData, clickFn ) {
+			var initialValue = colorData.color.getInitialValue();
+
 			this.$ = new CKEDITOR.dom.element( 'td' );
 
-			this.color = CKEDITOR.tools._isValidColorFormat( colorData.color ) ? colorData.color : '';
+			this.color = colorData.color;
 			this.clickFn = clickFn;
-			this.label = colorData.label || ColorBox.colorNames( editor )[ this.color ] || this.color;
+			this.label = colorData.label || ColorBox.colorNames( editor )[ initialValue ] || initialValue;
 
 			this.setHtml();
 		},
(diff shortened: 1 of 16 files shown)


plugins/colorbutton/plugin.js
```diff
@@ -503,22 +516,29 @@
...
-					// Browsers tend to represent transparent color as rgba(0, 0, 0, 0), so we need to filter out this value.
-					isTransparentColor = transparentRegex.test( color ),
-					colorInstance;
-
-				// For colors with alpha channel we need to use CKEDITOR.tools.color normalization (#4351).
-				if ( isAlphaColor && !isTransparentColor ) {
-					colorInstance = new CKEDITOR.tools.color( color );
-
-					return CKEDITOR.tools.normalizeHex( colorInstance.getHex() || '' ).replace( /#/g, '' );
+				var colorInstance = ( color instanceof CKEDITOR.tools.color ) ? color : new CKEDITOR.tools.color( color, '' ),
```
sculpt0r, line 519: API comment for this method defines `color` as a string. But now, we are verifying if it could be an instance of our `color` class. We should change API comments for that.

plugins/colorbutton/plugin.js
```diff
@@ -503,22 +516,29 @@
...
+					isTransparent = Number( colorInstance._.alpha ) === 0;
+
+				if ( isTransparent ) {
+					return 'transparent';
```
sculpt0r, line 523: Any somehow transparent color should be converted to 'transparent'? Because API comments say this method returns hex without '#'. The previous version seems to skip this case at all?

plugins/colorbutton/plugin.js
```diff
@@ -503,22 +516,29 @@
...
 				}
 
 				// Replace 3-character hexadecimal notation with a 6-character hexadecimal notation (#1008).
-				return CKEDITOR.tools.normalizeHex( '#' + CKEDITOR.tools.convertRgbToHex( color || '' ) ).replace( /#/g, '' );
+				// It also covers other cases, like colors with alpha channel (#4351).
+				return CKEDITOR.tools.normalizeHex( colorInstance.getHex() || '' );
```
sculpt0r, line 528: This function is already marked as deprecated. We already have some ambiguous `normalizations` around. The one is here in `ColorBox` and another in `tools`. So I'm starting to wonder - what is the normalization process because it leads to very different results. It is related to parsing values from CSS attributes. So maybe we should separate the so-called `normalization` and extracting value from the CSS attribute?

sculpt0r (changes requested):
It looks like the root issue is fixed now (https://github.com/ckeditor/ckeditor4/issues/4333) :+1: :+1: 

>However I'm not sure if it wouldn't be better to introduce options parameter. It would be more flexible

Well, it sounds like a good idea. When you look at some functions, that didn't use an object as an option. They end with lots of code just to provide compatibility later. For example:

https://github.com/ckeditor/ckeditor4/blob/8e5575f3c7cc25f28af67d8fa3eb96b7a62b34d7/core/editor.js#L1125-L1135

It looks for now that there won't be any more thing to add, but... you never know that - until you have to add another one :)

> I've also changed the way in which colors in Color History are rendered.

I found a little inconsistency here. When you explicitly type lowercase 6-HEX - it's lowercase in history. If you type it uppercase - it's uppercased.

Also, If we are allowing 6-HEX in history, why we even want them to be shortened? It looks like that creates some additional code that could be skipped? I know, we are allowing 3-HEX-es, but for sake of `unify` we could decide on one format to be displayed. Therefore, even 3-HEX will be converted to 6-HEX.

Overall I see, that a few '#' and `substr` disappears :+1:

plugins/colorbutton/plugin.js
Comandeer: Yes, the previous version returns incorrect hex value for transparent colors. This version returns `transparent` explicitly.

plugins/colorbutton/plugin.js
Comandeer: The issue is that removing this normalization will basically nuke tests… `CKEDITOR.tools.color` returns uppercase hexes and whole tests assume they are lowercase. So I'm not sure if it's worth it to refactor all tests.

Comandeer: Or not, nothing happened :thinking: Strange, because it suggests that there's yet another normalization on the way…

Comandeer:
> Also, If we are allowing 6-HEX in history, why we even want them to be shortened? It looks like that creates some additional code that could be skipped? I know, we are allowing 3-HEX-es, but for sake of `unify` we could decide on one format to be displayed. Therefore, even 3-HEX will be converted to 6-HEX.

The issue is that user could want to use the original format of the color (see http://localhost:1030/tests/plugins/colorbutton/config#tests%2Fplugins%2Fcolorbutton%2Fconfig%20test%20config.colorButton\_foreStyle%20with%20color%20attribute), not the normalized one. And the current flow de facto disables the ability to force normalization for history entries (it would require creating separate `ColorBox` class for history).

And that's why I'm so unhappy with the whole refactor: the code is still messy. And with changes from this PR it becomes even messier…

sculpt0r (changes requested):
I added a few requests for `color` class extensions and few other thoughts about passing the entire value. Also, I've made some diagram which helps me with the code flow (https://miro.com/app/board/o9J_lApbiDQ=/)


If my comments would appear to be invalid due to back incompatibility, I think we will be finishing with this only with a few detailed comments I also left. :crossed_fingers: .

Indeed there is a lot going on. It takes a while before I catch up how it works.

The last thing: I think we should look at [colordialog](https://github.com/ckeditor/ckeditor4/blob/70d7394134f0a20ae5be213c87a46e25401c5d82/plugins/colordialog/plugin.js) - there is also some magic with color validation and RegExps...

plugins/colorbutton/plugin.js
```diff
@@ -499,26 +512,34 @@
 			 * * HSL/HSLA colors (e.g. `hsl( 100, 50%, 20%)` or `hsla( 100, 50%, 20%, 10%)`).
 			 *
 			 * @private
-			 * @param {String} color
+			 * @param {String/CKEDITOR.tools.color} color
 			 * @returns {String} Returns color in hex format, but without the hash at the beginning, e.g. `ff0000` for red.
+			 * In case of transparent colors, it returns `'transparent'` string.
 			 */
 			normalizeColor: function( color ) {
```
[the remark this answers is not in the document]
Comandeer: I'm not sure about changing name. It's not used only for getting color suitable for `[style]` attribute, e.g. https://github.com/ckeditor/ckeditor4/blob/70d7394134f0a20ae5be213c87a46e25401c5d82/plugins/colorbutton/plugin.js#L180

plugins/colorbutton/plugin.js
```diff
@@ -499,26 +512,34 @@
...
+					return 'transparent';
 				}
 
 				// Replace 3-character hexadecimal notation with a 6-character hexadecimal notation (#1008).
-				return CKEDITOR.tools.normalizeHex( '#' + CKEDITOR.tools.convertRgbToHex( color || '' ) ).replace( /#/g, '' );
+				// It also covers other cases, like colors with alpha channel (#4351).
+				return colorInstance.getHex() || '';
+			},
+
+			getAppliableColor: function( color ) {
```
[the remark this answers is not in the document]
Comandeer: I thought that would be more coherent with existing `normalizeColor`. But there isn't any real need for that, it can be moved as a private function into `setHtml`.

Comandeer:
Rebased onto latest `major`.

f1ames:
After some talks with @Comandeer and @sculpt0r we have decided to freeze this PR for now. Any reasonable changes would require refactoring big part of the existing solution which would be really time consuming and due to other priorities we can't proceed with this ATM.

Issue #4728, opened 2021-05-25
Incorrect adding anchors 


KarolDawidziuk:
## Type of report

Bug

## Provide detailed reproduction steps (if any)

1. Go to [link](https://ckeditor.com/latest/samples/index.html).
1. Add bold text style to `world!` and add the anchor.
1. Select whole line of `Hello world!` and try add another anchor.

### Expected result

The whole line should have one anchor

### Actual result

First word `Hello` have one anchor and `world!` have two anchors.

![image](https://user-images.githubusercontent.com/18471998/119488308-ace00180-bd5a-11eb-8bc9-9a619bec31b1.png)

## Other details

* Browser: IE/Safari/Firefox/Chrome/Opera
* OS: macOS 11.1
* CKEditor version: 4.16.1
* Installed CKEditor plugins: link

jacekbogdanski:
Closed in #4733

Issue #3863, opened 2020-02-16
Anchor is duplicated when edited


AlexanderMur:
## Type of report

Bug

## Provide detailed reproduction steps (if any)

1. go to https://ckeditor.com/latest/samples/index.html
2. make some bold text
3. make it anchor
4. try to edit anchor - double click 
5. press ok

example: https://i.gyazo.com/fa442d930a3b59eec754ca0b5956b8b5.mp4

Every time when you anchor is edited with style like bold underline etc. - this makes a doubled anchor instead of just editing.

### Expected result
anchor is edited

### Actual result
anchor is doubled

## Other details

* Browser: Chrome 80.0.3987.106
* OS: Windows 10 1903
* CKEditor version: 4

Comandeer:
I can confirm the issue. Seems like internal `getSelectedAnchor` function assumes that text can be surrounded only by the link itself, not considering more complex situations like:

```html
<a><strong>Text</strong></a>
```

AlexanderMur:
How can I fix this without modifying source code?

deyankush:
Any updates on this defect? I am facing the same issue. This was working in v4.7. Since we migrated to v4.13 its stopped working.

Dumluregn:
Hello @deyankush, thank you for bringing to our attention that it's a regression. It was indeed introduced in `4.13.0`, which means we probably try to handle it in the near future.

vinaydhakre:
@Dumluregn - I am facing the same issue. Could you take this defect on priority so that we can get this fix ASAP?

Dumluregn:
Unfortunately we've just froze the branches before release, so this issue will have to wait for the next one. However if nothing unexpected happens we should be able to fix this for then.

vaibhav198989:
@Dumluregn when is the next release expected ?

deyankush:
> Hello @deyankush, thank you for bringing to our attention that it's a regression. It was indeed introduced in `4.13.0`, which means we probably try to handle it in the near future.

Is it possible to get a patch on 4.13?

Dumluregn:
@vaibhav198989 we do not have a date set, probably in a few months.
@deyankush we do not provide backwards patches, but if you have a commercial license you can reach [our support team](https://ckeditor.com/contact/) for workaround or help in your own development of this fix.

deyankush:
[...]

PR #4733, opened 2021-05-27
Fix: prevent duplicate anchors in text with styles

KarolDawidziuk:

Bug fix

### This PR contains

- [x] Unit tests
- [x] Manual tests

```
* [#4728](https://github.com/ckeditor/ckeditor4/issues/4728): Fixed: [Link](https://ckeditor.com/cke4/addon/link) Multiple anchors in one line and multi-line with text style.
* [#3863](https://github.com/ckeditor/ckeditor4/issues/3863): Fixed: [Link](https://ckeditor.com/cke4/addon/link) Multiple anchors in single word with text style.
```

## What changes did you make?

The main problem was the text with styles. ex. bolded word. 
I've created a function that search in selected range a existing or nested anchors and clean it to prevent doubled it.

Closes [#4728](https://github.com/ckeditor/ckeditor4/issues/4728),
Closes [#3863](https://github.com/ckeditor/ckeditor4/issues/3863)


plugins/link/dialogs/anchor.js
```diff
@@ -30,12 +30,38 @@ CKEDITOR.dialog.add( 'anchor', function( editor ) {
 			element = element.getParent();
 		}
 
+		// If anchor is exist and has some styles go to the closest parent <a> tag. (#3863)
```
f1ames, line 33:
```suggestion
		// If anchor exists and has any styles find the closest parent <a> tag. (#3863)
```

plugins/link/dialogs/anchor.js
```diff
@@ -76,6 +102,7 @@ CKEDITOR.dialog.add( 'anchor', function( editor ) {
 					if ( CKEDITOR.env.ie && CKEDITOR.env.version < 9 )
 						attributes[ 'class' ] = 'cke_anchor';
 
+					checkNestedAnchors( range );
```
f1ames, line 105: Can you add a comment here too which explains which issue it solves?

plugins/link/dialogs/anchor.js
```diff
@@ -30,12 +30,38 @@ CKEDITOR.dialog.add( 'anchor', function( editor ) {
...
+		}
+
 		if ( element && element.type === CKEDITOR.NODE_ELEMENT &&
 			( element.data( 'cke-real-element-type' ) === 'anchor' || element.is( 'a' ) ) ) {
 			return element;
 		}
 	}
 
+	// Checks if range contains some nested anchors and remove it.
+	function checkNestedAnchors( range ) {
```
f1ames, line 45: This functions removes all the anchors within the current range so the name is a little misleading. Maybe something like `removeAnchorsWithinRange()`?

f1ames: And with a descriptive name the comment above will not be needed.

f1ames: And with a descriptive name the comment above will not be needed.

plugins/link/dialogs/anchor.js
```diff
@@ -30,12 +30,38 @@ CKEDITOR.dialog.add( 'anchor', function( editor ) {
...
+		var newRange = range.clone();
+		newRange.enlarge( CKEDITOR.ENLARGE_ELEMENT );
```
f1ames, line 47: You can make it a one-liner `...clone().enlarge( ... );` and then merge with below vars block.

plugins/link/dialogs/anchor.js
```diff
@@ -30,12 +30,38 @@ CKEDITOR.dialog.add( 'anchor', function( editor ) {
...
+			bookmark = range.createBookmark();
+
+		while ( element ) {
+		// Remove anchor if it exist
+			if ( element.type === CKEDITOR.NODE_ELEMENT && element.getAttribute( 'data-cke-saved-name' ) ) {
+				element.remove( true );
+			}
+			element = walker.next();
+		}
+		// restore range
```
f1ames, line 61: This comment is not needed since the line below is self explanatory.

plugins/link/dialogs/anchor.js
```diff
@@ -30,12 +30,38 @@ CKEDITOR.dialog.add( 'anchor', function( editor ) {
...
+		var newRange = range.clone();
+		newRange.enlarge( CKEDITOR.ENLARGE_ELEMENT );
+
+		var walker = new CKEDITOR.dom.walker( newRange ),
+			element = newRange.collapsed ? newRange.startContainer : walker.next(),
+			// Create bookmark to save actual range before removing anchors
+			bookmark = range.createBookmark();
+
+		while ( element ) {
+		// Remove anchor if it exist
```
f1ames, line 55: This comment is not needed since the line below is self explanatory.

plugins/link/dialogs/anchor.js
```diff
@@ -30,12 +30,38 @@ CKEDITOR.dialog.add( 'anchor', function( editor ) {
...
 	}
 
+	// Checks if range contains some nested anchors and remove it.
+	function checkNestedAnchors( range ) {
+		var newRange = range.clone();
+		newRange.enlarge( CKEDITOR.ENLARGE_ELEMENT );
+
+		var walker = new CKEDITOR.dom.walker( newRange ),
+			element = newRange.collapsed ? newRange.startContainer : walker.next(),
+			// Create bookmark to save actual range before removing anchors
```
f1ames, line 51: This comment is not needed since the line below is self explanatory :wink:

tests/plugins/link/anchor.js
```diff
@@ -79,6 +79,46 @@
 					assert.areSame( anchor, dialog.getModel( editor ) );
 				} );
 			} );
+		},
+
+		// (#3863)
+		'test preserve duplicate anchors after editing word with custom style': function() {
```
f1ames, line 85:
```suggestion
		'test prevent duplicated anchors after editing a word with custom style': function() {
```

?

f1ames: Could you add another similar test with more nesting like e.g. `strong > i > span`?

tests/plugins/link/anchor.js
```diff
@@ -79,6 +79,46 @@
...
+					dialog.getButton( 'ok' ).click();
+
+					assert.areSame( anchor, dialog.getModel( editor ) );
+				} );
+
+			} );
+		},
+
+		// (#4728)
+		'test prevent duplicate anchors in selected multi words with styles': function() {
```
f1ames, line 109:
```suggestion
		'test prevent duplicated anchors after editing multiple words with styles': function() {
```

Maybe something like :point\_up:

tests/plugins/link/anchor.js
```diff
@@ -79,6 +79,46 @@
...
+				range.select();
+
+				bot.dialog( 'anchor', function( dialog ) {
+					dialog.setValueOf( 'info', 'txtName', 'duplicate-test' );
+					dialog.getButton( 'ok' ).click();
+
+					assert.areSame( anchor, dialog.getModel( editor ) );
+				} );
+
+			} );
```
f1ames, line 105: This test could be written the same way as [the 2nd one you have added](https://github.com/ckeditor/ckeditor4/pull/4733/files#diff-79b36452737f4e4748cde2f72e7ab1eb42745722781f4214899d2fa3ced4584fR109) (which is simpler I guess). Was there any particular reason for doing it like this?

tests/plugins/link/anchor.js
```diff
@@ -79,6 +79,46 @@
...
+				range.selectNodeContents( textNode );
+				range.select();
+
+				bot.dialog( 'anchor', function( dialog ) {
+					dialog.setValueOf( 'info', 'txtName', 'duplicate-test' );
+					dialog.getButton( 'ok' ).click();
+
+					assert.areSame( anchor, dialog.getModel( editor ) );
+				} );
+
```
f1ames, line 104: Unnecessary empty line :wink:

f1ames (changes requested):
Looking good :+1: 

Since this PR fixes multiple anchor issues it would be good to provide kind of generic manual unit test where one can experiment with anchors. You could go with heavily styled content and simple steps to create an anchor over selection having style text inside.

I recall we talked F2F about issue with adding anchor to multiline selection - have you extracted this to a separate issue already because it's not covered here from what I see?

KarolDawidziuk:
I added more unit tests and created one general manual test for more specific tests.

"I recall we talked F2F about issue with adding anchor to multiline selection - have you extracted this to a separate issue already because it's not covered here from what I see?"
Yes, but this problem is with non-standard text styles, so I figured we wouldn't need such a test. I added such a situation in one of the tests for greater coverage :)

sculpt0r:
Rebased on newest master.

plugins/link/dialogs/anchor.js
sculpt0r: TBH I don't see a way how this could be inlined. `enlarge` returns `undefined`, so inlining it makes `newRange` undefined. I'm marking this conversation as resolved.

tests/plugins/link/manual/anchorcustomstyles.md
```diff
@@ -0,0 +1,18 @@
+@bender-tags: 4.16.1, bug, link, 4728
+@bender-ui: collapsed
+@bender-ckeditor-plugins: link, toolbar, wysiwygarea, basicstyles, sourcearea, clipboard, enterkey, link, list, liststyle, tabletools, tableselection, undo, format
```
sculpt0r, line 3:
```suggestion
@bender-ckeditor-plugins: link, toolbar, wysiwygarea, basicstyles, sourcearea, clipboard, enterkey, link, list, liststyle, tabletools, tableselection, undo, format, elementspath
```

I would add `elementspath` to plugins. It shows what kind of element is in current caret position, which is important from the perspective of this issue.

sculpt0r (changes requested):
I resolved most of the inline comments from @f1ames since you correct them.

However, I found that still one of them has no response: https://github.com/ckeditor/ckeditor4/pull/4733#discussion_r641538887

Also, I'm not sure what exactly you were talking about, but from this comment: 
https://github.com/ckeditor/ckeditor4/pull/4733#issuecomment-850994301 I concluded that we should have a separate issue for anchors applied to multiline selection? Could you explain it a little bit more? I see one unit test for that case for now.

I also was able to fail during the manual test you provided with such scenarios:
- Scenario one
  - select `ading styl` from the sentence `Test heading styled words`
  - add anchor with name e.g. 'a'
  - we have two anchors in text with the same `id` which is not the correct situation, but we can extract it to another issue eventually
  - also we have two anchors, while when we select any two non-styled words, it's only one anchor for both of them
- Scenario two
  - select `ading styl` from the sentence `Test heading styled words`
  - add anchor with name e.g. 'a'
  - select `heading styled` from the sentence `Test heading styled words`
  - we have three anchors: one for `heading`, `styl` and `ed` - we should have one anchor for `heading styled`. That is the behaviour while you try to do a similar thing on non-styled words
  
  Summary:
  - please add at least two more cases to verify that my scenarios are correct, then try to find the solution for that. Ask me anytime if you need some more explanations.
  - please try to find any non-standard selections that are working on the non-styled text, but fails with styled text. Make a test case for each unique case. If you think that some of them are not covering the original issue, we can think about extracting them as separate issue.
  - please explain to me the situation with multiline selection, because I also find a scenario when such selection fails
  - please add a test with case that is described by @f1ames previously

KarolDawidziuk:
Thanks @sculpt0r for detailed review and bug cases.

>Also, I'm not sure what exactly you were talking about, but from this comment:
#4733 (comment) I concluded that we should have a separate issue for anchors applied to multiline selection?

I mean, the problem with multi-line selection exists, but this PR may fix this without creating a new one.

This changes included:
* Adding one more test case described by @flames
* Fix: remove more nested anchors into the range.

Referring to the your failing test cases, I think we should create a new issue for this. This problem with adding anchors to partially selected styled text is a little different as the adding anchors in part does not work in the core `CKEDITOR.style.applyToRange()` function and I think that it could be a more complex topic.

I was find one partially solution to this problem but I want to know what you think about this.

Steps:
```
1. Extract actual selected range and save it to future restoring.
2. Remove the custom styles from selected range. Then the, `applyToRange()` method can correctly add the anchor to selected range without custom styles.
3. Append selected `document.fragment` version from step 1. 
```

This solution works for me but.. There is one bigger issue when I select a multiline text. After that, the all content is removed and moved to the end of the editable. I think I know what the problem is, but I wondering if this solution will be to hacky or maybe there is easier way to handle it.

sculpt0r (commented):
To finalize topics with bugs, let's extract **three** separate issues for the anchor plugin:
- related to partially selected multi-styled, inline words. I describe it a little bit earlier, but please adjust the steps that they will be more useful in reproduction without the manual test introduced here.
- related to having two `id` with the same value for the separate anchors. We probably should provide uniq id in html that is generated by the editor.
- related to selection multi lines

They will be easiest to review and solve as separate issues.

sculpt0r (changes requested):
Good job :+1: 

Just some cleanup stuff related to extracting more issues:
- please see the previous comment: https://github.com/ckeditor/ckeditor4/pull/4733#pullrequestreview-709359189
- I think we can convert the "monkey-clicking" manual test into a test case that covers #3863. So please adjust the scenario and editor content. Please take a look at the indentation in `html` file, it looks like there are spaces instead of tabs.

KarolDawidziuk:
Thanks @sculpt0r for the review!
It's ready for the another one.

Changes:
* I've added three issues mentioned in:  https://github.com/ckeditor/ckeditor4/pull/4733#pullrequestreview-709359189.
* Converting 'monkey-clicking' manual test to test with scenarios covering  #4728 and #3863.
* Indentation fix in the manual `html` file.
* Rebased into latest `master`.

sculpt0r (changes requested):
Thank you for the corrects.

I left a few inline comments - only polishing - so :+1: 

As for the manual test: I somehow missed the fact that we have two cases here. I also added some inline comments there for general information, but what is more important:
- please split this test into two separate manual tests for each scenario and name those files accordingly.

It is best if our test can fail due to only one reason. It is also described in [our guide](https://github.com/ckeditor/ckeditor4/blob/major/dev/docs/codestyle.md#31-manual-tests).

KarolDawidziuk:
Thanks for review!
It's ready for the another.

At the time when we decided to split it up this issue into another three, I removed the manual playground test for anchor testing and added test cases to cover issues mentioned in #4728 and #3863. Also I added custom message when test fails.

sculpt0r:
Rebased on master

sculpt0r (approved):
LGTM 🎉 
After rebase I cannot find what was wrong with your rebasing 🤔 
We will try to investigate it in another case...

At this point it looks good :) Good job! 👍

sculpt0r:
@f1ames we need your approve. Could you take a look?

f1ames:
@sculpt0r I just dismissed my review as you took it over. Feel free to proceed with merging this PR.

PR #4746, opened 2021-06-09
Fix: [Chrome] &nbsp; is inserted instead of space

### LINKED ISSUE #3819: [Chrome] &nbsp; is inserted instead of space
## Type of report

Bug

1. Type a space between two words somewhere in the middle of a sentence: 
  `<h1>Hello| world!</h1>` -> `<h1>Hello | world!</h1>` - the caret becomes between two spaces.
2. Press backspace to delete one space on the left.
3. Type a space and new word "foobar".

### Expected result

`<h1>Hello foobar world!</h1>`

### Actual result

`<h1>Hello foobar&nbsp;world!</h1>`

## Other details

* Browser: Google Chrome 80.0.3987.16
* OS: linux, win, macos
* CKEditor version: 4.14.0, https://nightly.ckeditor.com/20-01-26-07-04/full/samples/

It doesn't reproduce in FF.
It doesn't reproduce in Chrome in a raw `<div contenteditable>`.
This bug is almost the same as https://dev.ckeditor.com/ticket/11415, but reproduction steps are different.

sculpt0r:

Bug fix

### This PR contains

- [X] Unit tests
- [X] Manual tests

```
* [#3819](https://github.com/ckeditor/ckeditor4/issues/3819): Fix: [Chrome] &nbsp; is inserted instead of space
```

## What changes did you make?

Typing more than one space leads to a string of alternating spaces and `&nbsp;`. It is possible to split such text node. It is happening in: https://github.com/ckeditor/ckeditor4/blob/8dbbbc651afa01e9240ed47260c695b9bc8b7ff7/core/dom/range.js#L1189

In this case, the first node ends with `&nbsp;` and the second node begins with `&nbsp;`. Even if they were normal spaces before, they are converted to `&nbsp;`, because we can't have an empty visible character at the edge of such node.

After this split, we are in a situation where we have two text nodes and caret placed between them. If we start typing, we are actually adding text to the first node. If there is a `&nbsp;` it is converted to normal space. But the `&nbsp;` from the second node is still there (and causes the problem here).

The entire situation is happening only on WebKit browsers, because of fix for some old issue: https://github.com/ckeditor/ckeditor4/blob/8dbbbc651afa01e9240ed47260c695b9bc8b7ff7/core/editable.js#L1195-L1197

Inside of this listener, we are checking the delete and backspace keys and if they are pressed, performing some merges, which finally invoke `range.trim()` with parameters that leads to text node splitting on the collapsed range: https://github.com/ckeditor/ckeditor4/blob/8dbbbc651afa01e9240ed47260c695b9bc8b7ff7/core/dom/range.js#L1209

I've made some approaches to find out why this so-called `trim` method needs to make a split on text node but ended up with the fact that it is used in so many cases inside the `range` class. So if I touch it anyhow, it broke some of the tests. The `range` is a crucial part of the selection, properly marking styles, and so on.

I also, try to remove trim invocation from higher methods, but they are also used in other cases.

Also, I try to merge split nodes after the trim, but some tests make assertions on node amount - so this approach also fails.

However, remove trimming doesn't affect these particular reproduction steps.

Finally, after F2F with @f1ames - we decide to fix only this particular case. And since we know, what is the source of it. We know that the trim invocation could be skipped if it is in the invocation stack beginning at the WebKit listener.

So I add a single argument that is passed all the way from the listener to https://github.com/ckeditor/ckeditor4/blob/8dbbbc651afa01e9240ed47260c695b9bc8b7ff7/core/dom/range.js#L2404 or https://github.com/ckeditor/ckeditor4/blob/8dbbbc651afa01e9240ed47260c695b9bc8b7ff7/core/dom/range.js#L2443 and just skips trimming in this case. The default value is `false` so all other use cases are not affected by this argument.

I added a description for `range.trim()` after my analysis of how this method works. I think it's better than no description at all. Also, I remove some comments from there because the code does something different than they said.

Added manual tests for delete and backspace situations in `/tests/core/editable/keystrokes/delbackspacequirks/manual`:
- nbspdeleting
- nbspbackspacing

and automatic tests which check if we have text nodes starting with whitespace after delete or backspace key down event. I decided to check this fact instead of verifying the number of nodes because this is the real issue here, not the number of nodes:
- `/tests/core/editable/keystrokes/delbackspacequirks/collapsed.js`

Closes #3819 .


core/editable.js
```diff
@@ -1215,10 +1215,11 @@
 
 						var backspace = key == 8,
 							range = sel.getRanges()[ 0 ],
-							startPath = range.startPath();
+							startPath = range.startPath(),
+							skipRangeTrimming = true;
 
 						if ( range.collapsed ) {
-							if ( !mergeBlocksCollapsedSelection( editor, range, backspace, startPath ) ) {
+							if ( !mergeBlocksCollapsedSelection( editor, range, backspace, startPath, skipRangeTrimming ) ) {
```
Dumluregn, line 1222: This is the only call of the `mergeBlocksCollapsedSelection()`, so we can initialize `skipRangeTrimming` parameter inside it.

tests/core/editable/keystrokes/delbackspacequirks/manual/nbspbackspacing.html
```diff
@@ -0,0 +1,11 @@
+<textarea id="editor" cols="30" rows="10">
+	<p>Hello World!</p>
+</textarea>
+
```
Dumluregn, line 4: Let's add a paragraph below the editor tracking the editor source code on `change` event to improve debugging experience.

tests/core/editable/keystrokes/delbackspacequirks/manual/nbspbackspacing.html
```diff
@@ -0,0 +1,11 @@
...
+<script>
+	if ( !CKEDITOR.env.webkit ) {
```
Dumluregn, line 6: I wasn't able to reproduce this issue on Safari, so I guess it's better to use `CKEDITOR.env.chrome` here? (it will still accept the new Edge and it should as the issue is also present there).

tests/core/editable/keystrokes/delbackspacequirks/manual/nbspdeleting.html
```diff
@@ -0,0 +1,11 @@
+<textarea id="editor" cols="30" rows="10">
+	<p>Hello World!</p>
+</textarea>
+
+<script>
+	if ( !CKEDITOR.env.webkit ) {
```
Dumluregn, line 6: Also here - use `CKEDITOR.env.chrome`. And this test should be ignored on mobile devices as they usually don't have `delete` key.

tests/core/editable/keystrokes/delbackspacequirks/manual/nbspdeleting.md
```diff
@@ -0,0 +1,14 @@
+@bender-tags: 4.16.2, bug, range, 3819
+@bender-ui: collapsed
+@bender-ckeditor-plugins: wysiwygarea, toolbar, sourcearea
+
+1. Put caret right after the word `Hello`.
+2. Type a space. The caret is now between two visual spaces.
+3. Press the left arrow to put caret before two visual spaces.
```
Dumluregn, line 7: Correct me if I'm wrong but this steps seems unnecessary.

tests/core/editable/keystrokes/delbackspacequirks/manual/nbspdeleting.md
```diff
@@ -0,0 +1,14 @@
...
+4. Use the delete key to remove one space from the right side.
```
Dumluregn, line 8: Since the `delete` key on MacBooks works like `backspace` on Windows, please add a note that on mac you should press `fn+delete` to spare googling for testers.

tests/core/editable/keystrokes/delbackspacequirks/manual/nbspdeleting.html
```diff
@@ -0,0 +1,11 @@
+<textarea id="editor" cols="30" rows="10">
+	<p>Hello World!</p>
+</textarea>
+
```
Dumluregn, line 4: In this test let's add a paragraph below the editor tracking the editor source code on change event to improve debugging experience too.

tests/core/editable/keystrokes/delbackspacequirks/collapsed.js
```diff
@@ -33,6 +33,62 @@
...
+
+				editable.fire( 'keydown', new CKEDITOR.dom.event( { keyCode: BACKSPACE } ) );
+
+				var foundIndex = CKEDITOR.tools.getIndex( pContent.getChildren().toArray(), function( node ) {
+					var firstCharCode = node.getText().charCodeAt( 0 );
+					return firstCharCode === 32 || firstCharCode === 160;
+				} );
+
+				assert.areEqual( -1, foundIndex, 'There should not be a node starting with whitespace after backspace key' );
+			} );
```
Dumluregn, line 61: Both unit tests are almost the same so we could extract them to the helper function.

Dumluregn (changes requested):
The proposed solution works well, is clear and simple 👍🏻 There are only some minor fixes and adjustments needed in the tests.

I was able to reproduce the issue in Edge so I'll add the proper label to the issue.

tests/core/editable/keystrokes/delbackspacequirks/manual/nbspbackspacing.html
Dumluregn: Oh, I see it is reproducible in Safari and I've omitted it because in Safari `14.1` devtools the `&nbsp;` is displayed simply as ` ` 👎🏻 so it can only be verified through the source mode. Therefore this comment is invalid, please leave the flag as is.

tests/core/editable/keystrokes/delbackspacequirks/manual/nbspdeleting.html
Dumluregn: See https://github.com/ckeditor/ckeditor4/pull/4746#discussion_r650813470 - so only add ignoring on mobile.

sculpt0r:

- I correct all your inline comments
- Clean up comments for `trim` and other methods & 
 rename one parameter in: https://github.com/ckeditor/ckeditor4/pull/4746/commits/932e63768ce907e254bffa9d7bfa86cd68aeb3e4 & https://github.com/ckeditor/ckeditor4/pull/4746/commits/8d859a49e23c51fa8ce363b8cc37701f630abfef - I forgot to push them before :see_no_evil: :cry: 
- As for https://github.com/ckeditor/ckeditor4/pull/4746#discussion_r650786898, I simplify the delete step description. Because those steps are necessary, and the `space on right` I mean the space on the right side of the caret. Now it should be more clear that we just want to delete a space.

sculpt0r:
As we agreed with @Dumluregn 
- unit test now verifies if there was a split performed
- manual tests verify deleting/backspacing first/second visual space.

core/dom/range.js
```diff
@@ -1182,29 +1182,31 @@ CKEDITOR.dom.range = function( root ) {
 		},
 
 		/**
+		 * Change start container to its parent or new node after start container split.
+		 * If range is collapsed, perform collapse to the start and ends.
+		 *
+		 * If not collapse earlier, change end container to its parent or new node after end container split.
+		 *
+		 * Works on container if it is {@link CKEDITOR#NODE_TEXT text node}, range is collapsed or start/end is not ignored.
+		 *
```
Dumluregn, line 1191: Cool that you've covered that. OTOH for me it's a bit too detailed and still not fully understandable without looking into code. So instead of describing each consecutive step of algorithm I'd rather try to grasp the essence of what is the result and close it in 1-2 sentences.

core/editable.js
```diff
@@ -2556,7 +2557,7 @@
 		};
 	} )();
 
-	function mergeBlocksCollapsedSelection( editor, range, backspace, startPath ) {
+	function mergeBlocksCollapsedSelection( editor, range, backspace, startPath, skipRangeTrimming ) {
```
Dumluregn, line 2560: By https://github.com/ckeditor/ckeditor4/pull/4746#discussion_r650770440 I meant removing this `skipRangeTrimming` parameter from function signature and instead either just pass `true` later:
```js
if ( !range[ backspace ? 'checkStartOfBlock' : 'checkEndOfBlock' ]( true ) )
```

or initialise the boolean variable inside this function:
```js
var skipRangeTrimming = true;

if ( !range[ backspace ? 'checkStartOfBlock' : 'checkEndOfBlock' ]( skipRangeTrimming ) )
```

tests/core/editable/keystrokes/delbackspacequirks/_helpers/tools.js
```diff
@@ -11,6 +11,34 @@ var quirksTools = ( function() {
...
+					pContent = editable.findOne( 'p' ),
+					initialChildCount = pContent.getChildCount(),
+					range = bot.editor.createRange();
+
+				// Move caret to proper place in text.
+				range.setStart( pContent.getFirst(), caretPosition );
+				range.select();
+
+				editable.fire( 'keydown', new CKEDITOR.dom.event( { keyCode: keyCode } ) );
+				var childCountAfterKeyUsage = pContent.getChildCount();
```
Dumluregn, line 34: Let's move this `var` to the rest above.

tests/core/editable/keystrokes/delbackspacequirks/manual/nbspbackspacing.html
```diff
@@ -0,0 +1,18 @@
+<textarea id="editor" cols="30" rows="10">
+	<p>One two three</p>
+</textarea>
+<p style="font-family: monospace;"><strong>Result:</strong><span id="sourceView"></span></p>
+
+<script>
+	if ( !CKEDITOR.env.webkit ) {
+		assert.ignore();
```
Dumluregn, line 8: In manual tests we have to use `bender.ignore()`.

tests/core/editable/keystrokes/delbackspacequirks/manual/nbspdeleting.html
```diff
@@ -0,0 +1,19 @@
+<textarea id="editor" cols="30" rows="10">
+	<p>One two three</p>
+</textarea>
+<p style="font-family: monospace;"><strong>Result:</strong><span id="sourceView"></span></p>
+
+<script>
+	// Ignore mobile due to lack of delete key.
+	if ( !CKEDITOR.env.webkit || CKEDITOR.env.mobile ) {
+		assert.ignore();
```
Dumluregn, line 9: In manual tests we have to use `bender.ignore()`.

Dumluregn (changes requested):
Almost there, some small docs and refactors left.

core/dom/range.js
```diff
@@ -1182,29 +1182,31 @@ CKEDITOR.dom.range = function( root ) {
...
-		 * @todo precise desc/algorithm
 		 */
 		trim: function( ignoreStart, ignoreEnd ) {
 			var startContainer = this.startContainer,
 				startOffset = this.startOffset,
 				collapsed = this.collapsed;
+
 			if ( ( !ignoreStart || collapsed ) && startContainer && startContainer.type == CKEDITOR.NODE_TEXT ) {
-				// If the offset is zero, we just insert the new node before
-				// the start.
```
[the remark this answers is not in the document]
sculpt0r: But that is the case. They are incorrect ;(

They are about inserting node, which is not happening. In this case, the pointer to the start container is just moved to the parent element. Basically, what I was trying to sum up in the method description, is happening. So the only insertion-like thing is happening when the text node is split.

core/dom/range.js
sculpt0r: It is hard to make summary of this method, but I'll try again :D

core/editable.js
sculpt0r: That is true, but I want to be 100% sure, that this boolean comes from a callback attached to `key` event. Right now, that is the only use case, but if someone ever thinks about usage `mergeBlocksCollapsedSelection` will not affect this logic.

sculpt0r:
@Dumluregn - ready for another review
I correct most of your inline comments. However, I left a few answers explaining, why I don't correct some of them( https://github.com/ckeditor/ckeditor4/pull/4746#discussion_r654199811, https://github.com/ckeditor/ckeditor4/pull/4746#discussion_r654183562)

Dumluregn (approved):
LGTM! 🎉 

I've extracted a similar issue, but involving using the `enter` key, so the solution may be completely different there: https://github.com/ckeditor/ckeditor4/issues/4770 (but not for sure).

PR #4757, opened 2021-06-13
Fix incorrect nesting of font styles

### LINKED ISSUE #4555: More nesting problems when using classes for font size
## Type of report

Bug

#1951 deals with text of one size that's nested inside text of another size, and attempting to change the size of the whole thing. This appears to be a new bug introduced in 4.14.0. It shows up when applying a size to some text and then applying a different size to the same text.

https://codepen.io/davidrees/pen/bGBKYQa

1. Select some text
2. Apply the available font sizes one at a time: small, medium, large

### Expected result

```html
<span class="small">The quick brown fox tires of having his maneuvers discussed.</span>
```

### Actual result

```html
<span class="large">
  <span class="medium">
    <span class="small">The quick brown fox tires of having his maneuvers discussed.</span>
  </span>
</span>
```

At first, as you move from small, to medium, to large, you continue to see small text, since the span with the small class is innermost. But if you then apply small again, rather than wrapping another span around the whole thing, it shuffles the order. So you still have three nested spans, but now they're in the order small-large-medium. So re-applying small actually results in making medium effective, since it's now the innermost span. And then if you re-apply medium, you effectively get large, etc.

## Other details

* Browser: Chrome 88
* OS: macOS 10.15.7
* CKEditor version: 4.14.0
* Installed CKEditor plugins: none

Comandeer:

Bug fix

### This PR contains

- [x] Unit tests
- [x] Manual tests

```
* [#4555](https://github.com/ckeditor/ckeditor4/issues/4555): Fixed: [Font](https://ckeditor.com/cke4/addon/font) styles with attributes are not applied correctly when used multiple times over the same selection.
```

## What changes did you make?

The culprit was the `2dcda53bdcc36793925d6da8cff044e1c01615f5` commit. The brought back `isEqualStyle` function checked only if the styles are the same, ignoring other attributes. So I've added logic to check also other attributes.

Closes #4555.

### DIFF
--- a/plugins/font/plugin.js
+++ b/plugins/font/plugin.js
@@ -263,7 +263,37 @@
 			return false;
 		}
 
-		return CKEDITOR.style.getStyleText( styleA.getDefinition() ) === CKEDITOR.style.getStyleText( styleB.getDefinition() );
+		var hasEqualAttributes = compareAttributes( styleA, styleB ),
+			hasEqualStyles = compareStyles( styleA, styleB );
+
+		return hasEqualAttributes && hasEqualStyles;
+
+		function compareAttributes( styleA, styleB ) {
+			var styleADefinition = styleA.getDefinition(),
+				styleBDefinition = styleB.getDefinition(),
+				styleAAttributes = CKEDITOR.tools.object.entries( styleADefinition.attributes ),
+				styleBAttributes = CKEDITOR.tools.object.entries( styleBDefinition.attributes );
+
+			if ( styleAAttributes.length !== styleBAttributes.length ) {
+				return false;
+			}
+
+			return CKEDITOR.tools.array.every( styleAAttributes, function( attributeA, i ) {
+				var attributeB = styleBAttributes[ i ],
+					attributeAName = attributeA[ 0 ],
+					attributeAValue = attributeA[ 1 ],
+					attributeBName = attributeB[ 0 ],
+					attributeBValue = attributeB[ 1 ],
+					isSameName = attributeAName === attributeBName,
+					isSameValue = attributeAValue === attributeBValue;
+
+				return isSameName && isSameValue;
+			} );
+		}
+
+		function compareStyles( styleA, styleB ) {
+			return CKEDITOR.style.getStyleText( styleA.getDefinition() ) === CKEDITOR.style.getStyleText( styleB.getDefinition() );
+		}
 	}
 
 	//  * @param {Object} options
--- a/plugins/font/plugin.js
+++ b/plugins/font/plugin.js
@@ -271,24 +271,10 @@
 		function compareAttributes( styleA, styleB ) {
 			var styleADefinition = styleA.getDefinition(),
 				styleBDefinition = styleB.getDefinition(),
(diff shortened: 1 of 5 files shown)


tests/plugins/font/customfontdefinition.js
```diff
@@ -7,40 +7,69 @@
...
+				fontSize_style: {
+					element: 'span',
+					attributes: dataAttributes
+				}
 			}
 		}
 	};
 
 	bender.test( {
 		'test setdata with font tag with face': function() {
```
Dumluregn, line 61: This test stopped working on IE9 (and only there) with an error
```
Unexpected error: Unable to get value of the property 'html': object is null or undefined
Expected: undefined (undefined)
Actual:   undefined (undefined)
```

So it doesn't seem connected with a fix, but rather with tests refactoring - but nothing important really changed 🤔  Could you please verify if it fails for you too?

tests/plugins/font/customfontdefinition.js
```diff
@@ -96,17 +125,17 @@
 		},
 
 		'test setdata with font tag with size': function() {
```
Dumluregn, line 127: This test stopped working on IE9 (and only there) with an error
```
Unexpected error: Unable to get value of the property 'html': object is null or undefined
Expected: undefined (undefined)
Actual:   undefined (undefined)
```

So it doesn't seem connected with a fix, but rather with tests refactoring - but nothing important really changed 🤔  Could you please verify if it fails for you too?

plugins/font/plugin.js
```diff
@@ -263,7 +263,37 @@
...
+				var attributeB = styleBAttributes[ i ],
+					attributeAName = attributeA[ 0 ],
+					attributeAValue = attributeA[ 1 ],
+					attributeBName = attributeB[ 0 ],
+					attributeBValue = attributeB[ 1 ],
+					isSameName = attributeAName === attributeBName,
+					isSameValue = attributeAValue === attributeBValue;
+
+				return isSameName && isSameValue;
+			} );
```
Dumluregn, line 291: This looks like a candidate for `CKEDITOR.tools.object.compareShallow()` method, WDYT? The additional benefit is that we could test this issue more thoroughly.

Dumluregn (changes requested):
Looks good :+1: Only some minor stuff.

plugins/font/plugin.js
Comandeer: Actually it seems that we have a method for shallow object comparison – [`CKEDITOR.tools.objectCompare()`](https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_tools.html#method-objectCompare). I didn't find it because I was looking into `CKEDITOR.tools.object`.

I'll use that method here to simplify the code.

Comandeer:
I've updated the code used for comparing attributes.

As for tests failing in IE9, I couldn't reproduce it.

Dumluregn (approved):
LGTM! 🎉

Tests failing on Safari are also failing on master (see https://github.com/ckeditor/ckeditor4/issues/4759).

Issue #4444, opened 2020-12-18
Print and Preview don't work on Firefox when using editor from CDN


jswiderski:
## Type of report

Bug

## Provide detailed reproduction steps (if any)

1. Open https://ckeditor.com/docs/ckeditor4/latest/examples/fullpreset.html
2. Press Print or Preview button

### Expected result

*What is the expected result of the above steps?*

Preview should open in a window normally.

*What is the actual result of the above steps?*

Blank window opens. 
Error is thrown in the console: Permission denied to access property "_cke_htmlToLoad" on cross-origin object

## Other details

* Browser: Firefox
* OS: Any
* CKEditor version: 4.0+
* Installed CKEditor plugins: Preview, Print

I have tried setting ` Header set Access-Control-Allow-Origin "*"` but it didn't work. This could be Firefox issue:
https://bugzilla.mozilla.org/show_bug.cgi?id=911444
https://github.com/mozilla/pdf.js/issues/5397

f1ames:
Fixed in https://github.com/ckeditor/ckeditor4/pull/4758.

jacekbogdanski:
Unfortunately, we won't be able to provide the fix in `4.16.2` release. I'm sorry for the confusion, we will move this issue to the next release instead.

PR #4758, opened 2021-06-13
Fix: Correct open Preview in FF using editor via CDN

KarolDawidziuk:

Bug fix

### This PR contains

- [ ] Unit tests
- [x] Manual tests

```
* [#4444](https://github.com/ckeditor/ckeditor4/issues/4444): Fixed: [Print](https://ckeditor.com/cke4/addon/print) / [Preview](https://ckeditor.com/cke4/addon/preview) Print and Preview are not opened correctly on FF due to cross-origin.
```

## What changes did you make?

FF opens the preview in a separate window to avoid an anchor problems. Unfortunately, this problem persists even without this fix. Using a section that uses `fixDomain` fixed the preview problem. Additionally, I've split this issue into another one regarding to problem with print.  

Closes [#4444](https://github.com/ckeditor/ckeditor4/issues/4444).


Comandeer:
Rebased onto latest `master`.

Comandeer (changes requested):
The fix works correctly, but it seems that the PR contains two separate fixes for different bugs (see my inline comments).

plugins/preview/plugin.js
```diff
@@ -155,6 +156,22 @@
 
 			return html;
 		}
+
+		// FF tries to start printing before page is loaded. The browser has wait for this. (#4444)
+		function checkPrintListener() {
+			if ( editor.plugins.preview.setPrintListener ) {
```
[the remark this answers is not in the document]
KarolDawidziuk: > Commenting out this function doesn't seem to change anything in a way in which printing works in Fx 🤔 However I'm testing on macOS, the behaviour can be different in other OSes.
I'm tested it on macOS and VBox Win 7 and still it's not working without fix.

>  If that's the case, please report the issue as it's connected with different plugin (print) and its handling of executing printing.
This problem with run print is related with change in [this](https://github.com/ckeditor/ckeditor4/pull/4758/files#diff-36cb2ca37dc3086bbe191b661d4f8923ff5872b765b8d537439a568da622580aR202) line. That's why I fix this in one PR.

KarolDawidziuk:
Thank's @Comandeer for the review.
It's ready for another turn.

tests/plugins/preview/manual/previewcdn.html
```diff
@@ -0,0 +1,19 @@
...
+	var iframe = document.getElementById('cdn'),
+		doc = iframe.contentWindow.document,
+		template = '<h1>Sample <span style="color: red;">document</span></h1>' +
+			'<p>Lorem ipsum</p>' +
+			'<div style="page-break-after:always"><span style="display:none">&nbsp;</span></div>' +
+			'<p>dolor sit amet.</p>';
+
+	doc.open();
+	// CDN points into commit revision with bug fix.
+	doc.write( '<script src="https://raw.githack.com/ckeditor/ckeditor4/be63bc46be15478259dfbca94de6a019b0b0145c/ckeditor.js"><\/script>' );
```
Comandeer, line 15: I'm not happy with this solution. It requires updating the commit hash on every change inside the PR.

Yet I don't see a better solution :thinking:

Comandeer (changes requested):
I'm still for removing the whole part fixing the print in Firefox. It is a separate bug that should be reported as blocked on this one. Additionally it requires much more refactoring than it's currently done:

1. Current fix for print breaks boundaries between `preview` and `print` plugin. Probably `preview` plugin should expose some kind of public API to allow adding callback fired when preview window is loaded.
2. The fix breaks preview if the user first prints and then opens preview. In such a case previe option will also display print dialog. It's connected with the fact that the preview window is created only once and then reused every time users click a "Print Preview" or "Print" button.
3. The fix adds much complexity and seems messy. Replacing it with public API for adding callback should make the whole thing far more readable. Especially as we already have triggering print on preview `load` event, so we basically duplicating the logic.

So I'm for reporting the print issue and leaving only the preview fix here.

Additionally, manual tests kill IE8 for some reason.

KarolDawidziuk:
Thanks for review @Comandeer.
After your advice, I've created a separated [issue](https://github.com/ckeditor/ckeditor4/issues/4790) regarding to problem with print on FF and this PR include fix only for open Preview.  

Beside of this I found another way to open editor from CDN without using iframe that which caused of failing test on IE(the browser was freezing). It's look like a little hacky way but for now it's only working solution on IE that I found.

tests/plugins/preview/manual/previewcdn.html
```diff
@@ -0,0 +1,72 @@
...
+	// 1. Load local editor by default with bender.
+	// 2. Remove existing instance and add script with editor from CDN.
+	// 3. Wait for the new editor to be loaded from CDN and initialize it.
+	// 4. Add icons in IE <11 from local.
+	var path = 'https://raw.githack.com/ckeditor/ckeditor4/903fa05e6bde756a7f4e618b6d24b4db21da5ab7';
+
+	CKEDITOR.replace( 'editor' );
+	CKEDITOR.once( 'instanceReady', function() {
+		var scripts = document.querySelectorAll( 'script' ),
+			head = document.head || document.getElementsByTagName("head")[0];
```
Comandeer, line 22:
```suggestion
			head = document.head || document.getElementsByTagName( 'head' )[ 0 ];
```

However it would be more elegant to use our `CKEDITOR.dom` methods instead of native ones.

tests/plugins/preview/manual/previewcdn.html
```diff
@@ -0,0 +1,72 @@
...
+	<p>Lorem ipsum</p>
+	<div style="page-break-after:always">
+		<span style="display:none">&nbsp;</span>
+	</div>
+	<p>dolor sit amet.</p>
+</div>
+
+<script>
+	// 1. Load local editor by default with bender.
+	// 2. Remove existing instance and add script with editor from CDN.
```
Comandeer, line 14: Are these steps necessary?

Comandeer (changes requested):
Fix works good, I've added only some inline comments about manual test.

Additionally please fix issue with double commits.

tests/plugins/preview/manual/previewcdn.html
KarolDawidziuk: Good point. Thanks 👍

tests/plugins/preview/manual/previewcdn.html
KarolDawidziuk: Yes, because bender loads all scripts by default every time, so I delete all files and namespace so that the new instance doesn't accidentally use local files. Additionally, we are sure that the new editor instance is loaded from CDN.

KarolDawidziuk:
Thanks @Comandeer for review. 
 It's ready for the next one.

f1ames:
Rebased onto latest `master`.

tests/plugins/preview/manual/previewcdn.html
```diff
@@ -0,0 +1,72 @@
...
+		newScript.type = 'text/javascript';
+		newScript.src = path + '/ckeditor.js';
+		document.head.appendChild( newScript );
+
+		var interval = setInterval( function() {
+			try {
+				if( CKEDITOR.status === 'loaded' ){
+					clearInterval( interval );
+
+					CKEDITOR.replace( 'editor_cdn', { plugins: 'wysiwygarea, print, pagebreak, toolbar, clipboard, image, floatingspace, link, basicstyles' } );
```
f1ames, line 44: This is a bit problematic since we have basic, standard and full presets and manual tests run for such preset use default plugins. So for `full all` used during testing phase we would like to have all plugins which are default for this preset.

The solution could be reading `plugins` list from the initial editor instance (before destroying) and then pass it to new editor config. WDYT?

tests/plugins/preview/manual/previewcdn.html
```diff
@@ -0,0 +1,72 @@
...
+	</div>
+	<p>dolor sit amet.</p>
+</div>
+
+<script>
+	// 1. Load local editor by default with bender.
+	// 2. Remove existing instance and add script with editor from CDN.
+	// 3. Wait for the new editor to be loaded from CDN and initialize it.
+	// 4. Add icons in IE <11 from local.
+	var path = 'https://raw.githack.com/ckeditor/ckeditor4/903fa05e6bde756a7f4e618b6d24b4db21da5ab7';
```
f1ames, line 17: This PR will be merged to `master` soon so maybe we could replace (just before merging) this URL to point to `master` branch?

Still this would mean we are not testing the current local editor version which might be confusing and is not wehat we usually want to achieve with manual tests :thinking:

f1ames: I wonder couldn't we use `http://0.0.0.0:1030/apps/ckeditor/ckeditor.js` or `http://127.0.0.1:1030/apps/ckeditor/ckeditor.js` or `http://localhost:1030/apps/ckeditor/ckeditor.js` to simulate CORS (depending what is the current host used)?

From what I have checked (from `0.0.0.0`) it will give us CORS issue (so the same think we would like to simulate here I suppose):
image

And then we will have test using local version.

tests/plugins/preview/manual/previewcdn.md
```diff
@@ -0,0 +1,16 @@
...
+@bender-ui: collapsed
+@bender-ckeditor-plugins: wysiwygarea, print, font, colorbutton, format, clipboard, pagebreak, toolbar, floatingspace, link, image2
+
+## Steps
+1. Wait for the editor to load.
+1. Click preview button.
+1. Open browser console in the new tab.
+
+## Expected result
+New browser tab with preview are opened correctly with no error in browser console.
```
f1ames, line 11: > New browser tab with preview are opened correctly

This should be expected result after 2nd step - you click preview and new tab/window with correct content is shown.

> with no error in browser console

And this is expected result after 3rd step.

tests/plugins/preview/manual/previewcdn.md
```diff
@@ -0,0 +1,16 @@
+@bender-tags: 4.16.2, 4444, bug
+@bender-ui: collapsed
+@bender-ckeditor-plugins: wysiwygarea, print, font, colorbutton, format, clipboard, pagebreak, toolbar, floatingspace, link, image2
```
f1ames, line 3: This test and PR is about `preview` plugin so I;m not sure if `print` is needed :thinking:

tests/plugins/preview/manual/previewcdn.md
```diff
@@ -0,0 +1,16 @@
...
+1. Click preview button.
+1. Open browser console in the new tab.
+
+## Expected result
+New browser tab with preview are opened correctly with no error in browser console.
+## Unexpected result
+Only a new tab with no content preview opens and an error occurs in the browser console: `Permission denied to access property "_cke_htmlToLoad" on cross-origin object`.
+
+## Notes
+You can add more content into editable or just copy existing to check if the printout still opens e.g. add 50 pages and click print.
```
f1ames, line 16: Again, why it's mentioning `print` if it's about `preview` plugin? :thinking:

f1ames (changes requested):
The fix looks good. There are still some things to polish in manual test.

tests/plugins/preview/manual/previewcdn.html
KarolDawidziuk: Good point. I didn't think about presets and your idea might work. Thanks!

tests/plugins/preview/manual/previewcdn.html
KarolDawidziuk: It's look like your suggestion with simulate CORS locally works. I just need to change it a little and it should work. Thanks!

tests/plugins/preview/manual/previewcdn.md
KarolDawidziuk: I'll correct it. Thanks.

tests/plugins/preview/manual/previewcdn.md
KarolDawidziuk: My bad. I left it after the first PR version which included a problem with print and preview.

tests/plugins/preview/manual/previewcdn.md
KarolDawidziuk: As I mentioned above i forgot remove it.

KarolDawidziuk:
Thanks @f1ames for the review. 
It's ready for the next turn.

Changes:
* Changed places of expected results.
* Changed `print` to `preview` in manual test.
* Mixed CORS simulation.

> I wonder couldn't we use `http://0.0.0.0:1030/apps/ckeditor/ckeditor.js` or `http://127.0.0.1:1030/apps/ckeditor/ckeditor.js` or `http://localhost:1030/apps/ckeditor/ckeditor.js` to simulate CORS (depending what is the current host used)?
> 
> From what I have checked (from `0.0.0.0`) it will give us CORS issue (so the same think we would like to simulate here I suppose):
> image
> 
> And then we will have test using local version.


Unfortunately, this will not work when we are going to use a virtual machine because we'll only have `10.0.2.2:<port>` host. So after our F2F talk I added mixed solution. 
When a person runs the test in the virtual machine, editor it will be created from the master branch from the githack, otherwise CORS from the local editor will be simulated.

f1ames (changes requested):
Manual test fails on IE8:
image

Apart from that it looks good. I did some small adjustments in manual test description and minor refactoring in test code itself.

KarolDawidziuk:
Hi @f1ames  thanks for review. 
It's ready for another one.

f1ames:
Rebased onto latest `master`.

Issue #4427, opened 2020-12-09
Enable Maximize plugin on iOS


jswiderski:
## Type of report

Bug

## Provide detailed reproduction steps (if any)

1. Open CKEditor Classic demo page on iPhone

### Expected result

Maximized button should be enabled.

### Actual result

Maximized button is disabled.

## Other details

About 9 years ago we have disabled Maximize button on iOS because the toolbar was getting out of viewport and it wasn't possible to come back to normal mode again:
- https://dev.ckeditor.com/ticket/8307
- https://github.com/ckeditor/ckeditor4/blob/major/plugins/maximize/plugin.js#L125-L126

From what we have initially tested the problem is no more, so we could made some extra testing to see if we can change current code to `modes: { wysiwyg: true, source: true },`

* Browser: Safari
* OS: iOS
* CKEditor version: 4.0+
* Installed CKEditor plugins: Maximize

MeghaAparanji:
Is this working? If we add modes: { wysiwyg: true, source: true } maximize button will be enabled in CKEditor on iOS but on click f maximize button page shows blank, is there any workaround for this?

PR #4767, opened 2021-06-17
Allow maximize editor on iOS

KarolDawidziuk:

Bug fix

### This PR contains

- [x] Unit tests
- [x] Manual tests

```
* [#4427](https://github.com/ckeditor/ckeditor4/issues/4427): Fix: [iOS] Enable maximize plugin on IOS.
```

## What changes did you make?

I unlocked the maximize button of the editor on iOS.
The editor behaves differently than on other systems. Now focus is activated every time we toggle maximize. Other browsers remember state of focus. This is related to this [fix](https://dev.ckeditor.com/ticket/12381)(problem with tap on dissabled buttons) that was present in iOS 6-7 and as of version iOS8+ this problem no longer occurs, so we can eaisily change it.

Closes [#4427](https://github.com/ckeditor/ckeditor4/issues/4427).

### DIFF
--- a/plugins/maximize/plugin.js
+++ b/plugins/maximize/plugin.js
@@ -91,6 +91,13 @@
 		}
 	}
 
+	function checkIOS( container ) {
+		if ( !container ) {
+			return CKEDITOR.env.iOS;
+		}
+		return CKEDITOR.env.iOS && container.hasClass( 'cke_maximized' );
+	}
+
 	CKEDITOR.plugins.add( 'maximize', {
 		// jscs:disable maximumLineLength
 		lang: 'af,ar,az,bg,bn,bs,ca,cs,cy,da,de,de-ch,el,en,en-au,en-ca,en-gb,eo,es,es-mx,et,eu,fa,fi,fo,fr,fr-ca,gl,gu,he,hi,hr,hu,id,is,it,ja,ka,km,ko,ku,lt,lv,mk,mn,ms,nb,nl,no,oc,pl,pt,pt-br,ro,ru,si,sk,sl,sq,sr,sr-latn,sv,th,tr,tt,ug,uk,vi,zh,zh-cn', // %REMOVE_LINE_CORE%
@@ -113,9 +120,20 @@
 			var outerScroll;
 
 			// Saved resize handler function.
-			function resizeHandler() {
-				var viewPaneSize = mainWindow.getViewPaneSize();
-				editor.resize( viewPaneSize.width, viewPaneSize.height, null, true );
+			function resizeHandler( height ) {
+				var viewPaneSize = mainWindow.getViewPaneSize(),
+					panelHeight = checkIOS() && height ? height : viewPaneSize.height;
+
+				editor.resize( viewPaneSize.width, panelHeight, null, true );
+			}
+
+			function setContainerHeight( container ) {
+				// We must wait for end of autoZoom on iOS.
+				setTimeout( function() {
+					if ( checkIOS( container ) ) {
+						resizeHandler( window.innerHeight );
+					}
+				}, 150 );
 			}
 
 			// Retain state after mode switches.
@@ -129,7 +147,8 @@
 					var container = editor.container.getFirst( function( node ) {
 						return node.type == CKEDITOR.NODE_ELEMENT && node.hasClass( 'cke_inner' );
 					} );
-					var contents = editor.ui.space( 'contents' );
+					var contents = editor.ui.space( 'contents' ),
+						savedHeight = 0;
 
 					// Save current selection and scroll position in editing area.
 					if ( editor.mode == 'wysiwyg' ) {
@@ -140,6 +159,22 @@
 						var $textarea = editor.editable().$;
 						savedSelection = !CKEDITOR.env.ie && [ $textarea.selectionStart, $textarea.selectionEnd ];
 						savedScroll = [ $textarea.scrollLeft, $textarea.scrollTop ];
+					}
+
+					// Safari automatically enlarges the page after activating focus.
+					// We need to calculate the new height of the iframe container. (#4427)
+					editor.on( 'focus', function() {
+						setContainerHeight( container );
+					} );
+
+					// visualViewport is experimental feature so we need add a simple feature detection.
+					if ( window.visualViewport && checkIOS( container ) ) {
+						window.visualViewport.addEventListener( 'resize', function() {
+							if ( window.innerHeight !== savedHeight ) {
+								resizeHandler( window.innerHeight );
+								savedHeight = window.innerHeight;
+							}
+						} );
 					}
 
 					// Go fullscreen if the state is off.
--- a/plugins/maximize/plugin.js
+++ b/plugins/maximize/plugin.js
@@ -122,8 +122,7 @@
 			var savedState = CKEDITOR.TRISTATE_OFF;
 
 			editor.addCommand( 'maximize', {
-				// Disabled on iOS (https://dev.ckeditor.com/ticket/8307).
-				modes: { wysiwyg: !CKEDITOR.env.iOS, source: !CKEDITOR.env.iOS },
+				modes: { wysiwyg: 1, source: 1 },
 				readOnly: 1,
 				editorFocus: false,
 				exec: function() {
--- a/plugins/maximize/plugin.js
+++ b/plugins/maximize/plugin.js
@@ -91,13 +91,6 @@
 		}
 	}
 
-	function checkIOS( container ) {
-		if ( !container ) {
-			return CKEDITOR.env.iOS;
-		}
-		return CKEDITOR.env.iOS && container.hasClass( 'cke_maximized' );
-	}
-
 	CKEDITOR.plugins.add( 'maximize', {
 		// jscs:disable maximumLineLength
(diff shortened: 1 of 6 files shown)


f1ames:
@ckeditor/qa-team could you take a look on this PR and if there are no regressions/issues on iOS after enabling maximize there?

LukaszGudel:
I've tested this branch on iPad with iOS 12.4.8 and overall it's looking good :+1: 

I have found only one issue when enabling maximized mode. Scroll is not smooth and it can jump to the focused element/caret or do not allow to scroll past focused element/caret, but changing the focus/caret placement `unlocks` the scroll. This scrolling issue does not occur when maximized mode is disabled.

While testing I was looking for manual tests with this feature and there is one at `/tests/plugins/tableresize/manual/maximizeddivarea` that is still disabled/ignored on iOS.

f1ames:
Thanks @LukaszGudel :+1: Could you provide specific repro steps for the issue you described:

> I have found only one issue when enabling maximized mode. Scroll is not smooth and it can jump to the focused element/caret or do not allow to scroll past focused element/caret, but changing the focus/caret placement "unlocks" the scroll. This scrolling issue does not occur when maximized mode is disabled.

@KarolDawidziuk could you take a look what's going on there?

LukaszGudel:
### Steps:
1. Open `test/plugins/widget/manual/block` test.
2. Toggle maximized mode.
3. Place a caret somewhere in the first paragraph.
4. Try to scroll to the end of the document.

### Result:
It's not possible to scroll to the end of the document. It blocks the scroll when a focused element is about to 'leave' the viewport. This issue also occurs when caret is placed somewhere at the bottom and you try to scroll to the top.

KarolDawidziuk:
Thanks @f1ames and @LukaszGudel for catch and reproduce steps!

This issue occurred because iOS automatically zooms in on the content when the editor activates the focus. More specifically, it was about calculating the height of the iframe that contains the content. Auto zoom was not detected by `resize` event, therefore the frame size was not changed.

@f1ames it's ready for next review.

plugins/maximize/plugin.js
```diff
@@ -113,24 +120,62 @@
...
+				modes: { wysiwyg: 1, source: 1 },
 				readOnly: 1,
 				editorFocus: false,
 				exec: function() {
 					var container = editor.container.getFirst( function( node ) {
 						return node.type == CKEDITOR.NODE_ELEMENT && node.hasClass( 'cke_inner' );
 					} );
-					var contents = editor.ui.space( 'contents' );
+					var contents = editor.ui.space( 'contents' ),
+						savedHeight = 0;
```
f1ames, line 178:
```suggestion
						previousWindowHeight = 0;
```
?

plugins/maximize/plugin.js
```diff
@@ -246,6 +297,21 @@
 						} );
 					}
 
+					// visualViewport is experimental feature so we need add a simple feature detection.
+					if ( window.visualViewport && checkIOS( container ) ) {
+						window.visualViewport.addEventListener( 'resize', function() {
+							if ( window.innerHeight !== savedHeight ) {
+								resizeHandler( window.innerHeight );
+
+								if ( savedHeight !== 0 && window.innerHeight > savedHeight ) {
```
f1ames, line 306: Could you add some comment with explanation why checking for `window.innerHeight > savedHeight` is used here?

plugins/maximize/plugin.js
```diff
@@ -113,24 +120,62 @@
...
-				var viewPaneSize = mainWindow.getViewPaneSize();
-				editor.resize( viewPaneSize.width, viewPaneSize.height, null, true );
+			function resizeHandler( height ) {
+				var viewPaneSize = mainWindow.getViewPaneSize(),
+					panelHeight = checkIOS() && height ? height : viewPaneSize.height;
+
+				editor.resize( viewPaneSize.width, panelHeight, null, true );
+			}
+
+			function setContainerHeight( container ) {
```
f1ames, line 130: Again, this function changes container height but apart from that does some magic with scroll position so its name is only partially relevant here.

From what I see it is used only once in the code, so maybe could be inlined into `focus` handler?

plugins/maximize/plugin.js
```diff
@@ -113,24 +120,62 @@
...
+							setTimeout( function() {
+								html.$.scrollTo( 0, scrollPos.y + toolbarHeight );
+								window.scrollTo( scrollPos.x, 0 );
+							}, 10 );
+						}
+					}, 150 );
+				}
+			}
+
+			function restoreIOSPosition() {
```
f1ames, line 153:
```suggestion
			function restoreIframeScrollPosition() {
```
?

plugins/maximize/plugin.js
```diff
@@ -113,24 +120,62 @@
...
+					setTimeout( function() {
+						resizeHandler( window.innerHeight );
+
+						// After autoZoom we have to take the caret position and scroll the editable to the current position.
+						var html = editor.editable().getParent(),
+							scrollPos = mainWindow.getScrollPosition(),
+							toolbarHeight = html.getClientRect( true ).top;
+
+						if ( scrollPos.y !== 0 ) {
+							window.scrollTo( 0, 0 );
```
f1ames, line 142: Is this call really needed here? Since there is similar call few lines below - `window.scrollTo( scrollPos.x, 0 );`?

plugins/maximize/plugin.js
```diff
@@ -113,24 +120,62 @@
 			var outerScroll;
 
 			// Saved resize handler function.
-			function resizeHandler() {
-				var viewPaneSize = mainWindow.getViewPaneSize();
-				editor.resize( viewPaneSize.width, viewPaneSize.height, null, true );
+			function resizeHandler( height ) {
```
f1ames, line 123: I know this function was here before, but `resizeHandler` is not the most meaningful name. I mean **it only tells in what context the function is called** - here it's a resize listener. I know this is common convention like `clickHandler`, `focusHandler`, etc. But it's hard to say what resize it is related to - editor resize, window resize or maybe something else? And what the function does. Apart from that, looking at the code it is also [called like a regular function without the resize context](https://github.com/ckeditor/ckeditor4/pull/4767/files#diff-08c728d4cbb0c41381231354b02cb3511dc880624a3e8a6319e9d19d34760942R245) which makes it even more confusing.

The **main responsibility of a function name is to tell and describe what the function does**. So here it could be something like `setEditorSize` or `resizeEditor`.

f1ames (changes requested):
Looks good for a start, but the code needs some adjustments and regrouping. Please see review comments below.

Also Since it's a new feature it should target `major` branch (please remember about updating tags in manual tests too).

KarolDawidziuk:
Thanks @f1ames  for the review! 
It's ready for next one.

Changes: 
* Changed target branch from `master` to `major` - the cause of my mistake was being reported as a bug, not a feature.
* Introduced `editor.isMaximized()` with unit test.
* More comments.
* Regrouping code.
* Changed function and variable names to more understandable.

jacekbogdanski (changes requested):
> I've tested this branch on iPad with iOS 12.4.8 and overall it's looking good 👍

This fix is specifically created for iPhone iOS. IPad is not disabled on `master` branch and was working fine before the fix.

I took a quick look at how the feature now behaves:

## IPad 11

1. As long as we are opening maximize mode with a hidden keyboard, the feature works pretty well. Scrolling works as expected and the editor is responsive.
2. If we focus the editor before maximize button and scroll the browser window a bit to move it outside the initial viewport, the system keyboard is taking the place at the bottom of the editor:

https://user-images.githubusercontent.com/10279586/129850662-bca40507-241d-438f-900d-0d8ed3a10146.MP4

The issue is not reproducible on the `master` branch, so it looks like a potential regression.

## IPhone 12

When using IPhone, the feature behaves very similarly to what is described in https://dev.ckeditor.com/ticket/8307 and the scrolling seems to be totally broken:

https://user-images.githubusercontent.com/10279586/129852024-2cc2b562-6d1f-4cff-bafe-f0fb6a4dc7ad.MP4

jacekbogdanski:
Interestingly, maximize feature works pretty well already on `master` branch:

https://user-images.githubusercontent.com/10279586/129853741-ff99d141-7c8a-4c3e-8b31-42be7baa4b73.MP4

So I'm not totally sure what was the reasoning for the proposed changes. I wasn't reviewing this PR from the beginning, so it would be good to synchronize @KarolDawidziuk. I guess that the trick for scrolling iframe/window was aimed at https://github.com/ckeditor/ckeditor4/pull/4767#issuecomment-880481008 ?

KarolDawidziuk:
After F2F talk with @jacekbogdanski I've prepared a short screen-records for each of one for the better understanding of the problem.

**Note**
These screen-records are version of editor without fix from this PR.

1. **Scroll block**
@LukaszGudel catch. After allowing the editor to maximize, there was a problem with stuck scrolling. This is due to a bug that prevents scrolling when the height of the content frame is greater than the height of the screen.  

https://user-images.githubusercontent.com/18471998/130434479-abc1b06a-14b3-46ab-aab8-a5d97606c873.mp4

To fix this issue I calculate a new height of iframe when editor is maximized and the editable is focused.

2. **Wrong autofocus position**
After adding the fix to the above problem, we get another bug. 
When we have the editor maximized, but not focused, and we want to focus it, the area will be automatically zoomed to the cursor position (by iOS). This position is computed before shrinking editable. So in situation when before zoomed editable have 1500px height and we want activate cursor somewhere near 1300px, the page position will be shifted by 1300px instead of shifting the iframe content by the given height. (see screen-record).


https://user-images.githubusercontent.com/18471998/130450062-26e4fd70-cd02-4511-a32d-878540ee9d71.mp4

To fix this, after the `paste` event I set the iframe position on the y-axis and the entire editor on the x-axis.

PR #4800, opened 2021-07-23
Prevent invocation on not existing editable [FF]

### LINKED ISSUE #4782: source tab view out gives "Uncaught TypeError: b is null in Firefox"
Does this happen with "normal" installation too? No, it does not. It is only part of my ckeditor plus implementation.

## Type of report

Bug

  -  Open console.
  -  Open basic or standard or full editor form 
  -  Either / or new form or edit an entry
  -  Press source tab button and tab again to get out

### Expected result

No error :)

### Actual result

Firefox 89 (with lots of detail steps) throws: `Uncaught TypeError: b is null`. FF 90 says `Uncaught TypeError: can't access property "isInline", b is null` in updateAriaAttributesOnEditable function.
It does not seem to harm anything essential having this TypeError.
Chromium Version 93 does not have this issue.

if I change `b.isInline()` in
```js
addAriaAttributesToEditable:function(){var a=this.editor.editable(),b=this.view.element.getAttribute("id");a.isInline()&&(a.setAttribute("aria-controls",b),a.setAttribute("aria-activedescendant",""),a.setAttribute("aria-autocomplete","list"),a.setAttribute("aria-expanded","false"))},updateAriaAttributesOnEditable:function(a){var b=
this.editor.editable();a=a.data;b.isInline()&&(b.setAttribute("aria-expanded",a?"true":"false"),a||b.setAttribute("aria-activedescendant",""))},
```
to `(b?b.isInline():null)` as
```js
addAriaAttributesToEditable:function(){var a=this.editor.editable(),b=this.view.element.getAttribute("id");a.isInline()&&(a.setAttribute("aria-controls",b),a.setAttribute("aria-activedescendant",""),a.setAttribute("aria-autocomplete","list"),a.setAttribute("aria-expanded","false"))},updateAriaAttributesOnEditable:function(a){var b=
this.editor.editable();a=a.data;(b?b.isInline():null)&&(b.setAttribute("aria-expanded",a?"true":"false"),a||b.setAttribute("aria-activedescendant",""))},
```
the error is gone and all seems to work well.
I can't say where this error originates from.

This is my custom config:
https://github.com/ophian/additional_plugins/blob/master/serendipity_event_ckeditor/cke_config.js

## Other details

* Browser: FF 89 and FF 90 (in my plugin version of CKE in both)
* OS: Win and Linux
* CKEditor version: 4.16.1
* Installed CKEditor plugins: codesnippet and more (- I don't think this is reproducible for others without using my Styx environment.)

If someone with more insight could lead me on this, I would be happy. Maybe this null check in ckeditor is officially a need or not, or how to rewrite my configuration to not have this error thrown. Thank you!

sculpt0r:

Bug fix

### This PR contains

- [ ] Unit tests
- [X] Manual tests

```
* [#4782](https://github.com/ckeditor/ckeditor4/issues/4782): Fixed: TypeError in autocomplete for FF.
```

## What changes did you make?

I just added some prevention in one `if` statement. So in case the editable is `null` nothing happened. The source of the issue is described [here](https://github.com/ckeditor/ckeditor4/issues/4782#issuecomment-885500886). Also this event is fired again anyway after the editable is fully initialized. So ignoring it one time (actually only in FF) doesn't break the feature. 

I wasn't able to get this error when I switched modes with automatic tests. So I decided to use the manual one.

Closes #4782 .

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -15,7 +15,7 @@ Fixed Issues:
 * [#3638](https://github.com/ckeditor/ckeditor4/issues/3638): Fixed: Opening the same dialog twice causes it to become hidden under the dialog's page cover.
 * [#4247](https://github.com/ckeditor/ckeditor4/issues/4247): Fixed: [Color Button](https://ckeditor.com/cke4/addon/colorbutton)'s incorrect rendering on the first opening.
 * [#4555](https://github.com/ckeditor/ckeditor4/issues/4555): Fixed: [Font](https://ckeditor.com/cke4/addon/font) styles with attributes are not applied correctly when used multiple times over the same selection.
-* [#4782](https://github.com/ckeditor/ckeditor4/issues/4782): [FF] Fixed: TypeError in autocomplete.
+* [#4782](https://github.com/ckeditor/ckeditor4/issues/4782): [Firefox] Fixed: `TypeError` is thrown when switching to Source View and back while [Autocomplete](https://ckeditor.com/cke4/addon/autocomplete) plugin is enabled.
 
 ## CKEditor 4.16.1
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -15,6 +15,7 @@ Fixed Issues:
 * [#3638](https://github.com/ckeditor/ckeditor4/issues/3638): Fixed: Opening the same dialog twice causes it to become hidden under the dialog's page cover.
 * [#4247](https://github.com/ckeditor/ckeditor4/issues/4247): Fixed: [Color Button](https://ckeditor.com/cke4/addon/colorbutton)'s incorrect rendering on the first opening.
 * [#4555](https://github.com/ckeditor/ckeditor4/issues/4555): Fixed: [Font](https://ckeditor.com/cke4/addon/font) styles with attributes are not applied correctly when used multiple times over the same selection.
+* [#4782](https://github.com/ckeditor/ckeditor4/issues/4782): [FF] Fixed: TypeError in autocomplete.
 
 ## CKEditor 4.16.1
 
--- a/tests/plugins/emoji/manual/editablenullerror.html
+++ b/tests/plugins/emoji/manual/editablenullerror.html
@@ -1,7 +1,7 @@
 <textarea id="editor"></textarea>
 
 <script>
-	if ( !CKEDITOR.env.gecko || bender.tools.env.mobile ) {
+	if ( bender.tools.env.mobile ) {
 		bender.ignore();
 	}
 
--- a/tests/plugins/emoji/manual/editablenullerror.html
+++ b/tests/plugins/emoji/manual/editablenullerror.html
@@ -1,0 +1,9 @@
+<textarea id="editor"></textarea>
+
+<script>
+	if ( !CKEDITOR.env.gecko || bender.tools.env.mobile ) {
+		bender.ignore();
+	}
+
+	CKEDITOR.replace( 'editor' );
+</script>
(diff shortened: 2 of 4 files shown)


tests/plugins/emoji/manual/editablenullerror.html
```diff
@@ -0,0 +1,9 @@
+<textarea id="editor"></textarea>
+
+<script>
+	if ( !CKEDITOR.env.gecko || bender.tools.env.mobile ) {
+		bender.ignore();
+	}
```
f1ames, line 6: The fix was applied to all browsers while the test covers only FF :thinking: I know there were no issue in other browsers but testing them ensures there are no regressions. So if the code change affects all browsers the test should cover all browsers too.

f1ames (changes requested):
The fix itself looks reasonable and works fine :+1: I have only some doubts regarding manual test (see review comment).

> I wasn't able to get this error when I switched modes with automatic tests. So I decided to use the manual one.

:+1:

sculpt0r:
Hi, @f1ames - PR is ready for another review.

I've removed the gecko restriction from manual test. The mobile restriction is still there due to the dev console usage...

tests/plugins/emoji/manual/editablenullerror.md
```diff
@@ -0,0 +1,9 @@
+@bender-tags: 4.16.2, bug, emoji, 4782
+@bender-ckeditor-plugins: wysiwygarea, sourcearea, toolbar, emoji 
+@bender-ui: collapsed
+
+1. Open browser's dev console.
+1. Switch to `source` mode.
+1. Switch to `wysiwyg` mode.
+
+**Expected** No errors in the console.
```
f1ames, line 9: I somehow missed that in the previous review... and it's good that this test covers the initial issue (error) but maybe it could have one more step to open emoji panel so we can see if nothing breaks? ~Unless we are 100% sure that...~ Well, because this change affects how autocomplete panel (and so emoji) works so maybe let's have step covering regressions too?

CHANGES.md
```diff
@@ -14,6 +14,7 @@ Fixed Issues:
 * [#3638](https://github.com/ckeditor/ckeditor4/issues/3638): Fixed: Opening the same dialog twice causes it to become hidden under the dialog's page cover.
 * [#4247](https://github.com/ckeditor/ckeditor4/issues/4247): Fixed: [Color Button](https://ckeditor.com/cke4/addon/colorbutton)'s incorrect rendering on the first opening.
 * [#4555](https://github.com/ckeditor/ckeditor4/issues/4555): Fixed: [Font](https://ckeditor.com/cke4/addon/font) styles with attributes are not applied correctly when used multiple times over the same selection.
+* [#4782](https://github.com/ckeditor/ckeditor4/issues/4782): [FF] Fixed: TypeError in autocomplete.
```
f1ames, line 17:
```suggestion
* [#4782](https://github.com/ckeditor/ckeditor4/issues/4782): [Firefox] Fixed: `TypeError` is thrown when switching to Source View and back while [Autocomplete](https://ckeditor.com/cke4/addon/autocomplete) plugin is enabled.
```

f1ames (changes requested):
Almost there :wink: :+1:

sculpt0r:
Hi @f1ames,
according to our F2F conversation - I'm re-requesting you. I correct proposals to the changelog.

tests/plugins/emoji/manual/editablenullerror.md
f1ames: After F2F talks we decided to leave it as is :+1:

f1ames:
Rebased onto latest `master`.

f1ames (approved):
LGTM :+1:

Issue #4790, opened 2021-07-14
Incorrect fire Print on FF


KarolDawidziuk:
## Type of report

Bug

## Provide detailed reproduction steps (if any)

Open editor on FF and click on print button.

### Expected result

Preview and print tab should open correctly. No error in the console.

### Actual result

Preview are opened correctly but print preview is not available. No errors in the console.

## Other details

This happens because FF trying fire `print()` method before page was fully loaded. We need wait for that.
In the `preview` plugin we can create some kind of public API to allow adding callback that fired when the preview window is loaded.

* Browser:FF
* OS: macOS, Windows

KarolDawidziuk:
Blocked by #4444.

f1ames:
As #4444 is done we can proceed with this one.

f1ames:
@KarolDawidziuk could you give more details on this one? Following the description (with CKEditor 4 `4.16.1` and Firefox `90.0.2`) it seems to be working fine. I see no content loss. Maybe some additional condition needs to be fulfilled? :thinking: 

Preview window:
![image](https://user-images.githubusercontent.com/1061942/128204457-ad679cec-05f7-406a-9493-9bf0f1e2badc.png)

Print window:
![image](https://user-images.githubusercontent.com/1061942/128204510-3aac6e8a-2bd1-46e5-ba98-0195c75ada68.png)

KarolDawidziuk:
@f1ames It's strange because for me it's not working. There are no additional steps for reproducing it that were not mentioned in #4444. I tested it on macOS and Win7(vBox) and it fails. After [#4758](https://github.com/ckeditor/ckeditor4/pull/4758) fix `preview` works but `print` still not.

f1ames:
@KarolDawidziuk so if you mentioned #4444 it means it is CDN related too? It's not mentioned anywhere in this ticket or issue description so if that's the case it would be good to update repo steps / description here.

PR #4813, opened 2021-08-09
Fix fire print on Firefox

KarolDawidziuk:

Bug fix 

### This PR contains

- [x] Unit tests
- [ ] Manual tests

```
* [#<4790>](https://github.com/ckeditor/ckeditor4/issues/4790): Fixed: Incorrect fire Print on FF
```

## What changes did you make?

The problem was with the page load status in FF mode, because an attempt to print was triggered when the preview page was not ready. So I added a `onreadystatechange` callback to wait for the `complete` status.

Closes [#4790](https://github.com/ckeditor/ckeditor4/issues/4790).

### DIFF
--- a/plugins/preview/plugin.js
+++ b/plugins/preview/plugin.js
@@ -24,7 +24,9 @@
 				modes: { wysiwyg: 1 },
 				canUndo: false,
 				readOnly: 1,
-				exec: CKEDITOR.plugins.preview.createPreview
+				exec: function() {
+					CKEDITOR.plugins.preview.createPreview( editor );
+				}
 			} );
 			editor.ui.addButton && editor.ui.addButton( 'Preview', {
 				label: editor.lang.preview.preview,
@@ -51,8 +53,8 @@
 		 * @param {CKEDITOR.editor} editor The editor instance.
 		 * @returns {CKEDITOR.dom.window} A newly created window that contains the preview HTML.
 		 */
-		createPreview: function( editor ) {
-			var previewHtml = createPreviewHtml( editor ),
+		createPreview: function( editor, callback ) {
+			var previewHtml = createPreviewHtml( editor, callback ),
 				eventData = { dataValue: previewHtml },
 				windowDimensions = getWindowDimensions(),
 				// For IE we should use window.location rather than setting url in window.open (https://dev.ckeditor.com/ticket/11146).
@@ -94,7 +96,7 @@
 		}
 	};
 
-	function createPreviewHtml( editor ) {
+	function createPreviewHtml( editor, callback ) {
 		var pluginPath = CKEDITOR.plugins.getPath( 'preview' ),
 			config = editor.config,
 			title = editor.lang.preview.preview,
@@ -113,6 +115,7 @@
 				'<link rel="stylesheet" media="screen" href="' + pluginPath + 'styles/screen.css">' +
 			'</head>' + createBodyHtml() +
 				editor.getData() +
+				setPrintCallback( callback ) +
 			'</body></html>';
 
 		function generateBaseTag() {
@@ -154,6 +157,14 @@
 			}
 
 			return html;
+		}
+
+		function setPrintCallback( callback ) {
+			if ( !callback ) {
+				return '';
+			}
+
+			return '<script> window.onreadystatechange = firePrint </script>';
 		}
 	}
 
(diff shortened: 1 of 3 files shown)


plugins/preview/plugin.js
```diff
@@ -155,6 +158,14 @@
 
 			return html;
 		}
+
+		function setPrintCallback( callback ) {
+			if ( !callback ) {
+				return '';
+			}
+
+			return '<script> window.onreadystatechange = firePrint </script>';
```
Comandeer, line 167: It should be bound to `document`, because `readystatechange` event does not bubble up to `window`.

```suggestion
			return '<script> document.onreadystatechange = firePrint; </script>';
```

Comandeer (changes requested):
I'm thinking about refactoring the whole thing a little bit and moving most of its part into `preview` plugin (see my inline comments).

Additionally there are two failing tests inside `print` suite (the one you modified and one about immediately invoking print on `document.readyState === 'complete'`). However, instead of fixing, we should probably remove them and instead add some tests to `preview` plugin to check if passed calback has been executed in the right time.

KarolDawidziuk:
Thanks, @Comandeer for refactoring and review. 
It's ready for another one.

Changes:
* update docs,
* revert old version of unit test from `tests/../print.js`,
* removed two unit tests from `tests/../print.js`,
* added unit test.

Comandeer (changes requested):
Unfortunately our current approach does not work correctly in IE as it does not change `document.readyState` to `complete` if there are any images in the content. Additionally simply switching to `load` event instead also won't work, because Safari does not fire `load` for subsequent opening of preview window. It means that we have to introduce two different flows – one for IE, using (probably) `load` event and other one for all other browsers.

Comandeer:
Blocked by #4824.

KarolDawidziuk:
@Comandeer, I've added two ways to invoke the `print` callback. One for IE with using `onload` and `onreadystatechange` for all other browsers.

plugins/preview/plugin.js
```diff
@@ -74,27 +78,45 @@
...
+					if ( !CKEDITOR.env.ie && nativePreviewWindow.document.readyState === 'complete' ) {
+						callback( previewWindow );
+					}
+
+					if ( CKEDITOR.env.ie ) {
+						callback( previewWindow );
+					}
+				};
+
+				nativePreviewWindow.fireCallback();
```
Comandeer, line 112: Invoking it here means that in IE callback is invoked immediately, without waiting for `load` event (due to the second `if`).

Probably we will need to duplicate check of `readyState` outside of `fireCallback`.

Comandeer (changes requested):
Overally looks good, there's only one issue with the moment of invoking callback in IE (see inline comment).

Please also rebase the PR onto latest `master`.

KarolDawidziuk:

Now it's working properly for me on every browser, and it's ready for another review.  
Also, due to the rollback of my previous changes related to the CDN opening "preview"  problem  [#4444](https://github.com/ckeditor/ckeditor4/issues/4444) I am adding these changes to this PR.

Comandeer (changes requested):
Fix seems to work well, good job!

However there are still some small issues:

* see my inline comments;
* there is a mismatch between manual test `.md` and `.html` file, resulting in non-functional test – please fix it;
* it'd be good to add two more unit tests – one for IE and checking if callback is really fired on `load` and one for other browsers to check if it's fired on `readystatechange` with appropriate `readyState`;
* and traditionally – rebase onto latest `master`.

plugins/preview/plugin.js
```diff
@@ -155,6 +174,21 @@
...
+
+		function setPrintCallback( callback ) {
+			if ( !callback ) {
+				return '';
+			}
+
+			// On IE onreadystatechange does not change document.readyState to complete if there are any images in the content.
+			// So we need introduce a two flows. One for IE and second for all other browsers. (#4790)
+			if ( CKEDITOR.env.ie ) {
+				return '<script> window.focus(); window.onload = function() { fireCallback() }; </script>';
```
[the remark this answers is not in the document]
KarolDawidziuk: > Is it focusing window necessary? Fix seems to work correctly after deleting it.

I have cases where it doesn't work without it, but now after ~20 tests on 3 VMs looks like it's working without it, so I'll delete it.

PR #4824, opened 2021-08-11
Revert "Fix: Correct open Preview in FF using editor via CDN"

jacekbogdanski:
Unfortunately, due to a much more complicated fix for https://github.com/ckeditor/ckeditor4/issues/4790 than we expected we need to move this PR to the next release.

Reverts ckeditor/ckeditor4#4758


Comandeer (approved):
LGTM!

PR #4837, opened 2021-08-13
Improve performance for onpaste event

### LINKED ISSUE #4807: Improve the performance of pasting images in high resolution
## Type of report

Bug

1) Go to https://ckeditor.com/ckeditor-4/demo/
2) Paste a highres image with a size of 15-20 MB

On my computer it takes **7 seconds to paste a 17MB image** (excluding any uploading for optimization):
image

For comparison, the browser is able to handle this **in just 1 second** as demonstrated with this JSFiddle (same image used for testing):
https://jsfiddle.net/9boeaLtc/
image

With retina displays and highres cameras everywhere, images with a size of 5-10-20 megabytes are here, so this is a real problem.

### Expected result

Usable performance for large images.

### Actual result

Very poor performance for large images.

## Other details

* Browser: Chrome Version 92.0.4515.107 (Official Build) (x86_64)
* OS: macOs 10.13.6 (17G14033)
* PC: MacBook Pro (15-inch, 2017), 2,8 GHz Intel Core i7, 16 GB 2133 MHz LPDDR3
* CKEditor version: 4.16.1
* Installed CKEditor plugins:

FlowIT-JIT:

Bug fix

See issue and below.

### This PR contains

- [ ] Unit tests
- [X] Manual tests

Test as described in #4807.
First verify poor performance by pasting an image of 15-20 MB. Notice how Chrome logs out something like "[Violation] 'paste' handler took 9829ms"

Next, test with the fixed version and observe how performance have greately improved for the 'paste' event:
"[Violation] 'paste' handler took 1919ms". Not perfect, but _much_ better.

CKEditor 4 allows for pasting images as base64. Images with a size of 20 MB, represented as base64, makes CKEditor perform quite poorly. But for plugins using blob storage this pull request makes image pasting usable for large images.

```
* [#4807](https://github.com/ckeditor/ckeditor4/issues/4807): Improved performance for onpaste event.
```

## What changes did you make?

Reduced calls to DataTransfer.files by storing the files collection in a local variable.

Closes #4807.


sculpt0r:
Rebased on newest major

sculpt0r (changes requested):
Hi @FlowIT-JIT 

Thank you for your effort in finding this issue and its resolution. 

My local full test runs on Safari, Chrome, FF, and IE11 show that nothing was broke - so that is good 👍 

Verifying this PR with your manual steps shows that, indeed, performance was increased. However, we need here something more solid and 'formal'. Please take a look at our [tests/plugins](https://github.com/ckeditor/ckeditor4/tree/major/tests/plugins) directory to check out how our manual and automatic tests look like. We cannot introduce changes without formal tests. Because if they are properly marked with tags we have 100% guarantee that someone will check them before release. Also, automatic tests are run on CI due to changes in related code (like related plugins).

Please, take a look at how to run a testing environment on your machine: https://ckeditor.com/docs/ckeditor4/latest/guide/dev_tests.html. 

Then you can check [code style guide](https://github.com/ckeditor/ckeditor4/blob/major/dev/docs/codestyle.md) which also covers tests style guide.

We are preferring automatic tests - since they can be performed without human involvement. You can try to play with [spies](https://sinonjs.org/releases/v9.2.1/spies/) in test to verify execution time 🤔

FlowIT-JIT:
Hi @sculpt0r 
I'm currently occupied with another project, but I expect to be able to provide either automated or manual (most likely) test cases in the near future. Whether that be 2 weeks from now or 12 weeks, I'm not sure. So please be patient.

FlowIT-JIT:
I'm currently occupied with another project, but I expect to be able to provide either automated or manual (most likely) test cases in the near future. Whether that be 2 weeks from now or 12 weeks, I'm not sure. So please be patient.

jacekbogdanski:
Thanks, @FlowIT-JIT for the info. We will be checking out this PR from time to time and see if it gets any updates.

FlowIT-JIT:
@sculpt0r , @jacekbogdanski :
I'm not entirely sure what you want me to do.
Do you need test cases that demonstrates and proves the performance improvement? Or test cases that proves pasting has not been broken? The latter is already covered.
If you want something that proves the performance gain, then I don't see the value in adding such test cases to the repository as you probably don't want to run them again for future releases, once the fix has been verified and applied.
Also, this fix is not something the user can toggle, so we need two different versions of CKEditor to verify - am I allowed to reference e.g. a JSFiddle running the affected version of CKEditor?
Furthermore I'm not familiar with the bender tags used in the test definitions and I can't seem to find any documentation for them - at least not very elaborate (https://ckeditor.com/docs/ckeditor4/latest/guide/dev_tests.html#common-tags-used-in-tests).

Suggestion which lends from https://raw.githubusercontent.com/ckeditor/ckeditor4/major/tests/plugins/clipboard/manual/pasteimage.md :

**pasteimageperformance.md**
```
@bender-ui: collapsed
@bender-tags: 4.16.1, clipboard, bug, performance, 4807
@bender-ckeditor-plugins: wysiwygarea, toolbar, undo, basicstyles, image, clipboard, sourcearea

## Scenario:

This test proves the performance gain achieved with pull request #4837, compared to older versions of CKEditor, when pasting large images. Run the test below with CKEditor version 4.16.2 or older (slow paste), and compare the performance to the most recent version of CKEditor (faster paste).

 1. Open Chrome
 2. Open Developer Tools.
 3. Set Developer Tools to display Console, and make sure log output is _Verbose_ under _Default levels_.
 4. Open a page with CKEditor 4.16.2 or older in _Standard_ mode, e.g. 
 5. Paste an image with a size of 15-25 MB from your clipboard into the editor.
 6. Observe the browser console - it will emit something like: [Violation] 'paste' handler took 7616ms.
 7. Repeat step 5-6 with the most recent version of CKEditor.

**Expected**:

Image is pasted as base64 encoded file.
The violaten logged in the browser console is reduced significantly with the most recent versions of CKEditor.
```

FlowIT-JIT:
New attempt to provide proof of successful implementation. Not ideal as it requires the user to download test material, but I'm not sure how to otherwise help with testing.

**pasteimageperformance.md**
```
@bender-ui: collapsed
@bender-tags: 4.16.1, clipboard, bug, performance, 4807
@bender-ckeditor-plugins: wysiwygarea, toolbar, undo, basicstyles, image, clipboard, sourcearea

## Scenario:

This test proves the performance gain achieved with pull request #4837, compared to older versions of CKEditor, when pasting large images.

1. Download https://codemagic.dk/CKEditorDemo-2.zip
2. Extract CKEditorDemo-2.zip
3. Open index.html in Chrome (it will run locally).
4. Make sure "CKEditor version" is set to "Old CKEditor"
5. Paste an image with a size of 15-25 MB into the editor - start a stopwatch immediately
6. When the image is visible, stop the stopwatch immediately - take note of time usage
7. Change "CKEditor version" to "New CKEditor (improved)"
8. Repeat step 5-6.

**Expected**:

It should be significantly faster to paste an image using the more recent version of CKEditor (approx. 250% improvement)
```

Also note that creating this sample revealed another problem with the Major branch: #4874

sculpt0r:
Hi @FlowIT-JIT 
First of all, I rebased your branch onto the newest major. We prefer rebasing rather than merging. It is because after merge your PR looks like it introduces additional changes(which comes from the merge commit). After rebase, your PR is still clear and contains only the code you are working on.

I try to help you with the tests soon. Thank you for your tries :)

sculpt0r:
>Do you need test cases that demonstrates and proves the performance improvement? Or test cases that proves pasting has not been broken? The latter is already covered.

That is right - we want something that proves that there is a benefit from this PR.

>If you want something that proves the performance gain, then I don't see the value in adding such test cases to the repository as you probably don't want to run them again for future releases, once the fix has been verified and applied.
Also, this fix is not something the user can toggle, so we need two different versions of CKEditor to verify - am I allowed to reference e.g. a JSFiddle running the affected version of CKEditor?

Actually, we have our own procedures to verify each version before release - so we are running a lot of test runs for the future release. Also, if in the future someone else changes sth in the clipboard that may affect the performance it will be revealed by the tests introduced with this PR.

Thank you for your effort in adding new tests. Your first attempt with manual tests was almost perfect 👍

I took it and made small adjustments: https://github.com/ckeditor/ckeditor4/pull/4837/commits/ed29ce4cad0fadea8fce2502b708d40602cbb68c

- used `4.17.0` bender tag - since our target is the next major release
- Shrink scenario steps - if you run our testing environment some of them were actually satisfied 
- Adjusted a little bit `Expected` and `Unexpected` section.

In the `html` file: I need to ignore this test on mobile (it is hard to get the console there) and ignore all browsers different from chrome - since this bug and the entire fix is for chrome only.

@FlowIT-JIT Please look at this test (where it is placed & its structure) and think if this one satisfied you. Feel free to modify it if you want to and let me know when this PR will be ready for another review. I hope that helps.

FlowIT-JIT:
@sculpt0r Thank you very much for taking the time to elaborate on these things. Very enlightening 🙂

There is a small problem though. Whether your test case succeeds will very much depend on the user's hardware. On a slow/regular computer it will fail as it will almost certainly not be able to handle the past operation in 2000 ms. A fast computer might even fail for an image with a size of 25 MB. That's why I initially constructed a test based on comparison. It should reveal a performance gain of approx. 250% (wall time) from key stroke (CTRL+V) to visible image.

I'm not sure what to do with this. I think it is pretty beneficial being able to "diff" the behaviour of two versions of CKEditor which is possible with https://codemagic.dk/CKEditorDemo-2.zip, but I also understand that this is a huge package to include in the test suite.

What to do? 🤔

sculpt0r:
> @sculpt0r Thank you very much for taking the time to elaborate on these things. Very enlightening 🙂
> 
> There is a small problem though. Whether your test case succeeds will very much depend on the user's hardware. On a slow/regular computer it will fail as it will almost certainly not be able to handle the past operation in 2000 ms. A fast computer might even fail for an image with a size of 25 MB. That's why I initially constructed a test based on comparison. It should reveal a performance gain of approx. 250% (wall time) from key stroke (CTRL+V) to visible image.
> 
> I'm not sure what to do with this. I think it is pretty beneficial being able to "diff" the behaviour of two versions of CKEditor which is possible with https://codemagic.dk/CKEditorDemo-2.zip, but I also understand that this is a huge package to include in the test suite.
> 
> What to do? 🤔

@FlowIT-JIT  You're right. Including external packages seems to be 'hacky' way to me. However, we already have one test that simulates loading CKEditor4 from external sources: https://github.com/ckeditor/ckeditor4/blob/major/tests/plugins/preview/manual/previewcdn.html

Maybe you could try to reuse some of that logic, and create sth like:
- perform this test on the editor provided by bender (so in the normal way)
- create a button that after click will wipe out the editor and load it from https://github.com/ckeditor/ckeditor4/blob/42384fb985c39cc5f9db126e39f3af28bd4a51a2/tests/plugins/preview/manual/previewcdn.html#L98 but from https://github.com/ckeditor/ckeditor4/tree/4.16.2 (so we will have a stable released editor where the performance is poor)
Confront times in logs from both versions?

That should solve the problem with external packaging and cover the problem with hardware dependency. WDYT?

FlowIT-JIT:
> That should solve the problem with external packaging and cover the problem with hardware dependency. WDYT?

@sculpt0r Sounds good to me. I'll look into it. Thank you :-)

FlowIT-JIT:
@sculpt0r I got Bender up and running and began constructing the test case, but CKEditor 4.16.x does not ship with any plugins capable of handling pasted images. I could probably use `CKEDITOR.plugins.addExternal(...)` to register a plugin from a CDN/GitHack, but I doubt you want external dependencies like that. What are my options?
It's actually still possible to test paste performance without a plugin - the console will still emit the violation, but no image will be added to the editor. Perhaps that's more acceptable than relying on an external plugin.
I would appreciate your input on this - thanks.

sculpt0r:
Hi @FlowIT-JIT 
Nice that you are able to run bender :) To add plugin, just add it to the line with `@bender-ckeditor-plugins: wysiwygarea, toolbar, undo, basicstyles, image, clipboard, sourcearea`. So currently there is `image` plugin added. You can use `image2` for example. Bender will add a plugin to the test case from that list. Hope, that helps 👍

FlowIT-JIT:
@sculpt0r Thank you. As far as I can see, neither `Image` nor `Image2`provides support for pasting images. This is handled by the clipboard plugin in 4.17.x, but not supported in 4.16.x
And to make things worse, the approach you suggested based on https://github.com/ckeditor/ckeditor4/blob/major/tests/plugins/preview/manual/previewcdn.html no longer works. The second you nullify CKEDITOR to allow a different version to be loaded, an observer registered by CKEditor starts to throw errors and eventually causes the browser tab to crash. I have reported the problem here: #4878

But no worries. What I did instead was to dynamically create a new editor instanse in an iframe. This works great. But in order for image pasting to work, I still need a plugin to handle this. I merely copied the source code from an existing one that I wrote (https://github.com/Jemt/Fit.UI/blob/master/Resources/CKEditor/plugins/base64imagepaste/plugin.js). The code is LGPL so it should be fine.

Is this approach acceptable? I have just committed everything so it is available for you to see.

sculpt0r (changes requested):
Hi @FlowIT-JIT 
Thank you for the report on failing the CDN test. Actually, it fails due to a new feature that will be introduced in the next release. With this new feature - we introduce [additional config](https://github.com/ckeditor/ckeditor4/blob/1058d33757b3566bc87cac3971cad915d466c2c2/plugins/wysiwygarea/plugin.js#L838). We are already preparing fix for this test.

I very like your reproduction steps - they are exactly matching the testing case 👍 Good job :)

- If you still want to use a solution similar to the CDN test, just pass some empty elements from the HTML to mentioned config option. Make sure this element will never contain editor instance. It seems to be easier to read separate JS and HTML code - rather than code and HTML used as string literals.

- If you want to keep your solution - it looks good - but it is harder to read. However, I'm not sure about the custom plugin. We rather want to test the editor 'as native as possible. Because it is a common situation, that some external behaviors have an impact on the code and we can't track them.

From my perspective, a simple `image` plugin is enough to reproduce this issue. It does not paste the base64 format, but still - we receive quite a big waiting time.  So please switch the CDN version to `standard` and add `extraPlugins: 'image'` in the new editor instance config.

FlowIT-JIT:
@sculpt0r 

Thanks for the feedback :-)

I see you point in simplifying the code and get rid of the paste plugin I added.
I do see a couple of benefits with the iframe solution though:

- Both editors can run side by side
- They are isolated from each other - low risk of code interference (now and in the future)

I kept the iframe solution, removed the custom paste plugin I added, loaded "standard" from CDN (no need to add 'image' to extraPlugins - it's already loaded), and made some notes regarding expected behaviour when pasting image data from different sources.

sculpt0r (changes requested):
Hi @FlowIT-JIT,
Thank you for all corrections you have made. Also for the fast time of response.

>I kept the iframe solution, removed the custom paste plugin I added, loaded "standard" from CDN (no need to add 'image' to extraPlugins - it's already loaded), and made some notes regarding expected behaviour when pasting image data from different sources.

That is totally fine 👍 That is nice, so you include those additional notes :)

The entire PR looks almost ready 👍 👏  I left a lot of inline comments, but don't worry. Those are only for some cosmetic changes. Things to do:
- separate JS and HTML a little bit,
- correct string literal readability,
- improve variable naming
- move additionally notes before steps that are using them

Overall, good job! I think that after those changes we will be able to merge :) Please let me know if I can refer to your name as GH profile (https://github.com/FlowIT-JIT) in the changelog?

FlowIT-JIT:
Hi @sculpt0r

Thanks for taking the time to guide me through this. It's my first attempt at contributing to an open source project that I do not manage myself, so this has been educational 😊

I took the liberty to add the notes _below_ the steps related to them, rather than before. I found that more intuitive.

image

Hopefully we are all set now. Thanks again for your help 💪👊👍😊

sculpt0r (approved):
Hi @FlowIT-JIT 
Thank you for the collaboration :) I added the changelog with a link to your GH profile ;)

Now, this solution seems to be complete :) Because of some issue with our CI config, I have to make a full test run on my PC, but everything looks good :)

As for the first time, I found collaboration with you was very good. Your response time was very fast. Your approach to adjust the solution to our rules was also fine 👍 It is always hard to switch to someone else standards :)

So, thank you again :)

FlowIT-JIT:
Thanks @sculpt0r :-)

PR #4841, opened 2021-08-15
Integrate maximize plugin with History API

### LINKED ISSUE #4374: Integrate maximize plugin with history API
## Type of report

Feature request

Follow-up of ckeditor/ckeditor4-integrations-common#8. It seems the best way to tackle this issue (already reported for Angular and Vue integrations) would be to introduce such mechanism in the plugin itself so it will fix the issue with most of the SPA integrations - not only official Angular/React/Vue ones.

Comandeer:

New feature

### This PR contains

- [x] Unit tests
- [x] Manual tests

```
* [#4374](https://github.com/ckeditor/ckeditor4/issues/4374): Integrated [Maximize](https://ckeditor.com/cke4/addon/maximize) plugin with browser's History API.
```

## What changes did you make?

I've added a simple `popstate` event listener, which minimize editor if it's maximized. This way the editor returns to normal dimensions when user clicks back or forward button in the browser.

There are no unit tests, unfortunately, because some of browsers (Blink-based ones) require user gesture to allow pushing to history with enabling back/forward buttons at the same time.

Closes #4374.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -11,7 +11,8 @@ New Features:
 * [#4612](https://github.com/ckeditor/ckeditor4/issues/4612): Allow pasting images as Base64 from [clipboard](https://ckeditor.com/cke4/addon/clipboard) in all browsers except IE.
 * [#4681](https://github.com/ckeditor/ckeditor4/issues/4681): Allow drag and drop images as Base64.
 * [#4750](https://github.com/ckeditor/ckeditor4/issues/4750): Added notification for pasting and dropping unsupported file types into the editor.
-* [#3433](https://github.com/ckeditor/ckeditor4/issues/3433): Marked required fields in dialogs with asterisk (*) symbol.
+* [#3433](https://github.com/ckeditor/ckeditor4/issues/3433): Marked required fields in dialogs with asterisk (`*`) symbol.
+* [#4374](https://github.com/ckeditor/ckeditor4/issues/4374): Integrated [Maximize](https://ckeditor.com/cke4/addon/maximize) plugin with browser's History API.
 
 Fixed Issues:
 
--- a/core/dom/event.js
+++ b/core/dom/event.js
@@ -226,3 +226,13 @@ CKEDITOR.HISTORY_NATIVE = 1;
  * @member CKEDITOR
  */
 CKEDITOR.HISTORY_HASH = 2;
+
+/**
+ * Switch off integration with browser's "Go back" and "Go forward" buttons.
+ *
+ * @since 4.17.0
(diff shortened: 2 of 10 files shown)


jacekbogdanski (changes requested):
I've checked if this change fixes the issue in different integrations:

## Vue

Unfortunately, `mainWindow.on` event is not called at all for Vue integration. I've updated it to `mainWindow.$.addEventListener` and it started to trigger the event on changing routes, but still, it looks like VueRouter is triggering `popstate` event **after** it updates the view, so the issue still occurs. I see that Vue is exposing [Navigation Guards](https://router.vuejs.org/guide/advanced/navigation-guards.html#global-resolve-guards) which could fix the issue, but it would require a specific Vue integration fix, not a general one like this.

## Angular
I've had some issues running the demo which would allow us to reproduce the issue as it seems not reproducible with our demo. However, I see that `popstate` event is fired before Angular is changing the view, so I can guess that the fix proposed in this PR should work correctly. Although, it is something we need to make sure of before accepting changes.

## React
Again, we don't have any samples allowing us to reproduce the issue. I will try to validate it later. We didn't have a reported issue for React but I suppose it also exists for this integration.

jacekbogdanski:
I've checked this fix with React and it seems to be working correctly.

jacekbogdanski:
Let's add Angular sample https://github.com/ckeditor/ckeditor4-angular/issues/217 when working on this PR.

Comandeer:
It seems that whole our sample app in Angular is based on router, so it's probably not needed to add another one.

However I tested the app with the fix and it seems to work well.

Comandeer:
Testing Vue router reminded me that there's another popular solution for routing: hashbang navigation. I've added support for it and also support for switching off history integration via `config.maximize_historyIntegration` config option. Switching off could help in cases where custom integration is needed (like in Vue).

I've also added new constants that would be used alongside config variable mentioned above: `CKEDITOR.HISTORY_NATIVE`, `CKEDITOR.HISTORY_HASH` and `CKEDITOR.HISTORY_OFF`.

jacekbogdanski (changes requested):
Looks good! I can confirm that it fixes both React and Angular integrations. Nice catch with the hash route.

We are missing unit tests for these changes. It's hard to reproduce real browser behavior with unit testing but it's covered with manual tests. However, we probably could stub `popstate` and `hashchange` events and verify if indeed maximize plugin has been turned off. WDYT?

Comandeer:
I've added tests.

jacekbogdanski (approved):
🎉

PR #4849, opened 2021-08-18
Fix failing tests/core/filter/editor

### LINKED ISSUE #4821: Failing tests/core/filter/editor
## Type of report

**Note:** The tests are only failing on built version as they include `fakeobjects` plugin which uses an active editor filter.

IMG_6D723608FBC7-1

## Other details

* Browser: ios
* OS: …
* CKEditor version: 4.16.2

sculpt0r:
It's no longer valid since fakeobjects plugin is using ACF

Failing test fix

### This PR contains

- [ ] Unit tests
- [ ] Manual tests

```
skip
```

## What changes did you make?

*Give an overview…*

I removed the failing assertion. This assertion is no longer valid since `fakeobjects` plugin is using ACF. It was not possible to remove this plugin with `removePlugins` config option since this one is inlining in build process into `ckeditor` file.

There are still two assertions left that verify the usage of two filter types.

Closes #4821

### DIFF
--- a/tests/core/filter/editor.js
+++ b/tests/core/filter/editor.js
@@ -283,9 +283,6 @@
 					editor.setActiveFilter( filter );
 					editor.insertHtml( '<u class="active">A1</u>-<u class="main">A2</u>' );
 
-					filter.applyTo = function() {
-						assert.fail( 'active filter should not be used while getting data from editor' );
-					};
 					assert.areSame( '<p><u class="active">A1</u>-A2</p>', editor.getData(), 'data was filtered using active filter' );
 
 					bot.setData( '<p><u class="active">A1</u>-<u class="main">A2</u></p>', function() {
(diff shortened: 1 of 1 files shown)


jacekbogdanski (changes requested):
I was rather thinking about removing Fake Objects plugin (so probably also iframe, flash, pagebreak, etc as they use it under the hood) from these specific tests, so we still keep test coverage for this particular feature. WDYT?

sculpt0r:
Rebased on the newest master

sculpt0r:
> I was rather thinking about removing Fake Objects plugin (so probably also iframe, flash, pagebreak, etc as they use it under the hood) from these specific tests, so we still keep test coverage for this particular feature. WDYT?

Unfortunately, the code from fakeobjects plugin that is executed in `afterInit` method is executed even without plugins that require this plugin (even not directly: `flash,forms,iframe,autolink,uploadfile,link,pagebreak,autoembed`) 🤔 Which looks like we add some functionality even if there is no plugin that relays on it...

jacekbogdanski (approved):

PR #4851, opened 2021-08-18
Move template content to HTML files.

### LINKED ISSUE #4850: Store templates in html files
## Type of report

Feature request

Currently the templates of the plugin [Content Templates](https://ckeditor.com/cke4/addon/templates) are stored as an concatenated String in the template file (like default.js). When there are many and, above all, long templates, the file gets really confusing. The HTML of the templates should be stored into seperated HTML files. The js file then should fetch the contents of the HTML file.

Fynn96:

New feature 

### This PR contains

- [x] Unit tests
- [x] Manual tests

```
* [#4850](https://github.com/ckeditor/ckeditor4/issues/4850): Store HTML templates in seperated files.
```

## What changes did you make?

The HTML snippets were moved to HTML files in the new html folder. Now default.js fetches the content of the HTML files instead of setting just a large String as html property. As the fetch is asynchronous the methode insertTemplate() in templates.js was modified to wait for the response.

Closes #4850.


Fynn96:
My first contribution ever. I hope I did everything right. Unfortunately, I couldn't run tests because the Bender installation didn't work. Maybe someone can make a contribution.
Feel free to give tips about contributions if I did anaything wrong!

plugins/templates/dialogs/templates.js
```diff
@@ -59,29 +59,31 @@
...
-		function insertTemplate( html ) {
+		function insertTemplate( htmlFile ) {
 			var dialog = CKEDITOR.dialog.getCurrent(),
 				isReplace = dialog.getValueOf( 'selectTpl', 'chkInsertOpt' );
 
-			if ( isReplace ) {
-				editor.fire( 'saveSnapshot' );
-				// Everything should happen after the document is loaded (https://dev.ckeditor.com/ticket/4073).
-				editor.setData( html, function() {
+			htmlFile.then(html => {
```
Comandeer, line 66: Unfortunately this change is not backwards compatible. The new way of getting template should be added alongside the old one (so there should be two independent flows after the change).

plugins/templates/templates/default.js
```diff
@@ -14,74 +14,41 @@ CKEDITOR.addTemplates( 'default', {
...
-	} ]
-} );
+		html: getHtml('text-and-table.html')
+	}
+	]
+});
+
+// Get the contents of the given html file.
+function getHtml(fileName) {
+	let path = CKEDITOR.getUrl(CKEDITOR.plugins.getPath('templates') + 'templates/html/' + fileName);
```
Comandeer, line 36: CKEditor 4 uses ES3 as it supports Internet Explorer 8+. Please refer to our [style guide](https://github.com/ckeditor/ckeditor4/blob/major/dev/docs/codestyle.md).

plugins/templates/templates/default.js
```diff
@@ -14,74 +14,41 @@ CKEDITOR.addTemplates( 'default', {
...
+	return fetch(path)
```
Comandeer, line 37: Fetch API can't be used directly as it's not supported in Internet Explorer and the plugin should not become broken there after introducing such changes. There is [`ajax` plugin](https://ckeditor.com/cke4/addon/ajax) that should cover asynchronous requests, however it can't be loaded in here – it has to be loaded inside the plugin.

Actually the whole function that fetches HTML from external files should be moved into the plugin itself. The template can inform plugin that it stores its HTML in an external file by using e.g. `htmlFile` property instead of `html` one.

plugins/templates/templates/default.js
```diff
@@ -14,74 +14,41 @@ CKEDITOR.addTemplates( 'default', {
...
+				throw Error(response.statusText);
+			}
+			return response;
+		})
+		.then(response => response.arrayBuffer())
+		.then(buffer => {
+		    let decoder = new TextDecoder("iso-8859-1");
+		    let text = decoder.decode(buffer);
+		    return text;
+		})
```
Comandeer, line 49: There is no need to treat data as binary. Additionally it should be always treated as encoded in UTF-8, not ISO-8859-1. It can be accomplished by using [`CKEDITOR.ajax#loadText()` function](https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_ajax.html#method-loadText) from `ajax` plugin.

plugins/templates/templates/default.js
```diff
@@ -14,74 +14,41 @@ CKEDITOR.addTemplates( 'default', {
...
+		.catch(error => {
+			console.log(error);
+			return '';
+		});
+}
```
Comandeer, line 54: There should be a newline at the end of the file. Please refer to our style guide.

Comandeer (changes requested):
Hi, @Fynn96! Thanks for your contribution! However there are some issues with your current solution. The biggest one is the fact that changes you propose are not backwards compatible. Instead of replacing the old way of loading templates, the new way need to be added alongside the old one. The plugin should have two separate flows for loading templates and it can decide which one to use by e.g. looking at properties available in the template. If there is an `html` one, it would use the old way of loading data, if there is an `htmlFile` (or similar) one, it would use the new approach.

You also used ES6+, which, unfortunately, can't be used inside CKEditor 4 codebase as it [supports also Internet Explorer 8+](https://ckeditor.com/docs/ckeditor4/latest/guide/dev_browsers.html#officially-supported-browsers). The same stands for using newer APIs, like Fetch one. Please see my inline comments for more info.

The PR lacks also tests. We can't accept changes (especially bigger ones) without proper tests as it makes verifying if everything still works much more difficult. You can refer to [tests for other plugins](https://github.com/ckeditor/ckeditor4/tree/major/tests/plugins) to get some ideas how to write tests for your PR.

Comandeer:
@Fynn96, I see that you pushed some changes since my last review. Is this PR ready for another one?

Fynn96:
@Comandeer Yeah my changes are ready now but I have trouble with bender.js. I wrote you a mail so maybe you could help me a bit. I can request your review again but I am not able to make some tests for my changes.

Comandeer:
Can you describe the issue here? I received some e-mail about CKEditor 4 recently, but it was empty.

Fynn96:
Sure. 
I tried to install bender.js as it is described here: https://ckeditor.com/docs/ckeditor4/latest/guide/dev_tests.html. 
My problem is I can't run `bender server run` because I get the following error:
`(node:17588) Warning: Accessing non-existent property 'padLevels' of module exports inside circular dependency
(Use `node --trace-warnings ...` to show where the warning was created)
(node:17588) Warning: Accessing non-existent property 'padLevels' of module exports inside circular dependency
(node:17588) Warning: Accessing non-existent property 'padLevels' of module exports inside circular dependency
(node:17588) [DEP0128] DeprecationWarning: Invalid 'main' field in '[...]\ckeditor4\node_modules\uid\package.json' of 'yes'. Please either fix that or report it to the module author
Potentially unhandled rejection [1] ReferenceError: primordials is not defined
    at fs.js:47:5
    at req_ ([...]\ckeditor4\node_modules\natives\index.js:143:24)
    at Object.req [as require] ([...]\ckeditor4\node_modules\natives\index.js:55:10)
    at Object.<anonymous> ([...]\ckeditor4\node_modules\gerard\node_modules\graceful-fs\fs.js:1:37)
    at Module._compile (node:internal/modules/cjs/loader:1101:14)
    at Object.Module._extensions..js (node:internal/modules/cjs/loader:1153:10)
   at Module.load (node:internal/modules/cjs/loader:981:32)
    at Function.Module._load (node:internal/modules/cjs/loader:822:12)
    at Module.require (node:internal/modules/cjs/loader:1005:19)
    at require (node:internal/modules/cjs/helpers:94:18)
`

I already tried mulitple node.js versions and even tested it on Linux and Windows, but still getting the same error.

Comandeer:
Did you try with Node.js 10? Unfortunately Bender.js is _a little_ bit old and can be problematic on newer Node.js versions.

Fynn96:
The oldest version I tried was 13. Version 10 finally worked. Thank you. I will request a new review as soon as I have created my tests.
Maybe it would be better to update your Test environment setup page as there is mentioned you need the latest release of node.js. Could be helpful for other contributors :)

Comandeer:
> Maybe it would be better to update your Test environment setup page as there is mentioned you need the latest release of node.js. Could be helpful for other contributors :)

Good catch, thanks! We'll update it.

plugins/templates/dialogs/templates.js
```diff
@@ -51,7 +51,14 @@
 			item.getFirst().setHtml( html );
 
 			item.on( 'click', function() {
-				insertTemplate( template.html );
+				if( template.htmlFile ) {
+					// Load HTML of the file before inserting it.
+					CKEDITOR.ajax.loadText( template.htmlFile, function( html ) {
+						insertTemplate( html );
+					});
```
Comandeer, line 58:
```suggestion
					} );
```

Comandeer, line 58:
```suggestion
					} );
```

plugins/templates/templates/default.js
```diff
@@ -27,61 +27,13 @@ CKEDITOR.addTemplates( 'default', {
...
-				'<td></td>' +
-				'<td>' +
-					'Text 2' +
-				'</td>' +
-			'</tr>' +
-			'</table>' +
-			'<p>' +
-			'More text goes here.' +
-			'</p>'
+		htmlFile: CKEDITOR.getUrl( 'plugins/templates/templates/html/strange-template.html' )
```
Comandeer, line 30: It's probably better to leave default templates as they were earlier. Especially because there is a difference in the behaviour: old templates are inserted synchronously and the new ones – asynchronously.

tests/plugins/templates/manual/insertTemplate.md
```diff
@@ -0,0 +1,27 @@
+@bender-tags: 4.16.2, feature
```
Comandeer, line 1:
```suggestion
@bender-tags: 4.17.0, feature, 4850
```

It will be released in 4.17.0. Please update version tag accordingly and add issue reference.

tests/plugins/templates/_assets/test.js
```diff
@@ -0,0 +1,26 @@
+/**
+ * @license Copyright (c) 2003-2021, CKSource - Frederico Knabben. All rights reserved.
+ * For licensing, see LICENSE.md or https://ckeditor.com/legal/ckeditor-oss-license
+ */
+
+// Register a templates definition set named "default".
+CKEDITOR.addTemplates( 'test', {
+	// The name of sub folder which hold the shortcut preview images of the
+	// templates.
+	imagesPath: '/tests//plugins/templates/_assets/images/',
```
Comandeer, line 10: Couldn't this path be simplified to `./images/`?

```suggestion
	imagesPath: './images/',
```

If not, then please remove the double slashes as they can cause an error:

```suggestion
	imagesPath: '/tests/plugins/templates/_assets/images/',
```

tests/plugins/templates/_assets/test.js
```diff
@@ -0,0 +1,26 @@
...
+
+	// The templates definitions.
+	templates: [ {
+		title: 'Title and Text',
+		image: 'titleAndText.gif',
+		description: 'A title and a text',
+		htmlFile: '/tests//plugins/templates/_assets/html/titleAndText.html'
```
Comandeer, line 17: The same here:

```suggestion
		htmlFile: './html/titleAndText.html'
```

or

```suggestion
		htmlFile: '/tests/plugins/templates/_assets/html/titleAndText.html'
```

tests/plugins/templates/manual/insertTemplate.html
```diff
@@ -0,0 +1,10 @@
+<textarea id="editor">
+	<p>Lorem ipsum dolor sit amet, consetetur sadipscing elitr</p>
+</textarea>
+
+<script>
+	CKEDITOR.replace( 'editor', {
+		templates_files: ['/tests//plugins/templates/_assets/test.js'],
```
Comandeer, line 7: The same here:

```suggestion
		templates_files: ['../_assets/test.js'],
```

or

```suggestion
		templates_files: ['/tests/plugins/templates/_assets/test.js'],
```

tests/plugins/templates/templates.js
```diff
@@ -0,0 +1,26 @@
+/* bender-tags: editor,dialog */
+/* bender-ckeditor-plugins: wysiwygarea,toolbar,basicstyles,templates,format*/
+
+( function() {
+    'use strict';
+
+    bender.test( {
+        'Abort template insertion': function() {
```
Comandeer, line 8: We have a convention that every test should start with `test`, so it should look like:

```suggestion
        'test aborting template insertion': function() {
```

Please update the test name accordingly.

tests/plugins/templates/templates.js
```diff
@@ -0,0 +1,26 @@
...
+                }
+            }, function( bot ) {
+                bot.dialog( 'templates', function( dialog ) {
+
+                    dialog.getButton('cancel').click();
+                    assert.areEqual( '<p>^Lorem ipsum</p>', bot.htmlWithSelection(), 'Editor data has not been altered.' );
+
+                } );
+            } );
+        },
```
Comandeer, line 23:
```suggestion
        }
```

We do not use dangling commas. Please remove them.

tests/plugins/templates/manual/insertTemplate.md
```diff
@@ -0,0 +1,27 @@
...
+@bender-ui: collapsed
+@bender-ckeditor-plugins: wysiwygarea, toolbar, dialog, templates, image, table, format, basicstyles
+
+Insert template and replace all contents
+========================================
```
Comandeer, line 6: We prefer marking headings with hash and the description should start from the level 2 heading:

```suggestion
## Insert template and replace all contents
```

Please update the headings.

tests/plugins/templates/manual/insertTemplate.md
```diff
@@ -0,0 +1,27 @@
...
+
+1. Click the `Templates` Button.
+2. Select the `insertOption` checkbox at the bottom.
+3. Choose the template `Title and Text`.
+ 
+
+**Expected:** The Text inside of the editor will be replaced by the heading `I am a title` and the text `I am a text`.
+
+Insert template at the cursor position
+======================================
```
Comandeer, line 16:
```suggestion
## Insert template at the cursor position
```

Comandeer (changes requested):
@Fynn96, thanks for all corrections you have made!

However there are still some issues, please see my inline comments. Additionally:

* It'd be good to have additional automatic tests, e.g. for inserting template with `htmlFile` field. Such a test can mock `CKEDITOR.ajax` to return some HTML and check if the editor's content is altered in a correct way. The test for inserting template with `html` field will be also nice. Could I ask you to add them?
* The filenames of manual tests should be lowercase. Please, rename `insertTemplate` to `inserttemplate`.
* The fact that inserting templates from file is asynchronous makes the whole thing a little bit tricky – as someone can start editing content before template is inserted. I'm not sure how to solve it. Maybe similar to how we deal with it in plugins that upload images – so add some fake placeholder (e.g. template's thumbnail) and replace it with the template once it's loaded? What do you think? Additionally I'd like to know the opinion of @ckeditor/ckeditor-4-devs. The other question here is, regardless of what we decide about inserting such a template, there will be need to implement it. @Fynn96, do you want to implement it in this PR or should I report a followup and we (or you, if you want to) will do it later, after finishing this PR?

plugins/templates/templates/default.js
Fynn96: > It's probably better to leave default templates as they were earlier. Especially because there is a difference in the behaviour: old templates are inserted synchronously and the new ones – asynchronously.

Okay. So should I come up with one or two new templates that are stored as a htmlFile? Something like title and text or whatever. As a simple example how to use htmlFile.

Fynn96:
> * It'd be good to have additional automatic tests, e.g. for inserting template with `htmlFile` field. Such a test can mock `CKEDITOR.ajax` to return some HTML and check if the editor's content is altered in a correct way. The test for inserting template with `html` field will be also nice. Could I ask you to add them?

I tried but I have trouble with the click event. I saw many tests that so `dialog.getButton( 'ok' ).click();`. But in my case die dialog has no ok button and I don't simulate a cklick on one of the templates. Do you have a hint for me?

> * The fact that inserting templates from file is asynchronous makes the whole thing a little bit tricky – as someone can start editing content before template is inserted. I'm not sure how to solve it. Maybe similar to how we deal with it in plugins that upload images – so add some fake placeholder (e.g. template's thumbnail) and replace it with the template once it's loaded? What do you think? Additionally I'd like to know the opinion of @ckeditor/ckeditor-4-devs. The other question here is, regardless of what we decide about inserting such a template, there will be need to implement it. @Fynn96, do you want to implement it in this PR or should I report a followup and we (or you, if you want to) will do it later, after finishing this PR?

`CKEDITOR.ajax` is able do load the data synchronously but it will cause Warnings in the Browser Developer Console so I used the asychronous way. But you are right this might cause some problems. I really like the idea of a placeholder while loading the template. A thumbnail that was already used for the template list in the dialog would be a good idea. 

I think it would be better to move the placeholder part to a new pr as a followup and I would try to do this little task by myself as soon as you decided how to solve this issue.

tests/plugins/templates/_assets/test.js
Fynn96: Can't find the path with ./images ir ./html so i remove the double slash.

plugins/templates/templates/default.js
Fynn96: AS a suggestion I know added two new template files. Is this what you meant?

jacekbogdanski:
@Comandeer from what I see, currently dialog will be closed after the template is fetched by ajax call:
https://github.com/ckeditor/ckeditor4/blob/d217471b3546cdafa49549f1a9e3d95ecbfa48bf/plugins/templates/dialogs/templates.js#L56-L58
https://github.com/ckeditor/ckeditor4/blob/d217471b3546cdafa49549f1a9e3d95ecbfa48bf/plugins/templates/dialogs/templates.js#L69-L91
which means that there is no option to edit editor content during loading external template file, am I right?

https://user-images.githubusercontent.com/10279586/133568306-2017023f-44b1-421a-b7de-38156b9ec6d7.mov

It may be not the best UX (e.g. upload image one has much better) but assuming that the operation of ajax call is quick enough, it seems like there is no need for that improvement for templates plugin right now? We can extract the feature request to make templates plugin asynchronous like it's done for upload images and d&d eventually, but I'm not really sure if it's needed at the moment.

plugins/templates/templates/default.js
Comandeer: Actually I just thought about not touching this file at all. Tests should be enough to check if the new feature works.

Comandeer:
@Fynn96, 

> I tried but I have trouble with the click event. I saw many tests that do `dialog.getButton( 'ok' ).click();`. But in my case the dialog has no ok button and I don't know how to simulate a click on one of the templates. Do you have a hint for me?

There is [`dialog#getContentElement()`](https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_dialog.html#method-getContentElement) method that will allow you to get to the whole list of templates and then you can get to links with templates. The whole thing is convoluted, but clicking the first template should look like this:

```javascript
dialog.getContentElement( 'selectTpl', 'templatesList' ).getElement().findOne( 'a' ).$.click();
```

> I think it would be better to move the placeholder part to a new pr as a followup and I would try to do this little task by myself as soon as you decided how to solve this issue.

Super! However, as @jacekbogdanski mentioned, dialog is closed after inserting template. That reminded me that we have [`dialog#setState()`](https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_dialog.html#method-setState) method which displays a nice spinner for dialogs that do something (it can be seen in ["Embed" dialog](https://ckeditor.com/docs/ckeditor4/latest/examples/mediaembed.html)). So the whole issue with UX could be fixed by adding `setState` calls before and after the Ajax request:

```javascript
item.on( 'click', function() {
	if( template.htmlFile ) {
		var templatesDialog = CKEDITOR.dialog.getCurrent();

		templatesDialog.setState( CKEDITOR.DIALOG_STATE_BUSY );

		// Load HTML of the file before inserting it.
		CKEDITOR.ajax.loadText( template.htmlFile, function( html ) {
			insertTemplate( html );
			templatesDialog.setState( CKEDITOR.DIALOG_STATE_IDLE );
		} );
	} else {
		insertTemplate( template.html );
	}
} );
```

However, `dialog#setState()` incorrectly assumes that every dialog has an "Ok" button and fails if it's absent. I've created an issue for it (#4888) and should prepare the fix today. Hopefully, it will be merged this week. After merging it to `major`, you could rebase your branch and add the integration mentioned above and the UX issue will be solved.

To sum up:

*   @Fynn96, for now you can work on tests,
*   I'll prepare the fix for #4888,
*   after the fix lands on `major`, the integration will need to be added to `templates` plugin.

plugins/templates/templates/default.js
Fynn96: I thought default.js serves as an example file for developers to see how you can use this template plugin and insert new templates. Wouldn't it be better to have at least one example on how to use the htmlFile feature? I don't think the people who use this plugin would have a look at the tests to understand the feature.

Fynn96:
I have some questions at this point: 

1. I need two timeouts for my test: One to wait for the templatelist to load as it is not immedeately there when I open the dialog and one for the assertion because the inserted template will not be there immedeately, too. So my code looks like this:
```js
'test insert template complete': function() {
            bender.editorBot.create( {
                startupData: '<p>Lorem ipsum</p>',
                config: {
                    templates_files: ['/tests/plugins/templates/_assets/test.js'],
                    templates: 'test'
                }
            }, function( bot ) {
                bot.dialog( 'templates', function( dialog ) {

                    setTimeout( function() {
                        resume( function() {
                            dialog.getContentElement( 'selectTpl', 'templatesList' ).getElement().findOne( 'a' ).$.click();
                        } );
                    }, 100 );
                    wait();

                    setTimeout( function() {
                        resume( function() {
                            assert.areEqual( '<p>^Lorem ipsum</p>', bot.htmlWithSelection(), 'Editor data has been altered.' );
                        } );
                    }, 100 );
                    wait();

                } );
            } );
        },
```
Now I get the error `Test has no asserts.`. How can I solve this issue?
2. I have trouble creating multiple tests in my test file. To prevent the error of point 1 I inserted a simple assertion before the timeouts. It simply checked if the editor text is the same as before. The test alone works fine but as soon as I add another test block I get the error `Timeout: wait() called but resume() never called.`. How can I prevent it?
```js
bender.test( {  
        'test aborting template insertion': function() {
            bender.editorBot.create( {
                startupData: '<p>Lorem ipsum</p>',
                config: {
                    templates_files: ['/tests/plugins/templates/_assets/test.js'],
                    templates: 'test'
                }
            }, function( bot ) {
                bot.dialog( 'templates', function( dialog ) {

                    dialog.getButton('cancel').click();
                    assert.areEqual( '<p>^Lorem ipsum</p>', bot.htmlWithSelection(), 'Editor data has not been altered.' );

                } );
            } );
        },
        'test insert template complete': function() {
            bender.editorBot.create( {
                startupData: '<p>Lorem ipsum</p>',
                config: {
                    templates_files: ['/tests/plugins/templates/_assets/test.js'],
                    templates: 'test'
                }
            }, function( bot ) {
                bot.dialog( 'templates', function( dialog ) {

                    assert.areEqual( '<p>^Lorem ipsum</p>', bot.htmlWithSelection(), 'Editor data has been altered.' );
                    setTimeout( function() {
                        resume( function() {
                            dialog.getContentElement( 'selectTpl', 'templatesList' ).getElement().findOne( 'a' ).$.click();
                        } );
                    }, 100 );
                    wait();

                    setTimeout( function() {
                        resume( function() {
                            assert.areEqual( '<h1>^I am a title</h1><p>I am a text</p>', bot.htmlWithSelection(), 'Template has been inserted.' );
                        } );
                    }, 100 );
                    wait();

                } );
            } );
        }
    } );
```
3. As we talked about the ok button in the dialog I think there is some issue about the replace ckeckbox that concerns the usability. The reading direction is from top to bottom so normally a user would select a template before checking the checkbox. This would be no problem if there would be a ok button so the user could select a template, then select or deselect the checkbox and then press ok. However it is also better to prevent unnecessary clicks. If we want to still prevent the necessity of clickling ok we should rearange the dialog. The ckeckbox should be above the template list to be in the right order of the reading direction. What do you think about that? We could make a new issue about this if needed. At the current state I often saw users accidently click a template before even reading the checkbox and then replace the whole editor content even if they didn't want to because the checkbox is selected by default.

plugins/templates/templates/default.js
Comandeer: In my opinion, such an example should be added to docs, not to the code. I'd leave the file with default templates untouched as some users can be using them without any modifications in their projects and suddenly the list of available templates would become different after updating the editor.

Comandeer:
1. You will need to nest these `setTimeouts` and use `resume` only for the deepest one:

   ```javascript
   setTimeout( function() {
     	dialog.getContentElement( 'selectTpl', 'templatesList' ).getElement().findOne( 'a' ).$.click();
   
     	setTimeout( function() {
       	resume( function() {
         	assert.areEqual( '<p>^Lorem ipsum</p>', bot.htmlWithSelection(), 'Editor data has been altered.' );
       	} );
       }, 100 );
   }, 100 );
   
   wait();
   ```

2. See answer for 1. Nesting should do the trick.

3. Yes, it sounds reasonable to remove newlines from the end of templates.

4. No, unit tests are for machines and manual tests are for people. It's always better to check it personally if it really works.

5. Yes, please report a new issue for it as it's a completely different problem and should be solved separately.

Fynn96:
The nested timeout solves the assertion problem but still multiple tests are not possible at the moment. I pushed my tests so you can see my file. Maybe you can see were I did a mistake. When I comment out all tests except of one this will work (for every test). But all test together will cause `Timeout: wait() called but resume() never called.`. I don't think this has something to do with my timeouts because if I add the first test again, which has no timeouts, I get the same error even thought I don't even call `wait()`. This is what I tested: 

```js
bender.test( {  
        'test aborting template insertion': function() {
            bender.editorBot.create( {
                startupData: '<p>Lorem ipsum</p>',
                config: {
                    templates_files: ['/tests/plugins/templates/_assets/test.js'],
                    templates: 'test'
                }
            }, function( bot ) {
                bot.dialog( 'templates', function( dialog ) {

                    dialog.getButton('cancel').click();
                    assert.areEqual( '<p>^Lorem ipsum</p>', bot.htmlWithSelection(), 'Editor data has not been altered.' );

                } );
            } );
        },
        'test aborting template insertion 2': function() {
            bender.editorBot.create( {
                startupData: '<p>Lorem ipsum</p>',
                config: {
                    templates_files: ['/tests/plugins/templates/_assets/test.js'],
                    templates: 'test'
                }
            }, function( bot ) {
                bot.dialog( 'templates', function( dialog ) {

                    dialog.getButton('cancel').click();
                    assert.areEqual( '<p>^Lorem ipsum</p>', bot.htmlWithSelection(), 'Editor data has not been altered.' );

                } );
            } );
        }
} );
```

Comandeer:
I've updated the tests. The main issue was the fact that you didn't name editors created by `bender.editorBot.create()` and [tests failed due to conflicts in editors' naming](https://ckeditor.com/docs/ckeditor4/latest/guide/dev_errors.html#editor-element-conflict). I've also made adjustments to make tests compliant with our style guide and extracted getting templates from dialog to a separate function. Now everything seems to work fine.

Yesterday we merged the fix for #4888 to `major`. @Fynn96, can I ask you to integrate it with your changes, as was described in https://github.com/ckeditor/ckeditor4/pull/4851#issuecomment-920774692?

Fynn96:
Perfect, thank you! I will integrate it now.

Comandeer (approved):
LGTM! I've split manual test into two – one for replacing content with templates and one for inserting templates alongside content. Additionally I've added unit and manual tests for checking if the dialog's state really changes.

There was also some issue with the rebase (commits were duplicated). I've rebased the PR once more and that seems to fix the issue.

Thanks for your PR!

PR #4867, opened 2021-09-03
Flash plugin removal

### LINKED ISSUE #4866: Remove flash plugin
## Type of report

Task

Due to EOL ( 12/31/2020 ) and various vulnerabilities we should remove flash plugin. 

Steps:
* Removing flash plugin from presets and repository.
* Adjust documentation. It will be still accessible for older setups, but we need to point out that the plugin is removed from presets and is no longer supported.
* Leave documentation for options with `deprecated` note.

KarolDawidziuk:

Task

### This PR contains

- [x] Unit tests
- [x] Manual tests

```
* [#4866](https://github.com/ckeditor/ckeditor4/issues/#4866): Flash plugin removal.
```

## What changes did you make?

Due to EOL flash. I've removed it from the editor.

Changes:
* I only leave part of documentation in the flash plugin with a deprecated note.
* All tests related to the flash are removed.
* All other things like: lang files, configuration have been updated.

Additionally, I ran a testing phase on modern browsers and IE( IE8 and IE11 ), and there is no additional failing test compared to the current `major` branch.

Monkey clicking also did not reveal any additional errors.

Closes #4866.


sculpt0r (changes requested):
Hi @KarolDawidziuk 
Nice that you have found a lot of tricky places with flash plugin mentioned.

This part of the job looks good, however I found a few places more:
- https://github.com/ckeditor/ckeditor4/blob/70649e5cd8b1afedba2aeb18017aed4e6227c911/plugins/dialog/plugin.js#L3597
-  language files from `fakeobjects` plugin still contain flash key
- common language files also contain a few mentions for flash
- https://github.com/ckeditor/ckeditor4/blob/70649e5cd8b1afedba2aeb18017aed4e6227c911/plugins/filebrowser/plugin.js#L561 & https://github.com/ckeditor/ckeditor4/blob/70649e5cd8b1afedba2aeb18017aed4e6227c911/plugins/filebrowser/plugin.js#L527 - those also should be marked as deprecated
- https://github.com/ckeditor/ckeditor4/blob/major/plugins/htmlwriter/samples/outputforflash.html - this one also should be handled as well as files related to it.
- small mention here to remove: https://github.com/ckeditor/ckeditor4/blob/70649e5cd8b1afedba2aeb18017aed4e6227c911/plugins/widget/plugin.js#L4748
- some leftover to handle: https://github.com/ckeditor/ckeditor4/blob/70649e5cd8b1afedba2aeb18017aed4e6227c911/tests/core/htmldataprocessor/htmldataprocessor.html#L40

Also, I don't see any PR for documentation that looks to be needed (e.g. here: https://ckeditor.com/docs/ckeditor4/latest/features/mathtype.html#input-and-output)

KarolDawidziuk:
Thanks, @sculpt0r for the review and for missing places!

It's ready for another one.
Changes: 
* removed `flash` mentions from lang files
* updated docs
* removed unnecessary `flash` related files
* created 2 another PR in [docs](https://github.com/ckeditor/ckeditor4-docs/pull/370) and [docs-samples](https://github.com/ckeditor/ckeditor4-docs-samples/pull/24)

vokiel:
There might be also some changes needed in the [CKEditor 4 addons repository](https://ckeditor.com/cke4/addon/flash) and the online builder.

tests/tickets/17001/pasteobject.md
```diff
@@ -1,6 +1,6 @@
 @bender-tags: bug, 4.7.0, trac17001
 @bender-ui: collapsed
-@bender-ckeditor-plugins: wysiwygarea, toolbar, undo, basicstyles, pastefromword, flash
+@bender-ckeditor-plugins: wysiwygarea, toolbar, undo, basicstyles, pastefromword
```
sculpt0r, line 3: Actually, the entire test looks like we could remove it, but maybe @jacekbogdanski will have a different opinion.

sculpt0r (approved):
Ok, it looks good to me now. Just one small inline comment, but this one should be resolved with @jacekbogdanski . 

So approved from me 🎉 👍

tests/tickets/17001/pasteobject.md
jacekbogdanski: We still have a code which is checked by this manual test, see https://github.com/ckeditor/ckeditor4/commit/77664e89f
I think that we should rather change this test a bit instead of removing it:

1. Add `allowedContent: true` to the editor.
2. Add `sourcearea` and `htmlwriter` plugins.
3. Change test description to verify if there is `object` element in the source.

@KarolDawidziuk could you update the test?

skins/moono-lisa/dialog.css
```diff
@@ -1121,7 +1110,7 @@ a.cke_specialchar:active
 }
 
 .cke_dialog_image_url .cke_dialog_ui_hbox_last,
-.cke_dialog_flash_url .cke_dialog_ui_hbox_last
+.cke_dialog_ui_hbox_last
 {
 	vertical-align: bottom;
 }
```
jacekbogdanski, line 1116: Previously this CSS rule was changing styling of `.cke_dialog_ui_hbox_last` as a child of `.cke_dialog_flash_url` but right now it will be used for every element outside flash? If so, the whole rule should be rather removed.

jacekbogdanski (changes requested):
We need to cover some additional cases to finish this PR:
1. Add a manual test and unit test confirming what will happen if someone is still using `flash` plugin (including `toolbar` config option). Such situation shouldn't break the editor, but rather gently remind that the flash plugin has been deprecated. From what I see now, it's just throwing console errors without information why it happened: 
image
2. Handle such cases as above, so create a dummy flash plugin definition including an error that the flash plugin has been deprecated with [Error Code Reference](https://ckeditor.com/docs/ckeditor4/latest/guide/dev_errors.html) entry to our documentation. Something like:
```js
CKEDITOR.plugins.add( 'flash', {
	init: function() {
		CKEDITOR.error( 'editor-plugin-deprecated', { pluginName: 'flash' } );
	}
} );
```
We will need test coverage that the error has been thrown also. You can find similar examples in our codebase, e.g. here https://github.com/ckeditor/ckeditor4/blob/b32afc7a722b517d1e45c7e5caa9a374faf3801b/tests/core/creators/inline-textarea.js#L149-L161
3. Remove (probably the last one) occurrence of flash plugin https://github.com/ckeditor/ckeditor4/blob/b32afc7a722b517d1e45c7e5caa9a374faf3801b/plugins/filebrowser/plugin.js#L465-L478

Other than that, I've checked documentation builder and it seems like all entries are nicely deprecated 👍🏻 
image

jacekbogdanski:
Creating a dummy flash plugin is important because some of the editor users may already use custom editor configurations based on some wide preset and just removing specific plugins from the toolbar. It's not recommended approach as it's just better to remove some plugin entirely from the builder but may happen and we don't want to break these configurations with the new release.

KarolDawidziuk:
Blocked by [#21](https://github.com/ckeditor/ckeditor4-benderjs-runner/pull/21).

KarolDawidziuk:
After fixing https://github.com/ckeditor/ckeditor4-benderjs-runner/issues/20 this PR is no longer blocked.

tests/plugins/flash/manual/error.md
```diff
@@ -0,0 +1,16 @@
+@bender-tags: 4.17.0, feature, 4866
+@bender-ui: collapsed
+@bender-ckeditor-plugins: wysiwygarea, toolbar, flash
```
jacekbogdanski, line 3: It would be good to add some minimal editor configuration like basicstyles, link, and undo, so the test doesn't look broken (without toolbar buttons).

jacekbogdanski (changes requested):
@KarolDawidziuk could you address the issue reported in https://github.com/ckeditor/ckeditor4/pull/4867/files#r705033198 ?

KarolDawidziuk:
Thanks, @jacekbogdanski  for the review.

I've removed test mentioned in  #4867 (comment) because this test relates to the flash plugin and it's unnecessary. 
To cover the uncovered piece of code with the filter `object` elements  I've created two more unit tests.

jacekbogdanski (approved):
Good job @KarolDawidziuk and @sculpt0r 🎉

PR #4872, opened 2021-09-08
Fix htmldataprocessor unit test

### LINKED ISSUE #4822: Failing test: tests/core/htmldataprocessor/htmldataprocessor
Failing test: `test protected source in iframe`

```
Editor data does not match.
Expected: <!doctype html><html><head><title>foo</title></head><body><p><iframe name="aa">[[mytag]]</iframe></p></body></html> (string)
Actual:   <!doctype html><html><head><title>foo</title></head><body><p><iframe name="aa"></iframe></p></body></html> (string)  
```

## Other details

* Browser: Chrome 92.0.4515.131
* OS: Android 9

jacekbogdanski:

Failing test fix

- [ ] PR is consistent with the code style guide

```
skip
```

## What changes did you make?

We disabled any HTML content in iframe elements some time ago.

Closes #4822

### DIFF
--- a/tests/core/htmldataprocessor/htmldataprocessor.js
+++ b/tests/core/htmldataprocessor/htmldataprocessor.js
@@ -612,7 +612,7 @@
 
 		// Some elements should not have protected source markup inside. (https://dev.ckeditor.com/ticket/11223)
 		'test protected source in title': addProtectedSourceTC( '<p>[[mytag]]</p>', '[[mytag]]' ),
-		'test protected source in iframe': addProtectedSourceTC( '<p><iframe name="aa">[[mytag]]</iframe></p>' ),
+		'test protected source in iframe': addProtectedSourceTC( '<p><iframe name="aa"></iframe></p>' ),
 		'test protected source in textarea': addProtectedSourceTC( '<p><textarea name="aa">[[mytag]]</textarea></p>' ),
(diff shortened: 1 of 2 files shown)


Comandeer (changes requested):
Although the test passes, it does no longer test protected source in `iframe`. It just tests if editor can contain empty `iframe`. Additionally test failed only on built version, which means that some of plugins changes the behaviour of data processor – maybe we should move this behaviour directly to data processor and adjust tests accordingly?

jacekbogdanski:
As we discussed some time ago, the iframe plugin is supposed to remove all its content and its expected behavior. Most browsers are not supporting any content inside iframe and allowing it may create a security risk. Instead of changing the current test, I've removed `iframe` plugin from `htmldataprocessor` test and added test coverage to `iframe` plugin itself. Note that I've had to make some adjustments to the test suite itself, as it was working in some cases accidentally (by reusing previous tests HTML for editing dialogs).

Comandeer (changes requested):
Changes in `/tests/core/htmldataprocessor/htmldataprocessor` seems fine. However, the tests in `/tests/plugins/iframe/iframe` does not work in IE 8-10 after the changes.

Comandeer (approved):
LGTM!

PR #4896, opened 2021-09-20
Allow built-in image paste handling to be disabled (#4874).

### LINKED ISSUE #4874: Clipboard plugin's handling of images should be configurable - will break other plugins by default
## Type of report

Bug

In june 2021 the built-in clipboard plugin was extended to offer native support for pasting images as base64.
But since there is already multiple plugins out there offering similar functionality (some with additional features), pasting an image will now result in two images being inserted.

Code causing the problem starts here:
https://github.com/ckeditor/ckeditor4/blob/70649e5cd8b1afedba2aeb18017aed4e6227c911/plugins/clipboard/plugin.js#L149

While native support for pasting images is wonderful, the current implementation seems to break support for custom image paste plugins as there is no way to prevent the clipboard event from inserting the image too, without stopping the event chain. At least not exposed as public API. One can hack it using `CKEDITOR.plugins.clipboard.isCustomDataTypesSupported = false;`

Furthermore, base64 images can be huge and it may severely hurt performance, especially if using the _change_ event to continuously validate data or push changes to application state - although this use case is discouraged:
https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_editor.html#event-change
> If it is important not to get the change event fired too often, you should compare the previous and the current editor content inside the event listener. It is not recommended to do that on every change event.

In addition, the clipboard event even stops the event chain once the image has been processed, preventing other plugins registered later from receiving the image data.
https://github.com/ckeditor/ckeditor4/blob/70649e5cd8b1afedba2aeb18017aed4e6227c911/plugins/clipboard/plugin.js#L199

Why would you even use a custom image plugin to handle image pasting?
To automatically upload images immediately. To temporarily store images in blob storage rather than using base64, which greatly increases performance when working with large documents (we use this). To optimize/modify images before inserting them. To support pasting of multiple images at once (we use this). There could be many reasons.

A small sample has been put together which lets the user switch between CKEditor 4.16.1 and the current Major, to test image pasting in both of them.

image

Download: https://codemagic.dk/CKEditorDemo.zip

1. Open _CKEditor demo/index.html_ (can run locally)
2. Make sure _CKEditor version_ is set to "Old CKEditor"
3. Paste an image into the editor - everything is fine, we get exatly one image.

Now test with the current major.

1. Open _CKEditor demo/index.html_ (can run locally)
2. Make sure _CKEditor version_ is set to "New CKEditor (improved)"
3. Paste an image into the editor - BUG: image is inserted twice !

### Expected result

Clipboard plugin should allow for other plugins to handle image pasting

### Actual result

Clipboard plugin causes duplicate images

## Other details

* Browser: Chrome 93.0.4577.63 (Official Build) (x86_64)
* OS: macOS 10.13.6 (17G14042)
* CKEditor version: 4.16.1 and 4.17.x
* Installed CKEditor plugins: clipboard, misc.

FlowIT-JIT:
This will allow 3rd party plugins to handle images being pasted.
Related issue: https://github.com/ckeditor/ckeditor4/issues/4874

Bug fix

### This PR contains

- [x] Unit tests
- [X] Manual tests

```
* [#4874](https://github.com/ckeditor/ckeditor4/issues/4874): Allow 3rd party plugins to handle images being pasted.
```

## What changes did you make?

Added a new configuration option to _clipboard_ plugin which allows for the built-in image paste handling to be disabled, so that other plugins can intercept clipboard data and handle images being pasted.

```
CKEDITOR.replace( 'editor', {
	// Allow 3rd party plugins to handle image pasting
	clipboard_handleImagePasting: false,
	extraPlugins: "customImagePasteHandlerPlugin"
} );
```

Closes #4874.


plugins/clipboard/plugin.js
```diff
@@ -149,7 +149,7 @@
 			// Convert image file (if present) to base64 string for modern browsers except IE<10, as it does not support
 			// custom MIME types in clipboard (#4612).
 			// Do it as the first step as the conversion is asynchronous and should hold all further paste processing.
-			if ( CKEDITOR.plugins.clipboard.isCustomDataTypesSupported || CKEDITOR.plugins.clipboard.isFileApiSupported ) {
+			if ( ( CKEDITOR.plugins.clipboard.isCustomDataTypesSupported || CKEDITOR.plugins.clipboard.isFileApiSupported ) && editor.config.clipboard_handleImagePasting !== false ) {
```
Comandeer, line 152: This condition became too long and therefore unreadable. Could you extract its first part to some variable (e.g. `isImagePastedSupported`)?

Additionally `editor.config.clipboard_handleImagePasting` alone should be enough to check if the plugin is supposed to handle image pasting.

tests/plugins/clipboard/manual/pasteimagedisable.md
```diff
@@ -0,0 +1,17 @@
+@bender-ui: collapsed
+@bender-tags: 4.17.0, clipboard, 4874
+@bender-ckeditor-plugins: wysiwygarea, toolbar, undo, basicstyles, image, clipboard, sourcearea
+
+ 1. Paste an image into the editor.
+
+    Observe that the image is inserted and that no "Custom image paste handling" entry is written to "log output" on the page.
```
Comandeer, line 7: I'd split it into two tests: the one that will test the plugin with handling image pasting and one without it. It should be more readable than putting both of these checks inside one test. Could you split the test?

tests/plugins/clipboard/manual/pasteimagedisable.md
```diff
@@ -0,0 +1,17 @@
+@bender-ui: collapsed
+@bender-tags: 4.17.0, clipboard, 4874
```
Comandeer, line 2: Please, annotate that this test is for a bug:

```suggestion
@bender-tags: 4.17.0, clipboard, 4874, bug
```

Comandeer (changes requested):
Hi, @FlowIT-JIT, and thanks for your contribution! There are however some issues – please see my inline comments for more info.

Additionally, it will be good to add some unit tests – to check if setting this option to `false` disables built-in handling of pasting images and if the `paste` event is no longer stopped. Could you add such tests?

plugins/clipboard/plugin.js
FlowIT-JIT: I just committet a change that shortens the code by extracting the first condition into an `isImagePasteSupported` variable as suggested. I'm not too crazy about the idea having only clipboard_handleImagePasting determine whether to enable image handling, since it will allow a developer to enable the feature on unsupported browsers.

tests/plugins/clipboard/manual/pasteimagedisable.md
FlowIT-JIT: I added unit tests to cover scenarios where image paste handling has been disabled using the new `clipboard_handleImagePasting` flag - see tests/plugins/clipboard/pasteimagedisabled.
Native image paste handling (when enabled) is already covered by tests/plugins/clipboard/pasteimage.

Do you still want manual test coverage, or should I just remove it again ?

FlowIT-JIT: Oh, worth mentioning; I took the liberty to move some code from tests/plugins/clipboard/pasteimage.js into tests/plugins/clipboard/_helpers/pasting.js to reduce duplicate code. I needed the exact same code for the new unit test in tests/plugins/clipboard/pasteimagedisabled.js

FlowIT-JIT:
@Comandeer Thank you for reviewing my PR. I have made some adjustments based on your feedback.

plugins/clipboard/plugin.js
Comandeer: >  I'm not too crazy about the idea having only clipboard_handleImagePasting determine whether to enable image handling, since it will allow a developer to enable the feature on unsupported browsers.

Actually, I meant something different, but I see that my message could be ambiguous. I thought about shortening `editor.config.clipboard_handleImagePasting !== false` to `editor.config.clipboard_handleImagePasting`.

tests/plugins/clipboard/manual/pasteimagedisable.md
Comandeer: > Do you still want manual test coverage, or should I just remove it again ?

It's always better to have an additional manual test. For me manual tests are for people and unit ones – for machines.

FlowIT-JIT: Alright, I'll refacter the manual test as you suggested. Thanks 😊

plugins/clipboard/plugin.js
FlowIT-JIT: Oh, okay 😅😊 Got it!

tests/plugins/clipboard/manual/pasteimagedisable.md
FlowIT-JIT: There. The new manual test (tests/plugins/clipboard/manual/pasteimagedisable.md) is now only about testing the new `clipboard_handleImagePasting` option, as support for image pasting is already covered with tests/plugins/clipboard/manual/pasteimage.md.

tests/plugins/clipboard/manual/pasteimagedisable.html
```diff
@@ -0,0 +1,32 @@
...
+	<p></p>
+</div>
+
+<br><br>
+
+<div id="log">Log output:</div>
+
+<script>
+	( function() {
+		if ( !window.FileReader ) {
```
Comandeer, line 12: This test should be ignored in Internet Explorer 10+ as these browsers support File API, but do not support custom MIME types during pasting (only during drag and drop). Probably the most sensible way to ignore this test there would be to do some browser sniffing. Additionally, this test should be ignored on mobile as pasting images is not possible in the same way as on desktops.

So the check should look like this:

```suggestion
		if ( CKEDITOR.env.ie || bender.tools.env.mobile ) {
```

Could you update it?

plugins/clipboard/plugin.js
```diff
@@ -3471,3 +3472,13 @@
  * @member CKEDITOR.config
  */
 CKEDITOR.config.clipboard_notificationDuration = 10000;
+
+/**
+ * Whether to use clipboard plugin to handle image pasting, turning
```
Comandeer, line 3477:
```suggestion
 * Whether to use clipboard plugin to handle image pasting and dropping, turning
```

Comandeer (changes requested):
Thx for the updates!

Your changes work correctly, but it seems that some additional tests will be needed, as the new config variable also disables dragging and dropping images. Could you add a manual test and unit test for this case? You can probably reuse some code from `dropimage.js` test. I'm sorry for not noticing it earlier.

It'd be also good to add a unit test for the original case, so checking if some custom paste handler can receive `paste` data after disabling image handling. Could you also add such a test?

FlowIT-JIT:
> Thx for the updates!
> 
> Your changes work correctly, but it seems that some additional tests will be needed, as the new config variable also disables dragging and dropping images. Could you add a manual test and unit test for this case? You can probably reuse some code from `dropimage.js` test. I'm sorry for not noticing it earlier.
> 
> It'd be also good to add a unit test for the original case, so checking if some custom paste handler can receive `paste` data after disabling image handling. Could you also add such a test?

Thank you for additional feedback. I should probably have noticed drag and drop support myself - sorry about that. I will look into extending test coverage as suggested.

FlowIT-JIT:
I have now added a unit test and manual test to cover image drag and drop.
In addition I extended the unit tests to verify that a custom image handler plugin receives the image when pasting or dropping. I combined this check with the existing unit tests to reduce duplicate code.

Furthermore I renamed the new option from `clipboard_handleImagePasting` to `clipboard_handleImages` as it is not restricted to paste operations alone. In conjunction with this change I also renamed the unit test file from `pasteimagedisabled.js` to `clipboardimagedisabled.js` which execute tests for both paste and drag'n'drop.

Please note that I once again had to do a little refactoring to reduce duplicate code for the new unit test covering image dropping. I moved code from `tests/plugins/clipboard/dropimage.js` to `tests/plugins/clipboard/_helpers/pasting.js`, but this time had to do some minor changes to tear down logic, as outline below:

image

Prior to this change the cleanup was - probably not entirely logical nor correct - performed in setUp(..) as shown below:

image

Comandeer (approved):
LGTM! I've just updated tests a bit:

* drop test should run in IE10+, so I updated check accordingly,
* we use global `assert` in our tests – I've updated asserts accordingly in your tests,
* I've renamed `assertDropImage` to `assertImageDrop` to achieve better coherency with `assertImagePaste`.

Thanks for the PR!

FlowIT-JIT:
@Comandeer Perfect! Thank you 😊

PR #4910, opened 2021-09-27
Fix issue with double escaped autolink special characters

### LINKED ISSUE #4858: Pasting a link with & will is changed to &amp; which can breaks link
## Type of report

Bug

1. Visit CKEditor 4 demo article 
2. Copy the following link: https://www.google.com/search?q=ckeditor+4+demo&sxsrf=AOaemvIOMlAFDqLmu7CedntFOZM__buC4Q%3A1629916057505&source=hp&ei=mYsmYYbvG5H70AST5RQ&iflsig=ALs-wAMAAAAAYSaZqel7dIi7ht5uNwnupRKhKbR7VdtE&oq=ckeditor+4+demo&gs_lcp=Cgdnd3Mtd2l6EAMyBAgjECcyBQgAEIAEMgYIABAWEB4yBggAEBYQHjIGCAAQFhAeMgYIABAWEB46BwgjEOoCECc6BQgAEJECOgQIABBDOg4ILhCABBCxAxDHARCjAjoRCC4QgAQQsQMQgwEQxwEQ0QM6CwguEIAEELEDEIMBOgoIABCxAxCDARBDOggIABCABBCxAzoLCC4QgAQQxwEQowI6CAgAELEDEIMBOgsIABCABBCxAxCDAToICC4QsQMQgwE6CAguEIAEELEDOgsIABCABBCxAxDJAzoFCAAQkgM6BwguELEDEEM6DgguEIAEELEDEIMBEJMCOgUILhCABDoFCC4QsQM6DgguEIAEELEDEMcBEK8BOgoIABCABBCHAhAUOgsILhCABBDHARDRAzoHCAAQsQMQQ1CFFVirOWDYOmgCcAB4AIABe4gB9hCSAQQxNi44mAEAoAEBsAEK&sclient=gws-wiz&ved=0ahUKEwjGn_m85szyAhWRPZQKHZMyBQAQ4dUDCAo&uact=5
3. Paste into CKEditor

### Expected result

Recognized as a link with no changes to URL

### Actual result
URL changes to the following, where "&" is replaced with "& a m p ;" (but with no spaces, but preview of this text not rendering as expected)
https://www.google.com/search?q=ckeditor+4+demo&amp;sxsrf=AOaemvIOMlAFDqLmu7CedntFOZM__buC4Q%3A1629916057505&amp;source=hp&amp;ei=mYsmYYbvG5H70AST5RQ&amp;iflsig=ALs-wAMAAAAAYSaZqel7dIi7ht5uNwnupRKhKbR7VdtE&amp;oq=ckeditor+4+demo&amp;gs_lcp=Cgdnd3Mtd2l6EAMyBAgjECcyBQgAEIAEMgYIABAWEB4yBggAEBYQHjIGCAAQFhAeMgYIABAWEB46BwgjEOoCECc6BQgAEJECOgQIABBDOg4ILhCABBCxAxDHARCjAjoRCC4QgAQQsQMQgwEQxwEQ0QM6CwguEIAEELEDEIMBOgoIABCxAxCDARBDOggIABCABBCxAzoLCC4QgAQQxwEQowI6CAgAELEDEIMBOgsIABCABBCxAxCDAToICC4QsQMQgwE6CAguEIAEELEDOgsIABCABBCxAxDJAzoFCAAQkgM6BwguELEDEEM6DgguEIAEELEDEIMBEJMCOgUILhCABDoFCC4QsQM6DgguEIAEELEDEMcBEK8BOgoIABCABBCHAhAUOgsILhCABBDHARDRAzoHCAAQsQMQQ1CFFVirOWDYOmgCcAB4AIABe4gB9hCSAQQxNi44mAEAoAEBsAEK&amp;sclient=gws-wiz&amp;ved=0ahUKEwjGn_m85szyAhWRPZQKHZMyBQAQ4dUDCAo&amp;uact=5

## Other details

The above change to the URL does allow for it to be navigated to, but this is not the case for our customer scenario. In our case the URL will navigate to error page because the amp; being added is unexpected

* Browser: Chrome
* OS: Windows
* CKEditor version: CKEditor 4
* Installed CKEditor plugins: link

## EDIT

I have managed to reproduce this problem. When you copy actual link from the HTML page e.g. `https://www.google.com/search?q=google&client=firefox-b-d`  then it is pasted correctly -> You will have `&` in wysiwyg and `&amp;` in source mode.

If however you will paste the same link but you will copy it from e.g. Notepad, you will get `&amp;` in wysiwyg and `&amp;amp;` in source mode which in my opinion can break the link (the actual non-working link can be found in support issue no. 70206). 

obraz

obraz

jacekbogdanski:

Bug fix

### This PR contains

- [x] Unit tests
- [x] Manual tests

```
* [#4858](https://github.com/ckeditor/ckeditor4/issues/4858): Fixed: [Autolink](https://ckeditor.com/cke4/addon/autolink) plugin incorrectly escapes `&` characters when pasting links into the editor.
```

## What changes did you make?

Links were double escaped by executing 2x `setAttribute` method with already escaped `&` character to `&amp;`.

Closes #4858

### DIFF
--- a/tests/plugins/autolink/autolink.js
+++ b/tests/plugins/autolink/autolink.js
@@ -99,6 +99,14 @@
 			assertPasteEvent( this.editors.classic, { dataValue: pastedText }, { dataValue: pastedText, type: 'html' } );
 		},
 
+		// (#4858)
+		'test URL link with encoded characters': function() {
+			var pastedText = 'https://www.google.com/test/?one=one&amp;two=two&amp;three',
+				expected = '<a href="https://www.google.com/test/?one=one&amp;two=two&amp;three">https://www.google.com/test/?one=one&amp;two=two&amp;three</a>';
+
+			assertPasteEvent( this.editors.classic, { dataValue: pastedText }, { dataValue: expected, type: 'html' } );
+		},
+
 		'test mail link with text after': function() {
 			var pastedText = 'mail@example.com nope';
 
--- a/tests/plugins/autolink/autolink.js
+++ b/tests/plugins/autolink/autolink.js
@@ -102,9 +102,15 @@
 		// (#4858)
 		'test URL link with encoded characters': function() {
 			var pastedText = 'https://www.google.com/test/?one=one&amp;two=two&amp;three',
-				expected = '<a href="https://www.google.com/test/?one=one&amp;two=two&amp;three">https://www.google.com/test/?one=one&amp;two=two&amp;three</a>';
+				expected = '<a href="https://www.google.com/test/?one=one&amp;two=two&amp;three">https://www.google.com/test/?one=one&amp;two=two&amp;three</a>',
+				spy = sinon.spy( CKEDITOR.tools, 'htmlDecodeAttr' );
 
 			assertPasteEvent( this.editors.classic, { dataValue: pastedText }, { dataValue: expected, type: 'html' } );
+
+			spy.restore();
+
+			// Ensure that the test is correct.
+			assert.isTrue( spy.calledWithExactly( pastedText ), 'htmlDecodeAttr was called with incorrect input' );
 		},
 
 		'test mail link with text after': function() {
--- a/tests/plugins/autolink/manual/encoding.md
+++ b/tests/plugins/autolink/manual/encoding.md
@@ -1,13 +1,15 @@
 @bender-ui: collapsed
 @bender-tags: 4.17.0, bug, 4858
-@bender-ckeditor-plugins: clipboard, toolbar, wysiwygarea, elementspath, autolink
+@bender-ckeditor-plugins: clipboard, toolbar, wysiwygarea, elementspath, autolink, sourcearea, htmlwriter
 
-Paste the given URL to the editor and check whether it's turned into a link.
+Paste the given URL to the editor and check whether it's turned into a link. Make sure to also confirm links in source area.
 
 ## Expected
 
-Link is pasted as it is, without additional encoding.
+**Editor:** Link is pasted as it is, without additional encoding.
+**Source area:** `&` characters are single encoded with `&amp;`.
 
 ## Unexpected
 
-Special `&` characters are additionaly encoded to `&amp;`.
+**Editor:** Special `&` characters are additionaly encoded to `&amp;`.
+**Source area:** `&` characters are double encoded to `&amp;amp;`.
--- a/tests/plugins/autolink/manual/encoding.md
+++ b/tests/plugins/autolink/manual/encoding.md
@@ -7,9 +7,11 @@ Paste the given URL to the editor and check whether it's turned into a link. Mak
 ## Expected
 
 **Editor:** Link is pasted as it is, without additional encoding.
+
 **Source area:** `&` characters are single encoded with `&amp;`.
 
 ## Unexpected
 
 **Editor:** Special `&` characters are additionaly encoded to `&amp;`.
+
 **Source area:** `&` characters are double encoded to `&amp;amp;`.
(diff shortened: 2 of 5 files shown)


tests/plugins/autolink/manual/encoding.md
```diff
@@ -0,0 +1,15 @@
+@bender-ui: collapsed
+@bender-tags: 4.17.0, bug, 4858
+@bender-ckeditor-plugins: clipboard, toolbar, wysiwygarea, elementspath, autolink, sourcearea, htmlwriter
+
+Paste the given URL to the editor and check whether it's turned into a link. Make sure to also confirm links in source area.
+
+## Expected
+
+**Editor:** Link is pasted as it is, without additional encoding.
+**Source area:** `&` characters are single encoded with `&amp;`.
```
Comandeer, line 10:
```suggestion
**Editor:** Link is pasted as it is, without additional encoding.

**Source area:** `&` characters are single encoded with `&amp;`.
```

Without the empty line, these instructions are incorrectly formatted.

tests/plugins/autolink/manual/encoding.md
```diff
@@ -0,0 +1,15 @@
...
+
+## Unexpected
+
+**Editor:** Special `&` characters are additionaly encoded to `&amp;`.
+**Source area:** `&` characters are double encoded to `&amp;amp;`.
```
Comandeer, line 15:
```suggestion
**Editor:** Special `&` characters are additionaly encoded to `&amp;`.

**Source area:** `&` characters are double encoded to `&amp;amp;`.
```

Comandeer (changes requested):
Looks good, however there is one edge case that fails after the changes:

```
https://www.google.com/test/?one=one&amp;=1
```

If link contains intentional `&amp;` sequence, it's converted to `&amp;` instead of `&amp;amp;`. The same happens with `&lt;` and other entities mentioned in `allEscRegex` in `tools.js` file.

jacekbogdanski:
The problem with the issue you mentioned is that there is no way (at the moment) to differentiate both escaped and unescaped characters when manipulating HTML in `autolink` plugin. Feeding `autolink` plugin with:
* `https://www.google.com/test/?one=1&amp;two=2`
* `https://www.google.com/test/?one=1&two=2`
will always result in passing escaped HTML to https://github.com/ckeditor/ckeditor4/blob/9fcdc324f2fa1c1eb6297afbe1fa873ae4ad2873/plugins/autolink/plugin.js#L76-L104

Both clipboard and typing operate on already processed HTML by a browser using copybin or text matcher.

This issue also exists in the older `4.15.1` version before #4858 regression, so I propose to extract it as a separate ticket as it looks like something more complicated to fix but also very unlikely to happen.

tests/plugins/autolink/autolink.js
```diff
@@ -99,6 +99,14 @@
 			assertPasteEvent( this.editors.classic, { dataValue: pastedText }, { dataValue: pastedText, type: 'html' } );
 		},
 
+		// (#4858)
+		'test URL link with encoded characters': function() {
+			var pastedText = 'https://www.google.com/test/?one=one&amp;two=two&amp;three',
```
Comandeer, line 104: Shouldn't it contain `&`? Probably this test passes due to #4912.

Comandeer (changes requested):
LGTM! I've got just some doubts about unit test (see my inline comment).

tests/plugins/autolink/autolink.js
jacekbogdanski: No, `assertPasteEvent` is not escaping characters by itself as it skips some parts of our data processing, so we have to feed it with correct input. Anyway, I've added an additional check confirming that the test is valid.

Comandeer (approved):
LGTM!

PR #4940, opened 2021-10-24
Added default value to useComputedState in the CKEDITOR config object

### LINKED ISSUE #4918: Default value for useComputedState is undefined but treated as `true` value.
## Type of report

Bug

Documentation says that default value for [useComputedState](https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_config.html#cfg-useComputedState) is `true`. However, there is no place in the code that assigns `true` to this config option.

This option is used in `justify` plugin, which interprets the `undefined` value as `true`: https://github.com/ckeditor/ckeditor4/blob/728e1923987dbf1ad9804a35e4b7945788025535/plugins/justify/plugin.js#L12

This option is used in `bidi` plugin, which interprets the `undefined` value as `true`: 
https://github.com/ckeditor/ckeditor4/blob/728e1923987dbf1ad9804a35e4b7945788025535/plugins/bidi/plugin.js#L18

This option is used in `list` plugin, which interprets the `undefined` value as `true`: 
https://github.com/ckeditor/ckeditor4/blob/728e1923987dbf1ad9804a35e4b7945788025535/plugins/list/plugin.js#L367

Please note, that there are few other lines in those plugins, that use this 'trick'.

Code logic makes the value of `config.useComputedState` interpreted the same as the documentation says: as `true` value. However, If the user tries to use `config.useComputedState`, he will get `undefined` instead of `true`.

### Expected result

Plain `config.useComputedState` config option is equal to `true`.

### Actual result

Plain `config.useComputedState` config option is equal to `undefined` and interpreted as `true`.

## Other details

* Browser: -
* OS: -
* CKEditor version: 4.16.2 (it works like that since the option was introduced)

shabab477:
Issue #4918

### This PR contains

- [ ] Unit tests
- [ ] Manual tests

Skipped

## What changes did you make?

Added default config value for `useComputedState`

Closes #4918.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -9,6 +9,9 @@ Fixed issues:
 
 * [#547](https://github.com/ckeditor/ckeditor4/issues/547): Fixed: Element/Image drag to move within the table is no longer available.
 
+API changes:
+
+* [#4918](https://github.com/ckeditor/ckeditor4/issues/4918): Explicitly set [`config.useComputedState`](https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_config.html#cfg-useComputedState) default value to `true`. Thanks to [Shabab Karim](https://github.com/shabab477)!
 
 ## CKEditor 4.17.1
 


Comandeer (changes requested):
Hi, @shabab477! Thanks for your contribution.

The change looks good. However, we require that every change be accompanied by the test. Could I ask you to add a test to `/tests/config` directory? The test can be simple, e.g. creating the editor and then checking its `editor.config.useComputedState` value.

It'd be also good to remove the logic connected to treating undefined `config.useComputedState` variable the same way as the one set to `true` (see the issue's description for more info). Could you add it as a part of this PR?

shabab477:
Alright :+1:

tests/core/config/inline.js
```diff
@@ -22,6 +22,10 @@ bender.test( {
 		} );
 	},
 
+	test_startup_computed_state_value: function() {
+		assert.isTrue( this.editor.config.useComputedState === true, 'config.useComputedState should return true as default' );
```
Comandeer, line 26: There is no need to compare variable's value to `true`, asserting its value is enough in this case:

```suggestion
		assert.isTrue( this.editor.config.useComputedState, 'config.useComputedState should return true as default' );
```

Comandeer (changes requested):
Thanks for your fixes!

The test looks good. Could I ask you to add a similar one also to `tests/core/config/themed_ui.js`?

There are also several other places in plugins where the trick similar to the one with `undefined` is used, e.g. https://github.com/ckeditor/ckeditor4/blob/6111ac2d0896e1bc53060ea1404c63612a37b5b1/plugins/bidi/plugin.js#L83. Could I ask you to also fix them?

core/config.js
```diff
@@ -369,6 +369,22 @@ CKEDITOR.config = {
...
+	 *
+	 * If enabled in a Left-To-Right written document, the "Left Justify"
+	 * alignment button will be shown as active, even if the alignment style is not
+	 * explicitly applied to the current paragraph in the editor.
+	 *
+	 *		config.useComputedState = false;
+	*
+	* @since 3.4.0
+	* @cfg {Boolean} [useComputedState=true]
+	*/
```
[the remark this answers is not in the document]
shabab477: Done

tests/core/config/inline.js
shabab477: Done

shabab477:
Fixed usages of  `useComputedState` workarounds in the code for plugins:
- justify
- list
- bidi

Comandeer (approved):
LGTM! Thanks once again for your contribution.

PR #4943, opened 2021-10-28
Fix removing multiple lists

### LINKED ISSUE #4875: Not possible to delete two lists selected by text
## Type of report

Bug

1.  Open e.g. https://nightly.ckeditor.com/21-09-08-06-03/full/samples/
2.  Switch to source mode and paste the following HTML
```
<p>begin</p>

<ul>
	<li>t1</li>
	<li>t2</li>
	<li>t3</li>
</ul>

<p>middle</p>

<ul>
	<li>t21</li>
	<li>t22</li>
</ul>

```
3. Select everything from first list item (t1) to last list item in the second list (t22)
obraz
4. Press delete or backspace

### Expected result
Whole text gets deleted.

### Actual result

Only second list gets deleted.
obraz

## Other details

*   Browser: Any
*   OS: Any
*   CKEditor version: 4.0 (I could reproduce it in 4.0)
*   Installed CKEditor plugins: list

KarolDawidziuk:

Bug fix

### This PR contains

- [x] Unit tests
- [x] Manual tests

```
* [#4875](https://github.com/ckeditor/ckeditor4/issues/4875): Fixed: Not possible to delete two lists selected by text.
```

## What changes did you make?

The 'getSelectedTableList()' function checks if the entire list is selected. When we selected more than one list, the function only returned the range of the last returned list. There was a problem with the walker getting lost. 
So, I've created an additional function that checks if a selected range has more than one list and if it's true, a new range is returned with the updated `range.startContainer`. 

In addition, I had a problem with unit tests on IE where the selection was only partial, probably because IE performs additional operations during selection, therefore I created manual tests for these cases.

PS. Let me know if you have any additional test cases that I could add.

Closes #4875.

### DIFF
--- a/core/editable.js
+++ b/core/editable.js
@@ -1032,24 +1032,36 @@
 							block,
 							parent,
 							next,
-							rtl = keyCode == 8;
+							rtl = keyCode == 8,
+							isMultipleSelectedList = false;
 
-
-						if (
-								// [IE<11] Remove selected image/anchor/etc here to avoid going back in history. (https://dev.ckeditor.com/ticket/10055)
-								( CKEDITOR.env.ie && CKEDITOR.env.version < 11 && ( selected = sel.getSelectedElement() ) ) ||
+						// [IE<11] Remove selected image/anchor/etc here to avoid going back in history. (https://dev.ckeditor.com/ticket/10055)
+						if ( CKEDITOR.env.ie && CKEDITOR.env.version < 11 && sel.getSelectedElement() ) {
+							selected = sel.getSelectedElement();
+						// Check it selection contain list or table.
+						} else if ( isListOrTableInSelection( sel ) ) {
+							// (#4875)
+							var multipleListSelection = areMultipleListInRange( range );
+							if ( multipleListSelection ) {
+								selected = multipleListSelection;
+								isMultipleSelectedList = true;
+							} else if ( !isMultipleSelectedList ) {
 								// Remove the entire list/table on fully selected content. (https://dev.ckeditor.com/ticket/7645)
-								( selected = getSelectedTableList( sel ) ) ) {
+								selected = getSelectedTableList( sel );
+							}
+						}
+
+						if ( selected ) {
 							// Make undo snapshot.
 							editor.fire( 'saveSnapshot' );
-							if ( !selected.isMultipleSelectedList ) {
+							if ( isMultipleSelectedList ) {
+								selected.deleteContents();
+							} else {
 								// Delete any element that 'hasLayout' (e.g. hr,table) in IE8 will
 								// break up the selection, safely manage it here. (https://dev.ckeditor.com/ticket/4795)
-								range.moveToPosition( selected.selection, CKEDITOR.POSITION_BEFORE_START );
+								range.moveToPosition( selected, CKEDITOR.POSITION_BEFORE_START );
 								// Remove the control manually.
-								selected.selection.remove();
-							} else {
-								selected.selection.deleteContents();
+								selected.remove();
 							}
 
 							range.select();
@@ -1548,15 +1560,22 @@
 		return false;
 	}
 
+	function isListOrTableInSelection( selection ) {
+		var range = selection.getRanges()[ 0 ],
+			path = range.startPath(),
+			structural = { table: 1, ul: 1, ol: 1, dl: 1 };
+
+		return !!path.contains( structural );
+	}
+
 	// Check if the entire table/list contents is selected.
 	function getSelectedTableList( sel ) {
 		var selected,
 			range = sel.getRanges()[ 0 ],
 			editable = sel.root,
-			path = range.startPath(),
 			structural = { table: 1, ul: 1, ol: 1, dl: 1 };
 
-		if ( path.contains( structural ) ) {
+		if ( isListOrTableInSelection( sel ) ) {
 			// Clone the original range.
 			var walkerRng = range.clone();
 
@@ -1566,16 +1585,6 @@
 
 			// Create a new walker.
 			var walker = new CKEDITOR.dom.walker( walkerRng );
-
-			// Check if range include more than one list.
-			var multipleList = checkForMultipleList( range );
-
-			if ( multipleList ) {
-				return {
-					isMultipleSelectedList: true,
-					selection: multipleList
-				};
-			}
 
 			// Assign a new guard to the walker.
 			walker.guard = guard();
@@ -1612,10 +1621,7 @@
 				// Go forwards checking for selected structural node.
 				walker.checkForward();
 
-				return {
-					isMultipleSelectedList: false,
-					selection: selected
-				};
+				return selected;
 			}
 		}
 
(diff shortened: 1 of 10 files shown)


core/editable.js
```diff
@@ -1042,12 +1042,16 @@
 								( selected = getSelectedTableList( sel ) ) ) {
```
sculpt0r, line 1042: Please, notice that with recent changes - you also changed the value of this expression.

Previously, there is a case, when the returned value was `false`. Currently, the returned value is an object, which makes this expression `truthy`. Because this expression is one of the options in the `OR` statement - it will execute the first block of code, even when there is no `selected.selection`. So it makes a few test cases fail:
- `tests/core/editable/keystrokes/keystrokes`
- `tests/core/editable/keystrokes/delbackspacequirks/collapsed`

To me, the best practice will be to avoid assignments inside `if` statements. It also might be missed by the further readers, that a value changed inside `if` statement.

If you fixed this one - the old tests seem to be fixed, however, tests added in this PR are failing in some cases.

Could you take a look at this problem, analyze the cases and fix it?

Also, could you improve the styling of this `if` statement: brackets in proper line etc.

core/editable.js
```diff
@@ -1042,12 +1042,16 @@
...
 							editor.fire( 'saveSnapshot' );
+							if ( !selected.isMultipleSelectedList ) {
+								// Delete any element that 'hasLayout' (e.g. hr,table) in IE8 will
+								// break up the selection, safely manage it here. (https://dev.ckeditor.com/ticket/4795)
+								range.moveToPosition( selected.selection, CKEDITOR.POSITION_BEFORE_START );
+								// Remove the control manually.
+								selected.selection.remove();
+							} else {
+								selected.selection.deleteContents();
+							}
```
sculpt0r, line 1053: Negation statements always require a little bit more focus than plain value. So we could change this code as below:
```suggestion
							if ( selected.isMultipleSelectedList ) {
							    selected.selection.deleteContents();
							} else {
								// Delete any element that 'hasLayout' (e.g. hr,table) in IE8 will
								// break up the selection, safely manage it here. (https://dev.ckeditor.com/ticket/4795)
								range.moveToPosition( selected.selection, CKEDITOR.POSITION_BEFORE_START );
								// Remove the control manually.
								selected.selection.remove();
							}
```

core/editable.js
```diff
@@ -1563,6 +1567,16 @@
 			// Create a new walker.
 			var walker = new CKEDITOR.dom.walker( walkerRng );
 
+			// Check if range include more than one list.
+			var multipleList = checkForMultipleList( range );
```
sculpt0r, line 1571: I think we should rename the function, so the abowe comment could be removed. Also, the `check` in names gives no feedback about the results. The most popular conventions with `is/are`, `has`, `does` etc. are mostly the best. So we could have:  `areMultipleListInRage`, `doesRangeContainsMultipleLists` etc.

core/editable.js
```diff
@@ -1563,6 +1567,16 @@
...
+
+			if ( multipleList ) {
+				return {
```
sculpt0r, line 1574: I'm wondering if we could separate somehow verifying if there are multiple lists in the range from the old code? For example, extract a separate method for that? Returning an object with a flag inside seems to be confusing.

core/editable.js
```diff
@@ -1618,6 +1635,54 @@
...
+
+			if ( !isListItem || firstRangeEl.getPrevious() !== null ) {
+				return;
+			}
+
+			// Walk through all the items in the range to find lists.
+			while ( element ) {
+				var tagName = element.$.nodeName.toLowerCase();
+
+				CKEDITOR.tools.array.forEach( listsTypes, function( listType ) {
```
sculpt0r, line 1661: Probably you could use `some` method in this case as above?

core/editable.js
```diff
@@ -1618,6 +1635,54 @@
...
+					}
+				} );
+
+				element = walker.next();
+			}
+
+			if ( isInclude ) {
+				var listParent = null;
+				// Get the parent of the list.
+				CKEDITOR.tools.array.map( listsTypes, function( listType ) {
```
sculpt0r, line 1673: I'm not sure why the `map` is used here? Probably some oversight?  Also, this method favors the end elements from `listsTypes`. Are you sure that there will be no errors in more advanced nested samples? Maybe you could provide an additional test case for different nested lists?

sculpt0r (changes requested):
Hi @KarolDawidziuk 
Good job with tests cases! Nice, that you realized the error is in the core quite complex code 👍 🧠 

There is a problem with regressions in the old tests cases. Please read my inline comments.

After you resolve the issue with working old and new samples - we may move forward. Also, as I mentioned in one comment - please provide some examples with nested lists.

sculpt0r:
Rebased on the newest `next` branch to enable travis builds.

KarolDawidziuk:
Thanks, @sculpt0r for the review! 

It's ready for another one.
Changes:
* Fixed regression in failing unit tests
* improve styling if statements
* fixed remove support for some cases of nested lists
* moving the logic responsible for detecting and returning a range from lists to a separate function

Also, I added a unit test to cover case with nested lists that you mentioned. There was also a problem on FF with a more specific selection (like on IE) so I added this case as a manual test on FF and IE.

sculpt0r:
Rebased on newest master

tests/core/editable/manual/removemultiplelist.html
```diff
@@ -0,0 +1,20 @@
...
+		<li>t21</li>
+		<li>t22</li>
+	</ul>
+</textarea>
+
+
+<script>
+	CKEDITOR.replace('editor', {
+		height: 250
+	});
```
sculpt0r, line 19:
```suggestion
	} );
```

tests/core/editable/manual/removemultiplelist.html
```diff
@@ -0,0 +1,20 @@
...
+	<p>middle</p>
+	<ul>
+		<li>t21</li>
+		<li>t22</li>
+	</ul>
+</textarea>
+
+
+<script>
+	CKEDITOR.replace('editor', {
```
sculpt0r, line 17:
```suggestion
	CKEDITOR.replace( 'editor', {
```

tests/core/editable/removemultiplelist.js
```diff
@@ -0,0 +1,116 @@
...
+
+	var tests = {
+		'test remove full multiple list': function( editor, bot ) {
+			bender.tools.testInputOut( 'multiple_full_list', function( source, expected ) {
+				bender.tools.setHtmlWithSelection( editor, source );
+
+				editor.editable().fire( 'keydown', new CKEDITOR.dom.event( { keyCode: 46 } ) );
+
+				bender.assert.beautified.html( expected, bot.htmlWithSelection() );
+			} );
```
sculpt0r, line 21: This block of code is repetitive. It could be extracted to separate testing functions with arguments.

tests/core/editable/removemultiplelist.js
```diff
@@ -0,0 +1,116 @@
...
+
+	bender.editors = {
+		classic: {},
+		inline: {
+			creator: 'inline'
+		}
+	};
+
+	var tests = {
+		'test remove full multiple list': function( editor, bot ) {
```
sculpt0r, line 14: I would be more precise in all of those tests names. For example, in this case, instead of `full multiple list` I would use `multiple fully selected list` or something similar.

And for an overall convention, I would go with: `mutiple fully selected lists are removed when delete key is pressed`.

Please correct all tests names according to the above convention.

tests/core/editable/removemultiplelist.js
```diff
@@ -0,0 +1,116 @@
...
+			bender.tools.testInputOut( 'multiple_full_list', function( source, expected ) {
+				bender.tools.setHtmlWithSelection( editor, source );
+
+				editor.editable().fire( 'keydown', new CKEDITOR.dom.event( { keyCode: 46 } ) );
```
sculpt0r, line 18: It would be easier to understand what is happening, if we extract hardcoded `46` into a separate variable, like `DELETE_KEY_CODE` or similar.

Also, is there any difference between using the delete key and backspace? Maybe we should verify the behavior for both of them?

tests/core/editable/removemultiplelist.js
```diff
@@ -0,0 +1,116 @@
...
+
+				bender.assert.beautified.html( expected, bot.htmlWithSelection() );
```
sculpt0r, line 20: Please add if possible additional failing test message unique for each test case. I know it is kind of redundant work, but those messages allow us to precisely identify failing tests on CI.

sculpt0r (changes requested):
Wow! Huge work with solving the complex `if...else`! 👏 👍 

It looks solid and for certain more understandable now. Still, there are too many `if...else`'s but according to this task scope - it is great!

I left a few inline comments - please take a look at them. Mostly technical stuff - not about the solution itself ;)

I've made a rebase and one refactor: https://github.com/ckeditor/ckeditor4/pull/4943/commits/d92bbeeb208fd5e4651ee332247c17e894fd9535 which cover function extraction. It allows simplifying reading the higher function logic. Also, it allows to remove helper vars from the main logic and keep them only for the purposes of finding parent logic.

Could you tell me why the `removemultiplelistunsupported` is also desired for FF?

The solution also looks good for IE11, but on IE8 there is a couple of test cases failing for `tests/core/editable/removemultiplelist`. Looks like detailed differences in the output. If it would be troublesome to fix them, they could be ignored, but please verify which versions of IE failed at them.

Please correct the tests tags once again 🙈

tests/core/editable/removemultiplelist.js
KarolDawidziuk: I didn't notice any difference when deleting lists using `Delete` or `Backspace` key, but to be sure I added tests for both cases.

KarolDawidziuk:
Thanks again @sculpt0r for the review!
It's ready for another one. 

Changes:
* Refactor unit tests. I've added a check for removing lists by using `Delete` and `Backspace` keys.
* correct bender tags
* ignore failing tests on IE
* indentation fix in the manual tests

> Could you tell me why the removemultiplelistunsupported is also desired for FF?

This is because for some reason they fail in the same way as tests on IE. This is probably related to selection on FF and correcting it would probably require writing these tests in a different way so a faster way was just to move this particular test to a manual test.

sculpt0r (approved):
Hi @KarolDawidziuk ,
I finally accept the PR 🎉 👍 

However, please take a look at refactored once again automatic tests (https://github.com/ckeditor/ckeditor4/pull/4943/commits/226588318c95e20353fe5b55dc2c672e30cbecdc).

Your code made everything automatic, but it was hard to tell what is actually going on in tests. Now there is only one support function that performs the test. Each test name is still nicely related to the template. There is only one kind of redundancy with the keys that are used to delete a list. However, I think it is a more readable approach WDYT?

PR #4951, opened 2021-11-05
Allow basic d&d element/image in tables when tableselection is on

### LINKED ISSUE #547: Element/Image drag to move within table is no longer available

KarolDawidziuk:

 Bug fix

### This PR contains

- [ ] Unit tests
- [x] Manual tests

```
* [#547](https://github.com/ckeditor/ckeditor4/issues/547): Fixed: Element/Image drag to move within the table is no longer available.
```

## What changes did you make?

D&D in tables was unavailable because the `drag` event handler cancel this event. 
To allow it I just removed the part of the code responsible for the interrupting of the event and in the same handler updated the `fakeSelection.active` flag to stop table selection.

This fix only allows moving only selected text and images. Trying to select an entire cell with `tableselection` and moving it fails.

PS. This branch has this name because there is already one frozen PR for this issue.

Closes #547.

### DIFF
--- a/plugins/tableselection/plugin.js
+++ b/plugins/tableselection/plugin.js
@@ -381,6 +381,9 @@
 	function fakeSelectionDragHandler( evt ) {
 		var table = evt.data.getTarget().getAscendant( 'table', true );
 
+		// Prevent table selection during dragging. (#547)
+		fakeSelection.active = false;
+
 		// (#2945)
 		if ( table && table.hasAttribute( ignoredTableAttribute ) ) {
 			return;
@@ -391,10 +394,6 @@
 		if ( !cell || cell.hasClass( fakeSelectedClass ) ) {
 			return;
 		}
-
-		// We're not supporting dragging in our table selection for the time being.
-		evt.cancel();
-		evt.data.preventDefault();
 	}
 
 	function copyTable( editor, isCut ) {
@@ -786,6 +785,11 @@
 			pastedTableMap;
 
 		if ( !isCustomPaste( selection, selectedCells, pastedTable, boundarySelection ) ) {
+			return;
+		}
+
+		// #(547)
+		if ( evt.data.method === 'drop' ) {
 			return;
 		}
 
--- a/tests/plugins/tableselection/manual/draganddropimage.html
+++ b/tests/plugins/tableselection/manual/draganddropimage.html
@@ -1,0 +1,42 @@
+<textarea id="editor">
+	<table border="1" cellpadding="1" cellspacing="1" style="width:500px">
+		<tbody>
+			<tr>
+				<td><img alt="" src="https://via.placeholder.com/100" />&nbsp;</td>
+				<td>test</td>
+			</tr>
+			<tr>
+				<td>
+				<p>&nbsp;</p>
+				<table border="1" cellpadding="1" cellspacing="1" style="width:500px">
+					<tbody>
+						<tr>
+							<td>&nbsp;</td>
+							<td>&nbsp;</td>
+						</tr>
+					</tbody>
+				</table>
+
+				<p>&nbsp;</p>
+				</td>
+				<td>&nbsp;</td>
+			</tr>
+			<tr>
+				<td>&nbsp;</td>
+				<td>&nbsp;</td>
+			</tr>
+		</tbody>
+	</table>
+</textarea>
+
+<script>
+	if ( bender.tools.env.mobile ) {
+		bender.ignore();
+	}
+
+	bender.tools.ignoreUnsupportedEnvironment( 'tableselection' );
+
+	CKEDITOR.replace('editor', {
+		height: 350
+	} );
+</script>
--- a/tests/plugins/tableselection/manual/draganddropimage.html
+++ b/tests/plugins/tableselection/manual/draganddropimage.html
@@ -2,7 +2,7 @@
 	<table border="1" cellpadding="1" cellspacing="1" style="width:500px">
 		<tbody>
 			<tr>
-				<td><img alt="" src="https://via.placeholder.com/100" />&nbsp;</td>
+				<td><img alt="" src="http://via.placeholder.com/100" />&nbsp;</td>
 				<td>test</td>
 			</tr>
 			<tr>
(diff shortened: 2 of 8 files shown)


tests/plugins/tableselection/manual/draganddropimage.html
```diff
@@ -0,0 +1,42 @@
+<textarea id="editor">
+	<table border="1" cellpadding="1" cellspacing="1" style="width:500px">
+		<tbody>
+			<tr>
+				<td><img alt="" src="https://via.placeholder.com/100" />&nbsp;</td>
```
sculpt0r, line 5: What is the purpose of this `&nbsp`? Because without it - when I d&d image - it is pasted with additional borders 🤔 

If you leave it, then it is possible to d&d image inside the same cell, but after `&nbsp`. Then when you d&d image to another cell it is copied with additional borders. It looks like it is copied with the entire `table` construction inside the `td` element.

tests/plugins/tableselection/manual/draganddropimage.html
```diff
@@ -0,0 +1,42 @@
...
+				<td>&nbsp;</td>
+				<td>&nbsp;</td>
+			</tr>
+		</tbody>
+	</table>
+</textarea>
+
+<script>
+	if ( bender.tools.env.mobile ) {
+		bender.ignore();
```
sculpt0r, line 34: Please add `return` statement and wrap the entire code in IIFE. The same for the other manual test.

tests/plugins/tableselection/manual/draganddropimage.html
```diff
@@ -0,0 +1,42 @@
...
+	}
+
+	bender.tools.ignoreUnsupportedEnvironment( 'tableselection' );
+
+	CKEDITOR.replace('editor', {
```
sculpt0r, line 39:
```suggestion
	CKEDITOR.replace( 'editor', {
```

tests/plugins/tableselection/manual/draganddropimage.md
```diff
@@ -0,0 +1,17 @@
+@bender-tags: dialog, __next__, 547, bug
+@bender-ui: collapsed
+@bender-ckeditor-plugins: wysiwygarea, toolbar, table, tableselection, image, undo, sourcearea, basicstyles
```
sculpt0r, line 3: Please add `elementspath` plugin as it additionally shows where is the current selection. The same for the other manual test.

tests/plugins/tableselection/manual/draganddropimage.md
```diff
@@ -0,0 +1,17 @@
...
+
+**Expected**
+
+Image was moved properly, the content looking good and there is no error in the console.
+
+**Unxpected**
+
+Image was not moved.
+There's error in the console.
+After the image was moved, the table was modified - an extra cell was added.
```
sculpt0r, line 17: Please - for `expected` and `unexpected` use notation described in our guide: https://github.com/ckeditor/ckeditor4/blob/major/dev/docs/codestyle.md#31-manual-tests.

So please split sentences and use list formating for them.

tests/plugins/tableselection/manual/draganddroptext.html
```diff
@@ -0,0 +1,39 @@
...
+</textarea>
+
+<script>
+	if ( bender.tools.env.mobile ) {
+		bender.ignore();
+	}
+
+	bender.tools.ignoreUnsupportedEnvironment( 'tableselection' );
+
+	CKEDITOR.replace('editor');
```
sculpt0r, line 38:
```suggestion
	CKEDITOR.replace( 'editor' );
```

tests/plugins/tableselection/manual/draganddropimage.html
sculpt0r: However, it occurs only in chrome... 🤔

plugins/tableselection/plugin.js
```diff
@@ -789,6 +788,11 @@
 			return;
 		}
 
+		// #(547)
+		if ( evt.data.method === 'drop' ) {
+			return;
+		}
```
sculpt0r, line 794: I'm assuming this is needed to prevent this listener from executing whenever the 'paste' is performed because of the `drop` operation?

sculpt0r (changes requested):
Nice and clean solution 👍  I hope that disabling selection do not break anything ;) However all tests on chrome all green 🤞 

I left a few inline comments. The most important is the one with d&d issue on chrome.

I think we could also add a mixed test, so d&d for mixed images with text content.

Also, please use the `4.17.1` tag as @jacekbogdanski suggested.

sculpt0r:
Rebased on newest `next` branch.

sculpt0r:
Rebased on newest `next` branch to enable travis.

tests/plugins/tableselection/manual/draganddropimage.html
KarolDawidziuk: As you said, without `&nbsp;` d&d is breaking. I created a separate [issue](https://github.com/ckeditor/ckeditor4/issues/4952) for this case. This problem occurs when `tableselection` and `clipboard` plugins are enabled.

plugins/tableselection/plugin.js
KarolDawidziuk: Exactly. Without preventing d&d images breaks because of the logic in this function.

KarolDawidziuk:
Thanks, @sculpt0r for the review!
It's ready for another one :)

Changes: 
* New manual test for mixed content ( text + image )
* Indentation fix
* Ignore test fix
* Using proper bender tags

sculpt0r:
Rebased on newest mater.

sculpt0r (approved):
Thank you for the latest corrections and extracting the issue 👍 Besides it: the solution looks good 🎉  We just earned a feature 👏 

I only changed the tags to `4.17.2` since the latest release already is `4.17.1`

PR #5023, opened 2022-01-10
Cleanup MutationObserver

### LINKED ISSUE #5004: Cleanup MutationObserver
## Type of report

Task

Extracted from https://github.com/ckeditor/ckeditor4-vue/pull/128

During test adjustment in our vue integration, we encountered a few issues related to MutationObserver.  There is a case, when:
- the default value of `observableParent` is used
- the CKEDITOR namespace was removed
- modification on the page DOM takes place - which causes mutation observer to run its callbacks.

Currently, it leads to an error, that CKEDITOR is undefined. After observer registration: https://github.com/ckeditor/ckeditor4/blob/10b8a218bdfb04e5b4db759c4072bd0ca6ab3830/plugins/wysiwygarea/plugin.js#L136-L142 we kind of loses control. The observer keeps tracking the DOM, but we may not handle all cases of what is happening with the editor and the namespace. Observer callbacks should be safely run even when the namespace is removed. So we should **remove CKEDITOR namespace reference from the callbacks**. It is used only for array tools which may be replaced with casual `for` loop. 

That was the case where we totally lose control over user actions. However, we allow the Users to safely clean up the editor with the `destroy` method. However, the observer ignores the fact that the editor may be destroyed... So we register something 'globally' and there is no chance to stop it unless the mode changes: https://github.com/ckeditor/ckeditor4/blob/10b8a218bdfb04e5b4db759c4072bd0ca6ab3830/plugins/wysiwygarea/plugin.js#L83-L85 but if someone destroys the editor - it also should be disconnected. Probably it could be `destroy` event: https://github.com/ckeditor/ckeditor4/blob/10b8a218bdfb04e5b4db759c4072bd0ca6ab3830/core/editor.js#L926 

*What steps should be taken to fulfil the task?*
Summary from the above:
- remove CKEDITOR namespace from the observer callbacks
- disconnect observer on `destroy` event

We should 

## Other details

* Browser: non IE browser
* OS: -
* CKEditor version: 4.17.1
* Installed CKEditor plugins: wysiwygarea

sculpt0r:

Bug fix

### This PR contains

- [ ] Unit tests
- [ ] Manual tests

```
* [#5004](https://github.com/ckeditor/ckeditor4/issues/5004): Fixed: possible memory leak with MutationObserver
```

## What changes did you make?

*Give an overview…*
MutationObserver is now disconnected on the `destroy` editor event.

Removed `CKEDITOR` namespace - replaced with casual `for` loop.

Closes #5004.


KarolDawidziuk (changes requested):
Hi, @sculpt0r and thank you for the improvements (and explanation 😉 )
Overall, everything looks good, and the issue with `MutationObserver` errors after removing `CKEDITOR` namespace is gone. 
I added a simple code style suggestion. Please check the comments below.

Despite the existing tests with `MutationObserver` I would be in favor of adding one more, closely related to this problem with removing `CKEDITOR` namespace and modifying DOM. Please add one more test covering this case.

plugins/wysiwygarea/plugin.js
```diff
@@ -135,7 +140,12 @@
 
 				function observeEditor() {
 					mutationObserver = new MutationObserver( function( mutationsList ) {
-						CKEDITOR.tools.array.forEach( mutationsList, verifyIfAddsNodesWithEditor );
+						var length = mutationsList.length,
+							index;
+
+						for ( index = 0; index < length; index++ ) {
+							verifyIfAddsNodesWithEditor( mutationsList[ index ] );
+						}
```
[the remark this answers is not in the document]
sculpt0r: It is simple. However, with the [default config option](https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_config.html#cfg-observableParent) the entire document is being observed. If someone uses CKE4 in an environment where a lot is going on - the observer might be involved many times with a lot of 'change' events. This looks like an unnecessary micro-optimization, but I would keep it. Is that alright?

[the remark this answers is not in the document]
KarolDawidziuk: To be honest for me it's still looking less readable but if this can help a little with the performance, I can stay with it. 👍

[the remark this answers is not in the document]
jacekbogdanski: It seems like unnecessary micro-optimization indeed. Accessing properties like that have minimal impact on performance, especially that MutationObserver is available only in modern browsers anyway.

sculpt0r:
@KarolDawidziuk - ready for another round

- added manual test for the issue with the CKEDITOR namespace. It should be verifiable and easier to test it.
- I switch to the plain version of for loop

tests/plugins/wysiwygarea/manual/detachobserver.md
```diff
@@ -0,0 +1,11 @@
+@bender-tags: 4.17.2, feature, 4462
```
KarolDawidziuk, line 1:
```suggestion
@bender-tags: 4.17.2, feature, 5004
```
This PR will close #5004. Please change it.

tests/plugins/wysiwygarea/manual/detachobserver.md
```diff
@@ -0,0 +1,11 @@
...
+@bender-ui: collapsed
+@bender-ckeditor-plugins: wysiwygarea,toolbar,basicstyles,link,format,sourcearea,elementspath,undo
+
+1. Open dev console to track possible errors.
+2. Wait until the `Trigger test` button will be available.
+3. Click the button and watch console.
+
+	**Expected:** No errors in the console.
+
+	**Unexpected:** Error appears: `Uncaught ReferenceError: CKEDITOR is not defined`.
```
KarolDawidziuk, line 11:
```suggestion
**Expected:** No errors in the console.

**Unexpected:** Error appears: `Uncaught ReferenceError: CKEDITOR is not defined`.
```
This additional blank space is not correct with our [code style](https://github.com/ckeditor/ckeditor4/blob/major/dev/docs/codestyle.md#31-manual-tests). Please adjust it.

tests/plugins/wysiwygarea/manual/detachobserver.md
```diff
@@ -0,0 +1,11 @@
+@bender-tags: 4.17.2, feature, 4462
+@bender-ui: collapsed
+@bender-ckeditor-plugins: wysiwygarea,toolbar,basicstyles,link,format,sourcearea,elementspath,undo
+
+1. Open dev console to track possible errors.
```
KarolDawidziuk, line 5:
```suggestion
1. Open the dev console to track possible errors.
```
Or maybe
```suggestion
1. Open a browser console to track possible errors.
```

tests/plugins/wysiwygarea/manual/detachobserver.html
```diff
@@ -0,0 +1,33 @@
+<div id="editor">
+	<p>Lorem ipsum dolor sit amet.</p>
+</div>
+<button id="triggerButton" disabled>Trigger test</button>
+
+<script>
+	( function() {
+		if ( bender.tools.env.mobile ) {
```
KarolDawidziuk, line 8: `MutationObserver`is not supported in IE below 11. We should add it to this condition.

```suggestion
		if ( bender.tools.env.mobile && CKEDITOR.env.ie && CKEDITOR.env.version < 11 ) {
```

KarolDawidziuk (changes requested):
Hi, @sculpt0r, 
Thanks for the improvements. 

Overall everything looks good and this test will make it easier for us to test it during the testing phase. There are a few little things to improve. Check out my comments below.

I see that some of the tests are failing on FF. See https://app.travis-ci.com/github/ckeditor/ckeditor4/jobs/555409935.

It seems that the `destroy` event is trying to disconnect a `mutationObserver` that does not exist. Can you check it?

tests/plugins/wysiwygarea/manual/detachobserver.md
sculpt0r: Seems that we are using the `dev console` quite often - however I'll change it.

KarolDawidziuk (changes requested):
Thanks, @sculpt0r for the improvements.
They are looking good except that I have 4 unit tests that still failing in `tests/core/editor/detaching` locally on FF. 

After a single test run, failing only two: 

* `/tests/core/editor/detaching#tests%2Fcore%2Feditor%2Fdetaching%20test%20should%20not%20throw%20any%20error%20after%20detach%20and%20destroy%20after%20iframe%20%22load%22%20event%20-%20classic%20editor%20in%20textarea`
* `/tests/core/editor/detaching#tests%2Fcore%2Feditor%2Fdetaching%20test%20should%20not%20throw%20any%20error%20after%20detach%20and%20destroy%20after%20iframe%20%22load%22%20event%20-%20classic%20editor%20in%20div`

Can you check it?

sculpt0r:
@KarolDawidziuk - ready for another review
- indeed, those test fails also on my FF. It was fixed in https://github.com/ckeditor/ckeditor4/pull/5023/commits/0b6867c53d2a18aea92981a01b6fd44f5bc0e8ef. Because the editor not always was ready and the observer was registered, the destroy event comes in - in those tests. Also, this is an edge case which we just covered - nice catch!
- I also found, that `test recreating editor returns null from getSelection` fails randomly ~1/5 runs. So I give a little time to the editor before reattaching, so it can properly verify the selection value during the 'recreation' phase: https://github.com/ckeditor/ckeditor4/pull/5023/commits/3047a3b01e710a4d6fc15122f6460c5014db1966

KarolDawidziuk (approved):
Thanks, @sculpt0r for the improvements.
Now everything looks good and there are no failing tests. Good job 👍

tests/plugins/wysiwygarea/manual/detachobserver.md
```diff
@@ -0,0 +1,11 @@
+@bender-tags: 4.17.2, feature, 5004
```
Comandeer, line 1:
```suggestion
@bender-tags: 4.17.2, bug, 5004
```

tests/plugins/wysiwygarea/manual/detachobserver.html
```diff
@@ -0,0 +1,33 @@
...
+
+			var element = document.createElement( 'div' );
+			document.body.appendChild( element );
+		} );
+
+		CKEDITOR.replace( 'editor', {
+			on: {
+				instanceReady: function ( evt ) {
+					evt.editor.destroy();
+				},
```
Comandeer, line 24: WDYT about moving destroy logic into button's click action? It seems counterintuitive to me that a manual test does something without even my knowledge.

Additionally, will the error connected to the removed `CKEDITOR` namespace even appear if the observer should be disconnected right after the editor's destroy? Probably these two tests should be separate.

tests/plugins/wysiwygarea/detached.js
```diff
@@ -33,7 +33,9 @@
 				} );
 
 				editorContainer.remove();
-				editorContainerParent.append( editorContainer );
+				setTimeout( function() {
+					editorContainerParent.append( editorContainer );
+				}, 200 );
```
Comandeer, line 38: Why do we need a `setTimeout` here?

Comandeer (changes requested):
The change itself looks well. There is however an issue with the manual test (see my inline comment). Additionally, it'd be nice to have also some unit tests.

tests/plugins/wysiwygarea/manual/detachobserver.html
sculpt0r: I thought about it previously. However, the `detroy` is not tested here. The significant action takes place on the button click action (force changed to trigger MutationObserver). So - still you decide when exactly the critical logic will be executed.

I'll separate those two cases.

tests/plugins/wysiwygarea/detached.js
sculpt0r: That was a fix for unstable test: https://github.com/ckeditor/ckeditor4/pull/5023#issuecomment-1015156203 - however, I reverted it and FF seems to be fine... I try to restart CI a few times and we will see

sculpt0r: And yet, the test fails again randomly... It is because:
https://github.com/ckeditor/ckeditor4/blob/ee3949d19b95af2eef96ec561e647b2eb450bd64/plugins/wysiwygarea/plugin.js#L164-L177

- ☝️ line 170 leads to https://github.com/ckeditor/ckeditor4/blob/ee3949d19b95af2eef96ec561e647b2eb450bd64/core/editable.js#L555

- line 176 leads to document loads again and lead to the https://github.com/ckeditor/ckeditor4/blob/ee3949d19b95af2eef96ec561e647b2eb450bd64/plugins/wysiwygarea/plugin.js#L119

The test fails when the document did load before the assertion - and that sometimes happened before the resume is executed. But since I wrote all of this - I figure out - that simply getting selection before the `resume` is executed is just fine.

sculpt0r:
@Comandeer - ready for another review
- manual tests are splitted
- I used different solutions to stabilize the test
- I didn't create automatic tests because it is really troublesome. I mean I'm able to delete `CKEDITOR` and spy on console.error  in the test. But if you try to fail that test with the old code - the test just hangs - because of the errors in the mutation observer. So if we try to use this test as a regression indicator - it doesn't work well. And I can't find any sensible way to spy or sth the MutationObserver after the editor is destroyed.

Comandeer:
Rebased onto latest `master`.

Comandeer (approved):
LGTM! I've just slightly rephrased one of the manual tests.

PR #5035, opened 2022-01-13
Add manual test for cache keys in CDN

### LINKED ISSUE #5034: Add manual test verifying cache key loaded via CDN
## Type of report

Task

Followup ticket to #5003 

It would be good to see whether the cache key behaves correctly when the editor is loaded via CDN. We already have some tests imitating CDN:

`/tests/plugins/preview/manual/previewcdn`

We should check if we can use the same/similar logic and reuse it. Note that we want to see if CSS files are loaded correctly, so the script in the above manual test may require some update (like removing CSS files before loading CDN).

Comandeer:

Task

### This PR contains

- [ ] Unit tests
- [ ] Manual tests

skip

## What changes did you make?

I've added a manual test for cache keys and CDN. It uses the same technique proposed by @KarolDawidziuk in `/tests/plugins/preview/manual/previewcdn` however, as we need a built version, it points to CKEditor 4 Nightly or to a local version if it's a built one.

Closes #5034.

### DIFF
--- a/tests/core/ckeditor/manual/assetscachekeycdn.html
+++ b/tests/core/ckeditor/manual/assetscachekeycdn.html
@@ -1,0 +1,114 @@
+<div id="editor"></div>
+<button disabled id="loadButton">Load external editor</button>
+<div id="loading_box" style="margin: 20px 0"></div>
+<div id="editor_cdn">
+	<p>A simple editor</p>
+</div>
+<div id="fake_observable_item"></div>
+
+<script>
+	( function () {
+		if ( bender.tools.env.mobile || ( CKEDITOR.env.ie && CKEDITOR.env.version < 11 ) ) {
+			return bender.ignore();
+		}
+
+		// 1. Load local editor by default with bender.
+		// 2. Remove existing instance and add script with editor from CDN.
+		// 3. Wait for the new editor to be loaded from CDN and initialize it.
+		// 4. Add icons in IE <11 from local.
+		var editorVersion = '',
+			path = getCKEditorCORSLink( window.location ),
+			loadButton = CKEDITOR.document.getById( 'loadButton' );
+
+		CKEDITOR.replace( 'editor', {
+			// Set fake element as observableParent due to removing existing instance of the editor.
+			observableParent: document.getElementById( 'fake_observable_item' )
+		} );
+
+		CKEDITOR.once( 'instanceReady', function() {
+			loadButton.$.disabled = false;
+			loadButton.once( 'click', loadExternalEditor );
+		} );
+
+		function loadExternalEditor() {
+			var scripts = document.querySelectorAll( 'script' ),
+				styles = document.querySelectorAll( 'style' ),
+				head = CKEDITOR.document.getHead();
+
+			CKEDITOR.tools.array.forEach( styles, function ( style ) {
+				style.parentNode.removeChild( style );
+			} );
+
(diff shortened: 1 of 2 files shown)


tests/core/ckeditor/manual/assetscachekeycdn.html
```diff
@@ -0,0 +1,114 @@
...
+<script>
+	( function () {
+		if ( bender.tools.env.mobile || ( CKEDITOR.env.ie && CKEDITOR.env.version < 11 ) ) {
+			return bender.ignore();
+		}
+
+		// 1. Load local editor by default with bender.
+		// 2. Remove existing instance and add script with editor from CDN.
+		// 3. Wait for the new editor to be loaded from CDN and initialize it.
+		// 4. Add icons in IE <11 from local.
```
sculpt0r, line 18: Why are we ignore this test on IE < 11? And if so, maybe the point `4.` in comments and its corresponding logic is not needed?

tests/core/ckeditor/manual/assetscachekeycdn.html
```diff
@@ -0,0 +1,114 @@
...
+			// Set fake element as observableParent due to removing existing instance of the editor.
+			observableParent: document.getElementById( 'fake_observable_item' )
+		} );
+
+		CKEDITOR.once( 'instanceReady', function() {
+			loadButton.$.disabled = false;
+			loadButton.once( 'click', loadExternalEditor );
+		} );
+
+		function loadExternalEditor() {
```
sculpt0r, line 33: As you mentioned (https://github.com/ckeditor/ckeditor4/pull/5035#issue-1101523489), there is a lot of common code with `previewcdn`. It could be extracted to some helper file, but the common way in bender with `exported` and `global` in comments don't work. Not sure if linking via `<script>` is a good idea in bender... WDYT?

tests/core/ckeditor/manual/assetscachekeycdn.html
```diff
@@ -0,0 +1,114 @@
...
+							removePlugins: 'scayt,exportpdf'
+						} );
+						loadingBox.innerText = editorVersion + ' editor version was loaded successfully!';
+
+						// There is problem in IE <11 with receiving icons from githack so we must add it manually.
+						if ( CKEDITOR.env.ie && CKEDITOR.env.version < 11 ) {
+							CKEDITOR.once( 'instanceReady', function () {
+								fixButtonsIcons();
+							} );
+						}
```
sculpt0r, line 72: This test ignores IE < 11. Either we should remove the ignore condition or this code?

sculpt0r (changes requested):
@Comandeer - nicely done 👍 

Could you take a look at my inline comments? Besides that - it looks good.

tests/core/ckeditor/manual/assetscachekeycdn.html
Comandeer: We ignore it because older IEs don't have sensible network debugging, so the test there does not make any sense.

And probably you're right about the logic for icons.

tests/core/ckeditor/manual/assetscachekeycdn.html
Comandeer: TBH I'd keep them separated and come back to the idea if there is yet another CDN test.

Comandeer:
I removed logic dedicated to IE \< 11.

sculpt0r (approved):
LGTM 🎉

PR #5042, opened 2022-01-20
Update dev dependencies

jacekbogdanski:
Updated dev dependencies by patch/minor versions.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1076,11 +1076,10 @@
 			}
 		},
 		"node_modules/bl": {
-			"version": "1.2.2",
-			"resolved": "https://registry.npmjs.org/bl/-/bl-1.2.2.tgz",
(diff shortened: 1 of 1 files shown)


sculpt0r:
Rebased on the newest master

sculpt0r (approved):
Grunt commands seem to work fine. Bender is able to run tests. Don't see any issues. But, since it is a rare case - could you @Comandeer also look at this one?

Comandeer (commented):
I'm not sure if we want to bump only `package-lock.json`, leaving `package.json` totally untouched. Additionally, it's really hard to even know what was changed if only `package-lock.json` file is present as its format is nearly human-unreadable :/

sculpt0r:
> I'm not sure if we want to bump only `package-lock.json`, leaving `package.json` totally untouched. Additionally, it's really hard to even know what was changed if only `package-lock.json` file is present as its format is nearly human-unreadable :/

Most of the time, I just traverse up to the nearest top-level package to find out which top-level package uses it. And find out that it is bender and grunt. So I just try my day-to-day routine with them...

Comandeer (approved):
LGTM!

PR #5047, opened 2022-01-25
Add style for unfocused multiselection list

### LINKED ISSUE #5044: styles reset prevents seeing selected options in plugin dialog multiple select
## Bug
I created a plugin for CKEditor 4, with a widget, and a dialog that has a number of form inputs that affect the contents of the widget. Among the form inputs, there are some select controls with the `multiple` attribute. However, I have noticed that after selecting options from the multiple select, as soon as the select loses focus none of the selected options are showing as selected.

Upon examining closely, I noticed that rules from `editor.css` are overriding the browser styling on the select control. Specifically:

```css
/* user agent stylesheet */
select:-internal-list-box option:checked {
    background-color: -internal-light-dark(rgb(206, 206, 206), rgb(84, 84, 84));
    color: -internal-light-dark(rgb(16, 16, 16), rgb(255, 255, 255));
}
```
is being overriden by rules from this block:

```css
/* skins/moono-lisa/editor.css */
.cke_reset_all, .cke_reset_all *, .cke_reset_all a, .cke_reset_all textarea {
    margin: 0;
    padding: 0;
    border: 0;
    background: transparent;
    text-decoration: none;
    width: auto;
    height: auto;
    vertical-align: baseline;
    box-sizing: content-box;
    position: static;
    transition: none;
    border-collapse: collapse;
    font: normal normal normal 12px Arial,Helvetica,Tahoma,Verdana,Sans-Serif;
    color: #000;
    text-align: left;
    white-space: nowrap;
    cursor: auto;
    float: none;
}
```

Upon examining `editor.css`, I found that the above referenced block is imported from `skins/moono-lisa/reset.css`:

https://github.com/ckeditor/ckeditor4/blob/ee3949d19b95af2eef96ec561e647b2eb450bd64/skins/moono-lisa/reset.css#L44-L68

Specifically the problematic line is:

https://github.com/ckeditor/ckeditor4/blob/ee3949d19b95af2eef96ec561e647b2eb450bd64/skins/moono-lisa/reset.css#L51

Considering that there is no styling defined for selected options, the rule with `background: transparent` takes precedence and all options in the select control always have a transparent background whether they are selected or not (when the select control loses focus).

* CodePen: https://codepen.io/johnrdorazio/pen/poWXObX

1. Right click in the example ckeditor instance, and select "Open dialog"
2. There will be two multiple selects, the first one will be focused and selected elements will show. The second one is not focused, but selected elements are showing because I have applied a fix to this one (see CSS tab in the example).
3. With some options selected in the first multiple select, click anywhere outside of the select so it will lose focus.

### Expected result

The options selected in the first multiple select should show as being selected, just like the options in the second multiple select.

### Actual result

The selected options in the first multiple select do not show as being selected when the select has lost focus.

## Other details

* Browser: Chrome
* OS: Windows 10
* CKEditor version: 4.17.1
* Installed CKEditor plugins: widget

JohnRDOrazio:

Bug fix for issue #5044 

This patch contains a simple addition of CSS rules, to bring styling back to multiple select controls for when they lose focus. The browser styling is reset by the skins, so new styling needs to be reinstated for such specific cases.

### This PR contains

- [ ] Unit tests
- [x] Manual tests
- [ ] N/A
- 

```
* [#5044](https://github.com/ckeditor/ckeditor4/issues/5044):
  * add rule for selected options of multiple selects, to skins/kama/reset.css
  * add rule for selected options of multiple selects, to skins/moono-lisa/reset.css
  * add rule for selected options of multiple selects, to skins/moono/reset.css
```

## What changes did you make?

Added rule in the stylesheets of each skin, to reinstate styling of selected options in multiple selects.

Closes #5044.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -8,6 +8,7 @@ CKEditor 4 Changelog
 Other changes:
 
 * [#5087](https://github.com/ckeditor/ckeditor4/issues/5087): Deprecated jQuery API calls in jQuery adapter were replaced by undeprecated ones. Thanks to [Fran Boon](https://github.com/flavour)!
+* [#5044](https://github.com/ckeditor/ckeditor4/issues/5044): Fixed: Lack of style for selected items in unfocused multiselection list. Thanks to [John R. D'Orazio](https://github.com/JohnRDOrazio)!
 
 ## CKEditor 4.17.2
 
--- a/skins/kama/reset.css
+++ b/skins/kama/reset.css
@@ -109,6 +109,11 @@ https://ckeditor.com/docs/ckeditor4/latest/guide/skin_sdk_reset.html
 	box-sizing: border-box;
 }
 
+.cke_reset_all select[multiple] option:checked
+{
+	background-color: rgb(206, 206, 206);
+}
+
 .cke_reset_all table
 {
 	table-layout: auto;
--- a/skins/moono-lisa/reset.css
+++ b/skins/moono-lisa/reset.css
@@ -109,6 +109,11 @@ https://ckeditor.com/docs/ckeditor4/latest/guide/skin_sdk_reset.html
 	box-sizing: border-box;
 }
 
+.cke_reset_all select[multiple] option:checked
+{
+	background-color: rgb(206, 206, 206);
+}
+
 .cke_reset_all table
 {
 	table-layout: auto;
--- a/skins/moono/reset.css
+++ b/skins/moono/reset.css
@@ -109,6 +109,11 @@ https://ckeditor.com/docs/ckeditor4/latest/guide/skin_sdk_reset.html
 	box-sizing: border-box;
 }
 
+.cke_reset_all select[multiple] option:checked
+{
+	background-color: rgb(206, 206, 206);
+}
+
 .cke_reset_all table
 {
 	table-layout: auto;
--- a/tests/core/skin/manual/_assets/multiselectfix.css
+++ b/tests/core/skin/manual/_assets/multiselectfix.css
@@ -1,0 +1,4 @@
+.cke_reset_all select[multiple] option:checked
+{
+	background-color: rgb(206, 206, 206);
+}
--- a/tests/core/skin/manual/_assets/multiselectfix.css
+++ b/tests/core/skin/manual/_assets/multiselectfix.css
@@ -1,4 +1,0 @@
-.cke_reset_all select[multiple] option:checked
-{
-	background-color: rgb(206, 206, 206);
-}
--- a/tests/core/skin/manual/kamamultiselect.html
+++ b/tests/core/skin/manual/kamamultiselect.html
@@ -1,120 +1,50 @@
-<head>
-	<link rel="stylesheet" href="_assets/multiselectfix.css">
-</head>
-<body>
+<div id="editor"></div>
+<button id="openDialogButton" disabled>Open dialog</button>
 
-	<div id="editor"></div>
-
-	<script>
-		CKEDITOR.plugins.add( 'test', {
-			requires: 'widget',
-			init: function(editor) {
-				editor.addCommand( 'mydialog', new CKEDITOR.dialogCommand( 'mydialog' ) );
-				if ( editor.contextMenu ) {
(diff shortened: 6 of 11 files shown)


sculpt0r:
Thank you @JohnRDOrazio for the effort to create a fix in this PR. We will review it and provide information about the results.

JohnRDOrazio:
any way to prevent this from becoming "stale", while you are reviewing it?

sculpt0r:
I'll handle that @JohnRDOrazio  - thank you for reminder :)

sculpt0r (changes requested):
Hi @JohnRDOrazio - your solutions is working fine 👍 

However, we need at least one manual test for each skin. Could you prepare them? Take a look at other manual tests in the project: e.g. https://github.com/ckeditor/ckeditor4/tree/master/tests/core/skin/manual

also, please take a look at our [code guide](https://github.com/ckeditor/ckeditor4/blob/master/dev/docs/codestyle.md#31-manual-tests).

JohnRDOrazio:
I'm not very familiar with bender, but I believe I have created the requested tests? Can you confirm?

sculpt0r:
Rebased onto newest master.

sculpt0r (changes requested):
Hi @JohnRDOrazio 

Thank you for preparing the manual test - it was almost perfect ;)

I managed to shrink it a little bit: we need as small sample as possible to reproduce the issue. Less code to maintain and to read is always better. Please take a look at changes that I've made in https://github.com/ckeditor/ckeditor4/pull/5047/commits/699547287ab3d228da288c729a9b4fe3f9e19707

The previous version didn't open properly. I assume you didn't manage to run it on [bender environment](https://ckeditor.com/docs/ckeditor4/latest/guide/dev_tests.html)? However, I made some changes in `moonolisamultiselect` to guide you with changes:
- in `@bender-ckeditor-plugins` tag we need `wysiwygarea` plugin to initialize the editor properly
- in `@bender-tags` - we are putting the next release version number. It changed in the meantime to `4.17.3`
- removed the dialog logic - since it requires an additional plugin. All we need is to just show the dialog. There is far less code and logic if we just show it on a button click.
- removed all additional onChange logic from the dialog, also I left only a single list. It is enough to reproduce the issue.
- things from above allows simplifying editor initialization, also I removed unnecessary config options
- we use only `var` instead of `let` and `const` because CKE4 is supporting also old browsers like IE8
- the proper value for moonolisa skin in options is `moono-lisa` (with the dash inside)
- there is no need to put body and head elements in the manual test. Also, we don't need additional CSS file in samples (could you remove it?) The fix you provide in skins reset.css files is enough

Could you:
- do the same changes for two other skin tests?
- adjust the tests steps - so they match the new scenario? Wait for the button to be active, click on it to open the dialog and all the stuff with selecting list items and focus.
- make sure the code style is following [our code style](https://github.com/ckeditor/ckeditor4/blob/master/dev/docs/codestyle.md)

JohnRDOrazio:
@sculpt0r I simplified the other two tests as asked. I have no idea how to use the bender environment, I have never used it. I made the tests simply by a bit of guesswork, looking at other examples.

I have tried to setup bender, I installed the cli globally with `npm install -g benderjs-cli` and I ran `bender init` in my ckeditor test folder, and then `bender server run`, but I'm not seeing any manual tests in this interface. What steps should I take to view the tests in a bender environment?

sculpt0r:
Hi @JohnRDOrazio,
If you are seeing the bender testing page properly, then you just need to select the correct filter:
image

If I good understand - you managed to open such bender page?

JohnRDOrazio:
Got it! The bender page automatically had the `is:unit` filter, once I removed that I was able to see the manual tests. They all seem to be working.

sculpt0r:
Rebased onto newest master.

sculpt0r (approved):
Hi @JohnRDOrazio 

Everything looks and works properly - LGTM 🎉 

It's nice that you were able to run bender I change the rest of the tests.

Thank you for your commitment to CKE4 development! Thanks to you - there is one bug less in the editor :) I added you to the changelog :)

PR #5065, opened 2022-02-01
Fix for incorrect replacing with phrases with several spaces inside

### LINKED ISSUE #5061: Multiple spaces in `find&replace` as 'replaced with' turns to a single space
## Type of report

Bug

1. Open CKEditor4 sample: https://codepen.io/karoldawidziuk/pen/LYzJvdx
2. Click Find&Replace
3.  Provide text to be replaced, e.g. `h`
4. Provide text for replacement: use more than one space(e.g. 4 spaces)
5. Use Replace or Replace All button

### Expected result

The letter `h` should be replaced with four spaces. Spaces probably will be encoded as `&nbsp;` mixed with casual spaces.

### Actual result

Only a single space is used as a replacement.

## Other details
I can track the same behavior back to the `4.14.0` so i looks like no regression here

* Browser: all (Safari, FF, Chrome)
* OS: …
* CKEditor version: 4.17.1
* Installed CKEditor plugins: full-all

Comandeer:

Bug fix

### This PR contains

- [x] Unit tests
- [x] Manual tests

```
* [#5061](https://github.com/ckeditor/ckeditor4/issues/5061): Fixed: [Find / Replace](https://ckeditor.com/cke4/addon/find) plugin incorrectly handles multiple whitespace during replacing text.
```

## What changes did you make?

Creating a text node from any text removes any consecutive whitespace. To prevent that I've added a `replace()` that replaced every odd space with a non-breaking one.

The only concern I have is if we want to replace all spaces with the non-breaking one or only "normal" spaces (` `)? For now the code replaces all of them.

Closes #5061.

### DIFF
--- a/plugins/find/dialogs/find.js
+++ b/plugins/find/dialogs/find.js
@@ -489,15 +489,7 @@
 					// Turn off highlight for a while when saving snapshots.
 					this.matchRange.removeHighlight();
 					var domRange = this.matchRange.toDomRange(),
-						textWithPreservedSpaces = newString.replace( consecutiveWhitespaceRegex,
-							function( whitespace ) {
-								var newSpaces = CKEDITOR.tools.array.map( whitespace, function( space, i ) {
-									return i % 2 === 0 ? nonBreakingSpace : space;
-								} );
-
-								return newSpaces.join( '' );
-							} ),
-						text = editor.document.createText( textWithPreservedSpaces );
+						text = createTextNodeWithPreservedSpaces( editor, newString );
 
 					if ( !isReplaceAll ) {
 						// Save undo snaps before and after the replacement.
@@ -837,6 +829,19 @@
 				}
 			}
 		};
+
+		function createTextNodeWithPreservedSpaces( editor, text ) {
+			var textWithPreservedSpaces = text.replace( consecutiveWhitespaceRegex,
+				function( whitespace ) {
+					var newSpaces = CKEDITOR.tools.array.map( whitespace, function( space, i ) {
+						return i % 2 === 0 ? nonBreakingSpace : space;
+					} );
+
+					return newSpaces.join( '' );
+				} );
+
+			return editor.document.createText( textWithPreservedSpaces );
+		}
 	}
 
 	CKEDITOR.dialog.add( 'find', findDialog );
--- a/plugins/find/dialogs/find.js
+++ b/plugins/find/dialogs/find.js
@@ -387,6 +387,8 @@
 
 		var wordSeparatorRegex = /[.,"'?!;: \u0085\u00a0\u1680\u280e\u2028\u2029\u202f\u205f\u3000]/;
 		var spaceSeparatorRegex = /[\u0020\u00a0\u1680\u202f\u205f\u3000\u2000-\u200a]/;
+		var consecutiveWhitespaceRegex = /[\u0020\u00a0\u1680\u202f\u205f\u3000\u2000-\u200a]{2,}/g;
+		var nonBreakingSpace = '\xA0';
 
 		function isWordSeparator( c ) {
 			if ( !c )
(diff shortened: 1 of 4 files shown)


Comandeer:
It seems that this PR should be merged after #5066 as the other one is fixing failing `find` tests.

Comandeer:
Rebased onto latest `master`.

plugins/find/dialogs/find.js
```diff
@@ -486,8 +488,17 @@
...
-					var text = editor.document.createText( newString );
+					var domRange = this.matchRange.toDomRange(),
+						textWithPreservedSpaces = newString.replace( consecutiveWhitespaceRegex,
+							function( whitespace ) {
+								var newSpaces = CKEDITOR.tools.array.map( whitespace, function( space, i ) {
+									return i % 2 === 0 ? nonBreakingSpace : space;
+								} );
+
+								return newSpaces.join( '' );
+							} ),
```
KarolDawidziuk, line 499: Overall code works good but let's move the replace logic into a separate function. It will look clearer.

tests/plugins/find/find.js
```diff
@@ -263,6 +263,41 @@ bender.test( {
...
+	// (#5061)
+	'test replace all occurences of phrase with phrase with several spaces inside': function() {
+		var bot = this.editorBot;
+
+		bot.setHtmlWithSelection( '<p>[replace me]</p><p>replace me</p>' );
+
+		bot.dialog( 'replace', function( dialog ) {
+			dialog.setValueOf( 'replace', 'txtReplace', 'foo   bar' );
+			dialog.getContentElement( 'replace', 'btnReplaceAll' ).click();
+			dialog.getButton( 'cancel' ).click();
```
KarolDawidziuk, line 293: We can remove this line because we hide dialog after assertion. ( line 297 )

KarolDawidziuk (changes requested):
Hi, @Comandeer 
Overall everything is good and your fix works well.

There are some small style adjustments to add. Please see my inline comments.

tests/plugins/find/find.js
```diff
@@ -263,6 +263,41 @@ bender.test( {
...
+
+			assert.areSame( '<p>foo&nbsp; &nbsp;bar</p><p>foo&nbsp; &nbsp;bar</p>', bot.getData( false, true ) );
```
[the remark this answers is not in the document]
Comandeer: I've just copied the code from some previous tests :see\_no\_evil: I'll remove unnecessary parameters.

tests/plugins/find/manual/replacedoublespace.html
```diff
@@ -0,0 +1,9 @@
+<textarea id="editor">
+	<p>Replace me</p>
+</textarea>
+
+<script>
+	CKEDITOR.replace( 'editor', {
+		language: 'en'
```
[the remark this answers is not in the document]
Comandeer: No, it's not. Leftover from testing in my Chrome that makes the editor using Polish :upside\_down\_face:

tests/plugins/find/find.js
Comandeer: Actually, I've moved it before the assertion in both tests. Otherwise, if the assertion failed, the dialog wouldn't be closed.

Comandeer:
I've moved the logic into a separate function and updated tests – switching also from "Replace" to "Replace All" (as "Replace" need two clicks and "Replace All" only one) in them. Also rebased the PR onto latest `master`.

KarolDawidziuk (approved):
Thanks, @Comandeer for the improvements.
For me, everything looks good, so it's R+ from my side. 🎉

plugins/find/dialogs/find.js
```diff
@@ -826,6 +829,19 @@
 				}
 			}
 		};
+
+		function createTextNodeWithPreservedSpaces( editor, text ) {
+			var textWithPreservedSpaces = text.replace( consecutiveWhitespaceRegex,
+				function( whitespace ) {
+					var newSpaces = CKEDITOR.tools.array.map( whitespace, function( space, i ) {
```
jacekbogdanski, line 836: `CKEDITOR.tools.array.map` function takes an array as an argument, not a string. This code takes advantage of strings defining iterator, but it reduces code readability.

jacekbogdanski (changes requested):
We are missing a case when text includes additional space at the end of a phrase, e.g. `foobar<space>`. I've added a test in https://github.com/ckeditor/ckeditor4/pull/5065/commits/76d3d0131a3dc31fbb3b72221367a91bb6c32e07 confirming that issue.

Comandeer:
> We are missing a case when text includes additional space at the end of a phrase

Spaces at the end of phrases are preserved as long as the end of the phrase is not the same as the end of the paragraph. In such a case it seems to be a totally separate issue, connected with how `CKEDITOR.dom.range#insertNode()` works (see #5074). In that case, I propose to simply remove the failing test and resolve that issue separately.

jacekbogdanski:
Alright, thanks for checking this out and extracting a follow-up ticket. Let's wait for CI and if it's green we can merge this PR 🎉

jacekbogdanski (approved):

PR #5102, opened 2022-02-23
Updated deps and removed package-lock.json file

jacekbogdanski:
I've updated our deps and removed `package-lock.json` as we discussed some time ago that this file may actually force library consumers to use a specific version of the package that should be rather specified in `package.json` . There is a nice discussion regarding this issue: https://github.com/yarnpkg/yarn/issues/838#issuecomment-253362537

No `package-lock.json` means also a better understanding of potential breaking changes during development, so we actually want to see ASAP if some dependency starts to break it.


sculpt0r (changes requested):
I removed the entire `node_muldes` and reinstalled deps again.

- grunt based tasks work fine (linting, img minification, packing ckeditor_base.js)
- I tried to run `grunt less` from `dev/tasks/samples.js` - but it fails due to dependency in another repo `cksource-samples-framework` and I don't have access to it
image
However, it looks like an old error (probably we could extract an issue for it? @jacekbogdanski )
- bender and randomly ran tests also works fine

Beside that... Looks fine

sculpt0r (approved):
I extracted the issue: #5103 

The rest: LGTM 🎉

PR #5123, opened 2022-03-12
Expose toggle buttons in accessibility tree

### LINKED ISSUE #2444: Expose toggle buttons in accessibility tree
## Type of report

Feature request

There are many toggle buttons inside editor's toolbar, e.g. "Bold", "Italic", "Underline" etc. To correctly expose them as toggle buttons in a11y tree, they must have `[role=button]` and `[aria-pressed]` attribute.

Default state:

```html
<a href="#" role="button" aria-pressed="false">Bold</a>
```

Toggled/pressed state:

```html
<a href="#" role="button" aria-pressed="true">Bold</a>
```

However we don't add `[aria-pressed=false]` for not-pressed buttons, as we do not distinguish between ordinary buttons and toggle ones. We only add `[aria-pressed=true]` when toggle buttons are activated. We could introduce the new toggle button category, which will always have `[aria-pressed]` attribute, removing handling of this attribute from ordinary buttons at the same time.

### LINKED ISSUE #4543: [Accessibility] - The selected state for the toolbar buttons is not announced by screen reader
## Type of report

Bug

1. Launch ckeditor demo url https://ckeditor.com/ckeditor-4/demo/
2. Launch any screen reader like NVDA/JAWS/Voice Over(iOS).
3. Press bold and type some text.
4. Press ALT+F10 to navigate to toolbar using keyboard.
5. Navigate to Bold button and some other button which is not selected e.g. Italic.

### Expected result

For screen reader users proper information should be conveyed when they access toolbar buttons i.e. selected/not.

### Actual result

When screen reader users navigate to the selected item(Bold) and non selected item(Italic), screen reader is not conveying any information like Bold selected etc...

## Other details

* Browser: Chrome, Safari
* OS: Windows, iOS
* CKEditor version: 4

Comandeer:

New feature

### This PR contains

- [x] Unit tests
- [x] Manual tests

```
New features:

* [#2444](https://github.com/ckeditor/ckeditor4/issues/2444): Togglable toolbar buttons are now exposed as toggle buttons in the browser's accessibility tree.

Fixed Issues:

* [#4543](https://github.com/ckeditor/ckeditor4/issues/4543): The selected states of toolbar buttons are not announced by screen readers.
```

## What changes did you make?

I've changed the way of handling the `[aria-pressed]` attribute in the `button` plugin. After the changes, the attribute is always present on the button and only its value is changed (previously the attribute was removed when the button was not pressed). Additionally, I've added the `isToggle` option to the button's definition which indicates that the button is togglable (this option is needed to handle the initial state of the toggle buttons).

I've then added `isToogle` option to plugins using toggle buttons:

* basicstyles,
* bidi,
* blockquote,
* copyformatting,
* justify,
* list,
* sourcearea.

However, there are still some plugins that require additional attention:

* language and colorbutton – these plugins use menu buttons that are also togglable, however, screen readers do not support such combination, so the information about the state of the toggle button is lost,
* maximize – this plugin uses the toggle button but instead of conveying the state using the `[aria-pressed]` attribute, it changes the label of the button (from "Maximize" to "Minimize"; I'd suggest replacing the current method with the `[aria-pressed]` one, used in other plugins.

Closes #2444.
Closes #4543.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -5,6 +5,14 @@ CKEditor 4 Changelog
 
 ## CKEditor 4.18.1 [IN DEVELOPMENT]
 
+New features:
+
+* [#2444](https://github.com/ckeditor/ckeditor4/issues/2444): Togglable toolbar buttons are now exposed as toggle buttons in the browser's accessibility tree.
+
+Fixed Issues:
+
+* [#4543](https://github.com/ckeditor/ckeditor4/issues/4543): The selected states of toolbar buttons are not announced by screen readers.
+
 ## CKEditor 4.18.0
 
 **Security Updates:**
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -3,15 +3,13 @@ CKEditor 4 Changelog
 
 ## CKEditor 4.19.0 [IN DEVELOPMENT]
 
-## CKEditor 4.18.1 [IN DEVELOPMENT]
-
 New features:
 
 * [#2444](https://github.com/ckeditor/ckeditor4/issues/2444): Togglable toolbar buttons are now exposed as toggle buttons in the browser's accessibility tree.
 
 Fixed Issues:
 
-* [#4543](https://github.com/ckeditor/ckeditor4/issues/4543): The selected states of toolbar buttons are not announced by screen readers.
+* [#4543](https://github.com/ckeditor/ckeditor4/issues/4543): Toolbar buttons toggle state is not correctly announced by screen readers.
 
 ## CKEditor 4.18.0
 
--- a/plugins/button/plugin.js
+++ b/plugins/button/plugin.js
@@ -14,7 +14,8 @@
 		' aria-labelledby="{id}_label"' +
 		' aria-describedby="{id}_description"' +
 		' aria-haspopup="{hasArrow}"' +
-		' aria-disabled="{ariaDisabled}"';
+		' aria-disabled="{ariaDisabled}"' +
+		'{toggleHtml}';
 
 	// Some browsers don't cancel key events in the keydown but in the
 	// keypress.
@@ -51,7 +52,10 @@
 	( CKEDITOR.env.hc ? '&#9660;' : '' ) +
 		'</span>';
 
+	var templateToggle = ' aria-pressed="false"';
+
 	var btnArrowTpl = CKEDITOR.addTemplate( 'buttonArrow', templateArrow ),
+		toggleBtnTpl = CKEDITOR.addTemplate( 'toggleButton', templateToggle ),
 		btnTpl = CKEDITOR.addTemplate( 'button', template );
 
 	CKEDITOR.plugins.add( 'button', {
@@ -81,6 +85,7 @@
 		CKEDITOR.tools.extend( this, definition,
 		// Set defaults.
 		{
+			isToggle: definition.isToggle || false,
 			title: definition.label,
 			click: definition.click ||
 			function( editor ) {
@@ -308,7 +313,8 @@
 				focusFn: focusFn,
 				clickFn: clickFn,
 				style: CKEDITOR.skin.getIconStyle( iconPath, ( editor.lang.dir == 'rtl' ), overridePath, this.iconOffset ),
-				arrowHtml: this.hasArrow ? btnArrowTpl.output() : ''
+				arrowHtml: this.hasArrow ? btnArrowTpl.output() : '',
+				toggleHtml: this.isToggle ? toggleBtnTpl.output() : ''
 			};
 
 			btnTpl.output( params, output );
@@ -337,14 +343,11 @@
 				element.setState( state, 'cke_button' );
 				element.setAttribute( 'aria-disabled', state == CKEDITOR.TRISTATE_DISABLED );
 
-				if ( !this.hasArrow ) {
-					// Note: aria-pressed attribute should not be added to menuButton instances. (https://dev.ckeditor.com/ticket/11331)
-					if ( state === CKEDITOR.TRISTATE_ON ) {
-						element.setAttribute( 'aria-pressed', true );
-					} else {
-						element.removeAttribute( 'aria-pressed' );
-					}
-				} else {
+				if ( this.isToggle && !this.hasArrow ) {
+					// Note: aria-pressed attribute should not be added to menuButton instances. (https://dev.ckeditor.com/ticket/11331).
+					// Do not remove the attribute, set its value (#2444).
+					element.setAttribute( 'aria-pressed', state === CKEDITOR.TRISTATE_ON );
+				} else if ( this.hasArrow ) {
 					// Indicates that menu button is opened (#421).
 					element.setAttribute( 'aria-expanded', state == CKEDITOR.TRISTATE_ON );
 				}
@@ -437,6 +440,8 @@
 	 * 		} )
 	 * @param {String/Boolean} definition.hasArrow If Boolean, it indicates whether the button should have a dropdown. If a string, it acts
 	 * as a value of the button's `aria-haspopup` attribute. Since **4.11.0** it supports the string as a value.
+	 * @param {Boolean} [definition.isToggle=false] Indicates if the button should be treated as a toggle one
+	 * (button that can be switched on and off, e.g. the "Bold" button). This option is supported since the **4.18.1** version.
 	 */
 	CKEDITOR.ui.prototype.addButton = function( name, definition ) {
 		this.add( name, CKEDITOR.UI_BUTTON, definition );
--- a/plugins/button/plugin.js
+++ b/plugins/button/plugin.js
@@ -342,7 +342,7 @@
 
 				if ( this.isToggle && !this.hasArrow ) {
 					// Note: aria-pressed attribute should not be added to menuButton instances. (https://dev.ckeditor.com/ticket/11331).
-					// Do not remove the attribute, set its value (#2444).
+					// For other buttons, do not remove the attribute, instead set its value (#2444).
 					element.setAttribute( 'aria-pressed', state === CKEDITOR.TRISTATE_ON );
 				} else if ( this.hasArrow ) {
 					// Indicates that menu button is opened (#421).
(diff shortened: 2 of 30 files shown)


plugins/button/plugin.js
```diff
@@ -51,7 +52,10 @@
 	( CKEDITOR.env.hc ? '&#9660;' : '' ) +
 		'</span>';
 
+	var templateToggle = ' aria-pressed="false"';
+
 	var btnArrowTpl = CKEDITOR.addTemplate( 'buttonArrow', templateArrow ),
+		toggleBtnTpl = CKEDITOR.addTemplate( 'toggleButton', templateToggle ),
```
jacekbogdanski, line 58: I'm not sure the toggle template is a good candidate for moving it into the templates store. Do you see any use case where it would make sense to customize this template using the "template" event?

plugins/button/plugin.js
```diff
@@ -337,14 +343,11 @@
...
-					if ( state === CKEDITOR.TRISTATE_ON ) {
-						element.setAttribute( 'aria-pressed', true );
-					} else {
-						element.removeAttribute( 'aria-pressed' );
-					}
-				} else {
+				if ( this.isToggle && !this.hasArrow ) {
+					// Note: aria-pressed attribute should not be added to menuButton instances. (https://dev.ckeditor.com/ticket/11331).
+					// Do not remove the attribute, set its value (#2444).
+					element.setAttribute( 'aria-pressed', state === CKEDITOR.TRISTATE_ON );
```
jacekbogdanski, line 349: I'm not sure if it's not introducing a breaking change. If someone is already using a button (3rd party plugin) without an arrow, it supports the `aria-pressed` feature. Maybe we should instead keep the previous logic for the non-defined `isToggle` value?

plugins/maximize/plugin.js
```diff
@@ -248,17 +248,6 @@
...
-					// Toggle button label.
-					var button = this.uiItems[ 0 ];
-					// Only try to change the button if it exists (https://dev.ckeditor.com/ticket/6166)
-					if ( button ) {
-						var label = ( this.state == CKEDITOR.TRISTATE_OFF ) ? lang.maximize.maximize : lang.maximize.minimize;
-						var buttonNode = CKEDITOR.document.getById( button._.id );
-						buttonNode.getChild( 1 ).setHtml( label );
-						buttonNode.setAttribute( 'title', label );
-						buttonNode.setAttribute( 'href', 'javascript:void("' + label + '");' ); // jshint ignore:line
-					}
```
jacekbogdanski, line 260: Aren't we losing some information about the second behavior of the button with that change? Previously, this button was nicely labeled with `minimize` once maximized. Now, it looks like this change favors screen readers only. Is it possible to keep both an aria-pressed attribute and a label (in the context of accessibility)?

plugins/button/plugin.js
jacekbogdanski: Uh, alright, that wouldn't make much sense as we will be telling that minimalize is active.

jacekbogdanski (commented):
TBC

plugins/button/plugin.js
Comandeer: I've not even know we have such an event :see\_no\_evil: I simply used a template here as all other button's parts were templates. But you're right, there's no need to have a template for the single attribute.

Comandeer:
I've removed the unnecessary template for the `[aria-pressed]` attribute.

plugins/maximize/plugin.js
jacekbogdanski: Yeah, I think that makes sense, thanks for explanation 👍🏻

plugins/button/plugin.js
jacekbogdanski: Looks like it was setting `aria-pressed` for `!this.hasArrow` check? Although I don't see it to be working correctly before your changes on VoiceOver despite having `aria-pressed="true"` for toggled button - not sure why.

tests/plugins/button/_helpers/buttontools.js
```diff
@@ -0,0 +1,90 @@
...
+		for ( var attrName in expectedAttributes ) {
+			assert.isTrue( buttonElement.hasAttribute( attrName ), 'Button HTML element does not contain ' + attrName + ' attribute.' );
+
+			expectedValue = expectedAttributes[ attrName ];
+			attributeValue = buttonElement.getAttribute( attrName );
+
+			if ( expectedValue instanceof RegExp )
+				assert.isTrue( expectedValue.test( attributeValue ), 'Attribute ' + attrName + ' did not matched expected ' + expectedValue + ' regex' );
+			else
+				assert.areSame( expectedValue, attributeValue, 'Invalid value for attribute ' + attrName + '.' );
```
jacekbogdanski, line 28: Could you correct this if/else to use brackets?

jacekbogdanski (changes requested):
These changes look good but I was only able to do manual tests using VoiceOver as my VM machines decided to not support OS sounds. @sculpt0r could you give it a try and confirm that manual test for screen readers works correctly on JAWS + Firefox and NVDA?

There is one issue but I'm not sure if it's related to your changes. Toggling source area forces the screen reader (VoiceOver) to read some garbage about being inside a table (for manual test with a table) or inside a list (for nightly sample). I'm not sure, but it looks like screen reader focus goes somewhere else after toggling a button?

Comandeer:
> There is one issue but I'm not sure if it's related to your changes. Toggling source area forces the screen reader (VoiceOver) to read some garbage about being inside a table (for manual test with a table) or inside a list (for nightly sample). I'm not sure, but it looks like screen reader focus goes somewhere else after toggling a button?

That's weird :thinking: The focus does not go anywhere, it stays on the button. The issue you describe appears only in Chrome, I can't reproduce it in Firefox or Safari. It looks like a browser bug.

plugins/button/plugin.js
Comandeer: Because toggle buttons are identified by the presence of the `[aria-pressed]` attribute. If the attribute is not present, it's considered a normal button. So adding that attribute only for the `true` value causes that the button is a toggle one only when it's pressed (in Chrome it isn't at all).

Comandeer:
I've fixed the reported codestyle issues.

tests/plugins/maximize/manual/togglebutton.md
```diff
@@ -0,0 +1,20 @@
+@bender-tags: 4.18.1, 2444, feature, button
+@bender-ui: collapsed
+@bender-ckeditor-plugins: wysiwygarea, toolbar, maximize
+
+1. Wait for editor to fully load.
+1. Check the info about the "Maximize" button that is rendered inside the editor
```
sculpt0r, line 6:
```suggestion
1. Check the info about the "Maximize" button that is rendered inside the editor.
```

tests/plugins/button/_helpers/buttontools.js
```diff
@@ -0,0 +1,94 @@
+window.buttonTools = {
+	// Standard aria attributes for button element.
+	typicalButtonAttributes: {
```
sculpt0r, line 3: Instead of the comment, we might rename the property  to:
```suggestion
	buttonAriaAttributes: {
```

or even something longer
```suggestion
	buttonStandardAriaAttributes: {
```

When I first read it - I wonder what are the `typical` attributes.

plugins/button/plugin.js
```diff
@@ -308,7 +310,8 @@
 				focusFn: focusFn,
 				clickFn: clickFn,
 				style: CKEDITOR.skin.getIconStyle( iconPath, ( editor.lang.dir == 'rtl' ), overridePath, this.iconOffset ),
-				arrowHtml: this.hasArrow ? btnArrowTpl.output() : ''
+				arrowHtml: this.hasArrow ? btnArrowTpl.output() : '',
+				toggleHtml: this.isToggle ? 'aria-pressed="false"' : ''
```
sculpt0r, line 314: Since it is part responsible only for arai attribute, then maybe:
```suggestion
				toggleAriaHtml: this.isToggle ? 'aria-pressed="false"' : ''
```

sculpt0r (changes requested):
@Comandeer nice work with the test tooling 👍 

JAWS & NVDA works fine 🎉  I have only doubts about naming and other small stuff in a few places - please take a look at the inline comments.

Additionally - I think we should add a test with the language button in the toolbar. It is not the plain toggle button, but the code around `hasArrow` was changed and it will be good to verify whenever this button still works as we expected - also with screen readers.

plugins/button/plugin.js
```diff
@@ -337,14 +340,11 @@
...
-				if ( !this.hasArrow ) {
-					// Note: aria-pressed attribute should not be added to menuButton instances. (https://dev.ckeditor.com/ticket/11331)
-					if ( state === CKEDITOR.TRISTATE_ON ) {
-						element.setAttribute( 'aria-pressed', true );
-					} else {
-						element.removeAttribute( 'aria-pressed' );
-					}
-				} else {
+				if ( this.isToggle && !this.hasArrow ) {
+					// Note: aria-pressed attribute should not be added to menuButton instances. (https://dev.ckeditor.com/ticket/11331).
```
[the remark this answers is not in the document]
Comandeer: Ok, I've rephrased the second comment, it should be clearer now.

Comandeer:
Ok, I've refactored names a bit and added a manual test for screen readers and menu button.

sculpt0r (approved):
LGTM 🎉 

Thank you for the separated test for the menu button. Thanks to that, I found and extracted an issue with `aria-expanded`: #5144 

Added a changelog.

sculpt0r:
@Comandeer - we have to wait for @jacekbogdanski approval - since he also reviewed this PR. Without it, GH is blocking me with the merge.

CHANGES.md
```diff
@@ -5,6 +5,14 @@ CKEditor 4 Changelog
 
 ## CKEditor 4.18.1 [IN DEVELOPMENT]
 
+New features:
+
+* [#2444](https://github.com/ckeditor/ckeditor4/issues/2444): Togglable toolbar buttons are now exposed as toggle buttons in the browser's accessibility tree.
```
jacekbogdanski, line 8: A feature changelog entry should go to a major release.

jacekbogdanski (approved):

Issue #4750, opened 2021-06-09
Add a notification to unsupported drop image type


KarolDawidziuk:
## Type of report

Feature request

## Provide description of the new feature

When the user drops an unsupported image type, nothing happens. Maybe let's add a notification with the information that the drop does not support the given type of files. WDYT?

Additionally I made a simple proof of concept.

https://user-images.githubusercontent.com/18471998/121433946-ef215980-c97c-11eb-9c7c-d459f4bb8f7b.mov

CKEditorBot:
Closed in https://github.com/ckeditor/ckeditor4/pull/4775

PR #5134, opened 2022-03-22
Fix hardcoded supportedImageTypes in clipboard plugin

### LINKED ISSUE #5095: Hard-coded supportedImageTypes variable in clipboard plugin.
## Hard-coded supportedImageTypes

I am using **uploadfile** plugin to upload documents like PDF, doc, etc..

But the editor is giving notification that "This File Format is not Supported".

ss

What can I do to close this notification ?

KarolDawidziuk:

Bug fix

### This PR contains

- [x] Unit tests
- [x] Manual tests

```
* [#5095](https://github.com/ckeditor/ckeditor4/issues/5095): Fix hardcoded supportedImageTypes in clipboard plugin.
```

## What changes did you make?

The problem was with displaying notifications when a custom file uploader tried to add an unsupported file type not supported by the editor by default.
The editor supports by default such MIME types as `image/png`, `image/jpeg`, `image/gif`. 

To solve this issue I add a new property called: `ignoredImageMimeTypes` and method `setIgnoredImageMimeType()`. 

The method is responsible for adding MIME types of image files, the notification of which we want to ignore. 

There was a problem with IE because this browser does not always return the MIME type of the currently added file. For example, the `image/webp` MIME type is returned as an empty string. 
To solve this I take the file type from its name. The solution is not perfect, however, it is the only one that I came up with at the moment. 

Closes #5095.


plugins/clipboard/plugin.js
```diff
@@ -163,10 +190,19 @@
...
+							defaultSupportedImageTypes = [ 'image/png', 'image/jpeg', 'image/gif' ];
+
+						// Check if the current file type is supported by the default supported types.
+						// We need to check only the image types as other file types require an additional file uploader. (#5095)
+						if ( !isSupportedImageMimeType( file, defaultSupportedImageTypes ) ) {
+							var supportedMimeTypes = CKEDITOR.plugins.clipboard.ignoredImageMimeTypes.concat( defaultSupportedImageTypes );
 
-						if ( CKEDITOR.tools.indexOf( supportedImageTypes, file.type ) === -1 ) {
-							editor.showNotification( unsupportedTypeMsg, 'info', editor.config.clipboard_notificationDuration );
+							if ( isImageType( file ) && !isSupportedImageMimeType( file, supportedMimeTypes ) ) {
```
sculpt0r, line 201: So now, the User sees the notification only for unsupported image files - unless the User tells to ignore them too.

I would add other use cases. The developer will have four options:
- the default one: so display the notification for all extensions that are not supported by default
- display the notification for extensions other than (defaults + added by the user). So the User can tell specifically that he wants to support additionally only certain formats (e.g. `pdf`) <-- This one could be added.
- display the notification for all non-supported images - unless the user excludes them. So the user tell that he will handle all non-image files without explicit enumerating them. Or specific currently non-supporting formats.
- don't display the notification at all - something will handle all non-default extensions <-- This one could be added

That might require some fun with cfg options, but if the user decides to handle all non-default image extensions - he might use some flag for it and don't have to define all possible image extensions. Is that right - or there is something that might break this logic (e.g. in other plugins or sth)?

sculpt0r (changes requested):
@KarolDawidziuk - could you take a look at the comment about new use cases? If that is possible - could you implement them?

KarolDawidziuk:
Thank you @sculpt0r for the review.

I've made the improvements that you mentioned above and it's ready for the next round of review.

plugins/clipboard/plugin.js
```diff
@@ -114,7 +114,36 @@
...
+			'image/jp2': jpeg2000Extensions,
+			'image/xbm': xBitmapExtension,
+			'image/xbitmap': xBitmapExtension,
+			'image/jxr': jpegXRExtensions,
+			'image/tiff-fx': tiffExtensions,
+			'image/tiff': tiffExtensions,
+			'image/avif': heifExtensions,
+			'image/heif': heifExtensions,
+			'image/heic': heifExtensions
+		};
```
Comandeer, line 146: I'm wondering if switching solely to file extensions wouldn't be simpler – we wouldn't need to keep such a map up to date.

The only real downside I could see is files without extensions or files with wrong extensions (e.g. text file inside `.gif` file).

plugins/clipboard/plugin.js
```diff
@@ -163,10 +190,43 @@
...
-						var file = dataTransfer.getFile( 0 );
+						var file = dataTransfer.getFile( 0 ),
+							defaultSupportedImageTypes = [ 'image/png', 'image/jpeg', 'image/gif' ],
+							isDisabledNotification = CKEDITOR.plugins.clipboard.ignoreUnsupportedMimeTypeNotification,
+							additionalSupportedMimeTypes = CKEDITOR.plugins.clipboard.supportedMimeTypes,
+							ignoredImageMimeTypes = CKEDITOR.plugins.clipboard.ignoredImageMimeTypes;
+
+						// Check if the current file type is supported by the default supported types.
+						// We need to check only the image types as other file types require an additional file uploader. (#5095)
+						if ( !isSupportedMimeType( file, defaultSupportedImageTypes ) ) {
```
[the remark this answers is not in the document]
sculpt0r: So in pseudo code I would do sth like:
```js
config.notificationDisabled = false;
config.nonImageHandled = false;

config.ignoredImages = [];
config.extraSupported = [];
config.supportsNonImages = false;

if( file isNot ['png' or 'gif' or 'jpg'] ) {
  if( notificationDisabled ) {
    return; // Don't show notifications at all
  }
  
  if ( file is config.extraSupported ) {
    // User decides to support that particular extension
    return;
  }
  
  var isImage = file.isImage(); // so image different than `png`, `jpg`, `gif`
  
  if( isImage && file is config.ignoredImages) {
    // User decides to ignores notification for that not supported image type
    return;
  }

  if( !isImage && config.supportsNonImages ) {
    // User decides to supports all non images
    return;
  }

 }
  
  // by default code reaches this line
  displayNotyfication();

}
```

To me, the above is far more readable and transparent for the use cases that verify if the file is image almost each time. The above probably requires some helper functions, like 'isOnTheList` or similar - but not the misleading `isSupported` 🤔 

Call me anytime if this won't work as I expected or you have some duobts.

sculpt0r (changes requested):
@KarolDawidziuk - thank you for the improvements and for adding new cases. I looked at this PR again and I have some more fundamental problems. Please take a look at the inline comments. They are about tests, configuration setup, and the verification logic. That probably erases half of this PR 😓 

If you have any doubts about such an approach - write/call anytime ;)

Additionally take a look at: https://github.com/ckeditor/ckeditor4/pull/5134/files#r841528656 I  also think that we could use the simple extensions list/array.

sculpt0r:
Also - rebased on the newest master.

KarolDawidziuk:
Thank you @sculpt0r for the review. 😉 

I applied improvements from your comments and our F2F talks and this PR is ready for another review.

KarolDawidziuk:
Thank you @sculpt0r for the review. 😉 

I applied improvements from your comments and our F2F talks and this PR is ready for another review.

sculpt0r (changes requested):
@KarolDawidziuk - thank you for all the changes and f2f experimental review :)

I've made some corrects to this PR:
- https://github.com/ckeditor/ckeditor4/pull/5134/commits/eb6e0a27668f55f6457da28a9b978f9fe1b4943a: lets do not use shorthand versions in the documentation.
- https://github.com/ckeditor/ckeditor4/pull/5134/commits/9e5f3d881acff30131584f9dd981ae74805a9aea: we should define the default values explicitly instead of interpreting lack of config (`undefined`) as some default value. Whenever the User reads the docs and notices that some option is an array - the User might try to `push` something there. And he will get an unexpected error.
- https://github.com/ckeditor/ckeditor4/pull/5134/commits/769c2820b49d3ff3dabaec429da845967552a4c8: fix for the early return which you spotted
- https://github.com/ckeditor/ckeditor4/pull/5134/commits/98ccb78cef909bd7b95a94d79a0a5de61a18f58f, https://github.com/ckeditor/ckeditor4/pull/5134/commits/a149db6b5732199398cc4d49b4509f948b348d76, https://github.com/ckeditor/ckeditor4/pull/5134/commits/72c55fff0099f399e026189b94e8016791e733d4 I've made some corrects to one manual test. Lets do not use shorthand variables like `btn` instead of `button`. Also, added a `notice` - because it is easy to do the test in a hurry and easily gets confused whenever the notification is from the previous step or not. Also, rephrased 'notification is gone'. It sounds like the button should hide it. The last one: I splitted the last step - so it is more obvious how te test should be performed. 

Could you do the similar corrects for all other manual test? Also - please take a look at the others inline comments.

tests/plugins/clipboard/manual/dndnotificationoff.html
```diff
@@ -0,0 +1,32 @@
...
+		} );
+
+	editor.on( 'loaded', function() {
+		if( !CKEDITOR.plugins.clipboard.isFileApiSupported || bender.tools.env.mobile ){
+			return bender.ignore();
+		}
+
+		var btn = document.getElementById( 'turnNotifyOff' );
+		btn.addEventListener( 'click', function() {
+			CKEDITOR.plugins.clipboard.ignoreUnsupportedMimeTypeNotification = true;
```
sculpt0r, line 29: The same as in https://github.com/ckeditor/ckeditor4/pull/5134/files/72c55fff0099f399e026189b94e8016791e733d4#r854003585. leftover and probably duplicated test.

tests/plugins/clipboard/manual/dndcustomimagetypenotification.html
```diff
@@ -0,0 +1,32 @@
...
+		} );
+
+	editor.on( 'loaded', function() {
+		if( !CKEDITOR.plugins.clipboard.isFileApiSupported || bender.tools.env.mobile ){
+			return bender.ignore();
+		}
+
+		var btn = document.getElementById( 'addIgnoredType' );
+		btn.addEventListener( 'click', function() {
+			CKEDITOR.plugins.clipboard.setIgnoredImageMimeType( [ 'image/jp2' ] );
```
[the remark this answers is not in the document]
sculpt0r: Also: https://github.com/ckeditor/ckeditor4/pull/5134/files/72c55fff0099f399e026189b94e8016791e733d4#r854011496

KarolDawidziuk:
Thanks, @sculpt0r for the review and corrections.

I've removed the old manual tests and corrected the rest of the tests, and the PR is ready for the next review.

sculpt0r (approved):
LGTM 🎉  Thank you @KarolDawidziuk for the solution :)

Now it looks good, also the tests looks good :)

Now it's @jacekbogdanski turn :)

jacekbogdanski (changes requested):
I have some small doubts about the number of configuration options we are exposing with these new changes and if they are indeed fixing a regression introduced in 4.17.0.

I'm not totally confident why notification about unsupported file type has been introduced in 4.17.0 but it certainly breaks the upload file plugin. We started detecting if a file extension is of `png`, `jpeg` or `gif` and show notifications about unsupported file types if it's not. It looks like a very aggressive change looking at possible 3rd party integrations already incorporating d&d into the editor. So, I would assume that the proper fix for this issue, should:

1. Revert changes that show an unnecessary notification if a file type is not supported in an opt-out manner, so we are actually reverting a regression.
2. Provide an option to limit supported types opt-in by a separate configuration option defaulting to nil.
```js
config.clipboard_supportedFileTypes = [ 'png', 'jpg', 'gif', 'pdf' ]; // null by default disabling notification
config.clipboard_handleImages = true // it should still work as previous except showing notification
```

So, from the code perspective:
```mermaid
flowchart TD
A[supportedFileTypes] -- not defined --> B[ignore that option and allow for drop];
A[supportedFileTypes] -- defined --> C[check supported file type];
C -- exists --> B;
C -- is missing --> E[prevent drop];
B -- clipboard_handleImages == true --> F[use clipboard for drop if it's of the valid image type]
```

Do you see any cons of that approach, or do I missing something here? I didn't work on CKEditor 4 when this logic has been introduced so I'm not 100% if I understand fully what happened here, so please correct me if I'm wrong.

sculpt0r:
> I'm not totally confident why notification about unsupported file type has been introduced in 4.17.0 but it certainly breaks the upload file plugin. 

https://github.com/ckeditor/ckeditor4/issues/4750

Comandeer:
> I'm not totally confident why notification about unsupported file type has been introduced in 4.17.0 \[…\]

It was introduced as a UX enhancement. Without it, the whole d&d functionality looked broken (user drops unsupported image type and nothing happens).

IMO introducing `config.clipboard_supportedFileTypes` would not solve the issue as its name suggests it's responsible for a completely different thing: setting the list of types supported by the clipboard plugin.

I'm for introducing something like `config.clipboard_notificationEnabled` (or `config.clipboard_handleImageNotification`) to complement `config.clipboard_notificationDuration` but I agree that there's no need to introduce several, "atomic" config variables for each case. One option for these notifications could be enough.

Or… there's probably one more solution that would require writing some docs but won't require introducing anything to the code. We have the [`notificationShow` event](https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_editor.html#event-notificationShow) that can be used to cancel any notification. But that's basically turning a workaround into a feature :see\_no\_evil:

KarolDawidziuk:
> I'm not totally confident why notification about unsupported file type has been introduced in `4.17.0`

It was introduced after we allow drag&drop images into the editor in #4750.
Previously the editor did not show any message in the case when someone tried to add an unsupported file and notification was supposed to help in this. I assume that then we didn't check it out with a 3rd party integrations causing regression. 

`clipboard_supportedFileTypes` and `clipboard_handleImages` 
Sounds like these options will allow support of some fileTypes or images but we just want to hide notifications for them. 
This of course could be supported by some 3rd party plugin but here is not.

IMO looks more flexible in its current version. It is now set up so that no additional changes are needed, but can be extended it if we want.

If I'm not wrong. In your diagram, there is no notification 🤔  So, overall we're going back to the beginning and this fix is not needed as we want to get rid of notifications.

sculpt0r:
If I good understand: `config.clipboard_handleImages = true` this would be the default value for the editor. So pasting `png`, `jpg` and `gif` as base64 will be optional (but enabled by default)?

To fix the regression we only need to disable notifications, but that means we back to the issue reported in #4750 

With the single config option (that requires to define each supported extension): it should work - at least in some use cases. 

I have some doubts about such use cases:
1. The user wants to handle all extensions (beside the graphics that are pasted as base64) - does it means he have to explicit write down all extensions?
2. The user wants to handle all dropped files by himself (even those we are currently supports as base64)
3. The user wants to handle graphics as base64 - so they are pasted into the content, upload `pdf` & `zip` (aka attachments) and for the rest of them - display notification.

IMO, we have a few problems to deal with:
- what supported files means: does it mean we are supports them in clipboard plugin, or some plugin will handle such files?
- we probably wants to separate the files supported by clipboard from the other files that might be supported by some other plugin
- finally we might assume, that someone would like to also handle by himself all files (even those that are pasted as base64).

I believe that the set of configs we provide - it is the most flexible way - that allows users to setup the editor as he wish. Also, the default settings (so the user don't have to touch anything) remains the editor in the default state. And the last one: it is possible to setup the editor on "your way" with only one of those config options. You don't need to combine them to achieve on of the possible states.

jacekbogdanski:
As we discussed F2F with @KarolDawidziuk @sculpt0r @Comandeer about my concerns, we end up with a slightly different approach than proposed in this PR that will let us:

* Remove regression introduced in 4.17.0
* Keep current notifications for the clipboard if the editor has not been integrated with the upload widget plugin (and other upload plugins)
* Extend clipboard plugin API allowing to decide if file type can be supported by the editor which can, later on, be used by 3rd party integrators

A feature will be developed on a separate, https://github.com/ckeditor/ckeditor4/tree/t/5095b branch. Let's give it a try and see if a new approach will allow us to resign from these changes.

PR #5140, opened 2022-03-29
Fix cell selection and navigation by TAB key after adding a new cell

### LINKED ISSUE #4904: Table cell selection and navigation by TAB key behaviour inconsistent after adding a new row
## Type of report

Bug

1. Create table with default parameters and fill each cell with some text
2. Right click on any cell and add a new row or column
3. Observe that the cell in which we clicked is selected
4. Press TAB

Please see the attached gif:

error

### Expected result

On pressing TAB selection should navigate to the next cell as it does when manually selecting a cell by Triple Clicking in it.

Behaviour when Triple Clicking on any cell and then pressing TAB key:

expected

### Actual result

Editor looses focus and the cell selection remains the same.

## Other details

* Browser: Google Chrome Version 94.0.4606.54 (Official Build) (64-bit)
* OS: Linux Mint 19.1
* CKEditor version: 4.14.1 - 4.17.0 (Standard)
* Installed CKEditor plugins: Standard Package

KarolDawidziuk:

Bug fix

### This PR contains

- [ ] Unit tests
- [x] Manual tests

```
* [#4904](https://github.com/ckeditor/ckeditor4/issues/4904): Fix cell selection and navigation by TAB key after adding a new cell.
```

## What changes did you make?

There was a problem with getting empty `td` element from selection by `new CKEDITOR.dom.elementPath()` method.

When the cell was empty, `CKEDITOR.dom.elementPath` didn't add `td` element to path elements, which caused in bypassing of all logic responsible for shifting focus and selection to the next cell.

The fix was pretty simple. 
I just change `CKEDITOR.dom.elementPath()` to `editor.elementsPath()` and removed the piece of code responsible for avoiding selecting empty cells.

Closes #4904.


plugins/tab/plugin.js
```diff
@@ -65,9 +64,7 @@
 						} else if ( next ) {
 							next = new CKEDITOR.dom.element( next );
 							resultRange.moveToElementEditStart( next );
-							// Avoid selecting empty block makes the cursor blind.
-							if ( !( resultRange.checkStartOfBlock() && resultRange.checkEndOfBlock() ) )
```
Comandeer, line 69: Is this change necessary? If it's added only to force our fake selection after pressing the <kbd>Tab</kbd> key, I'd remove it. It causes issues in Firefox and IE11 in editors without the `tableselection` plugin. After pressing the <kbd>Tab</kbd> key, the cursor disappears.

tests/plugins/tab/errorinlineeditor.js
```diff
@@ -15,8 +15,6 @@
 			var editor = this.editor;
 
 			editor.focus();
-			var sel = editor.getSelection();
-			sel.removeAllRanges();
```
Comandeer, line 19: This test does not seem to test anything sensible after this change – it basically just presses the <kbd>Tab</kbd> key now. Additionally, the unmodified version of the test raises the very same error described in [https://dev.ckeditor.com/ticket/17028](https://dev.ckeditor.com/ticket/17028) – which is the issue tested here. Could you check if the issue reappeared after your changes and think if you can come up with some sensible way to test it?

Comandeer (changes requested):
The proposed fix resolves the issue, however it causes some problems in Firefox and IE11 (see my inline comments).

Additionally, some of your changes touch also editors without the `tableselection` plugin – could you prepare additional tests for such cases? Also, could you add unit tests?

plugins/tab/plugin.js
KarolDawidziuk: This change is not necessary. It was just my assumption with false selection. I did not realize that this was causing additional problems. So I will bring back the previous changes.

KarolDawidziuk:
Thank you @Comandeer for the review.

I applied the improvements mentioned in your comments and PR is ready for another review.

Also, I found that there was a problem with getting proper `elementPath` when `startNode` was getting from `selection.getCommonAncestor()` I changed it for `selection.getStartElement()` and everything works well.

tests/plugins/tableselection/tablecellfocus.js
```diff
@@ -0,0 +1,286 @@
+/* bender-tags: tableselection */
```
Comandeer, line 1: Tests in this file seem to pass also on the `master`. Could you check it?

tests/plugins/tableselection/tablecellfocus.js
```diff
@@ -0,0 +1,286 @@
...
+				editable = editor.editable();
+
+			bot.setHtmlWithSelection( '<table border="1">' +
+				'<tbody>' +
+					'<tr>' +
+						'<td>1_1</td>' +
+						'<td>1_2</td>' +
+					'</tr>' +
+					'<tr>' +
+						'<td>success!^</td>' +
```
Comandeer, line 259: Probably it could be caused by the fact that you use a collapsed selection which is probably translated to the native one, not our `tableselection` one. To activate the latter you'd need to select the whole cell.

Comandeer (changes requested):
The fix looks good, but there is an issue with the unit test (see my inline comment). Additionally, could you add tests for native selection? The change you introduced is also affecting the native one as well.

tests/plugins/tableselection/tablecellfocus.js
KarolDawidziuk: Fixed.

tests/plugins/tableselection/tablecellfocus.js
KarolDawidziuk: The collapsed selection was not causing the problem because there was a problem with moving focus, not operations on the table. Already fixed.

Comandeer (changes requested):
As we spoke yesterday, could you split the unit tests and place all tests in the appropriate directory (`tab` plugin one).

KarolDawidziuk:
Thank you @Comandeer for the review.

I split and moved tests to the `tab` plugin and PR is ready for another review.

Comandeer (changes requested):
Everything looks fine. However, there is some issue with `tests/plugins/tab/tablefocus` test in IE8-10 – the first test always passes and there is info about some uncaught exception (probably it's connected with non-functioning `bender.tools.ignoreUnsupportedEnvironment()`). Could you take a look?

KarolDawidziuk:
> Everything looks fine. However, there is some issue with tests/plugins/tab/tablefocus test in IE8-10 – the first test always passes and there is info about some uncaught exception (probably it's connected with non-functioning bender.tools.ignoreUnsupportedEnvironment()). Could you take a look?

The reason for this was simple. I forgot to add the `tableselection` plugin in the test header. Everything looks fine after that.

Comandeer (approved):
LGTM!

PR #5141, opened 2022-03-29
Fix editor failing in strict mode

### LINKED ISSUE #5049: Make CKEditor 4 works in strict mode
## Type of report

Bug

1. Simply add `"use strict";` at the beginning of the compiled ckeditor.js

### Expected result

It should work

### Actual result

An exception is thrown:
> TypeError: this is undefined
>     getClientRects http://localhost:8080/ckeditor.js:12586
>     &lt;anonymous> http://localhost:8080/ckeditor.js:12805
>     &lt;anonymous> http://localhost:8080/ckeditor.js:12934
>     &lt;anonymous> http://localhost:8080/ckeditor.js:64500

## Other details

The error is in the getClientRects function of the range.js source, where `this` is undefined in strict mode:
```
getClientRects: ( function() {
	if ( this.document.getSelection !== undefined ) {
		return function( isAbsolute ) {
			// We need to create native range so we can call native getClientRects.
			var range = this.root.getDocument().$.createRange(),
				rectList;

			range.setStart( this.startContainer.$, this.startOffset );
			range.setEnd( this.endContainer.$, this.endOffset );

			rectList = range.getClientRects();

			rectList = fixWidgetsRects( rectList, this );

			if ( !rectList.length ) {
				rectList = fixEmptyRectList( rectList, range, this );
			}

			return CKEDITOR.tools.array.map( rectList, function( item ) {
				return convertRect( item, isAbsolute, this );
			}, this );
		};
	} else {
		return function( isAbsolute ) {
			return [ convertRect( getRect( this.createBookmark() ), isAbsolute, this ) ];
		};
	}
```

There is probably other bugs to fix in strict mode

* Browser: Firefox 96
* OS: Windows 10
* CKEditor version: 4.17.1
* Installed CKEditor plugins: Full preset

sculpt0r:

Bug fix

### This PR contains

- [ ] Unit tests
- [X] Manual tests

```
* [#5049](https://github.com/ckeditor/ckeditor4/issues/5049): Fix editor fails in strict mode.
```

## What changes did you make?

*Give an overview…*

Removed `this` from IIFE in getClientRects. Without strict mode, it refers to a global `window`. In strict mode it is `undefined` - but actually, we need the global `document` in this place - so the problematic `this` could be removed.

Closes #5049 .

### DIFF
--- a/core/dom/range.js
+++ b/core/dom/range.js
@@ -3,6 +3,8 @@
  * For licensing, see LICENSE.md or https://ckeditor.com/legal/ckeditor-oss-license
  */
 
+'use strict';
+
 /**
  * Represents a delimited piece of content in a DOM Document.
  * It is contiguous in the sense that it can be characterized as selecting all
@@ -176,7 +178,6 @@ CKEDITOR.dom.range = function( root ) {
 	//   param (see range.deleteContents or range.extractContents).
 	//   See comments in mergeAndUpdate because this is lots of fun too.
 	function execContentsAction( range, action, docFrag, mergeThen, cloneId ) {
-		'use strict';
 
 		range.optimizeBookmark();
 
@@ -2910,7 +2911,6 @@ CKEDITOR.dom.range = function( root ) {
 		 * @returns {CKEDITOR.dom.rect[]}
 		 */
 		getClientRects: ( function() {
-			'use strict';
 			if ( document.getSelection !== undefined ) {
 				return function( isAbsolute ) {
 					// We need to create native range so we can call native getClientRects.
--- a/core/dom/range.js
+++ b/core/dom/range.js
@@ -2910,6 +2910,7 @@ CKEDITOR.dom.range = function( root ) {
 		 * @returns {CKEDITOR.dom.rect[]}
 		 */
 		getClientRects: ( function() {
+			'use strict';
 			if ( document.getSelection !== undefined ) {
 				return function( isAbsolute ) {
 					// We need to create native range so we can call native getClientRects.
--- a/core/dom/range.js
+++ b/core/dom/range.js
@@ -2910,7 +2910,7 @@ CKEDITOR.dom.range = function( root ) {
 		 * @returns {CKEDITOR.dom.rect[]}
 		 */
 		getClientRects: ( function() {
-			if ( this.document.getSelection !== undefined ) {
+			if ( document.getSelection !== undefined ) {
 				return function( isAbsolute ) {
 					// We need to create native range so we can call native getClientRects.
 					var range = this.root.getDocument().$.createRange(),
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -9,7 +9,7 @@ New features:
 
 Fixed Issues:
 
-* [#4543](https://github.com/ckeditor/ckeditor4/issues/4543): Toolbar buttons toggle state is not correctly announced by screen readers.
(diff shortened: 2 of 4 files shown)


core/dom/range.js
```diff
@@ -2910,7 +2910,8 @@ CKEDITOR.dom.range = function( root ) {
 		 * @returns {CKEDITOR.dom.rect[]}
 		 */
 		getClientRects: ( function() {
-			if ( this.document.getSelection !== undefined ) {
+			'use strict';
```
Comandeer, line 2913: If we add `'use strict'` pragma, we should probably add it to the whole file, not this function only.

Comandeer (changes requested):
The fix itself looks good.

However, I'm wondering if it isn't also a good moment to add `'use strict'` pragma to files that lack it. Most of the source code in CKE4 already has it but there are some files that are still in sloppy mode. WDYT about adding pragma there?

Maybe we could also add some more generic strict mode test for the built version of the editor. However, I'm not sure how to force the strict mode for the whole code 🤔.

core/dom/range.js
sculpt0r: I tried it, but... the `jshint` has some objectives 🤔 
image

sculpt0r:
> Maybe we could also add some more generic strict mode test for the built version of the editor. However, I'm not sure how to force the strict mode for the whole code 🤔.

It is easy to do it manually for the complete build - but in development mode - it would require adding it file by file 🤔

sculpt0r:
> Most of the source code in CKE4 already has it but there are some files that are still in sloppy mode. WDYT about adding pragma there?

I would say that until we encourage further issues we should wait with such change 🤔

Comandeer:
> I would say that until we encourage further issues we should wait with such change :thinking:

SGTM.

> It is easy to do it manually for the complete build - but in development mode - it would require adding it file by file :thinking:

Yes, it is easy to manually alter the built version. However, it sounds like a really hacky way to create a test. I don't have any sensible idea how to actually automate it – without altering Bender.js, of course.

core/dom/range.js
Comandeer: Weird… 🤔  But I see that there is one pragma inside that file, inside the `execContentsAction()` function. Maybe `jshint` incorrectly interprets it as a pragma for the whole file?

Alternatively, we could just skip adding that pragma at all and just add to come up with some sensible way of testing the editor in the strict mode – yet, it seems quite cumbersome (https://github.com/ckeditor/ckeditor4/pull/5141#issuecomment-1094931387).

sculpt0r: > But I see that there is one pragma inside that file, inside the execContentsAction() function. Maybe jshint incorrectly interprets it as a pragma for the whole file?

Looks like you are right 😄 

So I removed the 'in functions' strict mode declarations & added one for the entire file.

The manual test seems to be still valid (fails without a fix, and passes with fix).

sculpt0r: But I needed to adjust it - since the error is visible in the console right away - even if we postpone editor creation.

sculpt0r:
@Comandeer - I'm not sure, but maybe we could introduce 'strict' mode to the builder? So the final `ckeditor.js` will be altered with 'use strict'? However, that also sounds like a little overkill 🤔

sculpt0r:
@Comandeer - after our discussion with @jacekbogdanski - I kept this PR in the current shape. A manual test is preserved for the sake of the testing phase - proper steps for testing strict mode are described in our internal documentation.

Comandeer:
Rebased onto latest `master`.

Comandeer (approved):
LGTM!

PR #5146, opened 2022-04-03
Tweaks for representation of the editor in the accessibility tree

### LINKED ISSUE #1904: Readonly editors incorrectly marked up for Assistive Technology

Bug

1. Open example of readonly editor, e.g. https://jsfiddle.net/sjg91czp/

### Expected result

Editor's editable has `[aria-readonly=true]` attribute attached.

### Actual result

There is no such attribute.

## Other details

* CKEditor version: 4.9.1

### LINKED ISSUE #2445: Add an option to customize editor's application region label
## Type of report

Bug

1. Open `http://ckeditor.test/ckeditor-dev/samples/`.
2. Navigate into the editor.
3. Listen to JAWS announcing the editor.

### Expected result

The name of the editor is announced once.

### Actual result

The name of the editor is announced twice.

## Other details

* JAWS: 18

Iframe-based editor has its label in 3 different places:

* `[title]` attribute;
* `title` element;
* dedicated element before the editable.

Not surprising that the name's announced twice…

### LINKED ISSUE #2499: No indication given to the user that a WYSIWYG field already contains text
## Type of report

Bug

1    Create a node that has that has a ckeditor field.
2    Open JAWS v18.0.2945.
3    Edit that node.
4    Use the Tab key to navigate to the ckeditor field.

### Expected result

All textual information is expected to be announced. 

### Actual result

When editing a entity that already has content contained in it and a user tabs to a WYSIWYG field, JAWS users in chrome receive no indication that there is already content in the field.

• Testing was performed using JAWS v18.0.2945.

Notes browser behavior for text fields:
- Regular text field behavior: In both Chrome and IE, JAWS will read the text automatically when you tab to a regular text field
- Textarea field behavior: In both IE and Chrome, JAWS will announce that field has text, but will not read it.

WYSIWYG Field behavior:
-    In Chrome, JAWS reads nothing and does not indicate that there is text in the field at all. A user needs to use the insert+down arrow command to read the text. (This shortcut is not announced to the user in any way. This shortcut was discovered during testing using JAWS).
-    In IE it will auto read the WYSIWYG text, but only the first line of the field. In order for it to read the remaining text you will need to use the keyboard insert+down. (This shortcut is not announced to the user in any way. This shortcut was discovered during testing using JAWS)

## Other details

Originally reported here https://www.drupal.org/project/drupal/issues/3002056

* Browser: Chrome

### LINKED ISSUE #4052: Accessibility: Editor Control Type Incorrect

Comandeer:

Bug fix

### This PR contains

- [x] Unit tests
- [x] Manual tests

```
Fixed Issues:

* [#4052](https://github.com/ckeditor/ckeditor4/issues/4052): Fixed: Editor labels are read incorrectly by screen readers due to invalid editor control type for [Iframe Editing Area](https://ckeditor.com/cke4/addon/wysiwygarea) editors.
* [#1904](https://github.com/ckeditor/ckeditor4/issues/1904): Fixed: Screen readers are not announcing read-only editor state due to missing proper aria-labels.
* [#2445](https://github.com/ckeditor/ckeditor4/issues/2445): Fixed: The same editor's label is read several times by screen readers when focusing an editor.

API Changes:

* [#2445](https://github.com/ckeditor/ckeditor4/issues/2445): Added [`config.applicationTitle`](https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_config.html#cfg-applicationTitle) alongside [`CKEDITOR.editor#applicationTitle`](https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_editor.html#property-title) to allow customizing editor's application region label.
```

## What changes did you make?

I've added missing `[role=textbox]` and `[aria-multiline]` attributes for `wysiwygarea` editors. I've also added `[aria-readonly]` attribute to editors to indicate the status of the "readonliness" of the editor. However, for iframe-based editors it required adding also the `[tabindex=0]` attribute to be actually read by screen readers. This change was implemented for all browsers except Firefox which has a [bug connected with this attribute and `iframe`s](https://bugzilla.mozilla.org/show_bug.cgi?id=1483828). It also means that marking read-only `iframe`s does not work in Firefox.

I didn't touch #2445 because I have no idea how to differentiate labels for applications and editables. But the changes introduced for read-only editors indicate such a need (NVDA now reads the same label _thrice_ – probably due to it being the label of the application, the label of the editor, and the title of the editor's `iframe`). Probably it's something that we need to discuss further.

Proposed changes should also fix #2499 but it needs further verification.

Closes #4052.
Closes #1904.
Closes #2445.

### DIFF
--- a/core/editor.js
+++ b/core/editor.js
@@ -1895,8 +1895,8 @@ CKEDITOR.ELEMENT_MODE_INLINE = 3;
 
 /**
  * Customizes the {@link CKEDITOR.editor#applicationTtle human-readable title} of the application for this
- * editor. This title is used as a label for the whole website's region containing editor with its toolbars and other
- * UI parts. This title impacts various
+ * editor. This title is used as a label for the whole website's region containing the editor with its toolbars and other
+ * UI parts. Application title impacts various
  * [accessibility aspects](#!/guide/dev_a11y-section-announcing-the-editor-on-the-page),
  * e.g. it is commonly used by screen readers for distinguishing editor instances and for navigation.
  * Accepted values are a string or `false`.
--- a/core/editor.js
+++ b/core/editor.js
@@ -424,7 +424,8 @@
 
 	function loadLang( editor ) {
 		CKEDITOR.lang.load( editor.config.language, editor.config.defaultLanguage, function( languageCode, lang ) {
-			var configTitle = editor.config.title;
+			var configTitle = editor.config.title,
+				configApplicationTitle = editor.config.applicationTitle;
 
 			/**
 			 * The code for the language resources that have been loaded
@@ -462,6 +463,21 @@
 			 * @property {String/Boolean}
 			 */
 			editor.title = typeof configTitle == 'string' || configTitle === false ? configTitle : [ editor.lang.editor, editor.name ].join( ', ' );
+
+			/**
+			 * Indicates the human-readable title of this editor's application (the website's region
+			 * that contains the editor and its whole UI). Although this is a read-only property,
+			 * it can be initialized with {@link CKEDITOR.config#applicationTitle}.
+			 *
+			 * **Note:** Please do not confuse this property with {@link CKEDITOR.editor#name editor.name}
+			 * which identifies the instance in the {@link CKEDITOR#instances} literal.
+			 *
+			 * @since 4.19.0
+			 * @readonly
+			 * @property {String/Boolean}
+			 */
+			editor.applicationTitle = typeof configApplicationTitle == 'string' || configApplicationTitle === false ?
+				configApplicationTitle : [ editor.lang.application, editor.name ].join( ', ' );
 
 			if ( !editor.config.contentsLangDirection ) {
 				// Fallback to either the editable element direction or editor UI direction depending on creators.
@@ -1867,12 +1883,50 @@ CKEDITOR.ELEMENT_MODE_INLINE = 3;
  *		config.title = false;
  *
  * See also:
- *
- * * CKEDITOR.editor#name
  * * CKEDITOR.editor#title
+ * * CKEDITOR.editor#name
+ * * CKEDITOR.editor#applicationTitle
+ * * CKEDITOR.config#applicationTitle
  *
  * @since 4.2.0
  * @cfg {String/Boolean} [title=based on editor.name]
+ * @member CKEDITOR.config
+ */
+
+/**
+ * Customizes the {@link CKEDITOR.editor#applicationTtle human-readable title} of the application for this
+ * editor. This title is used as a label for the whole website's region containing editor with its toolbars and other
+ * UI parts. This title impacts various
+ * [accessibility aspects](#!/guide/dev_a11y-section-announcing-the-editor-on-the-page),
+ * e.g. it is commonly used by screen readers for distinguishing editor instances and for navigation.
+ * Accepted values are a string or `false`.
+ *
+ * **Note:** When `config.applicationTitle` is set globally, the same value will be applied to all editor instances
+ * loaded with this config. This may adversely affect accessibility as screen reader users will be unable
+ * to distinguish particular editor instances and navigate between them.
+ *
+ * **Note:** Setting `config.title = false` may also impair accessibility in a similar way.
+ *
+ * **Note:** Please do not confuse this property with {@link CKEDITOR.editor#name}
+ * which identifies the instance in the {@link CKEDITOR#instances} literal.
+ *
+ *		// Sets the application title to 'My WYSIWYG'. The original title of the element (if it exists)
+ *		// will be restored once the editor instance is destroyed.
+ *		config.title = 'My WYSIWYG';
+ *
+ *		// Do not touch the title. If the element already has a title, it remains unchanged.
+ *		// Also if no `title` attribute exists, nothing new will be added.
+ *		config.title = false;
+ *
+ * See also:
+ *
+ * * CKEDITOR.editor#applicationTitle
+ * * CKEDITOR.editor#name
+ * * CKEDITOR.editor#title
+ * * CKEDITOR.config#title
+ *
+ * @since 4.19.0
+ * @cfg {String/Boolean} [applicationTitle=based on editor.name]
  * @member CKEDITOR.config
  */
 
--- a/core/editor.js
+++ b/core/editor.js
@@ -1905,18 +1905,16 @@ CKEDITOR.ELEMENT_MODE_INLINE = 3;
  * loaded with this config. This may adversely affect accessibility as screen reader users will be unable
  * to distinguish particular editor instances and navigate between them.
  *
- * **Note:** Setting `config.title = false` may also impair accessibility in a similar way.
+ * **Note:** Setting `config.applicationTitle = false` may also impair accessibility in a similar way.
  *
  * **Note:** Please do not confuse this property with {@link CKEDITOR.editor#name}
  * which identifies the instance in the {@link CKEDITOR#instances} literal.
  *
- *		// Sets the application title to 'My WYSIWYG'. The original title of the element (if it exists)
- *		// will be restored once the editor instance is destroyed.
- *		config.title = 'My WYSIWYG';
+ *		// Sets the application title to 'My WYSIWYG'.
+ *		config.applicationTitle = 'My WYSIWYG';
  *
- *		// Do not touch the title. If the element already has a title, it remains unchanged.
- *		// Also if no `title` attribute exists, nothing new will be added.
- *		config.title = false;
+ *		// Do not add the application title.
+ *		config.applicationTitle = false;
  *
  * See also:
  *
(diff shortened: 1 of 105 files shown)


jacekbogdanski (commented):
## NVDA

First thing is that I'm not actually able to reproduce https://github.com/ckeditor/ckeditor4/issues/4052
> Control type is not informational (I believe it is announced as 'group')

I've downloaded the latest version of NVDA, and it seems to properly read the editor labels (putting aside fact that the editor labels seem to be read a bit redundantly). So it's hard to tell me how these changes impacted reading labels as I don't see that many differences, except of course read-only support. @Comandeer did you manage to reproduce the original issue?

Currently, NVDA will read focused editor as:

Rich Text Editor, [editor name]
Rich Text Editor, [editor name] Frame
Rich Text Editor, [editor name] Document
Edit Multi Line [text]

In the case of a11yhelp plugin:

Rich Text Editor, [editor name]
Rich Text Editor, [editor name] Frame, press Alt + 0 for help
Rich Text Editor, [editor name] Document
Edit Multi Line [text]

I'm wondering if that much information is really needed. I'm not an accessibility expert, but wouldn't be more helpful to just read something like:

Rich Text Editor, [editor name]
Editing Area
[text]

In the case of a11yhelp plugin:
Rich Text Editor, [editor name], press Alt + 0 for help
Editing Area
[text]

It seems more helpful to skip that much information for me but maybe missing some important cases here. @Comandeer could you help me to understand why we are giving so much information here and duplicating the message?

I suppose that it may be related to switching between another focusable element (like toolbar) in the editor, but in that case, reading Editing Area would still give the user feedback on what's happening.

## VoiceOver
I didn't notice any difference except for different reading for read-only (which I suppose is good). It seems that VoiceOver doesn't care about reading Iframe titles at all.

## JAWS
Running JAWS resulted in freezing my VM, so I'm going to get back to it soon.

jacekbogdanski:
To sum up our discussion with @Comandeer F2F, we will try to come up with a simplified version of these labels:

Rich Text Editor, [editor name]
Editor [editor name] Frame, press Alt + 0 for help
Editing Area (let's check if this label can be skipped without making the editor less accessible)
[text]

It should reduce repetitive text while keeping a good level of accessibility for the different parts of the editor. Note that labels may be slightly different depending on results from another code iteration.

Comandeer:
Labels were far more difficult than I expected. I had to introduce `editor#applicationTitle` property alongside `CKEDITOR.config#applicationTitle` variable (as counterparts to `editor#title` and `CKEDITOR.config#title`). Additionally, I had to… add yet another label (`[aria-label]` attribute to the iframe's `body` for VoiceOver to correctly read the editor).

So there are two distinctive labels:

1.  application one ("Rich Text Editor, \<name>"),
2.  editor one ("Editor, \<name>").

The second one is used on editables and `iframe`s. Unfortunately, this label is repeated twice in NVDA and JAWS (once for the frame itself and once for the document inside it) but it's probably the best approach.

tests/core/editable/manual/ariareadonlyscreenreader.md
```diff
@@ -0,0 +1,16 @@
+@bender-tags: 4.18.1, bug, 1904
```
jacekbogdanski, line 1:
```suggestion
@bender-tags: 4.19.0, bug, 1904
```

tests/plugins/wysiwygarea/manual/ariareadonlyscreenreader.md
```diff
@@ -0,0 +1,16 @@
+@bender-tags: 4.18.1, bug, 1904
```
jacekbogdanski, line 1:
```suggestion
@bender-tags: 4.19.0, bug, 1904
```

tests/plugins/wysiwygarea/manual/tabindex.md
```diff
@@ -0,0 +1,11 @@
+@bender-tags: 4.18.1, bug, 1904
```
jacekbogdanski, line 1:
```suggestion
@bender-tags: 4.19.0, bug, 1904
```

jacekbogdanski (changes requested):
I'm often getting duplicated information about the help button with NVDA and JAWS with the Iframe editor. It reads something like:

1. Rich Text Editor, [editor name]
2. Editor [editor name] Frame
3. Press alt + 0 for help.
4. Editor [editor name] document
5. Press alt + 0 for help.

I'm not sure if it's valid from the accessibility point of view, but maybe it makes sense to read it only once?

I've also run into a bug where NVDA is reading labels literally twice - but I'm not sure if it's caused by a screen reader or the editor. Did you notice that issue during your tests?

I'm not fully confident that we should read so much information. I understand that technically it may make the editor less accessible by reading only some part of the editor (e.g. only frame) but it's a lot of information for a use case when the focus goes straight to editable. Anyway, I don't feel enough confident about how often there is a case when labeling all parts of the editor is important (and what % of our users it covers), so I'm leaving that up to @Comandeer.

tests/core/creators/manual/themeduilabels.md
```diff
@@ -0,0 +1,23 @@
...
+1. Focus the editor.
+
+	**Expected** The editor is correctly announced ("Editor hublabubla").
+
+	**Unexpected** The editor is incorrectly announced or not announced at all.
+
+Sample announcements for the `iframe`-based editor:
+
+* VoiceOver: "edit text, &lt;content and selection info&gt;. Editor, wysiwygarea, group"
+* NVDA: "Rich Text Editor, wysiwygarea, Editor, wysiwygarea frame, Editor, wysiwygarea document editable"
```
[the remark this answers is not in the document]
Comandeer: Hmm, did you test NVDA with Chrome? I can see that in Chrome it's read the way you described but in Firefox – the way I described.

It seems that it would be better to annotate browsers used for each screen reader.

Comandeer:
> I've also run into a bug where NVDA is reading labels literally twice - but I'm not sure if it's caused by a screen reader or the editor. Did you notice that issue during your tests?

The issue was not present in Firefox and in Chrome it's present only after refocusing the editor therefore I didn't catch it :/ It seems to be a screen reader error as the HTML code is correct and it works correctly in Firefox.

> I'm not sure if it's valid from the accessibility point of view, but maybe it makes sense to read it only once?

That's another thing that does not seem to be fixable… This info is added only once, as the `[aria-describedby]` attribute on the `iframe`, however, Firefox exposes the same description on both the `iframe` itself and the document inside the frame.

Comandeer:
I've updated tests and API docs for `config.applicationTitle`. Unfortunately, other raised issues don't seem to be fixable on our side.

core/editor.js
```diff
@@ -1867,15 +1883,51 @@ CKEDITOR.ELEMENT_MODE_INLINE = 3;
...
  * @member CKEDITOR.config
  */
 
+/**
+ * Customizes the {@link CKEDITOR.editor#applicationTtle human-readable title} of the application for this
+ * editor. This title is used as a label for the whole website's region containing editor with its toolbars and other
+ * UI parts. This title impacts various
+ * [accessibility aspects](#!/guide/dev_a11y-section-announcing-the-editor-on-the-page),
+ * e.g. it is commonly used by screen readers for distinguishing editor instances and for navigation.
+ * Accepted values are a string or `false`.
```
jacekbogdanski, line 1902:
```suggestion
 * Customizes the {@link CKEDITOR.editor#applicationTtle human-readable title} of the application for this
 * editor. This title is used as a label for the whole website's region containing the editor with its toolbars and other
 * UI parts. Application title impacts various
 * [accessibility aspects](#!/guide/dev_a11y-section-announcing-the-editor-on-the-page),
 * e.g. it is commonly used by screen readers for distinguishing editor instances and for navigation.
 * Accepted values are a string or `false`.
```

jacekbogdanski (changes requested):
Looks very well, and the explanation given in https://github.com/ckeditor/ckeditor4/issues/2445#issuecomment-1115068142 should make it clear why we choose that approach.

Could you add a manual test with:
* disabled or changed `config.applicationTitle` property
* disabled or changed `config.title` property?

Comandeer:
I've added manual tests for labels, six in total:

*   for disabled labels:
    *   `themeduidisabledapplicationlabels`,
    *   `themeduidisablededitorlabels`,
    *   `inlinedisabledapplicationlabels`,
    *   `inlinedisablededitorlabels`,
*   for custom labels:
    *   `themeduicustomlabels`,
    *   `inlinecustomlabels`.

I've also found a bug when an inline editable got the `[aria-label]` with the `false` string as a value instead of not getting this attribute at all. Fixed it and added unit tests for disabled labels as well.

Comandeer:
I've pushed one more fix.

The `iframe` has the `title` element with the same label as the one in `iframe[title]` (it seems that this fallback is needed for IE). However, if there is no `[title]` attribute on the `iframe` the `title`'s content is set to `null` string. I've changed it to `&nbsp;` (non-breaking space). I didn't remove the `title` completely as it would cause the screen reader to read the `iframe`'s source URL, making the whole experience far less accessible.

jacekbogdanski (approved):
Good job @Comandeer, these changes give much better control over editor accessibility compliance for integrators.

PR #5155, opened 2022-04-13
Fix for unfocusable toolbar buttons in VO on iOS

### LINKED ISSUE #4855: [Accessibility] VoiceOver - Unable to focus RichCombo toolbar buttons like (Font, Font Size, etc...)
## Type of report

Bug

1. Launch ckeditor demo URL in IPad safari browser https://ckeditor.com/ckeditor-4/demo/ 
2. Activate VoiceOver screen reader
3. Tap on Rich Combo toolbar button.

### Actual result

Voice Over focus is immediately transferred to content editing area. Accessible users who rely on Voice Over screen reader are unable to change Font. 

## Other details
This is a regression. This was working fine in CKEditor 4.11.4 version and we recently upgraded 4.16.0 and we are observing this issue
 
* Browser: Safari
* OS: iOS
* CKEditor version: 4.16.0
* Installed CKEditor plugins: RichCombo

Comandeer:

Bug fix

### This PR contains

- [x] Unit tests
- [x] Manual tests

```
* [#4855](https://github.com/ckeditor/ckeditor4/issues/4855): Fixed: buttons are unfocusable while using the editor with VoiceOver on iOS.
```

## What changes did you make?

I've removed the code mentioned in https://github.com/ckeditor/ckeditor4/issues/4855#issuecomment-1097851400 and checked if the issue with the focus on startup is still present in JAWS. As expected, it's no longer reproducible. Yet, to be perfectly sure, I've added the manual test for JAWS.

Closes #4855.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -14,6 +14,7 @@ Fixed Issues:
 * [#3394](https://github.com/ckeditor/ckeditor4/issues/3394): Fixed: [Enhanced image](https://ckeditor.com/cke4/addon/image2) plugin dialog is not supporting URL with query string parameters. Thanks to [Simon Urli](https://github.com/surli)!
 * [#4904](https://github.com/ckeditor/ckeditor4/issues/4904): Fixed: Selection in table via keyboard is broken after adding new table row.
 * [#5049](https://github.com/ckeditor/ckeditor4/issues/5049): Fixed: Editor fails in strict mode.
+* [#4855](https://github.com/ckeditor/ckeditor4/issues/4855): [iOS] Fixed: Focusing toolbar buttons with an enabled VoiceOver screen reader causes moving browser focus into an editable area and interrupts button functionality.
 
 API changes:
 


Comandeer:
I've removed that variable as it's not needed anymore. However, I'm wondering if it's not a breaking change :thinking: After all it was exposed on the editor's instance. OTOH, it's probably a private API as `editor.toolbox` is not documented at all.

jacekbogdanski (changes requested):
I think we should be ok, I've done some quick googling including StackOverflow and I didn't notice any important references to this property. It has not been documented and should be treated as a private API, so any potential breaking changes may happen only if someone incorrectly integrated the editor.

There is one issue though, I mislook that previously thinking that again PFW tests are failing on FF, but seems like there are some new tests that are not stable after your changes:
image

I can also confirm that D&D on Firefox stopped working properly:

https://user-images.githubusercontent.com/10279586/163379295-691c363e-f669-4230-ac0c-60f57fa18839.mov

Note that after D&D focus goes to the toolbar instead of the editor. It looks like the original code comment was misleading and the issue is not only related to JAWS.

Comandeer:
I've modified the hack for Firefox to move the focus to the toolbar and then back to the editor. I've also added tests for this case.

jacekbogdanski (approved):
Well, workarounds like to get out of control easily :) I didn't encounter any other issues with focus after your changes, so good job 👍🏻

PR #5168, opened 2022-04-25
Allow preserving line breaks at the end of inline elements

### LINKED ISSUE #4986: CKEditor removes trailing BR tags
## Type of report

Bug

NOTE: Could be related to https://dev.ckeditor.com/ticket/11392. 

1. Clean editor contents
2. Type Test, apply font size 72px and press Shift+Enter 3 times
3. Type Test and press Shift+Enter 3 times
4. Type Test and press Shift+Enter 3 times
5. Switch to source mode

### Expected result

```
<p><span style="font-size:72px;">Test<br><br><br>Test<br><br><br>Test<br><br><br></span><br></p>
```

### Actual result

BR tags from the last line are removed 

```
<p><span style="font-size:72px">Test<br /><br /><br />Test<br /><br /><br />Test</span><br /><br /><br />&nbsp;</p>
```

## Other details

* Browser: Any
* OS: Any
* CKEditor version: 4.0

NOTE: The issue is related to how editor works, how issues with styling and caret position have been solved back in the day thus it may be hard to provide a fix.

jacekbogdanski:

New feature

### This PR contains

- [x] Unit tests
- [x] Manual tests

```
* [#4986](https://github.com/ckeditor/ckeditor4/issues/4986): Added [`config.shiftLineBreaks`](https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_config.html#cfg-shiftLineBreaks) allowing to preserve inline elements formatting on shift enter.
```

## What changes did you make?

Added feature option allowing to manipulate how `CKEDITOR.htmlParser.fragment.fromHtml` treats line breaks inside inline elements.

Closes #4986

### DIFF
--- a/core/htmlparser/fragment.js
+++ b/core/htmlparser/fragment.js
@@ -677,11 +677,11 @@ CKEDITOR.htmlParser.fragment = function() {
 	};
 
 	/**
-	 * Indicates if break lines (`br`) should be moved outside inline elements.
+	 * Indicates if line breaks (`br`) should be moved outside inline elements.
 	 *
 	 * By default, all children `br` elements, placed at the end of an inline element,
 	 * are shifted outside that element. Shifted elements are attached at the end of the parent block element.
-	 * It allows to produce more clean HTML output without abundance of
+	 * It allows producing more clean HTML output without an abundance of
 	 * orphaned styling markers. This logic can be changed by disabling shifting line breaks or providing
 	 * a custom function allowing to conditionally choose proper behavior.
 	 *
--- a/core/htmlparser/fragment.js
+++ b/core/htmlparser/fragment.js
@@ -143,6 +143,35 @@ CKEDITOR.htmlParser.fragment = function() {
 				addElement( pendingBRs.shift(), currentNode );
 		}
 
+		function shiftBRsPosition() {
+			var shiftLineBreak = CKEDITOR.config.shiftLineBreak;
+
+			if ( shiftLineBreak === true || !pendingBRs.length ) {
+				return;
+			}
+
+			if ( typeof shiftLineBreak !== 'function' ) {
+				sendPendingBRs();
+				return;
+			}
+
+			var result = shiftLineBreak( pendingBRs[ pendingBRs.length - 1 ] );
+
+			if ( result === true ) {
+				return;
+			}
+
+			sendPendingBRs();
+
+			if ( result instanceof CKEDITOR.htmlParser.text ) {
+				currentNode.add( result );
+			}
+
+			if ( result instanceof CKEDITOR.htmlParser.element ) {
+				addElement( result, currentNode );
+			}
+		}
+
 		// Rtrim empty spaces on block end boundary. (https://dev.ckeditor.com/ticket/3585)
 		function removeTailWhitespace( element ) {
 			if ( element._.isBlockLike && element.name != 'pre' && element.name != 'textarea' ) {
@@ -380,8 +409,11 @@ CKEDITOR.htmlParser.fragment = function() {
 
 				currentNode = candidate;
 
-				if ( candidate._.isBlockLike )
+				if ( candidate._.isBlockLike ) {
 					sendPendingBRs();
+				} else {
+					shiftBRsPosition();
+				}
 
 				addElement( candidate, candidate.parent );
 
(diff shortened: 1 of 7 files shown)


core/htmlparser/fragment.js
```diff
@@ -643,4 +675,67 @@ CKEDITOR.htmlParser.fragment = function() {
 			return context || {};
 		}
 	};
+
+	/**
+	 * Indicates if break lines (`br`) should be moved outside inline elements.
```
jacekbogdanski, line 680:
```suggestion
	 * Indicates if line breaks (`br`) should be moved outside inline elements.
```

core/htmlparser/fragment.js
```diff
@@ -643,4 +675,67 @@ CKEDITOR.htmlParser.fragment = function() {
...
+	 *
+	 * By default, all children `br` elements, placed at the end of an inline element,
+	 * are shifted outside that element. Shifted elements are attached at the end of the parent block element.
+	 * It allows to produce more clean HTML output without abundance of
+	 * orphaned styling markers. This logic can be changed by disabling shifting line breaks or providing
+	 * a custom function allowing to conditionally choose proper behavior.
```
jacekbogdanski, line 686:
```suggestion
	 * By default, all children `br` elements, placed at the end of an inline element,
	 * are shifted outside that element. Shifted elements are attached at the end of the parent block element.
	 * It allows producing more clean HTML output without an abundance of
	 * orphaned styling markers. This logic can be changed by disabling shifting line breaks or providing
	 * a custom function allowing to conditionally choose proper behavior.
```

tests/core/htmlparser/manual/preserveinlinestyling.html
```diff
@@ -0,0 +1,30 @@
...
+				>hello, world!<br /><br /><br /></span
+		></span>
+	</p>
+</div>
+
+<script>
+	// IE allows to set cursor position exactly in BR tag.
+	if ( CKEDITOR.env.ie ) {
+		bender.ignore();
+	}
```
sculpt0r, line 14: We could wrap the entire script in IIFE and return the bender.ignore(); to prevent further editor creation, like in https://github.com/ckeditor/ckeditor4/blob/master/tests/core/editable/manual/enterbr.html

tests/core/htmlparser/manual/preserveinlinestyling.html
```diff
@@ -0,0 +1,30 @@
...
+
+	CKEDITOR.config.shiftLineBreaks = function( currentNode ) {
+		if ( currentNode.hasClass( 'filling-br') ) {
+			return;
+		}
+
+		var br =  new CKEDITOR.htmlParser.element( 'br' );
+
+		br.addClass( 'filling-br');
```
sculpt0r, line 23:
```suggestion
		br.addClass( 'filling-br' );
```

tests/core/htmlparser/manual/preserveinlinestyling.html
```diff
@@ -0,0 +1,30 @@
...
+</div>
+
+<script>
+	// IE allows to set cursor position exactly in BR tag.
+	if ( CKEDITOR.env.ie ) {
+		bender.ignore();
+	}
+
+	CKEDITOR.config.shiftLineBreaks = function( currentNode ) {
+		if ( currentNode.hasClass( 'filling-br') ) {
```
sculpt0r, line 17:
```suggestion
		if ( currentNode.hasClass( 'filling-br' ) ) {
```

sculpt0r (changes requested):
I've almost finished the review: mostly small things... But then I encountered the `preserveinlinestyling` manual test. You used the `extraAllowedContent` there.
- if you remove this cfg option - the be class will appear in the source mode (aka. in the content produced by the `getData()`) which makes a little inconsistency. And leads to a situation when each toggle of source mode produces an additional be at the end. However, we could say that it is the user's responsibility to keep the config organized - so it won't break the editor content 🤔 
- but, based on the above user is able to add extra things to the output bypassing the ACF 🤔  Like instead of `br` - the user might create a `script` element that will appear in the output - even if ACF cuts it off from the WYSIWYG mode. Maybe we could apply another round of filtering for the elements produced by the user? The same person might as well just turned off the ACF, without bypassing it with this config 🤔 

Besides the above - I'm sending all the inline comments made so far.

jacekbogdanski:
> if you remove this cfg option - the be class will appear in the source mode (aka. in the content produced by the getData()) which makes a little inconsistency. And leads to a situation when each toggle of source mode produces an additional be at the end. However, we could say that it is the user's responsibility to keep the config organized - so it won't break the editor content 

That's just a manual test with a very simplified, example use case. It's not something that really should bother us (especially that you actually modified test itself).

> but, based on the above user is able to add extra things to the output bypassing the ACF 🤔 Like instead of br - the user might create a script element that will appear in the output - even if ACF cuts it off from the WYSIWYG mode. Maybe we could apply another round of filtering for the elements produced by the user? The same person might as well just turned off the ACF, without bypassing it with this config 🤔

How the user will be able to manipulate the editor config?

core/htmlparser/fragment.js
```diff
@@ -143,6 +143,35 @@ CKEDITOR.htmlParser.fragment = function() {
 				addElement( pendingBRs.shift(), currentNode );
 		}
 
+		function shiftBRsPosition() {
+			var shiftLineBreaks = CKEDITOR.config.shiftLineBreaks;
```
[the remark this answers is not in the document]
jacekbogdanski: Yeah, I know about that limitation, however, we would need to change `fromHtml` signature to accept additional options or implement this feature completely differently. But it's a good idea to provide information that it's a global config only 👍🏻

sculpt0r:
> It's not something that really should bother us (especially that you actually modified test itself).

Just wanted to point out that incorrectly used by the integrator might lead to some differences. But, yes, it doesn't look like a big thing.

> How the user will be able to manipulate the editor config?

The integrator: obviously by setting this new option - but he might do any other mistakes as well - so probably not related 🤔 
The (end) user: by the console, but in such case he might also change the payload that is sent to the integrators backend - without digging in editor config 🤔 

If the eventual inconsistency between source mode and wysiwyg mode is not bothering us - that is fine. Seems like they are mostly possible only if the config of the editor is invalidately prepared 🤔

tests/core/htmlparser/manual/preserveinlinestyling.html
jacekbogdanski: I don't really see a reason for that. There is just a single editor and the test you linked is rather an exception from our standard approach and a rule. Sometimes a developer wants to perform a manual test on an unsupported browser even if it shouldn't be considered a valid case if that test failed.

sculpt0r: 😅  Actually this is a way I was taught to write manual tests ;p I'm not super 'on it' - but so far I tried to review each PR to use that early return. If I good remember it comes from some case when the editor tries to initialize despite the test should be ignored and some error was thrown - so the 'ignored' was not visible in the UI. So it looks like the test failed on some browser.

However, we can live without such construction 👍

sculpt0r:
Rebased on the newest master.

sculpt0r (changes requested):
Almost merged :)

Two final things. I tried to verify cases that are used in the config description and have two thoughts:
- Could we add another manual test for not shifting br's from the inline marker (like `strong`)? Also, this test could use a simple `false` value as a config option. The initial report is about span elements - however, since we provide an example with `strong` - let's also test it manually
- Mistakenly I tried to use the cfg opinion as an instance config option - which has no effect. Bassing on the last reference, we are making a 'backward' assign for such global config values: https://github.com/ckeditor/ckeditor4/blob/a415e4012d4219f92a33990911a8ff31c7ab79bc/core/editor.js#L400 However - there are two different options: `skin` as a member of `CKEDITOR` and `skinName` as a member of `CKEDITOR.config`. The global config options is marked as only `CKEDITOR` member: https://github.com/ckeditor/ckeditor4/blob/5fe059002f8c207df80b44ef999f3d1eec280694/ckeditor.js#L45

I would add a backward assignment to global `CKEDITOR.shiftLineBreaks`. However, If you think that we want the effect of a single global cfg option only. Without allowing to use it as instance cfg: let's change the API docs comment - so it will point the member to `CKEDITOR` and that also requires changes in the changelog entry.

sculpt0r (approved):
Since we discuss my latest comment F2F: LGTM 🎉 

Sorry for not noticing that my comment about the test is actually already in the PR 🙈

PR #5170, opened 2022-04-26
Prevent removal body element with widget

### LINKED ISSUE #5125: Deleting a widget with the keyboard removes the whole body
## Bug

Since CKEditor version 4.16.1, deleting a widget using the keyboard (BACKSPACE or DELETE) can lead to the whole document `body` element being removed. Unsurprisingly, the editor isn't usable anymore after that.

For reproduction purposes, the JSFiddle sample below adds a very simple widget consisting of just one `img` tag. Just place the cursor behind the "deleteme" text and then delete the text and the image using BACKSPACE. When CKEditor determines the next element to apply the "fake selection" to, the outcome will be the body, which will then be removed.
According to my own investigation, this was probably caused by the fix done in https://github.com/ckeditor/ckeditor4/issues/1572. 

The error is still there in the latest version. In 4.16.0 it still works fine.

http://jsfiddle.net/transporthummel/xnm3sebc/5/

### Expected result

The widget is selected by the fake selection and then removed.

### Actual result

Instead of selecting the widget, the document body is removed, making the editor unusable.

KarolDawidziuk:

Bug fix

### This PR contains

- [ ] Unit tests
- [x] Manual tests

```
* [#5125](https://github.com/ckeditor/ckeditor4/issues/5125): Fixed: Deleting a widget with the keyboard removes the whole body.
```

## What changes did you make?

There was an issue with `isEmptyElement` method which only checked if an element contain text. In the fix, I just extended the condition to check if the item doesn't also contain a widget.

Closes #5125.

### DIFF
--- a/core/selection.js
+++ b/core/selection.js
@@ -608,11 +608,6 @@
 				// Allow removal of empty paragraphs (#1572).
 				startElement = sel.getStartElement();
 
-				// Prevent removal of the body element. (#5125)
-				if ( startElement.getName() === 'body' || isWidget( startElement.getFirst() ) ) {
-					return;
-				}
-
 				if ( isEmptyBlock( startElement ) && isDeleteAction( keystroke ) ) {
 					startElement.remove();
 
@@ -633,7 +628,8 @@
 		function isEmptyElement( element ) {
 			var text = element.$.textContent === undefined ? element.$.innerText : element.$.textContent;
 
-			return text === '';
+			// Check if the element does not contain a widget to prevent removal of the body element. (#5125)
+			return text === '' && !isWidget( element.getFirst() );
 		}
 
 		function isEmptyBlock( block ) {
--- a/core/selection.js
+++ b/core/selection.js
@@ -608,6 +608,11 @@
 				// Allow removal of empty paragraphs (#1572).
 				startElement = sel.getStartElement();
 
+				// Prevent removal of the body element. (#5125)
+				if ( startElement.getName() === 'body' || isWidget( startElement.getFirst() ) ) {
+					return;
+				}
+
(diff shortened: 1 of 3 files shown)


sculpt0r (changes requested):
The manual test does not reproduce the issue - because even if I remove the fix - it still passes. The issue was reported with a reproducible sample that use the configuration:
```
autoParagraph: false,
allowedContent: true
```

However it seems that the `allowedConten` could be removed - but the `autoParagraph` is the key here. It seems that the fix is working either way - but just to be sure, could you:
- **recheck how this config option influences the editor behavior in this bug context?**
- add this config option to the manual test
- if it is possible: add some automatic tests? We have some auto-test that triggers the deleting event - so maybe it will be possible also in this case? Please remember about the  `autoParagraph` config option also in those tests.

KarolDawidziuk:

I updated the manual test by adding the missing config option and this PR is ready for another review.

> recheck how this config option influences the editor behavior in this bug context?

This config option automatically handles wrapping blocks around the content inside the document body so disabling it causes the reported issue.

> if it is possible: add some automatic tests? We have some auto-test that triggers the deleting event - so maybe it will be possible also in this case? Please remember about the `autoParagraph` config option also in those tests.

Unfortunately, I have not been able to create a unit test that covers this case. There is a problem with firing the `backspace` key which I was unable to avoid.
Every time I tried to fire `backspace`, the content was not removed, despite setting the selection using `bender` methods and manually.

sculpt0r (approved):
LGTM 🎉 

I also made an approach to automate this test, but without success. Chrome, FF, Safari and IE11 seem to be fine with this manual.

PR #5174, opened 2022-04-29
Fix hardcoded supported image type in clipboard plugin

### LINKED ISSUE #5095: Hard-coded supportedImageTypes variable in clipboard plugin.

KarolDawidziuk:

Bug fix

### This PR contains

- [ ] Unit tests
- [x] Manual tests

```
Bug fixes:

* [#5095](https://github.com/ckeditor/ckeditor4/issues/5095): Fixed: The [clipboard](https://ckeditor.com/cke4/addon/clipboard) plugin shows notification about unsupported file format when a file type is different than `jpg`, `gif`, `png`, not respecting [supported types](https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_fileTools_uploadWidgetDefinition.html#property-supportedTypes) by the [Upload Widget](https://ckeditor.com/cke4/addon/uploadwidget) plugin.

Api changes:

* [#5095](https://github.com/ckeditor/ckeditor4/issues/5095): Added [CKEDITOR.plugins.clipboard.addFileMatcher](https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_plugins_clipboard.html#method-addFileMatcher) function allowing to define supported file formats by the [clipboard](https://ckeditor.com/cke4/addon/clipboard) plugin. Unsupported file format will result in notification that it cannot be dropped or pasted into the editor.

```

## What changes did you make?

It's a second approach for handling this bug. 
For more information please see https://github.com/ckeditor/ckeditor4/pull/5134#issuecomment-1110606621

In this PR we decided to change the way to handle unsupported file types using the clipboard plugin and additional file uploaders.

Firstly, we decided to add `supportedClipboardMatchers` as an array of callbacks to check if the current file type or extension is supported by plugins and if not, display proper notification about the unsupported file type.
I changed the notification message to more generic and I add an additional for the case when the user will want to put a file without an extension.

Proposed approach:

1. Clipboard plugin on the `paste` method will check if the file is different than the default one.
2. If the file is not supported, check if some of the registered matches can support provided file.
3. If the file still is not supported, display proper notification about the unsupported file type. 
4. In case when some of matches can support file, do not display the notification.

Additionally, I added the `isExtensionSupported` method in `filetools` because the existing `isTypeSupported` check only if the MIME type exists. In the case when someone will use some custom file type or file without an extension this will not work. 

The current matcher logic in `uploadwidget` plugin by default tries to get the MIME type of the file and check if it is handled according to the set regexp in widget definition, but if the file type does not exist, the matcher will compare the file extension starting from the last for in the file name with regexp.

Closes #5095.


jacekbogdanski (changes requested):
This solution is a bit overcomplicated, we don't really need to start supporting file extensions right now, as looking at the original implementation of clipboard and upload file/upload widget plugins, it never was considered in the first place.

I will give a hand with this PR since we are close to the release and we plan to include this fix in 4.19.0.

plugins/clipboard/plugin.js
```diff
@@ -146,75 +148,96 @@
...
 
-						fileReader.readAsDataURL( file );
+				// (#5095)
+				if ( !isFileTypeSupported( file ) ) {
+					displayUnsupportedFileTypeNotification( file.type );
+					return;
+				}
 
-						latestId = dataObj.dataTransfer.id;
+				// Convert image file (if present) to base64 string.
```
Comandeer, line 181: I'm wondering if it would not be more readable and semantically correct to split these two things (showing notification and the whole image pasting):

```javascript
editor.on( 'paste', function( evt ) {
	// Notification showing logic.
}, null, null, 1 );

editor.on( 'paste', function( evt ) {
	// Image handling logic.
}, null, null, 1 );
```

plugins/clipboard/plugin.js
```diff
@@ -431,6 +454,10 @@
 					editor.openDialog( 'paste', evt.data );
 				}, 0 );
 			} );
+		},
+
+		addNotificationFileMatcher: function( matcher ) {
```
Comandeer, line 459: There is an inconsistency between the method and property names. Maybe we should rename this method?

```suggestion
		addSupportedFileMatcher: function( matcher ) {
```

Additionally, this method lacks proper API docs.

plugins/clipboard/lang/en.js
```diff
@@ -11,5 +11,6 @@ CKEDITOR.plugins.setLang( 'clipboard', 'en', {
 	pasteNotification: 'Press %1 to paste. Your browser doesn‘t support pasting with the toolbar button or context menu option.',
 	pasteArea: 'Paste Area',
 	pasteMsg: 'Paste your content inside the area below and press OK.',
-	fileFormatNotSupportedNotification: 'This file format is not supported. You can try with one of the supported formats: ${formats}.'
+	fileFormatNotSupportedNotification: 'File format ${format} is not supported.',
```
Comandeer, line 14: I wonder if we can't bold the format :thinking: It would make the notification a little bit more readable.

However, I'm not sure if the bold should be here or in the code that replaces the format in the message.

Comandeer (changes requested):
Looks good, I've left some inline comments.

But there is one case that is not supported by the current solution: dropping/pasting several files at once. It is supported by `uploadwidget`, so there can be a case when the user drops several files – some of them are supported, some of them are not supported. The current solution would not display the notification at all.

plugins/clipboard/plugin.js
jacekbogdanski: I've moved it to `CKEDITOR.plugins.clipboard` namespace, otherwise, AFAIK docs are not properly generated for plugin instances.

plugins/clipboard/lang/en.js
jacekbogdanski: We shouldn't place HTML inside language files as it may confuse language translators. I've changed that to italic as it just looks better from my perspective, but that's subjective so we can discuss that further.

jacekbogdanski:
I've corrected manual tests and review requests. @Comandeer could you take a look before we start working on unit tests?

plugins/clipboard/plugin.js
```diff
@@ -1685,6 +1740,30 @@
...
+		 * Adds a file matcher verifying if a file should be supported via clipboard operations.
+		 *
+		 * In the case of pasting or dragging and dropping unsupported files,
+		 * the clipboard plugin will show a notification informing a user that a given file type is not supported.
+		 *
+		 * ```javascript
+		 * CKEDITOR.plugins.clipboard.addFileMatcher( editor, function( file ) {
+		 * 		var supportedImageTypes = [ 'image/png', 'image/jpeg', 'image/gif' ];
+		 * 		return CKEDITOR.tools.indexOf( supportedImageTypes, file.type ) !== -1;
+		 * } );
```
Comandeer, line 1753: Shouldn't it be indented only once? :thinking: 

```suggestion
		 * CKEDITOR.plugins.clipboard.addFileMatcher( editor, function( file ) {
		 * 	var supportedImageTypes = [ 'image/png', 'image/jpeg', 'image/gif' ];
		 * 	return CKEDITOR.tools.indexOf( supportedImageTypes, file.type ) !== -1;
		 * } );
```

tests/plugins/uploadwidget/manual/fileformatnotification.html
```diff
@@ -0,0 +1,22 @@
...
+			} );
+		}
+	} );
+
+	CKEDITOR.replace( 'editor', {
+		language: 'en',
+		extraPlugins: 'uploadplugin',
+		on: {
+			instanceReady: function( evt ) {
+				bender.tools.ignoreUnsupportedEnvironment( 'uploadimage' );
```
Comandeer, line 18: Shouldn't it use the method for the `uploadwidget` plugin?

```suggestion
				bender.tools.ignoreUnsupportedEnvironment( 'uploadwidget' );
```

The same issue is present in other manual tests.

tests/plugins/uploadwidget/manual/fileformatnotification.md
```diff
@@ -0,0 +1,19 @@
+@bender-tags: 4.19.0, bug, 5095
+@bender-ui: collapsed
+@bender-ckeditor-plugins: wysiwygarea, toolbar, uploadwidget, floatingspace, toolbar, sourcearea
+
+1. Drag and drop, [JPEG2000](../_assets/logo.jp2) into the editor.
```
Comandeer, line 5: This test works incorrectly on Windows 7 and 10 because it does not recognize `.jp2` files correctly, not providing the browser with the MIME type of the file. Therefore, it's de facto not possible to test on Windows.

Comandeer (changes requested):
The code itself looks good. However, there is issue with manual tests and Windows (see my inline comments).

tests/plugins/uploadwidget/manual/fileformatnotification.html
jacekbogdanski: Yes, it should match the specific plugin namespace, in reality, these are using the same condition for unsupported environment check but it's misleading to use a different plugin name than the system under test.

Comandeer:
Rebased onto latest `master`.

plugins/clipboard/lang/cs.js
```diff
@@ -11,5 +11,6 @@ CKEDITOR.plugins.setLang( 'clipboard', 'cs', {
 	pasteNotification: 'Stiskněte %1 pro vložení. Váš prohlížeč nepodporuje vkládání pomocí tlačítka na panelu nástrojů nebo volby kontextového menu.',
 	pasteArea: 'Oblast vkládání',
 	pasteMsg: 'Vložte svůj obsah do oblasti níže a stiskněte OK.',
-	fileFormatNotSupportedNotification: 'Tento formát souboru není podporovaný. Můžete to zkusit s jedním z podporovaných formátů: ${formats}.'
+	fileFormatNotSupportedNotification: 'Tento formát souboru není podporovaný. Můžete to zkusit s jedním z podporovaných formátů: ${formats}.',
```
Comandeer, line 14: This notification was not updated for some reason 🤔

plugins/clipboard/lang/sk.js
```diff
@@ -11,5 +11,6 @@ CKEDITOR.plugins.setLang( 'clipboard', 'sk', {
 	pasteNotification: 'Stlačte %1 na vloženie. Váš prehliadač nepodporuje vloženie prostredníctvom tlačidla v nástrojovej lište alebo voľby v kontextovom menu.',
 	pasteArea: 'Miesto pre vloženie',
 	pasteMsg: 'Vložte svoj obsah do nasledujúcej oblasti a stlačte OK.',
-	fileFormatNotSupportedNotification: 'Tento formát súboru nie je podporovaný. Môžete to skúsiť s jedným z podporovaných formátov: ${formats}.'
+	fileFormatNotSupportedNotification: 'Tento formát súboru nie je podporovaný. Môžete to skúsiť s jedným z podporovaných formátov: ${formats}.',
```
Comandeer, line 14: The same here.

It would not be a problem if the new notification wasn't reversed (it no longer lists the valid formats, it now lists the invalid ones).

plugins/clipboard/lang/zh.js
```diff
@@ -11,5 +11,6 @@ CKEDITOR.plugins.setLang( 'clipboard', 'zh', {
 	pasteNotification: '請按下「%1」貼上。您的瀏覽器不支援工具列按鈕或是內容功能表選項。',
 	pasteArea: '貼上區',
 	pasteMsg: '請將您的內容貼於下方區域中並按下「OK」。',
-	fileFormatNotSupportedNotification: '不支援此檔案格式。您可以試試看使用其中一種支援的檔案格式：${formats}。'
+	fileFormatNotSupportedNotification: '不支援此檔案格式。您可以試試看使用其中一種支援的檔案格式：${formats}。',
```
Comandeer, line 14: I guess the same issue is present here.

Comandeer (approved):
LGTM!

I've ignored some manual tests in IE:

* `tests/plugins/uploadimage/manual/unsupportedfileformat`,
* `tests/plugins/uploadwidget/manual/fileformatnotification`,
* `tests/plugins/uploadwidget/manual/unsupportedfileformatmultiple`.

It seems to not recognize MIME types even if the OS knows them, so tests do not work even with `.webp` files.

I've also added a unit test for one case (not displaying notification when paste event provides `dataValue` alongside files).

I also have some doubt about the changes in our language files (see my inline comments) but it does not seem critical.

jacekbogdanski:
@Comandeer I have no idea why these files weren't updated, but it doesn't make sense. Good catch, I've just changed them manually as the original msg will be misleading.

jacekbogdanski (approved):
Thanks, @Comandeer @KarolDawidziuk!

PR #5178, opened 2022-05-04
Fix strict mode: Move function declaration outside the switch statement

### LINKED ISSUE #5177: Editor crashes due to strict mode
## Type of report

Bug

1. Open local CKE4 sample.

### Expected result

The editor renders correctly.

### Actual result

The editor does not show up and there is an error in the console:

```
SCRIPT1047: In strict mode, function declarations cannot be nested inside a statement or block. They may only appear at the top level or directly inside a function body. 
range.js, line 1474 character 6
```

## Other details

* Browser: IE10
* OS: Windows 7
* CKEditor version: `master` (4.19.0)

The issue is probably caused by 1cebbe557df3ceb0d47cf4e14077548b72bb8281.

sculpt0r:
…cope

### This PR contains

- [ ] Unit tests
- [ ] Manual tests

```
* none
```

## What changes did you make?

*Give an overview…*

Closes #5177 .

### DIFF
--- a/core/dom/range.js
+++ b/core/dom/range.js
@@ -1259,7 +1259,57 @@ CKEDITOR.dom.range = function( root ) {
 		 * @param {Boolean} [excludeBrs=false] Whether include line-breaks when expanding.
 		 */
 		enlarge: function( unit, excludeBrs ) {
-			var leadingWhitespaceRegex = new RegExp( /[^\s\ufeff]/ );
+			var leadingWhitespaceRegex = new RegExp( /[^\s\ufeff]/ ),
+				boundary = this.root,
+				siblingText;
+
+			// Function check if there are only whitespaces from the given starting point
+			// (startContainer and startOffset) till the end on block.
+			// Examples ("[" is the start point):
+			//  - <p>foo[ </p>           - will return true,
+			//  - <p><b>foo[ </b> </p>   - will return true,
+			//  - <p>foo[ bar</p>        - will return false,
+			//  - <p><b>foo[ </b>bar</p> - will return false,
+			//  - <p>foo[ <b></b></p>    - will return false.
+			function onlyWhiteSpaces( startContainer, startOffset ) {
+				// We need to enlarge range if there is white space at the end of the block,
+				// because it is not displayed in WYSIWYG mode and user can not select it. So
+				// "<p>foo[bar] </p>" should be changed to "<p>foo[bar ]</p>". On the other hand
+				// we should do nothing if we are not at the end of the block, so this should not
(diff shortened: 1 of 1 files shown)


core/dom/range.js
```diff
@@ -1261,6 +1261,54 @@ CKEDITOR.dom.range = function( root ) {
...
+
+				while ( ( node = walker.next() ) ) {
+					if ( node.type != CKEDITOR.NODE_TEXT ) {
+						// Stop if you enter to any node (walker.next() will return node only
+						// it goes out, not if it is go into node).
+						return false;
+					} else {
+						// Trim the first node to startOffset.
+						if ( node != startContainer )
+							siblingText = node.getText();
```
Comandeer, line 1299: The `siblingText` variable is overridden here but it's also used outside of the function. Isn't there a case when not overriding the variable outside of this function would generate an incorrect result (like in line 1577)?

Comandeer (changes requested):
Looks good, I only have one doubt (see my inline comments).

core/dom/range.js
sculpt0r: TBH - I just put my trust in all tests we have so far... Also, those are the most complicated ones. It is even hard to name them correctly - since there are so many `if` cases to handling the `enlarging` the range 🤔 Eventually we should decide to create a separate PR for writing separate tests and refactor range functions. IMHO - since those logics are very old - we could say that they work stable. I'm not fully convinced whether we should make additional changes here without any explicit issue?

sculpt0r:
- Moved the function according to the docs

core/dom/range.js
Comandeer: I'm not able to come up with a case that would break it 🤔 And you're right, our test case for ranges is pretty extensive and should catch it anyway.

Comandeer (changes requested):
Although I couldn't break the new `onlyWhiteSpaces()` function, let's make the `siblingText` a variable scoped to the `CKEDITOR.dom.range#enlarge()` method just in case.

sculpt0r:
@Comandeer - corrections are ready:
- I've moved the `siblingText` declaration to the top of `enlarge` definition and it is removed from arguments in `onlyWhiteSpaces`

Also, I've move the `boundary` definition to the top of `enlarge`. Previously it was declared in the second `case` and reassigned in the `third`. It is unused in the first `case`. So overall it will be good to see the declaration before the switch statement. But it is still passed to the `onlyWhiteSpaces` function.

Comandeer (approved):
LGTM! I've just removed one leftover `siblingText` argument and rebased onto latest `master`.

PR #5193, opened 2022-05-10
Use raw non-breaking space for an empty `iframe`'s `title`

### LINKED ISSUE #5189: Failing test: `tests/core/creators/aria`
## Type of report

Failing test

1. Run test on IE11

### Expected result

Test is passing.

### Actual result

Test fails with: 

Editor's title label is incorrect
Expected: &nbsp; (string)
Actual: &amp;nbsp; (string)

## Other details

* Browser: IE11
* OS: Win10
* CKEditor version: 4.19.0 (pre release)
* Installed CKEditor plugins: development

Comandeer:

Bug fix

### This PR contains

- [x] Unit tests
- [ ] Manual tests

N/A

## What changes did you make?

Instead of using `&nbsp;` entity I've used a raw character for setting `iframe`'s `title` when there is no editor label.

Closes #5189.

### DIFF
--- a/plugins/wysiwygarea/plugin.js
+++ b/plugins/wysiwygarea/plugin.js
@@ -417,6 +417,7 @@
 			}, this );
 
 			// In case of lack of the title attribute, use non-breaking space.
+			// It needs to be as a raw character because HTML entity can cause issues in IE.
 			this._.docTitle = this.getWindow().getFrame().getAttribute( 'title' ) || '\xa0';
 		},
 
--- a/plugins/wysiwygarea/plugin.js
+++ b/plugins/wysiwygarea/plugin.js
@@ -416,7 +416,8 @@
 				CKEDITOR.tools.setTimeout( onDomReady, 0, this, win );
 			}, this );
 
-			this._.docTitle = this.getWindow().getFrame().getAttribute( 'title' ) || '&nbsp;';
+			// In case of lack of the title attribute, use non-breaking space.
+			this._.docTitle = this.getWindow().getFrame().getAttribute( 'title' ) || '\xa0';
 		},
 
 		base: CKEDITOR.editable,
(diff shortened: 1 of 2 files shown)


plugins/wysiwygarea/plugin.js
```diff
@@ -416,7 +416,8 @@
 				CKEDITOR.tools.setTimeout( onDomReady, 0, this, win );
 			}, this );
 
-			this._.docTitle = this.getWindow().getFrame().getAttribute( 'title' ) || '&nbsp;';
+			// In case of lack of the title attribute, use non-breaking space.
+			this._.docTitle = this.getWindow().getFrame().getAttribute( 'title' ) || '\xa0';
```
sculpt0r, line 420: I tried this with '&NonBreakingSpace;' but there is a problem with the initial `&` in the HTML entity representation. So it appears that the representation is the issue, not the 'space' or 'non-breaking space'? If you try it with `\xA0` the test also passes. Could you confirm? Additionally - could you decide whether a casual space is enough or the non-breaking one is needed? After that - please adjust the solution (and the test).

sculpt0r (changes requested):
@Comandeer - I have one question about the solution (please take a look at the inline comment). Also there is additional comment about the comment 🙈

plugins/wysiwygarea/plugin.js
Comandeer: I've added a comment.

The regular space can't be used here as it is treated by screen readers as no content, making them read the URL of the `iframe`.

Comandeer:
I've updated comments in code and tests.

sculpt0r (approved):
Thank you for the corrects. The tests works on IE11, 10 & 8 👍 

Sorry for miscommunication in https://github.com/ckeditor/ckeditor4/pull/5193/files#r869940326

PR #5203, opened 2022-05-12
Update paste notification of clipboard plugin

jacekbogdanski:

Task


KarolDawidziuk (approved):
LGTM! 🎉

sculpt0r (approved):
LGTM 🎉 

There is still optional `(s)` but then `are` is used (plural) - but I think we can live with it 👍 Also - I know you talk with Bartek about those texts ;)

Issue #5196, opened 2022-05-11
Document addRichCombo definition


jacekbogdanski:
## Type of report

Docs

## Provide a description of requested docs changes

[`addRichCombo`](https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_ui.html#method-addRichCombo) definition is poorly documented resulting in incorrect usage by 3rd party integrators: https://github.com/ckeditor/ckeditor4/issues/5142#issuecomment-1123406285

jacekbogdanski:
Related https://github.com/ckeditor/ckeditor4/issues/4585

Issue #4839, opened 2021-08-14
It isn't clear what a feature exactly is


Comandeer:
## Type of report

Docs

## Provide a description of requested docs changes

There is a concept of "feature" inside CKEditor 4, but it's not clear what it is. The only place that describes it is https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_feature.html however it seems to be not enough, e.g. there is no way to learn that [adding a button](https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_ui.html#method-addButton) can be joint with modifying ACF rules – as UI button is also a feature (see #4838).

Issue #4585, opened 2021-03-24
No docs for rich combo definition


Comandeer:
## Type of report

Docs

## Provide a description of requested docs changes

Currently there is no description of the rich combo definition. All docs simply states [it's an object](https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_ui_richCombo_handler.html#method-create).

PR #5224, opened 2022-05-24
Removing requiring dir attribute to be on a <span> element in language plugin

### LINKED ISSUE #5085: Language plugin requires dir attribute to be set on a <span> element
## Type of report

Bug

CKEditor requires the `dir` attribute to be present when setting the `lang` attribute on a `<span>`. If it is not present it will remove the `<span>` element.

To reproduce copy the following in source mode:

A) With both the `lang` and `dir` attributes => **OK**
```html
<p>In French Hello is <span dir="ltr" lang="fr">Salut</span>.</p>
```

B) With only the `lang` attribute => **NOT OK**

If the `dir` attribute is not present CKEditor will remove the `<span>` tag.
```html
<p>In French Hello is <span lang="fr">Salut</span>.</p>
```
Will be transformed in:
```html
<p>In French Hello is Salut.</p>
```

### Expected result

Setting the `lang` without specifying the `dir` attribute (as shown in the above test B), should be accepted as it is valid HTML code.

## Other details

* CKEditor version: CKEditor 4.17.1 (revision f6dd30807a)
* Installed CKEditor plugins:
- a11ychecker
- a11yhelp
- about
- balloonpanel
- clipboard
- codesnippet
- codeTag
- dialog
- image
- language
- link
- magicline
- menubutton
- pastefromgdocs
- pastefromlibreoffice
- pastefromword
- pastetools
- pbckcode
- scayt
- special
- table
- tableselection
- widget

KarolDawidziuk:

Feature

### This PR contains

- [x] Unit tests
- [x] Manual tests

```
* [#5085](https://github.com/ckeditor/ckeditor4/issues/5085): Fixed: The [language](https://ckeditor.com/cke4/addon/language) plugin removed the `span` element when it did not have a `dir` attribute.
```

## What changes did you make?

Language plugin required `dir` attribute on a span element and element without this attribute was removed. 
Fix finds all `span` elements with `lang` and without the `dir` attribute, or with an empty `dir` attribute. Then, based on the value in the `lang` attribute, the correct value for the `dir` attribute is determined.

Closes #5085.


tests/plugins/language/manual/addmissingdirattribute.md
```diff
@@ -0,0 +1,14 @@
+@bender-tags: 4.19.1, feature, 5085
```
Comandeer, line 1:
```suggestion
@bender-tags: 4.20.0, feature, 5085
```

jacekbogdanski:
I changed the original issue to a bug, so we can target that PR into `master` branch.

KarolDawidziuk:

I applied the changes mentioned in your comment and PR is ready for another review.
Also, I resolved comments about changing bender tags, as we decided to point this PR to the `master` branch instead of `next`.

Comandeer:
Rebased onto latest `master`.

plugins/language/plugin.js
```diff
@@ -47,6 +49,27 @@
...
+
+				CKEDITOR.tools.array.forEach( langElements, function( element ) {
+					var rtlLanguages = CKEDITOR.tools.object.keys( CKEDITOR.lang.rtl ),
+						isRtlLanguage = CKEDITOR.tools.array.indexOf( rtlLanguages, element.attributes.lang ) !== -1,
+						dirAttribute = isRtlLanguage ? 'rtl' : 'ltr';
+
+					element.attributes.dir = dirAttribute;
+				} );
+
+			}, null, null, 10 );
```
Comandeer, line 71: I was quite sure that we have some kind of API dedicated for such use cases. And it seems that we, indeed, have one! Although it's undocumented :confused: (see #4839, #5196, or #4585). Yet it can be reverse-engineered through the code of our other plugins, e.g. [`colorbutton`](https://github.com/ckeditor/ckeditor4/blob/3dec7a87364e14c6e40791575c7a537109a5eb05/plugins/colorbutton/plugin.js#L35-L51). Buttons added to the toolbar via the `editor.ui#add()` method can have a `contentTransformations` property that can – as its name suggests – transform content input to the editor. The syntax of transformations is described in detail in [the `editor.filter#addTransformations()` method's docs](https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_filter.html#method-addTransformations).

tests/plugins/language/language.js
```diff
@@ -144,6 +144,123 @@
...
+							if ( CKEDITOR.env.gecko || ( CKEDITOR.env.ie && CKEDITOR.env.version < 10 ) ) {
+								expected = '<p>foo <span dir="rtl" lang="' + language + '">bar</span></p>';
+							} else {
+								expected = '<p>foo <span lang="' + language + '" dir="rtl">bar</span></p>';
+							}
+
+							assert.areSame(
+								editor.editable().getData(), expected,
+								'The value of dir attribute for ' + language + ' language is incorrect.'
+							);
```
Comandeer, line 169: Can't it be simplified by using `assert.beautified.html()`? It has an option to sort attributes, so we won't need two distinctive versions with all the same but attributes order

tests/plugins/language/language.js
```diff
@@ -144,6 +144,123 @@
...
+							if ( CKEDITOR.env.gecko || ( CKEDITOR.env.ie && CKEDITOR.env.version < 10 ) ) {
+								expected = '<p>foo <span dir="ltr" lang="' + language + '">bar</span></p>';
+							} else {
+								expected = '<p>foo <span lang="' + language + '" dir="ltr">bar</span></p>';
+							}
+
+							assert.areSame(
+								editor.editable().getData(), expected,
+								'The value of dir attribute for ' + language + ' language is incorrect.'
+							);
```
Comandeer, line 200: The same here.

tests/plugins/language/language.js
```diff
@@ -144,6 +144,123 @@
...
+							if ( CKEDITOR.env.gecko || ( CKEDITOR.env.ie && CKEDITOR.env.version < 10 ) ) {
+								expected = '<p>foo <span dir="' + dir + '" lang="' + language + '">bar</span></p>';
+							} else {
+								expected = '<p>foo <span lang="' + language + '" dir="' + dir + '">bar</span></p>';
+							}
+
+							assert.areSame(
+								editor.editable().getData(), expected,
+								'The value of dir attribute for ' + language + ' language is incorrect.'
+							);
```
Comandeer, line 255: And here.

Comandeer (changes requested):
The code after the refactor looks really nice, good job! But I've made some research and was able to find an API that was designed for such use cases. Could you take a look?

KarolDawidziuk:
Thanks @Comandeer for the review.

I've moved the fix logic to the dedicated `contentTransformations` API, updated unit tests and PR is ready for another review.

plugins/language/plugin.js
KarolDawidziuk: I applied `contentTransformations` and it looks much better. Thank you. 😉

tests/plugins/language/language.js
KarolDawidziuk: Yes, it can be used here, good point. 👍  Tests are already updated.

Comandeer (changes requested):
Looks really good 👍🏻 !

Yet, it seems that `contentTransformations` has some quirk: they are not applied if the `allowedContent` is set to true (so basically: if the filter is off in the editor). And due to that, `span[lang]:not([dir])` are not recognized as language elements by the plugin. It's probably caused by the logic inside the `getCurrentLangElement()` method (it checks for both `[lang]` and `[dir]` attributes). As we already updated `requiredContent` for the plugin, we should also update the logic inside `getCurrentLangElement()` – WDYT?

KarolDawidziuk:

I updated `getCurrentLangElement()` method, added manual and unit test with removed filters to check if the language plugin correctly detects language elements and PR is ready for another check.

Comandeer (approved):
LGTM! I've just slightly reworded the names of newly added unit tests and changed the `feature` tag to the `bug` one in manual tests.

Issue #2528, opened 2018-10-29
Failing tests on Safari for emoji dropdown


msamsel:
## Type of report

Failing tests

## Provide description of the task

There are 4 ignored test on Safari which should work: `/tests/plugins/emoji/dropdown`.
Situation with those test should be analysed and correct to make them active

## Other details

* Browser: Safari
* OS: iOS
* CKEditor version: 4.11
* Installed CKEditor plugins: emoji

mlewand:
The tested case works when performed manually, but the tests are failing on Safari nonetheless. So that should be further investigated.

jacekbogdanski:
There are additional failing tests on iOS from this group:
`test emoji dropdown filter search results`
`test left and right arrow does not change focus on input element`

PR #5232, opened 2022-06-07
Fix failing emoji dropdown test

### LINKED ISSUE #5188: Failing `tests/plugins/emoji/dropdown` test
## Type of report

Failing test

1. Open http://localhost:1030/#tests/is:unit,name:dropdown
2. Run the test from the dashboard.

### Expected result

All passes.

### Actual result

`test navigation highlights proper section when scrolls` fails

## Other details

* Browser: Chrome 101.0.4951.54
* OS: macOS 12.3.1
* CKEditor version: 4.19.0
* Installed CKEditor plugins: emoji

sculpt0r:

Failing test fix

### This PR contains

- [X] Unit tests
- [ ] Manual tests

```
skip
```

## What changes did you make?

*Give an overview…*

I moved the problematic test case to second place in the test suite. The first one validates emojis caching and needs to be first. But it also does not perform any operations on the panel. It looks like other test cases have an influence on each other. That sounds like trouble, but we rely on those tests for a long time - so I won't dig here more.

Also, I try to restore the same test case for `Safari` - I've run the test multiple times on Browserstack and locally, within the dashboard panel, and as a separate test case. I didn't encounter any issues with it. But I found it failing on ios: previously test case was ignored and reported in #2528.

I also tried to run it on FF - but it also still failing randomly (#2831).

Closes #5188 .

### DIFF
--- a/tests/plugins/emoji/dropdown.js
+++ b/tests/plugins/emoji/dropdown.js
@@ -44,6 +44,29 @@
 			} );
 		},
 
+		// This test case fail if performed later in this suite (#5188).
+		'test navigation highlights proper section when scrolls': function() {
+			var bot = this.editorBot;
+
+			waitForEmoji( bot.editor, function() {
+				bot.panel( 'EmojiPanel', function( panel ) {
+					var doc = panel._.iframe.getFrameDocument(),
+						testElement = doc.findOne( 'a[data-cke-emoji-name="star"]' );
+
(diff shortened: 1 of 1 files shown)


jacekbogdanski (approved):
That's not totally proper approach, but at the moment it makes no sense to spend more time on the issue as tests seem to perform correctly.

PR #5238, opened 2022-06-09
Fix #5158 - assure that calculator element is always attached to DOM

### LINKED ISSUE #5158: CKEDITOR.tools.convertToPx breaks with Hotwire Turbo / Turbolinks, et al
## Type of report

`CKEDITOR.tools.convertToPx` does not properly support frameworks that replace HTML body but don't reload Javascript libraries, such as Hotwire Turbo.

More accurately, anything that clears or replaces the DOM body will break this `CKEDITOR.tools.convertToPx`

This surfaces when using the autogrow plugin, but anything that relies on `editor.resize()` will trigger this behavior. Instantiate an editor, clear the DOM body, create new HTML for an editor and instantiate it, then call resize on it.

### Expected result

The editor should resize to the size specified.

### Actual result

The editor resizes to a zero height.

## Other details

This is because `CKEDITOR.tools.convertToPx` declares a variable `calculator` which it uses to memoize an element appended to the body used for the calculation.

But if anything clears the DOM body, this "calculator" element no longer exists, even though the variable still points to the disembodied element. This causes the incorrect calculation that—at least in Chrome—returns zero.

The fix is to also check `calculator.$.isConnected` to see whether it is connected to the DOM:

https://github.com/ckeditor/ckeditor4/blob/c0dbd14301d5639e600a6857de0478d742147342/core/tools.js#L978-L983

```
if ( !calculator ) {
	calculator = CKEDITOR.dom.element.createFromHtml( '<div style="position:absolute;left:-9999px;' +
		'top:-9999px;margin:0px;padding:0px;border:0px;"' +
		'></div>', CKEDITOR.document );
	CKEDITOR.document.getBody().append( calculator );
}
else
{
	if ( !calculator.$.isConnected ) {
		// Re-attach to DOM if it got disconnected.
		CKEDITOR.document.getBody().append( calculator );
	}
}
```

sculpt0r:

### This PR contains

- [X] Unit tests
- [ ] Manual tests

```
* [#5158](https://github.com/ckeditor/ckeditor4/issues/5158): Fixed: [tools.convertToPx()](https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_tools.html#method-convertToPx) keeps returning valid results despite DOM manipulations which might delete helper calculator object.
```

## What changes did you make?

I used our API [method](https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_dom_element.html#method-isDetached) to detect whether calculator is really under the DOM. If not: it is reattached.

The only thing I'm worried ☝️ about is there a valid use case that might entirely destroy the underlying object - so we rather should recreate the entire element instead of just reattaching?

Added an automatic test that corresponds to the issue description (https://github.com/ckeditor/ckeditor4/pull/5238/commits/90f916d906a5a193bbced65a275b460e404087f8) - it is failing as expected( without the fix).

And then the fix makes it work. Also added assertion if the conversion is correct. Because if sth goes wrong inside the method because of the introduced attaching the first and second results in test case might be equal (the first assertion passed) - but they will be equal to `0` which is not a correct result. That is why the second assertion was added.

Closes #5158.

### DIFF
--- a/core/tools.js
+++ b/core/tools.js
@@ -1002,14 +1002,12 @@
 				boundingClientRect;
 
 			return function( cssLength ) {
-				if ( !calculator ) {
+				// Recreate calculator whenever it was externally manipulated (#5158).
+				if ( !calculator || calculator.isDetached() ) {
 					calculator = CKEDITOR.dom.element.createFromHtml( '<div style="position:absolute;left:-9999px;' +
 						'top:-9999px;margin:0px;padding:0px;border:0px;"' +
 						'></div>', CKEDITOR.document );
-				}
 
-				// (#5158)
-				if ( calculator.isDetached() ) {
 					CKEDITOR.document.getBody().append( calculator );
 				}
 
--- a/core/tools.js
+++ b/core/tools.js
@@ -1006,6 +1006,10 @@
 					calculator = CKEDITOR.dom.element.createFromHtml( '<div style="position:absolute;left:-9999px;' +
 						'top:-9999px;margin:0px;padding:0px;border:0px;"' +
 						'></div>', CKEDITOR.document );
+				}
+
+				// (#5158)
+				if ( calculator.isDetached() ) {
 					CKEDITOR.document.getBody().append( calculator );
 				}
 
--- a/tests/core/tools.js
+++ b/tests/core/tools.js
@@ -1113,7 +1113,9 @@
 
 		// (#5158)
 		'test convertToPx works after calculator element was removed': function() {
-			var firstResult = CKEDITOR.tools.convertToPx( '10px' );
+			// Attach calculator element to the DOM.
+			CKEDITOR.tools.convertToPx( '10px' );
+
 			// Based on convertToPx implementation
 			// calculator is the last element under `body` after `convertToPx` invocation.
 			var bodyChildren = CKEDITOR.document.getBody().getChildren(),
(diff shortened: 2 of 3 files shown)


core/tools.js
```diff
@@ -979,6 +979,10 @@
 					calculator = CKEDITOR.dom.element.createFromHtml( '<div style="position:absolute;left:-9999px;' +
 						'top:-9999px;margin:0px;padding:0px;border:0px;"' +
 						'></div>', CKEDITOR.document );
+				}
+
+				// (#5158)
+				if ( calculator.isDetached() ) {
```
jacekbogdanski, line 985: I think that at this point we can assume that calculator has been somehow manipulated without our supervision. So, we can assume that it's broken and should be re-created instead.

jacekbogdanski (changes requested):

sculpt0r:
Rebased on the newest master.

sculpt0r:
@jacekbogdanski - ready for another review :)
- The calculator is now recreated also whenever it was removed from the DOM

tests/core/tools.js
```diff
@@ -1111,6 +1111,21 @@
...
+			// Based on convertToPx implementation
+			// calculator is the last element under `body` after `convertToPx` invocation.
+			var bodyChildren = CKEDITOR.document.getBody().getChildren(),
+				calculator = bodyChildren.getItem( bodyChildren.count() - 1 );
+
+			calculator.remove();
+
+			var result = CKEDITOR.tools.convertToPx( '10px' );
+			assert.areSame( firstResult, result, 'convertToPx() returns different values when helper calculator was removed.' );
+			assert.areSame( result, 10, 'calculator was not properly attached to DOM.' );
```
jacekbogdanski: The step comparing the first result and the second one is unnecessary. Also, the proper order of assertion method parameters is _expected_, _actual_.

jacekbogdanski (approved):
Looks good, just minor issues (I will correct them).

PR #5242, opened 2022-06-15
Fix merging cells with rowspan

### LINKED ISSUE #4284: Error thrown when merging cells with rowspan
## Type of report

Bug

1. Open https://jsfiddle.net/Comandeer/qj5367zy/
2. Open console.
3. Select column with a rowspanned cell and one other column.
4. Open context menu and choose "Cell" → "Merge cells" option.

### Expected result

Cells are merged, undo step is created.

### Actual result

Cells are merged, however error is thrown and no undo step is created:

```
ckeditor.js:154 Uncaught TypeError: Cannot read property 'type' of null
    at a (ckeditor.js:154)
    at CKEDITOR.dom.range.createBookmark2 (ckeditor.js:156)
    at Array.createBookmarks2 (ckeditor.js:515)
    at CKEDITOR.dom.selection.createBookmarks2 (ckeditor.js:468)
    at new CKEDITOR.plugins.undo.Image (ckeditor.js:1125)
    at CKEDITOR.plugins.undo.UndoManager.save (ckeditor.js:1119)
    at a.c (ckeditor.js:1114)
    at a.d (ckeditor.js:10)
    at a.<anonymous> (ckeditor.js:11)
    at a.CKEDITOR.editor.CKEDITOR.editor.fire (ckeditor.js:13)
```

## Other details

* CKEditor version: 4.7.0+
* Installed CKEditor plugins: undo, tableselection

What's the most interesting, it's the fact that it works correctly when whole standard preset is used: https://jsfiddle.net/Comandeer/2qfhuc8z/

KarolDawidziuk:
Bug fix

### This PR contains

- [x] Unit tests
- [x] Manual tests

```
* [#4284](https://github.com/ckeditor/ckeditor4/issues/4284): Fixed: [Tableselection](https://ckeditor.com/cke4/addon/tableselection) Merging cells with a rowspan was throwning error and did not create undo step.
```

## What changes did you make?

This issue occurred when the `undo` plugin was loaded before the `tableselection` plugin and undo `afterCommandExec` was called before tableselection `afterCommandExec`. This caused trying to create a snapshot with the wrong and not updated range of merged cells. 

To fix this, I updated the `afterCommandExec` priority in the `tableselection` plugin to a higher priority than the default.

Closes #4284.

### DIFF
--- a/plugins/tableselection/plugin.js
+++ b/plugins/tableselection/plugin.js
@@ -922,7 +922,7 @@
 			if ( CKEDITOR.tools.array.indexOf( cmds, evt.data.name ) !== -1 ) {
 				callback( editor, evt.data );
 			}
-		} );
+		}, null, null, 9 );
 	}
 
 	/**
--- a/plugins/tableselection/plugin.js
+++ b/plugins/tableselection/plugin.js
@@ -921,7 +921,10 @@
 		editor.on( 'afterCommandExec', function( evt ) {
 			if ( CKEDITOR.tools.array.indexOf( cmds, evt.data.name ) !== -1 ) {
 				callback( editor, evt.data );
+
 			}
+		// This listener is connected with undo plugin listener and require a higher priority
+		// than the listener in undo plugin to create a correct undo step (#4284).
 		}, null, null, 9 );
 	}
 
--- a/tests/plugins/tabletools/tabletools.js
+++ b/tests/plugins/tabletools/tabletools.js
@@ -265,7 +265,6 @@
 		},
 
 		'test merge cells with a rowspan should create undo step': function() {
-
 			bender.editorBot.create( {
 				name: 'editor_merge_rowspanned_cells',
 				config: {
@@ -282,8 +281,6 @@
 
 					var output = bot.getData( true ),
 						undo = editor.getCommand( 'undo' );
-
-					output = output.replace( /\u00a0/g, '&nbsp;' );
 
 					assert.areSame( bender.tools.compatHtml( expected ), output );
 					assert.isTrue( undo.state === CKEDITOR.TRISTATE_OFF );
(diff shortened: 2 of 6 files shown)


plugins/tableselection/plugin.js
```diff
@@ -922,7 +922,7 @@
 			if ( CKEDITOR.tools.array.indexOf( cmds, evt.data.name ) !== -1 ) {
 				callback( editor, evt.data );
 			}
-		} );
+		}, null, null, 9 );
```
sculpt0r, line 925: We definitely need a comment here:
- that this listener is connected with `undo` plugin listener, and it needs a higher priority than the listener in `undo`
- to reference the issue this PR resolve

```suggestion
		// Maybe put the comment here?
		}, null, null, 9 );
```

tests/plugins/tabletools/tabletools.js
```diff
@@ -262,6 +262,33 @@
...
+				var editor = bot.editor;
+
+				bender.tools.testInputOut( 'merge-rowspanned-cells', function( source, expected ) {
+					bot.setHtmlWithSelection( source );
+					bot.execCommand( 'cellMerge' );
+
+					var output = bot.getData( true ),
+						undo = editor.getCommand( 'undo' );
+
+					output = output.replace( /\u00a0/g, '&nbsp;' );
```
sculpt0r, line 286: This line of code seems to be unnecessary. Why do we need this replacement? The test looks fine without it.

sculpt0r (changes requested):
@KarolDawidziuk

The fix itself and tests look fine. However, I left a few inline comments and the question. Could you take a look at them?

tests/plugins/tabletools/tabletools.js
KarolDawidziuk: Yes, it looks like it's not needed. Removed

PR #5254, opened 2022-07-05
Fix: dialog validators  accept value only from inner getValue method

### LINKED ISSUE #4473: Dialog validators does not work if there is no 'getValue()' method provided in the current context
## Type of report

Bug

The `dialog` plugin provides it's own set of validators, for example `cssLength`:

https://github.com/ckeditor/ckeditor4/blob/5380706bb4ceb6092dadb2b7067d6c14efec6b26/plugins/dialog/plugin.js#L3341-L3345

And the API docs states they can be simply used as follows:

```js
CKEDITOR.dialog.validate.cssLength( 'error!' )( '10pt' );
```

However, such calls throws an error:

```
Uncaught TypeError: b.replace is not a function
    at Object.trim (ckeditor.js:30)
    at Array.<anonymous> (ckeditor.js:663)
    at ckeditor.js:663
    at <anonymous>:1:47
```

### Expected result

It works as described in API docs.

### Actual result

Error is thrown.

## Other details

Well, **that's the interesting part**... Many of the validators use special `functions()` method which combines validators:

https://github.com/ckeditor/ckeditor4/blob/5380706bb4ceb6092dadb2b7067d6c14efec6b26/plugins/dialog/plugin.js#L3219-L3258

When the validator is created (like in `cssLength` above), `functions()` returns another function (let's call it `validatorFn()`) which has access to initial `args`. This means calling `CKEDITOR.dialog.validate.cssLength( 'error!' )` internally calls:

```js
.functions( `cssLengthInternalFn`, `error!` )
```

https://github.com/ckeditor/ckeditor4/blob/5380706bb4ceb6092dadb2b7067d6c14efec6b26/plugins/dialog/plugin.js#L3342-L3344

Which returns `validatorFn()` mentioned above. This functions accesses the value to be validated in two ways. Using `getValue()` method from the current context or accessing `args[ 0 ]` (the ones from `functions()` call):

https://github.com/ckeditor/ckeditor4/blob/5380706bb4ceb6092dadb2b7067d6c14efec6b26/plugins/dialog/plugin.js#L3226

If `getValue()` method is present in the current context everything works fine. But if not, `args[ 0 ]` are used as a value to be validated. However, in the context of `.functions( cssLengthInternalFn, 'error!' )` call, `args[ 0 ]` are `cssLengthInternalFn` (so the validator function). **And so `args[ 0 ]` are both used as validator function and value to be validated** (not a string so `Uncaught TypeError: b.replace is not a function` is thrown).

**I wonder if the API docs are just confusing and it was never meant to be used without extending dialog and adding `getValue()` there, but the comment there states it should work**:

https://github.com/ckeditor/ckeditor4/blob/5380706bb4ceb6092dadb2b7067d6c14efec6b26/plugins/dialog/plugin.js#L3222-L3226

* Browser: All
* OS: All
* CKEditor version: At least `4.10.0` but looking at git history it might be `4.0.0` as well :thinking: 
* Installed CKEditor plugins: `dialog`

sculpt0r:

Bug fix

### This PR contains

- [X] Unit tests
- [ ] Manual tests

```
 [#4473](https://github.com/ckeditor/ckeditor4/issues/4473): Fix: [dialog.validate methods](https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_dialog_validate.html) accept value only from `getValue()` method in given context. Affected validators:
	* [equals](https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_dialog_validate.html#method-equals)
	* [notEqual](https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_dialog_validate.html#method-notEqual)
	* [cssLength](https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_dialog_validate.html#method-cssLength)
	* [htmlLength](https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_dialog_validate.html#method-htmlLength)
	* [inlineStyle](https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_dialog_validate.html#method-inlineStyle)
```

## What changes did you make?

*Give an overview…*
This PR added tests for all [dialog.validate](https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_dialog_validate.html) except [functions](https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_dialog_validate.html#method-functions) - since it will be covered in #5248.

Each validator validates examples given in the documentation. To be sure that our API works in the same manner as the documentation says.

The **fix**:
Instead of trying to get value from arguments cached in a variable (during the first call of `functions`) the value is once again searched in `arguments` when the returned validator function was called. Please note that the originally cached arguments are preserved since they are important to preserve composition parameters.

Thanks to that fix, it was also possible to simplify `validatorinlinestyle` tests.

Also thanks to test based on doc examples it appears that there were buggy samples for [regex validator](https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_dialog_validate.html#method-regex) - so they are corrected now.

Closes #4473 .

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -16,14 +16,13 @@ Fixed Issues:
 * [#5234](https://github.com/ckeditor/ckeditor4/issues/5234): Fixed: [Easy Image](https://ckeditor.com/cke4/addon/easyimage) can't be uploaded using toolbar button.
 * [#438](https://github.com/ckeditor/ckeditor4/issues/438): Fixed: It is impossible to navigate to the [elementspath](https://ckeditor.com/cke4/addon/elementspath) from the [toolbar](https://ckeditor.com/cke4/addon/toolbar) by keyboard and vice versa.
 * [#4449](https://github.com/ckeditor/ckeditor4/issues/4449): Fixed: [`dialog.validate#functions`](https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_dialog_validate.html#method-functions) incorrectly composes functions that returns an optional error message, like e.g. [`dialog.validate.number`](https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_dialog_validate.html#method-number) due to unnecessary return type coercion.
-* [#4473](https://github.com/ckeditor/ckeditor4/issues/4473): Fix: [dialog.validate methods](https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_dialog_validate.html) does not accept parameter value. Affected validators:
-	* [equals](https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_dialog_validate.html#method-equals)
-	* [notEqual](https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_dialog_validate.html#method-notEqual)
-	* [cssLength](https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_dialog_validate.html#method-cssLength)
-	* [htmlLength](https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_dialog_validate.html#method-htmlLength)
-	* [inlineStyle](https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_dialog_validate.html#method-inlineStyle)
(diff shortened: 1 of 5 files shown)


jacekbogdanski (approved):
Looks good. I've corrected tests and slightly updated code changes.

jacekbogdanski (approved):

PR #5255, opened 2022-07-05
Fix incorrect value of `[aria-expanded]` on menubuttons

### LINKED ISSUE #5144: Incorrect `aria-expanded` on language button after language selection
## Type of report

Bug

1. Open [CKE4 with language plugin](https://ckeditor.com/docs/ckeditor4/latest/examples/language.html)
2. Inspect the language toolbar button and find its `aria-expanded` attribute. At this moment it is set to `false`
3. Use the language button and select some language.

### Expected result

The `aria-expanded` attribute should be set to `false` since the submenu is folded. Also, screen readers shouldn't read this button at the current state as `expanded`.

### Actual result

The `aria-expanded` attribute is set to `true`. Screen readers read this button as expanded when any language is selected.

## Other details

* Browser: -
* OS: -
* CKEditor version: `4.18.0` (reproducible even back to `4.13.0` - so no regression)
* Installed CKEditor plugins: language

Comandeer:

Bug fix

### This PR contains

- [x] Unit tests
- [x] Manual tests

```
* [#5144](https://github.com/ckeditor/ckeditor4/issues/5144): Fixed: [menu buttons](https://ckeditor.com/cke4/addon/menubutton) and [panel buttons](https://ckeditor.com/cke4/addon/panelbutton) incorrectly indicates the open status of their associated pop-up menus.
```

## What changes did you make?

As explained in https://github.com/ckeditor/ckeditor4/issues/5144#issuecomment-1174959858 I had to find another way to detect if the menu is open. Fortunately, this information is available as an internal property of the button, `button._.on`.

I was thinking about exposing this property as a public one but finally decided not to do that. This decision really simplified the solution but at the same time – made it a little bit dirty (as the internal property is now used directly inside the `button` plugin)

Closes #5144.


sculpt0r:
Rebased on the newest master.

plugins/button/plugin.js
```diff
@@ -345,8 +345,8 @@
 					// For other buttons, do not remove the attribute, instead set its value (#2444).
 					element.setAttribute( 'aria-pressed', state === CKEDITOR.TRISTATE_ON );
 				} else if ( this.hasArrow ) {
-					// Indicates that menu button is opened (#421).
-					element.setAttribute( 'aria-expanded', state == CKEDITOR.TRISTATE_ON );
+					// Indicates that menu button is opened (#421, #5144).
+					element.setAttribute( 'aria-expanded', this._.on ? 'true' : 'false' );
```
sculpt0r, line 349: Currently, the `on` property is introduced in the `menubutton` plugin. The current hierarchy looks like this:
|--button
|---- menubutton <- defines and operates on internal `on` property

So now, in the lower hierarchy plugin, we are using the property defined in the higher hierarchy. This property means something in the `button` only if it is actually at least `menubutton` instance 🤔 So we have a button with an arrow, but we are unable to detect whether this button is expanded or not without a higher-level plugin (`menubutton`). That seems to be invalid in terms of proper inheritance.

What is more, the `panelbutton` plugin sets this property after invoking setState:
https://github.com/ckeditor/ckeditor4/blob/058f65bedfa674d166320dfed3e1f4fd17e430a8/plugins/panelbutton/plugin.js#L91-L93

This effectively means `colorbutoon` has `aria-expanded` always on false. The `language` button sets the `_.on` property on `1` and right after that on `0` **after** selecting language, which also makes this effectively that the `aria-expanded` is set on `false`.

Also, the aria-expanded does not change to `true` whether the menu actually opens. And even so, that is senseless for screen reader since the only way reader can read the state of the button is to leave focus (then menu is collapsed again) and focus it again (so it is folded at this moment).

Screen reader properly reads the attribute, but maybe... changing the aria attribute doesn't have a sense for such menus and always should be `false` or even removed? However, if it makes sense to update the `aria-attribute`:
- please correct the behaviour of `panelbutton`
- please correct the behaviour whenever the menubutton is expanded
- maybe we should declare `on` property in the `button` plugin: https://github.com/ckeditor/ckeditor4/blob/c08cfedb607fab87866052e4aff2ec8facfb115d/plugins/button/plugin.js#L93

sculpt0r (changes requested):
@Comandeer I left one significant inline comment. There are two ways to choose there (or you can see even more?). I believe you can choose the right one based on your a11y knowledge. However, if I'm incorrect at some point - please let me know :p So we can establish the approach together 👍

Comandeer:
I've completely rewritten the solution. I've moved the logic responsible for setting the `[aria-expanded]` attribute to the `menubutton` and `panelbutton` plugins. The only bit left inside the `button` plugin is adding the initial value of the `[aria-expanded]` attribute if the `button.hasArrow` property is set.

plugins/button/plugin.js
Comandeer: > Screen reader properly reads the attribute, but maybe... changing the aria attribute doesn't have a sense for such menus and always should be `false` or even removed?

The attribute should be added and reflect the current status of the menu associated with the button – that's how it's [described in the spec](https://w3c.github.io/aria/#aria-expanded). Let's do it properly.

jacekbogdanski (approved):
Good job!

PR #5291, opened 2022-07-20
Fix: selection optimization in list disallow full list removal

### LINKED ISSUE #4931: CKEditor 4 - Selecting whole content by pressing Control + A & press Delete on Keyboard doesn’t erase all bullet points On first try
Hi,
I have tried this using CKEditor 4 - Standard Package Version 4.16.1 and Standard Package Version 4.16.2 and found this bug on Chrome browser. Basically, under Chrome browser, when I am selecting whole content in CKEditor by pressing Control + A & press Delete on Keyboard then it doesn’t erase all bullet points On first try. We could delete whole content present on CKEditor in multiple attempts and not in one single attempt. It looks like there is some bug with this CKEditor control only in Chrome browser since it works fine on Firefox browser, no issue with that FF browser.
Can someone please advice how to resolve this issue on Chrome browser? Thanks!

Here it is sample content that I tried erasing it on CKEditor 4 control by pressing Control + A and then Press Delete button on Keyboard

image

sculpt0r:

Bug fix

### This PR contains

- [X] Unit tests
- [X] Manual tests

```
* [#4931](https://github.com/ckeditor/ckeditor4/issues/4931): Fix: selection optimization skips list elements and disallow list removal.
```

## What changes did you make?

*Give an overview…*

Besides all details I put in the issue comments: I found that we have additional selection optimization which is responsible for shrinking selection under some circumstances. I added an additional check if the end selection container is a `li` element and prevents optimization if so.

Closes #4931 .

### DIFF
--- a/core/selection/optimization.js
+++ b/core/selection/optimization.js
@@ -117,7 +117,10 @@
 			return false;
 		}
 
-		var isInList = range.endContainer.is && range.endContainer.is( 'li' );
+		// The endContainer might be a text inside li element (in IE8).
+		var isInList = range.endContainer.is ?
+			range.endContainer.is( 'li' ) :
+			range.endContainer.getParent().is && range.endContainer.getParent().is( 'li' );
 
 		// Prevent optimization in lists (#4931).
 		if ( isInList ) {
--- a/core/selection/optimization.js
+++ b/core/selection/optimization.js
@@ -117,7 +117,9 @@
 			return false;
 		}
 
-		if ( range.endOffset === 0 ) {
+		var isInList = range.endContainer.is && range.endContainer.is( 'li' );
+
+		if ( range.endOffset === 0 && !isInList )  {
 			return true;
 		}
 
@@ -125,7 +127,7 @@
 			endsInText = isText( range.endContainer ),
 			limit = startsInText ? range.startContainer.getLength() : range.startContainer.getChildCount();
 
-		return range.startOffset === limit || startsInText ^ endsInText;
+		return !isInList && ( range.startOffset === limit || startsInText ^ endsInText );
 	}
 
 	// Prevent infinite recurrency when the browser does not allow the expected selection.
--- a/core/selection/optimization.js
+++ b/core/selection/optimization.js
@@ -120,7 +120,11 @@
 		var isInList = range.endContainer.is && range.endContainer.is( 'li' );
 
 		// Prevent optimization in lists (#4931).
-		if ( range.endOffset === 0 && !isInList )  {
+		if ( isInList ) {
+			return false;
+		}
+
+		if ( range.endOffset === 0 )  {
 			return true;
 		}
 
@@ -128,7 +132,7 @@
 			endsInText = isText( range.endContainer ),
 			limit = startsInText ? range.startContainer.getLength() : range.startContainer.getChildCount();
 
-		return !isInList && ( range.startOffset === limit || startsInText ^ endsInText );
+		return range.startOffset === limit || startsInText ^ endsInText;
 	}
 
 	// Prevent infinite recurrency when the browser does not allow the expected selection.
--- a/core/selection/optimization.js
+++ b/core/selection/optimization.js
@@ -119,6 +119,7 @@
 
 		var isInList = range.endContainer.is && range.endContainer.is( 'li' );
 
+		// Prevent optimization in lists (#4931).
 		if ( range.endOffset === 0 && !isInList )  {
 			return true;
 		}
--- a/tests/core/selection/manual/optimizationlist.html
+++ b/tests/core/selection/manual/optimizationlist.html
@@ -1,110 +1,26 @@
 <h2>Iframe editor</h2>
 <div id="editor1">
-	<ul>
-		<li>1</li>
-		<li>2</li>
-		<li>3</li>
-		<li>4</li>
-		<li>5
-		<ul>
-			<li>6
-			<ul>
-				<li>7
-				<ul>
-					<li>8</li>
-				</ul>
-				</li>
-			</ul>
-			</li>
-			<li>9</li>
-		</ul>
-		</li>
-		<li>0</li>
-	</ul>
-
 	<ol>
-		<li>12</li>
-		<li>14</li>
-		<li>16</li>
-		<li>18</li>
-		<li>20
+		<li>1</li>
+		<li>2
 		<ol>
-			<li>22
-			<ol>
-				<li>24</li>
-			</ol>
-			</li>
-			<li>26
-			<ol>
-				<li>28
-				<ol>
-					<li>30
-					<ol>
-						<li>32</li>
-					</ol>
-					</li>
-					<li></li>
-				</ol>
-				</li>
-			</ol>
-			</li>
+			<li>1</li>
 		</ol>
 		</li>
+		<li>&nbsp;</li>
 	</ol>
 </div>
 
 <h2>Divarea editor</h2>
 <div id="editor2">
-	<ul>
-		<li>1</li>
-		<li>2</li>
-		<li>3</li>
-		<li>4</li>
-		<li>5
-		<ul>
-			<li>6
-			<ul>
-				<li>7
-				<ul>
-					<li>8</li>
-				</ul>
-				</li>
-			</ul>
-			</li>
-			<li>9</li>
-		</ul>
-		</li>
-		<li>0</li>
-	</ul>
-
 	<ol>
-		<li>12</li>
-		<li>14</li>
-		<li>16</li>
-		<li>18</li>
-		<li>20
+		<li>1</li>
+		<li>2
 		<ol>
-			<li>22
-			<ol>
-				<li>24</li>
-			</ol>
-			</li>
-			<li>26
-			<ol>
-				<li>28
-				<ol>
-					<li>30
-					<ol>
-						<li>32</li>
-					</ol>
-					</li>
-					<li></li>
-				</ol>
-				</li>
-			</ol>
-			</li>
+			<li>1</li>
 		</ol>
 		</li>
+		<li>&nbsp;</li>
 	</ol>
 </div>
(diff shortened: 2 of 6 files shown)


sculpt0r:
Rebased on the newest master.

core/selection/optimization.js
```diff
@@ -117,15 +117,17 @@
 			return false;
 		}
 
-		if ( range.endOffset === 0 ) {
+		var isInList = range.endContainer.is && range.endContainer.is( 'li' );
+
+		if ( range.endOffset === 0 && !isInList )  {
```
KarolDawidziuk, line 122: Let's add a comment with a reference to the issue that this PR is solving and I think that short information about preventing shrinking selection in lists would be also nice.

```suggestion
		var isInList = range.endContainer.is && range.endContainer.is( 'li' );
        
        // Prevent shrink selection in lists (#4931).
		if ( range.endOffset === 0 && !isInList )  {
```

Also, is this additional check `range.endContainer.is` is needed? 🤔

tests/core/selection/manual/optimizationlist.html
```diff
@@ -0,0 +1,39 @@
...
+	<ol>
+		<li>1</li>
+		<li>2
+		<ol>
+			<li>1</li>
+		</ol>
+		</li>
+		<li>&nbsp;</li>
+	</ol>
+
```
KarolDawidziuk, line 12: Let's remove this unnecessary empty line. This comment applies also to the second list below. 
```suggestion
```

tests/core/selection/manual/optimizationlist.md
```diff
@@ -0,0 +1,13 @@
+@bender-tags: selection, 4.19.2, bug, 4931
+@bender-ui: collapsed
+@bender-ckeditor-plugins: wysiwygarea, toolbar, elementspath, sourcearea, list, undo, div, table, image, basicstyles, format
```
KarolDawidziuk, line 3: Do we need to include this plugins: `div, table, image, format`?

tests/core/selection/manual/optimizationlist.md
```diff
@@ -0,0 +1,13 @@
...
+
+
+1. Select entire content via `Ctrl+A`
```
KarolDawidziuk, line 6: It's a very small detail, but let's add dots at the end of each sentence.
```suggestion
1. Select entire content via `Ctrl+A`.
```

KarolDawidziuk (changes requested):
Hi @sculpt0r 

I've made an early review of your PR.

Overall fix working very well but unfortunately, it's not working on Firefox.
It's looking like after selecting the whole content by using `CTRL+A` `range.startContainter` and `range.endContainer` are the same and they contain the whole `body` instead of the first and last element from the list(line in the other browsers). 🤔  Could you look at it?

There are some also small things that can be improved. Please see my inline comments.

I think that we missed some unit tests to cover this issue.
If I remember correctly we currently have an existing test with removing lists so we can add it here. Please see https://github.com/ckeditor/ckeditor4/blob/t/4931/tests/plugins/list/list.js#L1
It would be good to add some cases with removing the whole mixed lists, lists with some basic styles etc.

sculpt0r:
Rebased on the newest master.

core/selection/optimization.js
sculpt0r: I'll add a comment 👍 

The additional check is for the case when container is a text node - then it doesn't have `is` method which causes errors.

sculpt0r: I slightly change the comment you proposed, since at this function level we don't know nth about 'shrinking' itself.

sculpt0r:
> Overall fix working very well but unfortunately, it's not working on Firefox.
It's looking like after selecting the whole content by using CTRL+A range.startContainter and range.endContainer are the same and they contain the whole body instead of the first and last element from the list(line in the other browsers). 🤔 Could you at it?

If you take a look at those containers during select all action on FF - they are always indicating `body` element. So it looks like some other bug with FF, but this time on 'delete' action. I changed the content in the manual test. It still proves this issue  (it is even closer to the content from the report). But it won't fail in FF. With the new content selection in FF also remains set on `body` - but this time the entire list is removed. Probably we should extract another issue for that - WDYT? In other case - that require a new 'debug' session to find out what cause this behaviour on FF.

> I think that we missed some unit tests to cover this issue.
If I remember correctly we currently have an existing tests with removing lists so we can add it here.

There are cases of 'removing' lists - but with 'deselecting' list feature - not with del/backspace key. Also, this issue is more about 'selection'. Because removing lists remains the same - the problem was that the selection was improper. So I would stick just to tests related to selection optimization.

KarolDawidziuk (approved):
Thanks, @sculpt0r for the improvements.

LGTM! 🎉 

I just add some style adjustments in the manual test by myself. 

> Probably we should extract another issue for that - WDYT? In other case - that require a new 'debug' session to find out what cause this behaviour on FF.

I think that we can extract it as a follow-up to see what's happened there. On the other hand, further delving into the issue may take some time and at this moment, we do not know how long it may take, so I'd like to hear the second reviewer's opinion about it.


#### It was an early review. In the next step, this PR will be checked again by a more experienced developer.

Comandeer (changes requested):
It seems that the fix works differently depending on the browser:

1.  In Firefox it works partially – the second editor in the manual test does not work (this was already discussed in the thread and the initial consensus is to move it to a follow-up).
2.  In Chrome the fix works correctly but only when selection is made via <kbd>Ctrl</kbd>+<kbd>A</kbd>. If the selection is made via our `selectall` plugin, Chrome starts behaving like Firefox: the last item in the second editor is not removed.
3.  In Safari the fix works correctly but something odd happens to the caret – it disappears after the deletion of the list.
4.  In IE8 and IE11 fix seems to work correctly.

Due to the fact that the issue in Firefox can also be reproduced in Chrome after using the `selectall` plugin, I would be for investigating it a little (even if only to check if the selection in these two cases is the same). It would be also good to check what happens in Safari and if we can fix it somehow.

Additionally, it be also good to add a unit test with a selection causing the issue (so selecting the list with the empty element at the end).

Please also see my inline comment about possible refactor.

sculpt0r:
> Due to the fact that the issue in Firefox can also be reproduced in Chrome after using the selectall plugin, I would be for investigating it a little (even if only to check if the selection in these two cases is the same).

Select all plugin in div editor in both browsers (FF & Chrome) makes selection range.startContainer and endContainer set on `div` with editor content:

Screenshot 2022-08-23 at 09 14 42
 
 So the selection optimization early returns with `false`:
 https://github.com/ckeditor/ckeditor4/blob/a7b0d54376a794f65bce64086bd5f6bb51cb02d9/core/selection/optimization.js#L116-L117
 
 Not removing a single list element is not a problem with selection optimization.

sculpt0r:
> In Safari the fix works correctly but something odd happens to the caret – it disappears after the deletion of the list.

The selection is set on the body after content deletion:
image

However, that happens only in this manual test 🤔 not happening in our dev samples:
 
image

So the selection remains the same, but the carret is not visible 🤔

sculpt0r:

- I'm going to create a follow-ups
- Please see the above comments as a response to your recent review.
- Added automatic test
- Refactor code according to the suggestion
- Just in case (random manual test performing) added select all plugin to the manual test

Comandeer (changes requested):
@sculpt0r, could you check why the fix does not work with the `selectall` plugin? What's even more interesting, it seems to fail only for ordered lists, not unordered ones :thinking:

sculpt0r:
Rebased on the newest master

sculpt0r:
@Comandeer - as we spoke - please recheck the PR 🤞

sculpt0r:
Rebased on the newest master.

sculpt0r:
@Comandeer - as we spoke - I adjusted the manual test and reverted `ctrl+A` binding. Could you take a look at this PR again?

Comandeer:
Rebased onto the latest `master`.

tests/core/selection/manual/optimizationlist.md
```diff
@@ -0,0 +1,14 @@
+@bender-tags: selection, 4.20.0, bug, 4931
+@bender-ui: collapsed
+@bender-ckeditor-plugins: wysiwygarea, toolbar, elementspath, sourcearea, list, undo, selectall
```
Comandeer, line 3: The `selectall` plugin does not seem to be used in the test.

```suggestion
@bender-ckeditor-plugins: wysiwygarea, toolbar, elementspath, sourcearea, list, undo
```

tests/core/selection/manual/optimizationlist.md
```diff
@@ -0,0 +1,14 @@
...
+
+1. Select entire content via `Ctrl+A`.
+2. Delete content with backspace key.
+
+**Expected** The entire list content is removed from the editor.
+
+**Note** It is possible that list styling was not removed, so the first list item punctuation is in the content.
```
Comandeer, line 10: I'd rephrase this note a little because the results of unfixed code are very similar – the first list item marker is preserved. The only difference is the fact that unfixed code produces additional whitespace before it.

So maybe:

```suggestion
**Note** It is possible that list styling was not removed, so the first list item punctuation is in the content **without any preceding whitespace**.
```

Comandeer (changes requested):
The fix works fine. However, unit tests fail in IE8 – could you check it? Besides that, I have small remarks about the manual test.

sculpt0r:
Rebased on the newest master.

sculpt0r:
@Comandeer - I fixed the code for the IE8. It appears that the `endContainer` was not a `li` as in other browsers, but the text node inside the `li`. Because of that, the used method `is` does not exists on such node.

Also, adjusted the manual test. So the PR is ready for another review.

Comandeer:
Rebased onto the latest `master`.

Comandeer (approved):
LGTM!

PR #5303, opened 2022-08-02
Add config variable to control default state of Lock Ratio button

### LINKED ISSUE #5219: Control default state of Lock Ratio button
## Type of report

Feature request

Is there a way to set default state to "off" for "Lock Ratio" button

https://prnt.sc/PO7Tj9ADPvjU

Comandeer:

New feature

### This PR contains

- [x] Unit tests
- [x] Manual tests

```
* [#5219](https://github.com/ckeditor/ckeditor4/issues/5219): Added the [`config.image2_defaultLockRatio`](https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_config.html#cfg-image2_defaultLockRatio) config variable to allow setting the default value of the "Lock ratio" option in the [Enhanced Image](https://ckeditor.com/cke4/addon/image2) dialog.
```

## What changes did you make?

I've added the `config.image2_defaultLocRatio` boolean variable to allow user to control the default state of the "Lock ratio" button. The variable is set to `true` by default.

However, the change is not that straightforward. The "Lock ratio" had the second function in the image dialog – it informed the user if the loaded image has preserved its natural aspect ratio. In that case, the "Lock ratio" button was automatically set to "locked" after loading the image. Yet, it made this feature completely unusable – as even after setting the default value of the "Lock ratio" to unlocked, it was automatically switching to locked when any image was loaded. Due to that I've removed the whole mechanism of toggling lock ratio after loading the image. I'm not sure if it's the right thing to do – it is kind of a breaking change. Maybe the implementation should be updated, e.g. to preserve the after-load behavior if the config variable is not set at all?

Closes #5219.


jacekbogdanski:
@Comandeer what do you think about providing 3 state option instead? Like: 
```ts
CKEDITOR.config.image2_defaultLockRatio = OFF | ON | AUTO
```

Comandeer:
Yup, we would probably need to keep another set of tristate-like constants. In fact it's very similar to `true | false | undefined` yet I find the latter more intuitive to use – omit the option if you want the old/default behaviour, set it if you want to set a unchanging value to the "Lock ratio".

jacekbogdanski:
I think it's a fine compromise to avoid adding too many enums to the global scope. Could you implement the required changes? Right now we indeed introduce regression (and basically remove a valid feature from the image plugin).

Comandeer:
I've updated the config variable to take three possible values: `true`, `false` and `undefined`. The `undefined` is the default one, connected with the current behavior of "Lock ratio" (so indicating if the loaded image has the correct aspect ratio).

I've also rebased the branch onto latest `master`.

plugins/image2/plugin.js
```diff
@@ -378,8 +378,10 @@
 						width: image.getAttribute( 'width' ) || '',
 						height: image.getAttribute( 'height' ) || '',
 
-						// Lock ratio is on by default (https://dev.ckeditor.com/ticket/10833).
-						lock: this.ready ? helpers.checkHasNaturalRatio( image ) : true
+						// Lock ratio should respect the value of the config.image2_defaultLockRatio.
+						// If the variable is not set, then it defaults to true (#5219).
+						lock: editor.config.image2_defaultLockRatio !== undefined ?
+							editor.config.image2_defaultLockRatio : true
```
jacekbogdanski, line 384: We changed the original logic for that widget, but I'm not able to reproduce https://dev.ckeditor.com/ticket/10833. I suppose that's the same issue with using `checkHasNaturalRatio` so inability to properly test that method in modern browsers. WDYT about keeping the old behavior untouched just to make sure we won't introduce any regression with that addition?

tests/plugins/image2/manual/defaultlockratio.md
```diff
@@ -0,0 +1,20 @@
+@bender-tags: 4.20.0, feature, 5219
```
jacekbogdanski, line 1: Could you extend that test with 2 other editors and different values for the configuration option? It will let us test if everything is fine with the old behavior during the testing phase. I didn't notice any other dedicated manual test for the lock button, but we can just add the proper ticket number to the manual test if such a test is available (e.g. we could probably enable `aspectratio.html` test).

jacekbogdanski (changes requested):

Comandeer:
Rebased onto latest `master`.

Comandeer:
I've updated the manual tests. Now there are three separate tests for each value of the config variable (as testing steps are different enough for that). I've also enabled `aspectratio` test.

I've also reenabled the legacy lock behavior which will be used if the config variable is not set.

jacekbogdanski (approved):
Good job 👏🏻

PR #5306, opened 2022-08-09
Fix encoding entities

### LINKED ISSUE #4941: Some entities get wrongly encoded, when using entities_processNumerical=true
## Type of report

Bug

Open JSFiddle to see the problem: https://jsfiddle.net/veyd1gjn/10/

There is an Emoji in there 👍 and the config.entities_processNumerical is set to true.

When switching to Sourcecode of the Editor and back, the Emoji is destroyed, showing ��

In Sourcecode view it is encoded to: "&#65533;&#65533;"

The correct value would be "&#x1F44D;"

Maybe, it is because the number is larger than 0xFFFF ?

### Expected result

The expected result, that the encoded entity is correctly encoded and decoded.

### Actual result

The entity is destroyed. From 👍 to ��

## Other details

* Browser: Chrome
* OS: Windows 10
* CKEditor version: 4.5.0+
* Installed CKEditor plugins: entities

KarolDawidziuk:

Bug fix

### This PR contains

- [x] Unit tests
- [x] Manual tests

```
* [#4941](https://github.com/ckeditor/ckeditor4/issues/4941): Fix: Some entities get wrongly encoded, when using `entities_processNumerical = true`
```

## What changes did you make?

There was an issue with using `String#charCodeAt()` in the `entities` plugin that returns an integer from 0 to 65535 representing UTF-16 code unit at the given index. In the case when index was greater than 65535 method returns a surrogate pair. To fix this I used `codePointAt()` which returns a Unicode code point value at the given index.

Another issue was with the regex which was unable to match and replace given unicodes. To fix this I updated existing entities regex by adding a `u`- unicode option. 

Unfortunately, this is not working on IE because `codePointAt()` and Unicode option are not supported there. Although it is possible to correctly convert HTMLEntity based on the values from the surrogate pair(see [d148812](https://github.com/ckeditor/ckeditor4/commit/d148812ee33f5231e4193a1a2acbd96f3f81e489#diff-d14edc554ac7a26f0fe8a2787c381907902340da10665f40992a0e417dfb240aR113-R137)) I was unable to find a solution to support the Unicode option on IE.

Closes #4941.

### DIFF
--- a/plugins/entities/plugin.js
+++ b/plugins/entities/plugin.js
@@ -106,8 +106,33 @@
 			}
 
 			function getEntity( character ) {
-				return config.entities_processNumerical == 'force' || !entitiesTable[ character ] ? '&#' + character.charCodeAt( 0 ) + ';'
+				return config.entities_processNumerical == 'force' || !entitiesTable[ character ] ? '&#' + getCodePoint( character ) + ';'
 				: entitiesTable[ character ];
+			}
+
+			// (#4941)
+			function getCodePoint( character ) {
+				if ( !CKEDITOR.env.ie ) {
+					return character.codePointAt( 0 );
+				}
+
+				// IE does not support codePointAt.
+				var code = character.charCodeAt( 0 );
+				if ( code < 0xD800 || code > 0xDFFF ) {
+					return code;
+				}
+
+				var highSurrogate = code;
+				var lowSurrogate = character.charCodeAt( 1 );
+
+				// Check if high and lead surrogate are in the range.
+				if ( !( highSurrogate >= 0xD800 && highSurrogate <= 0xDBFF && lowSurrogate >= 0xDC00 && lowSurrogate <= 0xDFFF ) ) {
+					return code;
+				}
+
+				// Reverse mapping from a surrogate pair to a Unicode code point.
+				// Unicode 3.0.0 Chapter 3.7 Surrogate Pairs.
+				return ( highSurrogate - 0xD800 ) * 0x400 + lowSurrogate - 0xDC00 + 0x10000;
 			}
 
 			var dataProcessor = editor.dataProcessor,
@@ -143,7 +168,9 @@
 				if ( config.entities && config.entities_processNumerical )
 					entitiesRegex = '[^ -~]|' + entitiesRegex;
 
-				entitiesRegex = new RegExp( entitiesRegex, 'g' );
+
+				// IE does not support unicode option in the Regex constructor (#4941).
+				entitiesRegex = CKEDITOR.env.ie ? new RegExp( entitiesRegex, 'g' ) : new RegExp( entitiesRegex, 'gu' );
 
 				// Decode entities that the browsers has transformed
 				// at first place.
--- a/plugins/entities/plugin.js
+++ b/plugins/entities/plugin.js
@@ -118,54 +118,57 @@
 			var dataProcessor = editor.dataProcessor,
 				htmlFilter = dataProcessor && dataProcessor.htmlFilter;
 
-			if ( htmlFilter ) {
-				// Mandatory HTML basic entities.
-				var selectedEntities = [];
-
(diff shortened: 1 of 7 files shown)


sculpt0r (changes requested):
@KarolDawidziuk - I need to ask you to add some tests. I didn't look at the code much more (but we discussed it ;) ).

We missed tests (since the beginning, not only for this PR) for [`force`](https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_config.html#cfg-entities_processNumerical) option 🤔 It is mentioned in docs but our tests only use `true` value. So before we touch the code even more - let's establish the current logic with tests 🤞

KarolDawidziuk:
@sculpt0r I've added more unit tests with `force` option and also I added a separate manual test to cover that case and PR is ready for another check.

KarolDawidziuk:
@sculpt0r I've added more unit tests with `force` option and also I added a separate manual test to cover that case and PR is ready for another check.

sculpt0r:
Rebased on the newest master.

sculpt0r:
> I've added more unit tests with force option
Looks good - thank you for that :)

plugins/entities/plugin.js
```diff
@@ -143,7 +148,9 @@
 				if ( config.entities && config.entities_processNumerical )
 					entitiesRegex = '[^ -~]|' + entitiesRegex;
 
-				entitiesRegex = new RegExp( entitiesRegex, 'g' );
+
+				// IE does not support unicode option in the Regex constructor (#4941).
+				entitiesRegex = CKEDITOR.env.ie ? new RegExp( entitiesRegex, 'g' ) : new RegExp( entitiesRegex, 'gu' );
```
sculpt0r, line 153: Lets make this code less repeatable:
```suggestion
				entitiesRegex = new RegExp( entitiesRegex, CKEDITOR.env.ie ? 'g' : 'gu' );
```

tests/plugins/entities/entities.js
```diff
@@ -82,5 +82,115 @@ bender.test( {
...
+				expectedHtml =  '<p>&#192;&#193;&#206;&#207;&#208;</p>',
+				editor = bot.editor;
+
+			bot.setData( inputHtml, function() {
+				assert.areEqual( expectedHtml, editor.getData() );
+			} );
+		} );
+	},
+
+	// (#4941)
```
sculpt0r, line 141: Let's remove issue reference from tests that were added just for sake of this feature coverage. So this one and  the two others with `force` param.

tests/plugins/entities/manual/entitiesprocessnumericalforce.md
```diff
@@ -0,0 +1,10 @@
+@bender-tags: bug, 4.20.0, 4941
```
sculpt0r, line 1: I wouldn't use issue tag here. This test is just for sake of coverage entities feature.

sculpt0r (changes requested):
@KarolDawidziuk - FF, Safari, Chrome looks good 👍 IE also is not blown up, so another 👍 

Please take a look at my final inline comments. Sorry for dragging you through writing tests not explicitly related to the issue itself, but we never know when we come back here. So if we have already deep dive into this plugin - we might add tests that will cover this feature for future changes - hope you understand 😓

tests/plugins/entities/manual/entitiesprocessnumericalforce.md
KarolDawidziuk: I think that it's correct now because even if it's for coverage purposes, it's still broken while using `force` option, and this PR fixing it. WDYT?

KarolDawidziuk:

I've applied changes mentioned in your comments except a change with bender tags in the manual test and PR is ready for another check.

sculpt0r (approved):
Thanks for all new test cases 👍 

LGTM 🎉

PR #5312, opened 2022-08-22
Disallow adding subscript and superscript at the same time

### LINKED ISSUE #5215: Disallow subscript and superscript at the same time
## Type of report

Feature request

Currently, enabling subscript and superscript on the same content will result in producing the below HTML:

```
<p><sup><sub>foobar</sub></sup></p>
```

Many popular text editors (like MS Word) don't allow to style text with subscript and superscript at the same time. It's not recommended to change the current implementation, but we could introduce an opt-in configuration option that would give more freedom in choosing proper behavior:

1. Select text and apply subscript:
`<p><sub>foobar</sub></p>`
2. Select the same text and apply superscript:
`<p><sup>foobar</sup></p>`

KarolDawidziuk:

New feature

### This PR contains

- [x] Unit tests
- [x] Manual tests

```
* [#5215](https://github.com/ckeditor/ckeditor4/issues/5215): Added a new config option to disallow adding subscript and superscript at the same time.
```

## What changes did you make?

I've added a new config option `coreStyles_disallowSubscriptSuperscript` with default value sets to `false`. default configuration will allow adding subscript and superscript like before. To change this we need to set this option to `true`.

Closes #5215.

### DIFF
--- a/plugins/basicstyles/plugin.js
+++ b/plugins/basicstyles/plugin.js
@@ -107,32 +107,40 @@ CKEDITOR.plugins.add( 'basicstyles', {
 		var subscriptCommand = editor.getCommand( 'subscript' );
 		var superscriptCommand = editor.getCommand( 'superscript' );
 
+		// Prevent adding subscript and superscript only when both buttons exists. (#5215)
 		if ( subscriptCommand && superscriptCommand ) {
-			subscriptCommand.on( 'state', function() {
-				if ( superscriptCommand.state == CKEDITOR.TRISTATE_ON ) {
-					removeSubSup( 'sup' );
-					// superscriptCommand.setState( CKEDITOR.TRISTATE_OFF );
-					// editor.selectionChange();
-				}
-			} );
-
-			superscriptCommand.on( 'state', function() {
-				if ( subscriptCommand.state == CKEDITOR.TRISTATE_ON ) {
-					removeSubSup( 'sub' );
-					// subscriptCommand.setState( CKEDITOR.TRISTATE_OFF );
-					// editor.selectionChange();
+			editor.on( 'afterCommandExec', function( evt ) {
+				if ( evt.data.name === 'subscript' ) {
+					if ( superscriptCommand.state == CKEDITOR.TRISTATE_ON ) {
+						removeSubSup( 'sup' );
+					}
+				} else if ( evt.data.name === 'superscript' ) {
+					if ( subscriptCommand.state == CKEDITOR.TRISTATE_ON ) {
+						removeSubSup( 'sub' );
+					}
 				}
 			} );
 		}
 
-		function removeSubSup( element ) {
+		function removeSubSup( elementName ) {
 			var selection = editor.getSelection(),
 				range = selection.getRanges()[ 0 ],
-				style = new CKEDITOR.style( { element: element } );
+				style = new CKEDITOR.style( { element: elementName } );
 
-			style.removeFromRange( range, editor );
+			range.enlarge( CKEDITOR.ENLARGE_INLINE );
 
-			if ( element === 'sup' ) {
+			var walker = new CKEDITOR.dom.walker( range ),
+				element = range.startContainer;
+
+			// Remove all subscript or superscript style within the range.
+			while ( element ) {
+				if ( element.$.nodeName.toLowerCase() === elementName ) {
+					editor.removeStyle( style )
+				}
+				element = walker.next();
+			}
+
+			if ( elementName === 'sup' ) {
 				superscriptCommand.setState( CKEDITOR.TRISTATE_OFF );
 			} else {
 				subscriptCommand.setState( CKEDITOR.TRISTATE_OFF );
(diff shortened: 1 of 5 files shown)


tests/plugins/basicstyles/manual/togglesubsupscript.md
```diff
@@ -0,0 +1,19 @@
...
+
+**Expected** Selected text contain only `superscript` style.
+
+**Unexpected** Selected text contain `subscript` and `superscript` style.
+
+4. Click `undo` button.
+
+**Expected** There is only one step needed to back to enabled `subscript` button.
+
+**Unexpected** There is one more additional step needed to back to enabled `subscript` button.
```
sculpt0r, line 17: Seems to be redundant - could we remove it?

plugins/basicstyles/plugin.js
```diff
@@ -101,6 +101,41 @@ CKEDITOR.plugins.add( 'basicstyles', {
...
+						editor.fire( 'lockSnapshot' );
+					}
+				}
+
+				if ( evt.data.name === 'superscript' ) {
+					if ( subscriptCommand.state === CKEDITOR.TRISTATE_ON ) {
+						editor.execCommand( 'subscript' );
+						editor.fire( 'lockSnapshot' );
+					}
+				}
```
sculpt0r, line 128: Those two blocks seems to be identical, could we extract them to a single function with name adjusted to the abstraction level of this code, so:
```suggestion
				if ( evt.data.name === 'subscript' ) {
					offActiveCommand( superscriptCommand );
				}

				if ( evt.data.name === 'superscript' ) {
					offActiveCommand( subscriptCommand );
				}
```

and the function itself (improved based on [API](https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_command.html#method-exec)):
```js
function offActiveCommand( command ){
				if ( command.state === CKEDITOR.TRISTATE_ON ) {
					command.exec( editor );
					editor.fire( 'lockSnapshot' );
				}
			}
```

plugins/basicstyles/plugin.js
```diff
@@ -208,3 +243,12 @@ CKEDITOR.config.coreStyles_subscript = { element: 'sub' };
  * @member CKEDITOR.config
  */
 CKEDITOR.config.coreStyles_superscript = { element: 'sup' };
+
+/**
+ * Disallow adding subscript and superscript on one element at the same time.
```
sculpt0r, line 248:
```suggestion
 * Disallow setting subscript and superscript simultaneously on the same element.
```

sculpt0r (changes requested):
@KarolDawidziuk - looks like your solution is working very well :)

I left some comments regarding test cases and refactoring stuff. Please adjust my suggestions in parts that are not fit in your opinion.

plugins/basicstyles/plugin.js
```diff
@@ -101,6 +101,41 @@ CKEDITOR.plugins.add( 'basicstyles', {
 			[ CKEDITOR.CTRL + 73 /*I*/, 'italic' ],
 			[ CKEDITOR.CTRL + 85 /*U*/, 'underline' ]
 		] );
+	},
+
+	afterInit: function( editor ) {
+		var subscriptCommand = editor.getCommand( 'subscript' ),
+			superscriptCommand = editor.getCommand( 'superscript' ),
+			disallowAddSubSup = editor.config.coreStyles_disallowSubscriptSuperscript;
```
jacekbogdanski, line 109: We can't go with that name for the configuration option. It indicates that the option disables sub and sup script features, not make then exclusive.

tests/plugins/basicstyles/basicstyles.js
```diff
@@ -7,6 +7,8 @@ bender.editor = {
 	}
 };
 
+var isNotGeckoOrIe = !CKEDITOR.env.gecko && !CKEDITOR.env.ie;
```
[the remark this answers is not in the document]
KarolDawidziuk: That's interesting. It looks like it was some cache because when I rerun these tests without `isNotGeckoOrIe` in `getData` it works so, I'll remove it.

tests/plugins/basicstyles/basicstyles.js
```diff
@@ -25,5 +27,203 @@ bender.test( {
...
+		bender.editorBot.create( {
+			name: 'editor-subsup2',
+			config: {
+				coreStyles_disallowSubscriptSuperscript: true
+			}
+		}, function( bot ) {
+			var editor = bot.editor;
+
+			bot.setHtmlWithSelection( '<p>[foo] bar</p>' );
+			editor.execCommand( 'bold' );
```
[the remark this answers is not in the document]
KarolDawidziuk: It was on purpose to check if toggling does not change subscript and superscript behavior. I'll adjiust this test but IMO it would be good to leave the rest tests that toggle sub and sup as it now, but if you feel that it's unnecessary I can adjust it.

KarolDawidziuk:

I've applied the improvements mentioned in your comments and PR is ready for another review.

The one thing worth mentioning is the fact that I've renamed and reversed the behavior of the config option. As we decided with @jacekbogdanski in our F2F talk, in this particular case disabling this feature is not entirely good, because it should work like with this fix by default. If someone needs to add `sup` and `sub` at the same time, there will be an available config option for it.

sculpt0r (approved):
@KarolDawidziuk - thanks for corrects 👍 

I think we can have the automatic tests as you created them 👍 

Because of https://github.com/ckeditor/ckeditor4/pull/5312#issuecomment-1241536511 - I'm approving the current state of the config option. LGTM 🎉

PR #5325, opened 2022-09-14
Improvement for allowing subscript and superscript at the same time

KarolDawidziuk:
It's a follow-up to https://github.com/ckeditor/ckeditor4/pull/5312

Changes:
* Renamed config option to `coreStyles_toggleSubSup`.
* Fixed a bug that removed sup and superscript elements when pasting content containing these elements via sourcemode and then attempting to disable one of them in toolbar.
Screencast with bug:

https://user-images.githubusercontent.com/18471998/190137470-66757ee2-3ec0-4e3e-9605-88e3425fcf45.mov

* Added unit test to cover fix for the above issue.
* Updated changelog.


jacekbogdanski:
As we are close to the release, I let myself help with the initial implementation of that feature, as it was unnecessarily overcomplicated and I still managed to reproduce some issues with undo. @KarolDawidziuk could you check if the changes are ok?

jacekbogdanski (approved):

rklf:
We have now no way since ``coreStyles_toggleSubSup`` got removed in CKEditor 5 to disallow subscript and superscript at the same time? Nothing [here](https://ckeditor.com/docs/ckeditor5/latest/updating/ckeditor4/ckeditor4-configuration-compatibility.html) to achieve that in CKEditor 5.

jacekbogdanski:
@rklf CKEditor 5 is not a new version of the same editor, but an entirely separate product. The editor is compatible with CKEditor 4 in most cases (feature-wise) but some of them may not work the same. If you would like to have something similar implemented in CKEditor 5, please report it in [the CKEditor 5 issue tracker.](https://github.com/ckeditor/ckeditor5/issues)

PR #5296, opened 2022-07-27
Add `scope` attribute to table header cells

sculpt0r:
## What is the purpose of this pull request?

New feature
## Does your PR contain necessary tests?

All patches that change the editor code must include tests. You can always read more
on [PR testing](https://ckeditor.com/docs/ckeditor4/latest/guide/dev_contributing_code.html#tests),
[how to set the testing environment](https://ckeditor.com/docs/ckeditor4/latest/guide/dev_tests.html) and
[how to create tests](https://ckeditor.com/docs/ckeditor4/latest/guide/dev_tests.html#creating-your-own-test)
in the official CKEditor documentation.

### This PR contains

- [X] Unit tests
- [X] Manual tests

## Did you follow the CKEditor 4 code style guide?

Your code should follow the guidelines from the [CKEditor 4 code style guide](https://github.com/ckeditor/ckeditor4/blob/major/dev/docs/codestyle.md) which helps keep the entire codebase consistent.

- [X] PR is consistent with the code style guide

## What is the proposed changelog entry for this pull request?

```
* [#5084](https://github.com/ckeditor/ckeditor4/issues/5084): Introduce `scope` attribute to table header cells. `Header` cell type is split into two types: `Column Header` & `Row Header`.
```

## What changes did you make?

*Give an overview…*

## Which issues does your PR resolve?

Closes #5084.

PR #5328, opened 2022-09-15
Introduce `CKEDITOR.config.tabletools_scopedHeaders`

Comandeer:

New feature

### This PR contains

- [x] Unit tests
- [x] Manual tests

```
* [#5084](https://github.com/ckeditor/ckeditor4/issues/5084): Added the [`config.tabletools_scopedHeaders`](https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_config.html#cfg-tabletools_scopedHeaders) configuration option controlling the behaviour of table headers with and without the `[scope]` attribute.
```

## What changes did you make?

I've introduced the `CKEDITOR.config.tabletools_scopedHeaders` config variable which controls how the UI of "Cell Properties" dialog identify the `th` elements without the `[scope]` attribute:

* `true` means that such cells will be represented as an empty option and available values are "Data", "Column Header" and "Row Header",
* `false` means that such cells will be represented as "Header" option and available values are "Data" and "Header".

 I've also added detailed API docs for it.

### DIFF
--- a/plugins/tabletools/plugin.js
+++ b/plugins/tabletools/plugin.js
@@ -1271,6 +1271,46 @@ CKEDITOR.tools.buildTableMap = function( table, startRow, startCell, endRow, end
 };
 
 /**
+ * Indicates if table header elements (`th`) needs the
+ * [`scope` attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/th#attr-scope).
+ *
+ * This config variable changes the available values of the "Cell Type" field inside the
+ * "Cell Properties" dialog. If it's set to `false` (the default value), the "Cell Type" field
+ * will contain two options:
+ *
+ * * "Data",
+ * * "Header".
+ *
+ * If the option is set to `true`, the "Cell Type" field in the "Cell Properties" dialog
+ * will contain three possible values:
+ *
+ * * "Data",
+ * * "Column Header",
+ * * "Row Header".
+ *
+ * Additionally, if this config variable is set to `true` and there is a `th` element without the
+ * `scope` attribute in the editor's content, its "Cell Type" value will be set to an empty value.
+ * To fix this, the `th` element needs to be transformed to gain the `scope` attribute.
+ * The sample transformation that adds `[scope=col]` to all scopeless `th` elements is presented below:
+ *
+ * ```javascript
+ * editor.filter.addTransformations( [
+ * 	[
+ * 		{
+ * 			element: 'th',
+ * 			left: function( el ) {
+ * 				return !el.attributes.scope;
+ * 			},
+ * 			right: function( el ) {
+ * 				el.attributes.scope = 'col';
+ * 			}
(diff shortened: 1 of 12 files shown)


plugins/tabletools/plugin.js
```diff
@@ -1269,3 +1269,50 @@ CKEDITOR.tools.buildTableMap = function( table, startRow, startCell, endRow, end
...
+ * will contain three possible values:
+ *
+ * * "Data",
+ * * "Column Header",
+ * * "Row Header".
+ *
+ * Additionally, if this config variable is set to `true` and there is a `th` element without the
+ * `scope` attribute in the editor's content, its "Cell Type" value will be set to an empty value.
+ * To fix this, the `th` element needs to be transformed to gain the `scope` attribute.
+ * The sample transformation that adds `[scope=col]` to all scopeless `th` elements is presented below:
```
jacekbogdanski, line 1294:
```suggestion
 * Changes the available values of the "Cell Type" field inside the
 * "Cell Properties" dialog. If it's set to `false` (the default value), the "Cell Type" field
 * will contain two options:
 *
 * * "Data",
 * * "Header".
 *
 * If the option is set to `true`, the "Cell Type" field in the "Cell Properties" dialog
 * will contain three possible values:
 *
 * * "Data",
 * * "Column Header",
 * * "Row Header".
 *
 * Column and row header options updates table headers (`th`) with the
 * [`scope` attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/th#attr-scope)
 * that may improve accessibility experience in more complex tables. Read [w3.org guide about using the scope attribute to associate header cells and data cells in data tables](https://www.w3.org/WAI/WCAG21/Techniques/html/H63) to learn more.
 *
 * If this config variable is set to `true` and there is a `th` element without the
 * `scope` attribute in the editor's content, its "Cell Type" value will be set to an empty value.
 * To avoid that issue, tables with `th` elements need to be migrated.
 * The sample transformation that adds `[scope=col]` to all scopeless `th` elements is presented below:
```

jacekbogdanski: I could make some small mistakes due to editing in GH, so please review that suggestion before applying.

jacekbogdanski (changes requested):
Looks good 👍🏻 Just some minor issues:

Manual tests added in https://github.com/ckeditor/ckeditor4/pull/5296 requires an update:

* `tests/plugins/tabletools/manual/allowedcontent.html` (here, I would just probably revert changes)
* `tests/plugins/tabletools/manual/cellproperties.html` (ditto)
* `tests/plugins/tabletools/manual/cellpropertiescelltype.html` manual steps should be rewritten and we should enable scoped headers

Comandeer:
I've updated manual tests and API docs.

jacekbogdanski (approved):
Good job @Comandeer 👏🏻

PR #5331, opened 2022-09-20
Fire change event after upload completes

### LINKED ISSUE #1180: getData() wrong response right after image dropped

Bug 

## [Check if the issue is already reported]
I've searched for hours

User drops image to the editor, the image uploading successfully, the image tag is missing from the getData() response.

1. Drop image to the editor --> Upload successful, the image is appearing in the editor, the image's URL changing to the uploaded image's URL.
2. editor.getData() will print the content without the inserted image tag some reason. The complete 

`` html tag is missing from the `getData()` returned value. The rest part of content is okay, only the last dropped image's html tag is missing. 

If I drop an image to the editor, then I change anything in the editor, this issue disappears.
I suppose ckeditor forget to update something after the successful image uploading. 

I tried fire change,saveSnapshot,key events after 'fileUploadResponse' event and before getData() function, without any luck.

### Expected result

getData() function should return with the real content of ckeditor after image dropped

### Actual result

It returns without image

## Other details

* Browser: Chrome
* OS: W10
* CKEditor version: newest
* Installed CKEditor plugins: 
CKEDITOR.config.extraPlugins = 'sourcedialog,jsplus_image_editor,filebrowser,uploadimage';

rrg92:

This allows user to get the updated data after image is uploaded.

Today, there is a bug, that after the upload completes, the call to getData() returns an old HTML version (without the image tag updated...)

This solve the  issue #1180

### This PR contains

- [ ] Unit tests
- [x] Manual tests

Test:
- Create a simplest editor with image and imageupload pluging
- Configure upload to some valid upload server
- Add the following event listener:
```js
editor.on('change',function(e){  console.log(e.editor.getDate())  })
```
- Paste some image into editor (and observe)
- Before this fix, 'change' is not fired
- After this fix, 'change' is fired and user can handle new data.

```
* [#1180](https://github.com/ckeditor/ckeditor4/issues/1180): Fix bug when uploading pasting data and change event not fired correctly.
```

## What changes did you make?

Just added a call to fire method of editor instance, firing the "change" event.

Closes #1180

### DIFF
--- a/plugins/uploadimage/plugin.js
+++ b/plugins/uploadimage/plugin.js
@@ -88,6 +88,9 @@
 					this.replaceWith( '<img src="' + upload.url + '" ' +
 						'width="' + width + '" ' +
 						'height="' + height + '">' );
+					
+					// Fire change event in order to get updated content.
+					editor.fire('change');
 				}
 			} );
 


plugins/uploadimage/plugin.js
```diff
@@ -88,6 +88,9 @@
 					this.replaceWith( '<img src="' + upload.url + '" ' +
 						'width="' + width + '" ' +
 						'height="' + height + '">' );
+					
+					// Fire change event in order to get updated content.
```
KarolDawidziuk, line 92: Let's add an issue reference.
```suggestion
  // Fire change event in order to get updated content(#1180).
```

KarolDawidziuk (changes requested):
Hi @rrg92 and thanks for your contribution and sorry for the late reply.

Fix looks good and HTML with uploaded image is visible after using `editor.getData()` on `change` event.

Please remember that code that goes to the `master` branch requires test coverage.
I encourage you to add this part and I'll try to help with it.

Let's add unit tests inside the `uploadimage.js` file, there are a lot of existing tests that you can try to reuse in this case.

Manual tests are really helpful to check live if this fix does the job and to create a manual test we need to create two files with the same name but with another extension(`.html` and `.md`).

Helpful links:
* Guite to test environment: https://ckeditor.com/docs/ckeditor4/latest/guide/dev_tests.html
* Similar test with uploading images: https://github.com/ckeditor/ckeditor4/blob/master/tests/plugins/uploadwidget/manual/droponnotallowedparts.html#L3

Issue #4829, opened 2021-08-12
Undo reverse entire table content instead of single cell


sculpt0r:
## Type of report

Bug

## Provide detailed reproduction steps (if any)

1. Create a standard table
2. Write some text in a single cell
3. Jump to the next cell with the 'tab' key
4. Write text to each cell
5. Use undo button

### Expected result

*What is the expected result of the above steps?*
Text from the last cell is removed. All other cells contain text.
There is undo step for each cell.

### Actual result

*What is the actual result of the above steps?*
There is a single undo step that reverses the entire input from the table.

## Other details

* Browser: -
* OS: -
* CKEditor version: PROD (pre-released) 4.16.2
* Installed CKEditor plugins: table

jacekbogdanski:
Could be probably fixed similarly like https://github.com/ckeditor/ckeditor4/blob/90e680e17d9d385077e47edd5dde60efa607e8c7/plugins/undo/plugin.js#L1125-L1131

PR #5349, opened 2022-10-12
Fix for caret moving to the previous cell after tabbing into the next cell and then removing its content

### LINKED ISSUE #4802: Cursor moves to the previous cell inside table on pressing BACKSPACE after removing text.
### Type of Report

Bug

### Reproduction Steps

https://ckeditor.com/latest/samples/

1. Fill in both columns with some text. 
2. Click on the first cell then press the TAB. (Next cell will turn blue)
3. Now press BACKSPACE to remove the text inside this cell.
4. Observe the cursor.

Please see the attached gif.

ezgif com-gif-maker

### Expected Result

The cursor should stay in the same cell after removing text

### Actual Result

The cursor moves to the previous cell. 

### Other Details

- Browser: Version 91.0.4472.164 (Official Build) (64-bit)
- OS: Windows 10 Version 1090 (OS Build 18363.836)
- CKEditor version: 14.14.1
- Installed CKEditor plugins: table, tableselection, tab

### Note

The issue occurred after integrating TAB plugin

Comandeer:

Bug fix

### This PR contains

- [ ] Unit tests
- [x] Manual tests

```
* [#4802](https://github.com/ckeditor/ckeditor4/issues/4802): Fixed: Caret moves to the previous cell after tabbing into the next cell and then removing its content.
```

## What changes did you make?

I've added the`keyup` event listener that updates fake selection after pressing the <kbd>Tab</kbd> key.

I added only a manual test as I wasn't able to come up with a sensible automated one.

Closes #4802.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -7,6 +7,7 @@ Fixed Issues:
 
 * [#2881](https://github.com/ckeditor/ckeditor4/issues/2881): Fixed: Changing table headers from "Both" to "First column" in the [Table](https://ckeditor.com/cke4/addon/table) dialog does not change the first column cell correctly.
 * [#2996](https://github.com/ckeditor/ckeditor4/issues/2996): Fixed: Table header "scope" attribute is incorrect for the "Headers: both" option in the [Table](https://ckeditor.com/cke4/addon/table) dialog.
+* [#4802](https://github.com/ckeditor/ckeditor4/issues/4802): Fixed: [Tableselection](https://ckeditor.com/cke4/addon/tableselection) caret moves to the previous cell after tabbing into the next cell and then removing its content.
 
 ## CKEditor 4.20
 
--- a/plugins/tableselection/plugin.js
+++ b/plugins/tableselection/plugin.js
@@ -1149,6 +1149,21 @@
 				}
 			}
 
+			function getTableKeyUpListener( editor ) {
+				return function( evt ) {
+					var key = evt.data.getKey(),
+						selection = editor.getSelection();
+
+					// Handle only Tab key in table selection.
+					if ( key !== 9 || !selection.isInTable() ) {
+						return;
+					}
+
+					// Restore fake selection after pressing Tab (#4802).
+					restoreFakeSelection( editor );
+				};
+			}
+
 			function clearCellInRange( range ) {
 				var node = range.getEnclosedNode();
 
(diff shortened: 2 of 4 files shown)


KarolDawidziuk (changes requested):
Hi, @Comandeer 

I've made an early review of your PR.
Overall, great job 👍 code looks simple, clean and solves this issue.

There are some small things that we can improve so please see my inline comments.

There is a case when `undo` works incorrectly - removing content from more cells by using `tab` key and `backspace`. Undo step will restore all content instead of only one cell, but I see that we already have reported issue for that. https://github.com/ckeditor/ckeditor4/issues/4829 

As we talked, it would be good to add in some way unit tests because there is a chance that some future change will break this. 

I make a few approaches and it looks like actually we can handle it by using `undoEventDispatchTestsTools`. 
You can base on tests inside `keyboard.js` file, eg. https://github.com/ckeditor/ckeditor4/blob/t/4802/tests/plugins/tableselection/keyboard.js#L164-L179

Based on this, the part of using `tab` and removing next cell using `backspace` can looks like:
```JS
// Added `tab` to /* `bender-ckeditor-plugins` */ in the top of the file.

'some test': function( editor, bot ) {
  bender.tools.testInputOut( 'enterKey', function( source, expected ) {
  bender.tools.setHtmlWithSelection( editor, source );
  
  editor.getSelection().selectRanges( getRangesForCells( editor, [ 0 ] ) );				
  
  var keyTools = undoEventDispatchTestsTools( { editor: editor } );

  keyTools.key.keyEvent( 9 ); // it would be good to update keyCodesEnum by adding TAB.
  keyTools.key.keyEvent( keyTools.key.keyCodesEnum.BACKSPACE );
  
 // ... Some asserts
  } );
}

```

Comandeer:
@KarolDawidziuk, thanks for the review. I've updated the manual test but I didn't manage to create a working unit test.

The approach you suggested allows emulating keyboard events but it does not mimic the default behaviors (as they are fired only for trusted events, not synthetic ones). And setting the selection manually does not work – in such a case the test passes even when the Tab-handling logic is disabled. Due to that I skipped adding a unit test.

KarolDawidziuk (approved):
Thanks, @Comandeer for the improvements.
LGTM! 🎉 

Unfortunately, as you mention, we must skip unit tests in this case.

This was an early review. In the next step, this PR will be checked again by a more experienced developer.

PR #5364, opened 2022-11-07
Fix broken integration between `uploadimage` and `clipboard`

### LINKED ISSUE #5333: Broken integration between `uploadimage` and `clipboard` plugin
## Type of report

Bug

1. Open https://ckeditor.com/docs/ckeditor4/latest/examples/fileupload.html#uploading-dropped-and-pasted-images
2. Drag and drop any image into the editor.
3. Wait for the upload to finish and then check the source of the editor.

### Expected result

The image has preserved its original name.

### Actual result

The image has a generic name in the format of `image-<datetime>-<no>`.

## Other details

* Browser: N/A
* OS: N/A
* CKEditor version: 4.17.0+
* Installed CKEditor plugins: clipboard, uploadimage, uploadwidget

Similar to #4874 but much more subtle.

Due to how our clipboard logic works, handling images inside the `clipboard` plugin prevents the `uploadimage`/`uploadwidget` plugins from getting image files from the clipboard. This causes the `uploadimage` to use its fallback logic: https://github.com/ckeditor/ckeditor4/blob/52fa6b470589f6b086a29558ef88d67213024906/plugins/uploadimage/plugin.js#L94-L143

The solution introduced in the #4874 (setting the [`config.clipboard_handleImages`](https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_config.html#cfg-clipboard_handleImages) to `false`) works but the broken behavior _seems_ to work as long as the original name of the file is not important. It's so deceiving that we missed it for several releases in our docs… Due to that, I think we should revisit the issue and think about better integration between the `clipboard` and `uploadimage`/`uploadwidget` plugins.

Comandeer:

Bug fix

### This PR contains

- [x] Unit tests
- [x] Manual tests

```
* [#5333](https://github.com/ckeditor/ckeditor4/issues/5333): Fixed: original name of the uploaded image was not preserved in the [Upload Image](https://ckeditor.com/cke4/addon/uploadimage) plugin if the [Clipboard](https://ckeditor.com/cke4/addon/clipboard) plugin had enabled image handling.
```

## What changes did you make?

I've introduced the new error code, `clipboard-image-handling-disabled`, that is used when the `uploadimage` plugin detects that the `clipboard` plugin has enabled image handling. In that case the mentioned warning is emitted and the clipboard's handling logic is disabled (by setting the `editor.config.clipboard_handleImages` config variable to `false`).

It's probably not the _best_ solution for the issue but it's the fastest one and it guarantees that there will be no conflict between these two plugins.

~~If we decide that it's the right way to do it, I'll prepare also the second PR for docs to introduce the new error code there.~~ Docs PR: ckeditor/ckeditor4-docs#409

Closes #5333.

### DIFF
--- a/tests/plugins/uploadimage/manual/clipboardintegration.md
+++ b/tests/plugins/uploadimage/manual/clipboardintegration.md
@@ -1,0 +1,16 @@
+@bender-tags: 4.20.1, bug, 5333
+@bender-ui: collapsed
+@bender-ckeditor-plugins: wysiwygarea, toolbar, uploadwidget, uploadimage, image, floatingspace, toolbar, sourcearea
+
+1. Open the console.
+
+**Expected** There is `[CKEDITOR] Error code: clipboard-image-handling-disabled.` warning in the console after the editor is loaded.
+
+**Unexpected** There is no warning in the console.
+2. Drag and drop an image into the editor.
+
+**Expected** The original name of the file is displayed above the editor.
+
+**Unexpected** The name in the `image-<timestamp>-<n>.<ext>` format is displayed above the editor.
+
+3. Repeat step `2` for the paste method.
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -5,6 +5,7 @@ CKEditor 4 Changelog
 
 Fixed Issues:
 
+* [#5333](https://github.com/ckeditor/ckeditor4/issues/5333): Fixed: The original name of the uploaded image is not preserved by the [Upload Image](https://ckeditor.com/cke4/addon/uploadimage) plugin if the [Clipboard](https://ckeditor.com/cke4/addon/clipboard) plugin has enabled image handling.
 * [#2881](https://github.com/ckeditor/ckeditor4/issues/2881): Fixed: Changing table headers from "Both" to "First column" in the [Table](https://ckeditor.com/cke4/addon/table) dialog does not change the first column cell correctly.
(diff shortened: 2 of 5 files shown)


tests/plugins/uploadimage/manual/clipboardintegration.md
```diff
@@ -0,0 +1,16 @@
...
+**Expected** There is `[CKEDITOR] Error code: clipboard-image-handling-disabled.` warning in the console after the editor is loaded.
+
+**Unexpected** There is no warning in the console.
+2. Drag and drop an image into the editor.
+
+**Expected** The original name of the file is displayed above the editor.
+
+**Unexpected** The name in the `image-<timestamp>-<n>.<ext>` format is displayed above the editor.
+
+3. Repeat step `2` for the paste method.
```
jacekbogdanski, line 16: Will paste actually work? For me, it embeds the original image URL. Not sure if that case wouldn't require configuring the upload adapter with dialog:
image

jacekbogdanski (changes requested):
Looks like a good approach. We could probably come up with something more complicated like delegating that logic to the clipboard (similarly like we did for paste notification) but that doesn't seem necessary.

In the documentation, it would be good to write how to get rid of the warning.

tests/plugins/uploadimage/manual/clipboardintegration.md
Comandeer: It works for me on macOS 🤔 But AFAIR it could work differently on other OSes.

Comandeer:
I've prepared the PR for docs → ckeditor/ckeditor4-docs#409

PR #5368, opened 2022-11-22
Improve Tab navigation bahaviour with radio buttons inside the dialog 

### LINKED ISSUE #439: Incorrect Tab navigation behavior with radio buttons inside the dialog

Bug

http://dev.ckeditor.com/ticket/13839

1. Open Creating Captioned Images sample http://sdk.ckeditor.com/samples/captionedimage.html
1. Double click on the image to get 'Image Properties' dialog
1. Use Tab key to navigate inside the radio button group
1. Keep pressing Tab key

### Expected result

Tab key will enter the radio group.

	When Tab or Shift+Tab into a radio group, focus goes to the selected radio button. If none is selected, focus goes to the first radio button if Tab was pressed, or the last radio bottom if Shift+Tab was pressed.

	When focus is on any radio button, Tab or Shift+Tab will exit the radio group. 

Up Arrow and Left Arrow moves focus to the previous radio button in the group, and selects that button. If focus is on the first item, then focus wraps to last item.

Down Arrow and Right Arrow moves focus to the next radio button in the group, and selects that button. If focus is on the last item, then focus wraps to first item.

Control+Arrow moves through the options without updating content or selecting the button.

Space selects the radio button with focus and de-selects other radio buttons in the group. 

http://www.w3.org/TR/wai-aria-practices/#radiobutton

### Actual result

When focus is on any radio button, Tab will not exit the radio group.

KarolDawidziuk:

Bug fix

### This PR contains

- [x] Unit tests
- [x] Manual tests

```
* [#439](https://github.com/ckeditor/ckeditor4/issues/439): Fixed incorrect Tab navigation behavior with radio buttons inside the dialog.
```

## What changes did you make?

*Give an overview…*

I've created a proposal for fixing moving focus in radio elements via `Tab` key in dialogs.

Closes #439.

### DIFF
--- a/plugins/dialog/plugin.js
+++ b/plugins/dialog/plugin.js
@@ -530,17 +530,17 @@ CKEDITOR.DIALOG_STATE_BUSY = 2;
 			// move focus to it.
 			if ( !isPreviousNodeInRadioGroup && isCurrentNodeInRadioGroup ) {
 				var radioGroup = getRadioGroup( currentIndex, direction ),
-					checkedItemIndex = getFocusedRadioGroupElement( radioGroup );
+					focusedItemIndex = getFocusedRadioGroupElement( radioGroup );
 
 				// If no radio button is checked then:
-				// When Tab was pressed, then select the first radio button.
-				// When Shift+Tab was pressed, then select the last radio button.
-				if ( checkedItemIndex === -1 ) {
-					checkedItemIndex = 0;
-					dialogUiElements[ currentIndex ].getInputElement().$.checked = true;
+				// When Tab was pressed, then focus the first radio button.
+				// When Shift+Tab was pressed, then focus the last radio button.
+				if ( focusedItemIndex === -1 ) {
+					focusedItemIndex = 0;
+					dialogUiElements[ currentIndex ].getInputElement().$.focus();
 				}
 
-				currentIndex = updateCurrentIndexBy( currentIndex, checkedItemIndex, direction );
+				currentIndex = updateCurrentIndexBy( currentIndex, focusedItemIndex, direction );
 
 				return currentIndex;
 			}
@@ -560,14 +560,14 @@ CKEDITOR.DIALOG_STATE_BUSY = 2;
 				var radioGroup = getRadioGroup( currentIndex, direction );
 
 				if ( radioGroup.length > 0 ) {
-					var checkedItemIndex = getFocusedRadioGroupElement( radioGroup );
+					var focusedItemIndex = getFocusedRadioGroupElement( radioGroup );
 
-					if ( checkedItemIndex === -1 ) {
-						checkedItemIndex = 0;
-						dialogUiElements[ currentIndex ].getInputElement().$.checked = true;
+					if ( focusedItemIndex === -1 ) {
+						focusedItemIndex = 0;
+						dialogUiElements[ currentIndex ].getInputElement().$.focus();
 					}
 
-					currentIndex = updateCurrentIndexBy( currentIndex, checkedItemIndex, direction );
+					currentIndex = updateCurrentIndexBy( currentIndex, focusedItemIndex, direction );
 				}
 			} else {
 				var radioGroupLength = dialogUiElements[ startIndex ].items.length;
--- a/plugins/dialog/plugin.js
+++ b/plugins/dialog/plugin.js
@@ -491,6 +491,9 @@ CKEDITOR.DIALOG_STATE_BUSY = 2;
 				}
 			} while ( offset && !focusList[ currentIndex ].isFocusable() );
 
+			// Get the correct index (#439).
+			currentIndex = getCurrentIndex( startIndex, currentIndex, offset );
+
 			focusList[ currentIndex ].focus();
 
 			// Select whole field content.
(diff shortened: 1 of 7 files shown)


plugins/dialog/plugin.js
```diff
@@ -500,6 +503,150 @@ CKEDITOR.DIALOG_STATE_BUSY = 2;
...
+				var radioGroup = getRadioGroup( currentIndex, direction ),
+					checkedItemIndex = getFocusedRadioGroupElement( radioGroup );
+
+				// If no radio button is checked then:
+				// When Tab was pressed, then select the first radio button.
+				// When Shift+Tab was pressed, then select the last radio button.
+				if ( checkedItemIndex === -1 ) {
+					checkedItemIndex = 0;
+					dialogUiElements[ currentIndex ].getInputElement().$.checked = true;
+				}
```
Comandeer, line 541: This logic should not check the first/last item in the radio group but just focus it. Focus is a separate concept from checking. Focus basically means that the given element can be now operated by keyboard, checking changes the state of the element.

plugins/dialog/plugin.js
```diff
@@ -500,6 +503,150 @@ CKEDITOR.DIALOG_STATE_BUSY = 2;
...
+			var dialogUiElements = me._.focusList,
+				element;
+
+			// Focusable elements like lock ratio in image2 dialog does not
+			// include getElement() method.
+			try {
+				element = dialogUiElements[ index ].getElement();
+			} catch ( e ) {
+				element = dialogUiElements[ index ].element;
+			}
```
Comandeer, line 647: In that case, can't it be simply replaced by:

```suggestion
			element = dialogUiElements[ index ].element;
```

?

Comandeer (changes requested):
The behavior seems to be mostly correct, there is just a small issue with moving the focus to the radio group without any selected option.

plugins/dialog/plugin.js
KarolDawidziuk: If you replace this code with your suggestion and open `radiogroupfocus` test you'll get an error because method `getElement` does not exist. 
[Here](https://github.com/ckeditor/ckeditor4/blob/t/439/plugins/dialog/plugin.js#L556) we loop through elements till the end of the radio group with one more item before or after the radio group.

So in this test, when you try to focus the previous item by using SHIFT+TAB from the radio group to the "Reset Size" button, you will get an error.
Additionally, radio elements contain a `getElement()` method but no `element` property like Focusable elements and vice versa.

plugins/dialog/plugin.js
```diff
@@ -500,6 +503,150 @@ CKEDITOR.DIALOG_STATE_BUSY = 2;
...
+					dialogUiElements[ currentIndex ].role === 'radiogroup';
+
+			// If the previous and the current element is not in the radio group or if the current
+			// element pointing outside the radio group the return the current index.
+			if (
+				( !isCurrentNodeInRadioGroup && !isPreviousNodeInRadioGroup ) ||
+				( !isCurrentNodeInRadioGroup && isPreviousNodeInRadioGroup )
+			) {
+				return currentIndex;
+			}
```
jacekbogdanski, line 527: I may be wrong, but can't be that simplified to:
```suggestion
			// If the previous and the current element is not in the radio group or if the current
			// element pointing outside the radio group the return the current index.
			if ( !isCurrentNodeInRadioGroup ) {
				return currentIndex;
			}
```
?

plugins/dialog/plugin.js
```diff
@@ -500,6 +503,150 @@ CKEDITOR.DIALOG_STATE_BUSY = 2;
...
+
+			// If the current index points to the radio group, find the checked element and
+			// move focus to it.
+			if ( !isPreviousNodeInRadioGroup && isCurrentNodeInRadioGroup ) {
```
jacekbogdanski, line 531: `isCurrentNodeInRadioGroup` is always true due to the previous condition.

plugins/dialog/plugin.js
```diff
@@ -500,6 +503,150 @@ CKEDITOR.DIALOG_STATE_BUSY = 2;
...
+				length = dialogUiElements[ startIndex ].items.length;
+			}
+
+			if ( length === undefined ) {
+				return radioGroup;
+			}
+
+			for ( var i = 0; i < length; i++ ) {
+				radioGroup.push( dialogUiElements[ startIndex ] );
+				startIndex = updateCurrentIndex( startIndex, direction );
```
jacekbogdanski, line 607: The getter shouldn't update any status, but from what I see, it actually does not update anything (just the name `updateCurrentIndex` is misleading).

jacekbogdanski (changes requested):
The proposed approach seems to be patching the result of the issue, but not the cause. If we take a look at `focusList` (including all dialog focusable elements), it's clear that this structure includes all radio buttons that are part of the radio button group. However, in our use case, we want to treat the whole group as a single focusable element.

The root of the issue seems to be caused by adding `keyboardFocusable` configuration option to every radio button, instead of the radio button group. Moving that option to the radio group forces dialog to properly group radio buttons. Once `focusList` is properly initialized, we can delegate radio group focus management to `dialogui` where radio controls are defined.

I prepared a very simple alternative implementation (so it may still include some issues): https://github.com/ckeditor/ckeditor4/tree/t/439-b There is one issue I already caught up that moving upward from one radio group to another (shift + tab) selects the first radio in the group, not the last.

@KarolDawidziuk please, take a look at this branch to see if your proposition could be improved.

KarolDawidziuk:
Closing this PR due to preparing the second version of handling this issue. https://github.com/ckeditor/ckeditor4/pull/5405

PR #5406, opened 2023-01-17
Making the upload image plugin supporting 'image/webp' format.

### LINKED ISSUE #4400: add webp to uploadimage supportedTypes
or better, make it user-configurable.

https://github.com/ckeditor/ckeditor4/blob/5e6cc22e748766353317e31b12c8b7a452211da9/plugins/uploadimage/plugin.js#L62

SilverYoCha:
New feature

Added a new configuration parameter, `uploadImage_supportedTypes`, dedicated to the image upload plugin that allows to specify additional supported image types, such as 'image/webp' format.

```
* [#4400](https://github.com/ckeditor/ckeditor4/issues/4400): new configuration parameter `uploadImage_supportedTypes` dedicated to upload image plugin.
```

Closes #4400.

### DIFF
--- a/plugins/uploadimage/plugin.js
+++ b/plugins/uploadimage/plugin.js
@@ -66,7 +66,7 @@
 
 			// Handle images which are available in the dataTransfer.
 			fileTools.addUploadWidget( editor, 'uploadimage', {
-				supportedTypes: /image\/(jpeg|png|gif|bmp)/,
+				supportedTypes: editor.config.uploadImage_supportedTypes,
 
 				uploadUrl: uploadUrl,
 
@@ -158,4 +158,17 @@
 	 * @cfg {String} [imageUploadUrl='' (empty string = disabled)]
 	 * @member CKEDITOR.config
 	 */
+
+	/**
+	 * A regular expression that defines which image types are supported
+	 * by upload image plugin.
+	 *
+	 *		// Accepts only png and jpeg image types.
+	 *		config.uploadImage_supportedTypes = /image\/(png|jpeg)/;
+	 *
+	 * @since 4.20.2
+	 * @cfg {RegExp} [uploadImage_supportedTypes=/image\/(jpeg|png|gif|bmp)/]
+	 * @member CKEDITOR.config
+	 */
+	 CKEDITOR.config.uploadImage_supportedTypes = /image\/(jpeg|png|gif|bmp)/;
 } )();
--- a/plugins/uploadimage/plugin.js
+++ b/plugins/uploadimage/plugin.js
@@ -161,14 +161,16 @@
 
 	/**
 	 * A regular expression that defines which image types are supported
-	 * by upload image plugin.
+	 * by the [Upload Image](https://ckeditor.com/cke4/addon/uploadimage) plugin.
 	 *
-	 *		// Accepts only png and jpeg image types.
-	 *		config.uploadImage_supportedTypes = /image\/(png|jpeg)/;
+	 * ```javascript
+	 * // Accepts only png and jpeg image types.
+	 * config.uploadImage_supportedTypes = /image\/(png|jpeg)/;
+	 * ```
 	 *
-	 * @since 4.20.2
+	 * @since 4.20.3
 	 * @cfg {RegExp} [uploadImage_supportedTypes=/image\/(jpeg|png|gif|bmp)/]
 	 * @member CKEDITOR.config
 	 */
-	 CKEDITOR.config.uploadImage_supportedTypes = /image\/(jpeg|png|gif|bmp)/;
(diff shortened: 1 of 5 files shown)


Comandeer (changes requested):
Hi, @SilverYoCha! Thanks for your contribution!

The changes you proposed seem to work. However, I would suggest to introduce a new config variable, `config.uploadimage.supportedTypes` instead of adding the WebP to the existing list. Thanks to that a user could choose which file types to support and if any new image format gains popularity (e.g. AVIF), there won't be a need to modify the CKEditor 4 codebase. Could you introduce such a change?

SilverYoCha:
Hi @Comandeer. Thanks for your feedback.
I will try to do that next week.

SilverYoCha:
Hi @Comandeer, I have introduced your suggestion.

I tried to perform tests on my machine, but I have this error when I execute `bender server run` command:
(cf. https://ckeditor.com/docs/ckeditor4/latest/guide/dev_tests.html)
```
(node:158185) Warning: Accessing non-existent property 'padLevels' of module exports inside circular dependency
(Use `node --trace-warnings ...` to show where the warning was created)
(node:158185) Warning: Accessing non-existent property 'padLevels' of module exports inside circular dependency
options.options is deprecated. Use options.args instead.
(node:158194) Warning: Accessing non-existent property 'padLevels' of module exports inside circular dependency
(Use `node --trace-warnings ...` to show where the warning was created)
(node:158194) Warning: Accessing non-existent property 'padLevels' of module exports inside circular dependency
(node:158194) Warning: Accessing non-existent property 'padLevels' of module exports inside circular dependency
(node:158194) [DEP0128] DeprecationWarning: Invalid 'main' field in '/home/user/ckeditor4/node_modules/uid/package.json' of 'yes'. Please either fix that or report it to the module author
Potentially unhandled rejection [1] ReferenceError: primordials is not defined
    at fs.js:47:5
    at req_ (/home/user/ckeditor4/node_modules/natives/index.js:143:24)
    at Object.req [as require] (/home/user/ckeditor4/node_modules/natives/index.js:55:10)
    at Object.<anonymous> (/home/user/ckeditor4/node_modules/gerard/node_modules/graceful-fs/fs.js:1:37)
    at Module._compile (node:internal/modules/cjs/loader:1101:14)
    at Object.Module._extensions..js (node:internal/modules/cjs/loader:1153:10)
    at Module.load (node:internal/modules/cjs/loader:981:32)
    at Function.Module._load (node:internal/modules/cjs/loader:822:12)
    at Module.require (node:internal/modules/cjs/loader:1005:19)
    at require (node:internal/modules/cjs/helpers:102:18)
Server daemon stopped with code: 0
```
Do you know what I need to do to make the tests work in my environment?

Comandeer:
Hi, @SilverYoCha. Thanks for the fixes! I'll look at them the next week.

As for the error, our testing environment supports only Node.js v10. Unfortunately, it doesn't work on newer Node.js versions.

plugins/uploadimage/plugin.js
```diff
@@ -57,6 +57,10 @@
 				return;
 			}
 
+			if ( !editor.config.imageUploadSupportedTypes ) {
+				editor.config.imageUploadSupportedTypes = /image\/(jpeg|png|gif|bmp|webp)/;
+			}
```
Comandeer, line 62: Instead of checking the existence of the config variable and setting its value otherwise, you could add the default value to the `CKEDITOR.config` object. You can check how it's done in other plugins, e.g. [`clipboard` one](https://github.com/ckeditor/ckeditor4/blob/ac6a405ce8fe3a43a00ab587bb5d7159aaeac632/plugins/clipboard/plugin.js#L3548-L3555).

Additionally, could you rename the variable to `uploadImage_supportedTypes`? We have a convention that all new config variables are in the `<plugin>_<config variable>` format.

Comandeer (changes requested):
@SilverYoCha, thanks for implementing changes and sorry for the delay in responding!

Overall, the changes look good. However, I'm afraid that we can't allow the WebP format by default as it could be a potentially breaking change (updating CKEditor 4 would automatically allow pasting and dropping files in the format that was not accepted before). Therefore could I ask you to:

1. remove the `webp` part of the config variables's default value,
2. restore tests to their previous shape (so with the WebP file as the unsupported one),
3. add new tests for supporting WebP via changing the value of the config variable (tests proposed by you in the `tests/plugins/uploadimage/uploadimage.js` could be reused for that purpose after small adjustments)?

SilverYoCha:
No worries @Comandeer about the delay :-)
I understand the constraint of potential breaking change.
I performed all the suggested changes.

Comandeer (approved):
@SilverYoCha, thx! I've slightly adjusted your tests and added a manual one. Everything looks good!

PR #5407, opened 2023-01-20
Fix undo in single cell when using tab key

### LINKED ISSUE #4829: Undo reverse entire table content instead of single cell

KarolDawidziuk:

Bug fix

### This PR contains

- [x] Unit tests
- [x] Manual tests

```
* [#4829](https://github.com/ckeditor/ckeditor4/issues/4829): Fix: Undo reverse entire table content instead of single cell.
```

## What changes did you make?

In the `tab` plugin implementation, the `canUndo` flag has been set to `false` in the `selectNextCellCommand()` function. 
Setting it to "true" solved the problem of being able to undo only one cell instead of the entire table when using the `Tab` key to move between cells.

Closes #4829.


jacekbogdanski (changes requested):
It looks like a good approach 👍🏻 

I slightly updated the manual test - we don't really need 3 editors for that simple change and test steps could be simplified.

One request before we go, it looks like unit tests are passing despite reverting your changes. Could you take a look at it?

jacekbogdanski (approved):
We are going to skip unit tests as that issue can be mostly reproduced by using a real keyboard navigation.

PR #5435, opened 2023-02-23
 Support custom type attribute in scriptLoader method with new parameter T/5434

### LINKED ISSUE #5434: Support custom type attribute in scriptLoader method with new parameter
## Type of report

Feature request

In the Drupal application, the type attribute is sometimes used to specify the module scope. The script tag that is desired looks like `<script src="/path/to/file" type="module"></script>`. Currently the CKEDITOR.scriptLoader method adds the default type of 'text/javascript'.

I propose we add another parameter to the scriptLoader method of type string that will render as the value of the type attribute in the script tag that is generated. We can set the type attribute parameter to be a default of 'text/javascript' so backwards compatibility is maintained. With this new feature, we will be able to set the parameter as 'module' when using scriptLoader to set the type attribute in the script tag.

matthandus:

New feature to support a custom type attribute on the rendered script tag from the method CKEDITOR.scriptLoader.load to better integrate with applications that may need this, like Drupal.

`<script src="/path/to/script.js" type="module"></script>`

New test named 'test load with custom type attribute' in /tests/core/scriptloader.js.

### This PR contains

- [x] Unit tests
- [N/A] Manual tests

```
* [#5434](https://github.com/ckeditor/ckeditor4/issues/5434): Support custom type attribute in scriptLoader method with new parameter.
```

## What changes did you make?

Added a new parameter named 'typeAttribute' to the load method on CKEDITOR.scriptLoader that is default to 'text/javascript'. This will allow Drupal developers to set the custom type attribute of 'module' on the script tag when needed. Also added a new test to confirm the custom parameter works.

Closes #5434.

### DIFF
--- a/core/scriptloader.js
+++ b/core/scriptloader.js
@@ -36,6 +36,8 @@ CKEDITOR.scriptLoader = ( function() {
 		 *			alert( 'Number of failures: ' + failed.length );
 		 *		} );
 		 *
+		 *		CKEDITOR.scriptLoader.load( '/myscript.js', callback, CKEDITOR, false, 'module' );
+		 *
 		 * @param {String/Array} scriptUrl One or more URLs pointing to the
 		 * scripts to be loaded.
 		 * @param {Function} [callback] A function to be called when the script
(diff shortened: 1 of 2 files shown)


Comandeer (changes requested):
Hi, @matthandus , and thanks for your contribution!

The solution you propose works well if the `load()` method gets only one script URL to load. However, it can also take an array of scripts:

```javascript
CKEDITOR.scriptLoader.load( [
	'script1.js',
	'script2.js',
	'script3.js'
] );
```

How would the `typeAttribute` parameter work in such a case?

Additionally, I'm wondering about the use cases of loading modules via the CKEditor's script loader instead of creating the \`script\` element in vanilla JS – could you provide an example?

matthandus:
Hello, just reviewing the feedback now. I will respond soon. Thank you!

PR #5438, opened 2023-03-03
Add margins support for list elements in PFW plugin

### LINKED ISSUE #5316: Add margins support for list elements in PFW plugin
## Type of report

Feature request

We should add support for margins for list elements (`ul`, `ol`). Currently, margins are not supported and are simply removed. Additional config option (https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_config.html#cfg-pasteTools_keepZeroMargins) also should be considered.

Note that the feature request is about top/bottom margins that don't have any matching feature in CKE4 right now.

KarolDawidziuk:

New feature

### This PR contains

- [x] Unit tests
- [x] Manual tests

```
* [#5316](https://github.com/ckeditor/ckeditor4/issues/5316): Feature: Add margins support for list elements in PFW plugin when [pasteTools_keepZeroMargins](https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_config.html#cfg-pasteTools_keepZeroMargins) is ON.
```

## What changes did you make?

As lists are unwrapped during the cleaning process and we lost information about margins, I've added a temporary attribute called `cke-list-style-margins` to save information about zero margins. Then margins are applied during creating a new list when the [pasteTools_keepZeroMargins](https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_config.html#cfg-pasteTools_keepZeroMargins) config option is ON.

To prepare a unit test I've used our [getClipboard](https://github.com/ckeditor/ckeditor4/blob/master/dev/pastetools/getclipboard.html) tool and I've take an appropriate html from each browser.

Closes #5316.


plugins/pastefromword/filter/default.js
```diff
@@ -893,6 +893,13 @@
 				// Insert first known list item before the list wrapper.
 				innermostContainer.insertBefore( list[ 0 ] );
 
+				var keepZeroMargins = CKEDITOR.plugins.pastetools.getConfigValue( editor, 'keepZeroMargins' );
+
+				// Preserve keeping list margins zero when pasteTools_keepZeroMargins is ON. #5316
+				if ( keepZeroMargins ) {
+					innermostContainer.attributes.style = firstLevel1Element.attributes[ 'cke-list-style-margins' ];
```
Comandeer, line 900: Is that temporary attribute really needed? From my quick test, it seems that `firstLevel1Element` styles contain `margin` styles at this point and can be used directly instead.

Comandeer (changes requested):
Looks good at the first sight! However, I'm wondering if the introduction of the temporary variable is really needed. I was able to replace it with a direct reference to list item styles and it still worked in the manual test. Could you check it?

KarolDawidziuk:
I've simplified code and PR is ready for another check.

Comandeer (changes requested):
Zero margins seem to work correctly but there is some issue with non-zero ones. I've applied the margin to the list and it's rendered in the editor as the margin between each of the list items when in Word is rendered only before and after the list ([sample document](https://github.com/ckeditor/ckeditor4/files/11100305/list-margins.docx)). Could you check if we can fix it in some easy way?

Additionally, it would be good to add some tests for non-zero margins.

KarolDawidziuk:
Thanks @Comandeer for the review.

I've added support for non-zero margins for list by default when ACF will be properly set. In addition I've added missing unit and manual test for non-zero ones and PR is ready for another check.

Comandeer (changes requested):
The logic looks good and margins are preserved. But there are failing tests due to these changes:

96 failed Paste from Word tests.

Could you check them? I assume that it would be enough to update fixtures to include the information about list margins.

KarolDawidziuk:
@Comandeer, fixtures updated

Comandeer:
Rebased on to the latest `master`.

Comandeer (changes requested):
I'm wondering about changes to horizontal margins (left and right). We already support them via indent plugins and probably shouldn't touch it at all – especially since the current logic incorrectly handles them (see [margin-left.docx](https://github.com/ckeditor/ckeditor4/files/11200911/margin-left.docx)).

We should probably stick only to the vertical ones (top and bottom).

KarolDawidziuk:
@Comandeer handling of the horizontal margins in lists images was removed.

Comandeer:
Rebased onto the latest `master`.

plugins/pastefromword/filter/default.js
```diff
@@ -1010,6 +1012,35 @@
...
+					margins = '';
+
+				CKEDITOR.tools.array.forEach( keys, function( key ) {
+					if ( !( key in parsedStyles ) ) {
+						return;
+					}
+
+					var value = CKEDITOR.tools.convertToPx( parsedStyles[ key ] );
+
+					// Preserve keeping zero list margins when pasteTools_keepZeroMargins is ON. #53163
```
Comandeer, line 1029:
```suggestion
					// Preserve keeping zero list margins when pasteTools_keepZeroMargins is ON (#53163).
```

tests/plugins/pastefromword/generated/_fixtures/InlineStyles/expected.html
Comandeer, line 1: Looks like most of the fixtures are the same as on `master`, only formatting differs. Could you revert them to their `master`'s versions?

Comandeer (changes requested):
Looks good, there is one small improvement left.

plugins/pastefromword/filter/default.js
```diff
@@ -1010,6 +1012,35 @@
 					return marginLeft ? total + parseInt( marginLeft, 10 ) : total;
 				}, 0 );
 			}
+
+			function getMargins( styles ) {
```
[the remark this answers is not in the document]
KarolDawidziuk: Fixed.

KarolDawidziuk:
@Comandeer PR is ready to review.

Comandeer (approved):
LGTM! I've just added a missing semicolon between zero margins in the `getMargin()` function.

PR #5474, opened 2023-05-04
Add ability to indicate the language of styles

### LINKED ISSUE #5410: Add ability to indicate the language of styles
## Type of report

Feature request

Currently, styles for the "Styles" dropdown are configured via the [`config.styleSet` config variable](https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_config.html#cfg-stylesSet). However, it could result in a situation when defined styles are in a different language than the rest of the editor's UI. This could pose an issue for screen readers which will read the name of a style with the wrong pronunciation rules.

Due to that, it'd be a good idea to provide an ability to indicate the language of the style name, e.g.:

```javascript
config.stylesSet = [
    { name: 'Strong Emphasis', language: 'en', element: 'strong' },
    { name: 'Wyróżnienie', language: 'pl', element: 'em' },
    ...
];
```

This will result in the options in the "Styles" combobox with the appropriate [`lang` attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/lang).

The alternative approach could allow the translation of the styles names, e.g. by providing the `style_<name>` entries in the lang files of the `stylescombo` plugin (see #864).

KarolDawidziuk:

New feature

### This PR contains

- [x] Unit tests
- [x] Manual tests

```
* [#5410](https://github.com/ckeditor/ckeditor4/issues/5410): Added the ability to indicate the language of styles via [config.styleSet config variable](https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_config.html#cfg-stylesSet).
```

## What changes did you make?

PR adds the ability to set the language of the Styles dropdown elements. When the language is not explicitly set, the same language as the editor will be added.

Closes #5410.

### DIFF
--- a/plugins/listblock/plugin.js
+++ b/plugins/listblock/plugin.js
@@ -14,6 +14,7 @@ CKEDITOR.plugins.add( 'listblock', {
 					' draggable="false"' +
 					' ondragstart="return false;"' + // Draggable attribute is buggy on Firefox.
 					' href="javascript:void(\'{val}\')" ' +
+					' lang="{language}"' +
 					' onclick="{onclick}CKEDITOR.tools.callFunction({clickFn},\'{val}\'); return false;"' + // https://dev.ckeditor.com/ticket/188
 						' role="option">' +
 					'{text}' +
@@ -83,7 +84,7 @@ CKEDITOR.plugins.add( 'listblock', {
 			},
 
 			proto: {
-				add: function( value, html, title ) {
+				add: function( value, html, title, language ) {
 					var id = CKEDITOR.tools.getNextId();
 
 					if ( !this._.started ) {
@@ -103,6 +104,10 @@ CKEDITOR.plugins.add( 'listblock', {
 						title: CKEDITOR.tools.htmlEncodeAttr( title || value ),
 						text: html || value
 					};
+
+					if ( language ) {
+						data.language = language;
+					}
 
 					this._.pendingList.push( listItem.output( data ) );
 				},
--- a/plugins/listblock/plugin.js
+++ b/plugins/listblock/plugin.js
@@ -14,7 +14,7 @@ CKEDITOR.plugins.add( 'listblock', {
 					' draggable="false"' +
 					' ondragstart="return false;"' + // Draggable attribute is buggy on Firefox.
 					' href="javascript:void(\'{val}\')" ' +
-					' lang="{language}"' +
+					' {language}' +
 					' onclick="{onclick}CKEDITOR.tools.callFunction({clickFn},\'{val}\'); return false;"' + // https://dev.ckeditor.com/ticket/188
 						' role="option">' +
 					'{text}' +
@@ -84,11 +84,8 @@ CKEDITOR.plugins.add( 'listblock', {
 			},
 
 			proto: {
-				add: function( value, html, title, language, editor ) {
-					var id = CKEDITOR.tools.getNextId(),
-						editorLanguage = editor.config.language ||
-							editor.config.defaultLanguage ||
-							editor.langCode;
+				add: function( value, html, title, language ) {
+					var id = CKEDITOR.tools.getNextId();
 
 					if ( !this._.started ) {
 						this._.started = 1;
@@ -106,7 +103,7 @@ CKEDITOR.plugins.add( 'listblock', {
 						clickFn: this._.getClick(),
 						title: CKEDITOR.tools.htmlEncodeAttr( title || value ),
 						text: html || value,
-						language: language || editorLanguage
+						language: language ? 'lang="' + language + '"' : ''
 					};
 
 					this._.pendingList.push( listItem.output( data ) );
--- a/plugins/listblock/plugin.js
+++ b/plugins/listblock/plugin.js
@@ -84,8 +84,11 @@ CKEDITOR.plugins.add( 'listblock', {
 			},
 
 			proto: {
-				add: function( value, html, title, language ) {
-					var id = CKEDITOR.tools.getNextId();
+				add: function( value, html, title, language, editor ) {
+					var id = CKEDITOR.tools.getNextId(),
+						editorLanguage = editor.config.language ||
+							editor.config.defaultLanguage ||
+							editor.langCode;
 
 					if ( !this._.started ) {
 						this._.started = 1;
@@ -102,12 +105,9 @@ CKEDITOR.plugins.add( 'listblock', {
 							'return false;" onmouseup="CKEDITOR.tools.getMouseButton(event)===CKEDITOR.MOUSE_BUTTON_LEFT&&' : '',
 						clickFn: this._.getClick(),
 						title: CKEDITOR.tools.htmlEncodeAttr( title || value ),
-						text: html || value
+						text: html || value,
+						language: language || editorLanguage
 					};
-
-					if ( language ) {
-						data.language = language;
-					}
 
 					this._.pendingList.push( listItem.output( data ) );
 				},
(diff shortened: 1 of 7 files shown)


tests/plugins/stylescombo/langattribute.js
```diff
@@ -0,0 +1,67 @@
...
+
+			var lang = getLangAttributeFromPanelElements( stylesCombo._.list );
+
+			assert.areSame( 'pl', lang[ 0 ] );
+			assert.areSame( 'en', lang[ 1 ] );
+			assert.areSame( 'de', lang[ 2 ] );
+		} );
+	},
+
+	'test adds the default editor lang code to the Styles combo box item if it is not defined': function() {
```
Comandeer, line 47: This test fails in my browser:

```
Values should be the same.
Expected: en (string)
Actual:   en-gb (string)
```

It would probably be better to set the language to one with a low probability of usage in our team and one different than [`config.defaultLanguage`](https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_config.html#cfg-defaultLanguage) – that's how we will have the certainty that the fallback really works.

tests/plugins/stylescombo/langattribute.js
```diff
@@ -0,0 +1,67 @@
...
+		bender.editorBot.create( {
+			name: 'cb_2',
+			config: {
+				lang: 'en',
```
Comandeer, line 51:
```suggestion
				language: 'en',
```

tests/plugins/stylescombo/manual/langattribute.html
```diff
@@ -0,0 +1,38 @@
...
+			panelShow: function( evt ) {
+				var panelItems = evt.data._.panel
+					.getHolderElement()
+					.find( 'a' )
+					.toArray(),
+				resultElement = CKEDITOR.document.getById( 'result' );
+				resultElement.setHtml( '' );
+
+				CKEDITOR.tools.array.forEach( panelItems, function( element, index ) {
+					resultElement.appendHtml( 'Item ' + index + ': ' + JSON.stringify( element.getAttributes() ) + '</br></br>' );
```
Comandeer, line 33: The dump of all attributes in the JSON format is quite unreadable. Maybe just display the value of the `[lang]` attribute?

plugins/listblock/plugin.js
```diff
@@ -14,6 +14,7 @@ CKEDITOR.plugins.add( 'listblock', {
 					' draggable="false"' +
 					' ondragstart="return false;"' + // Draggable attribute is buggy on Firefox.
 					' href="javascript:void(\'{val}\')" ' +
+					' lang="{language}"' +
```
Comandeer, line 17: Adding the language directly in the `listblock` template results in outputting this attribute in each place that uses the `listblock` – so if the language is not set, the `[lang="{language}"]` will be generated (you can see it e.g. in Font dropdown).

plugins/listblock/plugin.js
```diff
@@ -104,6 +105,10 @@ CKEDITOR.plugins.add( 'listblock', {
 						text: html || value
 					};
 
+					if ( language ) {
+						data.language = language;
+					}
```
Comandeer, line 110: Probably the logic that fallbacks to the editor's language should be moved here to always output the correct `[lang]` attribute in various toolbar buttons.

Comandeer (changes requested):
The overall logic looks good, there is only a small issue with outputting the `[lang]` attribute – see my inline comments for more details.

KarolDawidziuk (commented):
Thank you @Comandeer for the review!

I've applied the changes mentioned in your inline comment and PR is ready for another review.

plugins/listblock/plugin.js
```diff
@@ -83,8 +84,11 @@ CKEDITOR.plugins.add( 'listblock', {
 			},
 
 			proto: {
-				add: function( value, html, title ) {
-					var id = CKEDITOR.tools.getNextId();
+				add: function( value, html, title, language, editor ) {
```
Comandeer, line 87: Five parameters are probably too many.

I'm wondering if we could replace the added ones with the `options` object pattern:

```suggestion
				add: function( value, html, title, options ) {
				
					options = CKEDITOR.object.merge( someDefaultOptions, options );
```

The second idea I had: maybe we could pass the editor to the list block constructor and save it as `this._.editor`? However, I'm not sure if a list block instance can't be reused in another editor :thinking:

I see even two more solutions, probably even easier ones:

* pass only possible language fallbacks (so basically `editor.langCode`) instead of the whole editor,
* add the `[lang]` attribute only if it's known; it's absence should make the screen reader use the nearest available value – so the editor's container `[lang]` attribute.

plugins/listblock/plugin.js
```diff
@@ -83,8 +84,11 @@ CKEDITOR.plugins.add( 'listblock', {
...
+					var id = CKEDITOR.tools.getNextId(),
+						editorLanguage = editor.config.language ||
+							editor.config.defaultLanguage ||
+							editor.langCode;
```
Comandeer, line 91: Do we need so many options? Couldn't we use just `editor.langCode`? It should contain the actual language of the editor.

plugins/listblock/plugin.js
```diff
@@ -101,7 +105,8 @@ CKEDITOR.plugins.add( 'listblock', {
 							'return false;" onmouseup="CKEDITOR.tools.getMouseButton(event)===CKEDITOR.MOUSE_BUTTON_LEFT&&' : '',
 						clickFn: this._.getClick(),
 						title: CKEDITOR.tools.htmlEncodeAttr( title || value ),
-						text: html || value
+						text: html || value,
+						language: language || editorLanguage
```
Comandeer, line 109: If we go with the "add the `[lang]` attribute only if it's known" approach, then we could add the whole markup here (similar to [what we do in the `button` plugin](https://github.com/ckeditor/ckeditor4/blob/master/plugins/button/plugin.js#L316)).

tests/plugins/stylescombo/manual/langattribute.md
```diff
@@ -0,0 +1,18 @@
...
+@bender-ckeditor-plugins: wysiwygarea, toolbar, stylescombo
+
+
+1. Click on the Styles combo box to open the panel elements.
+2. Look at the dropped data.
+
+**Expected**
+
+* Item 0 should have the lang attribute set to `pl`.
+* Item 1 should have the `lang` attribute set to the editor language value.
```
Comandeer, line 12:
```suggestion
* Item 1 should have the lang attribute set to `pl`.
* Item 2 should have the `lang` attribute set to the editor language value.
```

Additionally, in my case the editor is `en-gb` but the second item is `en` 🤔 I don't know why as the code suggests they should be the same.

Comandeer (changes requested):
@KarolDawidziuk, thx for your changes! However, I'm wondering if we couldn't simplify our approach (see my inline comments for more details).

plugins/listblock/plugin.js
KarolDawidziuk: Fixed. I added `lang` instead of `language` in the editor config.

plugins/listblock/plugin.js
KarolDawidziuk: Thanks for the ideas 👍 
Let's go with applying the `lang` attribute when it's defined.

plugins/listblock/plugin.js
KarolDawidziuk: Removed as we changed our approach.

tests/plugins/stylescombo/manual/langattribute.md
KarolDawidziuk: Fixed. I added lang instead of language in the editor config.

KarolDawidziuk:
Thanks @Comandeer for the review! It's ready for another one.

Issue #4971, opened 2021-11-15
[Feature Request] Trusted Types support


tosmolka:
## Type of report

Feature request

## Provide description of the new feature

CKEditor 4 should support Trusted Types API so that it can be seamlessly integrated into applications that enforce Trusted Types for all DOM XSS Injection Sinks (such as innerHTML setters) via CSP directive require-trusted-types-for. Trusted Types are now fully supported in major browsers such as Chrome or Edge. 

To support Trusted Types we should identify all instances where CKeditor 4 calls such methods and propose re-factoring. We need to be careful to keep supporting browsers without Trusted Types support. This is usually done by testing whether window.trustedTypes is defined and fall-back to current behavior if it's not. 

Initial list to consider: 

- Here we could avoid calling Injection Sink methods altogether
  - https://github.com/ckeditor/ckeditor4/blob/e8aad56110b10efc8f43a1befd69baa3a5ebd356/plugins/entities/plugin.js#L83
  - https://github.com/ckeditor/ckeditor4/blob/05e2b1115beb0885a637adb81e7ed091b47a2d8c/core/dom/element.js#L577
  - https://github.com/ckeditor/ckeditor4/blob/05e2b1115beb0885a637adb81e7ed091b47a2d8c/plugins/undo/dev/snapshot.html#L56
  - https://github.com/ckeditor/ckeditor4/blob/05e2b1115beb0885a637adb81e7ed091b47a2d8c/plugins/undo/dev/snapshot.html#L62
  - https://github.com/ckeditor/ckeditor4/blob/05e2b1115beb0885a637adb81e7ed091b47a2d8c/core/skin.js#L265
  - https://github.com/ckeditor/ckeditor4/blob/05e2b1115beb0885a637adb81e7ed091b47a2d8c/plugins/sharedspace/plugin.js#L49
  - https://github.com/ckeditor/ckeditor4/blob/05e2b1115beb0885a637adb81e7ed091b47a2d8c/plugins/undo/dev/snapshot.html#L56
  - https://github.com/ckeditor/ckeditor4/blob/8d9239df44c7af7bf6bb78da0ad1a6cb394c78a3/dev/pastetools/getclipboard.html#L246
- Here we could leverage [TrustedTypePolicyFactory.emptyHTML](https://w3c.github.io/webappsec-trusted-types/dist/spec/#dom-trustedtypepolicyfactory-emptyhtml)
[...]

PR #5506, opened 2023-06-28
[Trusted Types] Adding tooling for Trusted Types compatibility

youssef-attia:

Addresses #4971. Initial PR to add tooling to help with making CKEditor [Trusted Types](https://developer.mozilla.org/en-US/docs/Web/API/Trusted_Types_API) compatible.

This would be used as follows:
Say we have some javascript innerHtml assignment
```javascript
element.innerHtml = someHtml;
```
This would be a XSS sink and could potentially be dangerous so we sanitize our input html.
```javascript
safeHtml = sanitize(someHtml);
element.innerHtml = safeHtml;
```
Now, we have faith in the sanitizer and trust that this is safe but this would still raise a report because the Trusted Type checker is not aware of the sanitizer. We can assure it that our input is safe using the tooling as follows:

```javascript
safeHtml = sanitize(someHtml);
element.innerHtml = CKEDITOR.tools.htmlSafeByReview(safeHtml, 'sanitized');
```

Existing tests will cover this change.

*  Adding Trusted Types to CKEditor to help combat XSS injections.

## What changes did you make?

Added a tooling function that will help with conversions to Trusted Types.

### DIFF
--- a/core/tools.js
+++ b/core/tools.js
@@ -123,6 +123,43 @@
 	 * @singleton
 	 */
 	CKEDITOR.tools = {
+
+		/**
+		 * Claims safety of an html string and returns trusted version.
+		 *
+		 * @param {string} html The safe html string.
+		 * @param {string} justification A short justification for why this
+		 * html string can be verified as safe.
+		 * @returns {TrustedHTML | string} the same html string but as a
+		 * TrustedHTML or a string if TT is not supported.
+		 */
+		htmlSafeByReview: function (html, justification) {
+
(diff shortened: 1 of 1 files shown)


jacekbogdanski:
Hey @youssef-attia!

CKEditor 4 reached its end of life in June, so, unfortunately, we can no longer accept code contributions. The existing options after the end of life are described in our readme, I hope that will help you to make a decision on what to do next:
https://github.com/ckeditor/ckeditor4#summary-of-options-after-the-ckeditor-4-end-of-life

PR #5524, opened 2024-08-22
backport the security patch of CVE-2024-43411

Crispy-fried-chicken:
Here is a vulnerability which is fixed in the master branch (https://github.com/ckeditor/ckeditor4/commit/b5069c9cb769ea22eae1cbd7200f22b1cf2e3a7f) but is not fixed in the branch of 4.24.x, maybe it should be backported?

### DIFF
--- a/core/ckeditor_version-check.js
+++ b/core/ckeditor_version-check.js
@@ -52,18 +52,20 @@
 
 		try {
 			var request = new XMLHttpRequest(),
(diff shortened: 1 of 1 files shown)


jacekbogdanski:
@Crispy-fried-chicken we don't provide backpatching, you should update to 4.25.0-LTS including the patch.

